PR #18, opened 2017-02-03
Format ESP.getChipId as an 8 character hex instead of 7 digit decimal

don-willingham:
Copy of https://github.com/arendst/Sonoff-MQTT-OTA-Arduino/pull/289

I have two sonoff duals with chip id's of 15163478 and 15165864. Adding an eighth would work for now, but 10 should be required. Although unlikely, it's possible for UUID collisions. I have two, but as some people have greater numbers of devices, and likely purchased in batches, it's plausible the issue will come up.

ESP.getChipId() returns a 32 bit integer, which can range up to 2,147,483,647. That would require 10 decimal digits. When I was experimenting with simply adding 1 to the UUID for the second channel on a dual, the UUID for both channels would match. They didn't differ as the least significant digit(s) were truncated. My https://github.com/don-willingham/Sonoff-MQTT-OTA-Arduino/tree/dual-wemo branch appends another digit for channel number to the UUID's.

### DIFF
--- a/sonoff/xdrv_wemohue.ino
+++ b/sonoff/xdrv_wemohue.ino
@@ -52,14 +52,14 @@ const char WEMO_MSEARCH[] PROGMEM =
 
 String wemo_serial()
 {
-  char serial[15];
-  snprintf_P(serial, sizeof(serial), PSTR("201612K%07d"), ESP.getChipId());
+  char serial[16];
+  snprintf_P(serial, sizeof(serial), PSTR("201612K%08X"), ESP.getChipId());
   return String(serial);
 }
 
 String wemo_UUID()
 {
-  char uuid[26];
+  char uuid[27];
   snprintf_P(uuid, sizeof(uuid), PSTR("Socket-1_0-%s"), wemo_serial().c_str());
   return String(uuid);
 }


arendst:
Don did you check the result (8 hex digits). I ask as I failed to get all 8 digits once tried the same.

If so I will implement.

don-willingham:
Here's a test...
1. Flashed 3.9.10 to ESP-01
2. Enabled WeMo emulation
3. Browsed to /setup.xml URL, and it contains <serialNumber>201612K1292396</serialNumber>
4. Browsed to /in URL, and it contains ESP Chip id	1292396, oops, 7 digits, unlike the 8 of my sonoff.
5. Replaced ESP.getChipId() in xdrv_wemohue.ino with 15163478 from one of my sonoffs.
6. Browsed to /setup.xml again, and it contains <serialNumber>201612K1516347</serialNumber>, which is missing the 8.
7. Updated to this patch, replaced ESP.getChipId() with 15163478, and now its <serialNumber>201612K00E76056</serialNumber>
8. Google search "15163478 to hex" and it returns 0xE76056; but that's not padding with zeros, and we don't need the 0x, 

So it will depend on what ESP.getChipId() returns for each sonoff/esp-whatever. Since it's a 32 bit integer.

PR #25, opened 2017-02-04
Should we unify /RESULT with /STATE?

Hogesyx:


ThomDietrich:
RESULT is the switching result yielded after a command. STATE is the current state of the module, send every few minutes. STATE already includes RESULT. Both topics have their purpose, they should not be changed.
Why mixing different messages on one topic is a bad idea was already discussed here btw: https://github.com/arendst/Sonoff-Tasmota/issues/4

Hogesyx:
I am assuming /POWER is the result feedback, so why do we have both /POWER and /RESULT doing the same thing? By renaming /RESULT we give it a new purpose which is to update the new current /STATE

Hogesyx:
Also /tele/sonoff/STATE and /stat/sonoff/STATE are still different. We won't be mixing tele and stat at level 1 and we get to make STATE at level 3 do the same thing.

PR #48, opened 2017-02-09
HUE arg parsing using ArduinoJSON / Initial IR remote support (skeleton)

altelch:
Hi,

after a while 2 changes/additions. Switched to ArduinoJSON to handle HUE arg parsing what makes the code much cleaner especially when adding more features. I also need the JSON parsing for the comming IR extension so I think it makes sense to include some more "overhead". IR send will use the Sonoff pin header to connect an IR sender diod. Have a working setup which fits into the original sonoff and uses a marked "hole" on the cover which exactly fits the diod. My plan is to get a MQTT -> IR gateway to switch on devices which don't completly turn on after power is switched on (like some TV sets). In the end it will be a simple MQTT Harmony hub. I like the new Tasmota code, integrating new Hardware feels very clean.

Greetings,

        Heiko.


altelch:
Travis-CI seems to have an older version of IRremoteESP8266.

altelch:
Messed something up in IR code. JSON part is ok. Please wait.

PR #56, opened 2017-02-11
no quotes around Energy values in json

Nayar:
Problem:
When sending the logs into Elasticsearch through Graylog, the Energy_Power fields are stored as strings instead of numbers. It's not possible to graph the items.

The proper JSON way would be to not quote ints and floats.


arendst:
You seem to be right after I gleaned over the JSON definition ;-)

I will update all floats to non strings in the near future.

Thnx for attending me.

arendst:
Implemented in version 3.9.15 just released

PR #59, opened 2017-02-12
HUE arg parsing using ArduinoJSON / MQTTtoIR remote support

altelch:
Hi,

Switched to ArduinoJSON to handle HUE arg parsing what makes the code much cleaner especially when adding more features. I also need the JSON parsing for the IR extension so I think it makes sense to include some more "overhead". IR send will use the Sonoff pin header to connect an IR sender diod (will document hardware in Wiki later). Have a working setup which fits into the original sonoff and uses a marked "hole" on the cover which exactly fits the diod. Now I have a working MQTT -> IR gateway to switch on devices which don't completly turn on after power is switched on (like some TV sets). Sending cmnd/sonoff/irsend { "protocol": "NEC", "bits": 32, "data": 551502015 } for examle turns LG TV volume up. The codes can be found in the lirc-project. In my case the LG remote controller http://lirc.sourceforge.net/remotes/lg/AKB72915207 with a 32 bit code pre_data 0x20DF and KEY_VOLUMEUP 0x40BF -> 0x20DF40BF -> 551502015.

Greetings,

    Heiko.


altelch:
Here the HW description: https://github.com/altelch/SonoffIR
Would reference this in the Wiki if code makes into your branch.

Greetings,

      Heiko.

arendst:
Code will get into branch but I have to check the consequences first.

altelch:
Great, thanks for addition.

Greetings,   Heiko.

PR #75, opened 2017-02-16
Add optional sensor to GPIO4 for Sonoff Dual

danielztolnai:
GPIO4 is not used on the Dual board, so it is possible to use it for a sensor. Some soldering skills required.


arendst:
Will be implemented in version 3.9.17

anthonykeane:
Likewise GPIO2 and GPIO5 are also not used on the Dual board, so it is possible to use it for a sensor or even better switch. Some soldering skills required.

sonoff_dual
 --

anthonykeane:
Likewise GPIO2 and GPIO5 are also not used on the Dual board, so it is possible to use it for a sensor or even better switch. Some soldering skills required.

sonoff_dual

PR #89, opened 2017-02-20
HUE fixes and HSB support

altelch:
Hi,

fixed an error in JSON argument evaluation what caused wrong setting changes. Implemented full HSB support for HUE api calls (colormode hs).

Greetings,

   Heiko.


altelch:
Bwah, now travis-ci uses another IRremoteESP8266 than what is included in Arduino!? Doesn't matter for the HUE part.

altelch:
Ok, changed platformio.ini:
lib_deps = PubSubClient@~2.6, NeoPixelBus@~2.2.6, https://github.com/sebastienwa
rin/IRremoteESP8266.git, ArduinoJSON@~5.8.3

Now the correct versions are used and the built works again.

arendst:
First looked at your the PlatformIO library suggestion. If I try it locally it fails as it needs git installed to download and install IRremoteESP8266.git so I decided to add all libraries to the repository in the lib folder.

Now platformIO works great both locally and on Travis using the supplied libs.

I still need to look at your Hue changes but have to find some time for it...

arendst:
Implemented in version 3.9.20 just released

PR #102, opened 2017-02-25
initial support for wion devices

davidelang:
This adds the initial support for the wion series of devices.

created by @arendst , tested by @davidelang

Some of these devices have power monitoring as well that we do not yet have setup. This just enables the basic switching functionality.


arendst:
Implemented in version 3.9.22 just released

PR #108, opened 2017-02-26
fix bug pwm.

AlexTransit:
pwm1 pin <> pwm0 pin+1


arendst:
I think you miss the pin array **INDEXED** by GPIO_PWM0 and GPIO_PWM1.

As GPIO_PWM0 = 33 and GPIO_PWM1 = 34 (sonoff_template.h) and pin[GPIO_PWM0] is GPIO12 and pin[GPIO_PWM1] = GPIO14 as in Sonoff Led it works perfectly well.

PR #109, opened 2017-02-26
fix. use relay.

AlexTransit:
init and use relay if pin assigned.


arendst:
Did it my way. If you want more relays than are supported by the standard Sonoff defines in user_template.h just use the User Test define and add your relays there. It now updates Maxdevices by counting the relays.

AlexTransit:
i connect external led driver. (use gpio12 & gpio14)
change model to sonoff led. (relay use to powered driver)
device worked not correctly if power on
change settings can frozen or reboot device. 
i change pwm control, frozen and restart stopped.
i don't know that. 
may be overflow.
I decided to write to you.
unchanged module (sonoff led) is not working properly

Issue #96, opened 2017-02-22
HTU21 


Greg2002:
ver.3.9.20 20170221
I don't know why but HTU21 not works with I2C GPIO4,14 (BME280 and BH1750 - tested OK).
(HTU21 sensor tested with Arduino 3.3V Sparkfun Library and works fine).

altelch:
Whats the result of command I2Cscan? Is the sensor found?

arendst:
My HTU21 works fine. Sometimes it needs a power cycle to reset itself.

Greg2002:
I have a Si7021 sensor, not HTU21. With your library it will not work. I have an universal library from GitHub and it works fine with it. Thank you.

altelch:
Might be related to the device id. I check for 0x32 as this is the device id returned by the htu21. The Si7021 seems to return 0x15. Please try to change this in the HTU21 driver and test again.

altelch:
Added code in my fork. Please try: https://github.com/altelch/Sonoff-Tasmota.git
If it works I'll do a pull request.

Greg2002:
It works but give wrong data of Humidity. HTU Temperature: 
20.4°C
HTU Humidity: 
-0.7%
BMP Temperature: 
21.2°C
BMP Humidity: 
30.6%
BMP Pressure: 
745.4 mmHg
Illuminance: 
102 lx
This is due to wrong delay (HTU21 is faster). You can check a working library at https://github.com/enjoyneering/HTU21D

altelch:
Changed delay to 23ms. Please try again with my fork. If it works I'll see how to "fix" this is a better way.

Greg2002:
HTU Temperature: 
21.2°C
HTU Humidity: 
28.2%
BMP Temperature: 
21.5°C
BMP Humidity: 
27.2%
BMP Pressure: 
752.0 mmHg
Illuminance: 
105 lx
OFF

Greg2002:
Now it works

altelch:
Hi, could you please test again with my current code? Correct name should now be returned in logfile and the conversion times are set per sensor type.

Greg2002:
{"Time":"2017-02-27T15:28:30", "SI7021":{"Temperature":22.0, "Humidity":27.3}, "BME280":{"Temperature":21.4, "Humidity":27.4, "Pressure":755.0}, "BH1750":{"Illuminance":10}}

Greg2002:
But web server returns HTU
[...]

PR #118, opened 2017-02-27
Support for SI70xx in HTU21 driver

altelch:
Hi, the code now supports the 3 SI70xx types, too. Mainly setting convertion times and IDs to the correct values and the code works. See #96


davidelang:
a workflow note (general git advice, up to @arendst if he wants to enforce it 
for this project), there are a lot of merge and merge cleanup commits here. It 
makes for a cleaner history (and easier for everyone to review and understand) 
if you don't do develop-merge-develop-merge...

Instead, create one branch that you develop in, keep a pristine upstream/master 
branch that contains only what's published, and create a throwaway branch to 
test merges. After you test the merge, throw it away and create it again as 
needed (git has tools to remember how you solved merge conflicts so you don';t 
have to keep re-doing the fixes)

you can rebase your work on top of the current upstream/master so that you can 
take advantage of changes that have taken place upstream.

When you have an ugly history like this, you can do an interactive rebase 
(rebase -i <first commit>) and that will let you re-order your patches and 
squash/split them as needed.

For example all the merge/merge cleanups that are adjacent in the history should 
be able to be trivially squashed into one entry.

As you develop, you find things you missed, do work to fix things that you coudl 
have done better, etc. In an ideal git world, before you submit you would 
re-write the history (by re-ordering/sqashing/splitting changes) to make it look 
like you had perfect forsight rather than hindsight. That way people can review 
the patches and don't get confused with some of the dead ends that you hit.

As I say, it's not my call to require this sort of cleanup, but I think it would 
help. It looks like this commit lists a bunch of things that were done upstream 
that have no relation with your work.

David Lang
k

PR #124, opened 2017-03-01
Improve MQTT_MAX_PACKET_SIZE handling

davidelang:
Calculate MESSZ instead of hard-coding it, make failures due to too small MQTT_MAX_PACKET_SIZE be compile-time

the prior code checked for MQTT_MAX_PACKET_SIZE being too small at runtime, but that can result in an exception crash loop.

Instead, calculate MESSZ based on max packet size - headroom and then do a compile-time check so that problems are discovered before devices are flashed.

A block of includes needed to be moved higher in the file so that MQTT_MAX_PACKET_SIZE was available to do the calculation

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -4,7 +4,7 @@
  * ====================================================
  * Prerequisites:
  *   - Change libraries/PubSubClient/src/PubSubClient.h
- *       #define MQTT_MAX_PACKET_SIZE 400
+ *       #define MQTT_MAX_PACKET_SIZE 1400
  *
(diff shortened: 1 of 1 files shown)


arendst:
Thnx for the idea behind this test. I'll implement it the other way round as for some reason bigger MESSZ also increases the code and I want to be independent from a too large MQTT_MAX_PACKET_SIZE.

I can't find the cause why the code increases but I decided to do it the other way round where I increased MESSZ to 352 and test if MQTT_MAX_PACKET_SIZE is at least to 512.

Test with both IDE and platformIO and seems to work fine.

PR #196, opened 2017-03-13
Added host:port to MQTT connection failure message and changed to INFO

khcnz:


khcnz:
I messed up this PR with your 4.0.4 merge - will replace with a better one

PR #206, opened 2017-03-15
Added sonoff dev board and enabled ADC for electrodragon

khcnz:
Tested with a Sonoff Dev https://www.itead.cc/sonoff-dev.html


arendst:
Will be implemented in next release

PR #211, opened 2017-03-16
GPIO PWM

khcnz:
Hi Theo - would this be of use?

Added ability to set GPIO as PWM1-4. Added command to set PWM value by command. PWM values restored upon restart

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -191,6 +191,7 @@ struct SYSCFG {
 
   char          ntp_server[3][33];
   uint16_t      pulsetime[MAX_PULSETIMERS];
+  uint16_t      pwmvalue[4];
 
 } sysCfg;
 
--- a/sonoff/settings.ino
+++ b/sonoff/settings.ino
@@ -511,6 +511,8 @@ void CFG_DefaultSet2()
 
   CFG_DefaultSet_4_0_4();
   sysCfg.pulsetime[0] = APP_PULSETIME;
+  
+  CFG_DefaultSet_4_0_6();
 }
 
 void CFG_DefaultSet_3_2_4()
@@ -557,6 +559,11 @@ void CFG_DefaultSet_4_0_4()
       if (sysCfg.ntp_server[j][i] == ',') sysCfg.ntp_server[j][i] = '.';
   sysCfg.pulsetime[0] = sysCfg.ex_pulsetime;
   for (byte i = 1; i < MAX_PULSETIMERS; i++) sysCfg.pulsetime[i] = 0;
+}
+
+void CFG_DefaultSet_4_0_6()
+{
+  for (byte i = 1; i < 4; i++) sysCfg.pwmvalue[i] = 0;
 }
 
 void CFG_Default()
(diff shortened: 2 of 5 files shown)


arendst:
Did you test it's functionality? If a led string was connected to the PWM port did it blink when you power ON/Off the leds?

khcnz:
Tested by configuring LED1 as a PWM output manually in the template file. You can alter the value and clearly see the LED brightness change. I need to dig out my multi-meter..

I don't have a SNLED though so can't i didn't break something in xdrv_snfled

arendst:
I ask because I have problems with the Sonoff Led. As PWM is highly interrupt driven it's output is halted when I write updates to flash which need interrupts disabled (every power on/off is two updates and therefore flickers Sonoff Led). As a work around I do not disable interrupts for Sonoff Led with possible flash corruption that may be the cause of some issues already raised regarding Sonoff Led. 

Please test thoroughly with real LED load and see if flash updates are no problem.

khcnz:
Hrrm ok I will do some more research then - have you looked into the ICACHE_RAM_ATTR? would that help speed up the section of critical code in the flash save area? Also the PWM frequency could possibly be reduced (whats the minimum a human can see?) I will do some more research this weekend.

davidelang:
On Thu, 16 Mar 2017, khcnz wrote:

> Hrrm ok I will do some more research then - have you looked into the 
> ICACHE_RAM_ATTR? would that help speed up the section of critical code in the 
> flash save area?

It's not a matter of inefficient code, it's that writing to flash is inherently 
slow (it takes time to erase a block of flash and then write to it)

khcnz:
@davidelang sure but if you are reading from (for program memory) and writing to flash at the same time obviously it will be slower than if program is fully in ram. Haven't said that i haven't mapped out any of the timings yet so it could be a red herring.

There is an interesting and (I think) related thread here https://github.com/esp8266/Arduino/issues/836 where moving to NMI interrupts helps/solves the problem along with a couple of other improvements and running at 160MHz  https://github.com/esp8266/Arduino/pull/2299

khcnz:
@arendst tested now that i understand better what you mean - yes the LED (even LED1) flickers on switch on/off of power to main relay when writing to flash.

In line with the research aboveI updated my Arduino library to master which contains the fixes above and the problem is completely resolved even at very very low PWM levels (like 1 or 2). I would strongly suspect you could remove your workaround for the sonoff LED if you compile to master of Arduino - when is the next build of the underlying Arduino library expected, it seems to have been a long time since last release?

khcnz:
I am not 100% sure of the legitimacy (from a compiler point of view) of doing this but the fixes are fully self contained so I just put core_esp8266_wiring_pwm.c from master into the sonoff directory ( and using the standard 2.3.0 of Arduino library in the ide)

No flashing - appears to be completely resolved.

arendst:
I'll give it a try now. Looks promising!

arendst:
I'm testing with the master core_esp8266_wiring_pwm.c and have to agree that when dimmer is set to 100 it doesn't flicker when power on/off/on is performed. As soon as dimmer is lower than 90 the flickering starts to appear due to flash writes of power on/off/on. Also changed ANALOG_WRITE_FREQ in xdrv_snfled.ino from 432 to 1000 but still there are noticeable flashes when flash is accessed.

arendst:
Another test with the complete master version is very promissing! No flickering at all!

Now need to find out what other than core_esp8266_wiring_pwn.c has changed...

arendst:
Test3: using your suggestion of released Arduino-ESP8266 and master file core_esp8266_wiring_pwm.c in sonoff directory seems to work well with regards to flash updates but now for some reason to me yet unknown the power on/off doesn't seem to work. Investigating...

arendst:
Got It!!!! Apart from core_esp8266_wiring_pwm.c (issue 2299) you also need core_esp8266_timer.c and core_esp8266_wiring_digital.c (issue 2328) from the current master as of today, March 17, 2017.

I'll add them to the sonoff directory as from the other PR's noted I see no direct relation to these three files.

arendst:
Will be released in next version

khcnz:
Implemented :)

PR #219, opened 2017-03-17
Move BE_MINIMAL to a separate user_config_override.h file

davidelang:
also changes user_config_override.h to be a symlink

Change the user_config_override.h symlink to point at this file to create
the minimal build instead of having BE_MINIMAL as an ifdef. This should
make it less likely that people will tinker with it.

Ideally the include would check for the file existing instead of doing an unconditional include (eliminating the need for a symlink under the default conditions), but I don't know the C preprocessor well enough to figure out how to do that.


arendst:
symlink won't work on windows.

davidelang:
true, I forgot about that

I did some digging and it looks like there is no easy way to get the arduino 
IDE compiler chain to test if an include exists before trying to access it (or 
to have it not die if the include doesn't exist), so the best we can do is to 
have multiple files and copy them :-(

I was hoping to find a way to have a local config_override file that could 
contain my local stuff (network defs, etc) that wouldn't always conflict with 
the upstream version on a pull. It looks as if the best I can do is to have a 
local copy and copy it to overwrite the one that is included, then try to 
replace it with the upstream one before doing a push :-(

I guess I could do this all in a different branch, and keep the modified file as 
a local update to that branch.

any thoughts?

wiktorschmidt:
when using platformio it might be easiest to configure multiple envs here:
https://github.com/arendst/Sonoff-Tasmota/blob/master/platformio.ini

davidelang:
thoughts on this approach? should I update it to work with the new release?

davidelang:
@wiktorschmidt I expect that the command-line version would be much easier, I can even use makefiles :-)

but this needs to be able to work with the arduino IDE as long as it's being supported.

ascillato:
Hi,

Is this PR still valid?

PR #231, opened 2017-03-19
Fixups

davidelang:
pfirst a cleanup of trailing whitespace, then a cleanup of ifdef protections in cases where all items related to a topic are commented out, not just the top level one.

### DIFF
--- a/sonoff/_releasenotes.ino
+++ b/sonoff/_releasenotes.ino
@@ -67,7 +67,7 @@
  * 3.9.20 20170221
  * Add minimal basic authentication to Web Admin mode (#87)
  * Fix Hue and add HSB support (#89)
- * 
+ *
  * 3.9.19 20170219
  * Sonoff Led: Made GPIO04, 05 and 15 available for user
  * Sonoff Led: Add commands Fade, Speed, WakupDuration, Wakeup and LedTable
@@ -77,7 +77,7 @@
  * Fix Sonoff Led dimmer range (#16)
  * Change Sonoff Led command Dimmer to act on both cold and warm color
  * Add Sonoff Led command Color CCWW where CCWW are hexadecimal values fro 00 - FF
- * Reduce Sonoff Led flickering by disabling interrupts during flash save and disabling 
+ * Reduce Sonoff Led flickering by disabling interrupts during flash save and disabling
  *   Led during OTA upgrade and Web upload (#16)
  *
  * 3.9.17 20170217
@@ -148,7 +148,7 @@
  *
  * 3.9.3 20170127
  * Add confirmation before Restart via webpage
- * Expand Domoticz Configuration webpage with Key, Switch and Sensor Index and 
+ * Expand Domoticz Configuration webpage with Key, Switch and Sensor Index and
  *   add commands DomoticzSwitchIdx and DomoticzSensorIdx (#86) (#174) (#219)
  * Fix default DHT11 sensor driver selection
  * Fix LedPower status after button press (#279)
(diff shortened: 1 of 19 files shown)


davidelang:
I also started looking at the MQTT_USE option, but it would be extremely intrusive.

Does it make sense to pursue the ability to not use mqtt at all? While there is the theoretical web-only use, I think the only practical reason to have it is the be-minimal use to get just enough running to bootstrap an OTA update, and I'm wondering if that would be best handled by a complete fork that would have everything except wifi and enough of the webserver to admin it ripped out completely (no support for buttons/relays/etc at all)

arendst:
What tool do you use to get rid of trailing spaces?

davidelang:
I'm running on Linux and am an old unix sysadmin, so I use the command line 
tools available (they should be available as well on OS/X)

sed -i s/" $"//

then

grep -c " $" * |grep -v :0

to find any that still exist, repeat until the grep comes back clean

I also found that something I was doing was adding blank lines to the end of 
files and had to clean up a lot of those in one of my earlier patches.


not knowing how familiar you are with these tools, details below.

in these cases, the $ matches the end of line, so it's looking for anyplace 
there is a space immediately before the end of the line.

the sed replaces it with nothing (I could have gotten fancier and make it 
replace one or more spaces, but complex regular experessions will bite you, so 
for something like this it was easy enough to do up-arrow, enter a bunch of 
times :-)

in the grep, the -c tells it to count the lines rather than return matching 
lines, and then the second grep strips out any filenames that have no matches in 
them

arendst:
Thnx. Still working on Windows but have knowledge of Linux as I'm running all my servers on Debian. Will look into the spaces 'someday'.

davidelang:
any thoughts on the other parts of this series?

arendst:
I do not see why you want to use defines in default settings. That way the defaults for webserver will never be set once enabled.

I don't get the BE_MINIMAL rework either. Goal is to have a small version knowing the flash layout just to be used as first step in upgrading/uploading a new too large version. Doing it my way I do not have to manage different sources.

davidelang:
On Sun, 19 Mar 2017, Theo Arends wrote:

> I do not see why you want to use defines in default settings. That way the defaults for webserver will never be set once enabled.

enabling the webserver will be a new compile, that will include the defaults at 
that time.

I had a bit of trouble when I first started using this, I didn't care about the 
dom* server, so I just commented out all the dom* ifdef options. That caused 
build failures because the code assumes that the various parameters were always 
defined. It's not obvious that it is an error to comment out the other unneeded 
ifdefs.

> I don't get the BE_MINIMAL rework either. Goal is to have a small version knowing the flash layout just to be used as first step in upgrading/uploading a new too large version. Doing it my way I do not have to manage different sources.

be_minimal is an 'attractive nusense', it looks like something that is useful, 
but isn't appropriate for anyone but experts to use. I was trying to move it out 
of the way so that it's not visible easily, and anyone who does look at it will 
see that it's disabling everything.

It's bad to have people committing their passwords in user_config.h, so we 
should be encouraging people to use the user_config_override.h for their 
settings (or possibly we should change user_config.h to config_defaults.h and 
user user_config.h the way override is today??)

So I was looking at combining these into one effort.

be-minimal went away, to be replaced by a user_override that makes it obvious 
that everything is being disabled.

a -sample can be used to have everything work as-is

users can put in the appropriate ones for their system.

If we were able to use a symlink, or a makefile that copied a default into place 
to make an image (but didn't overwrite the existing file if it's been 
customized) would be an ideal situation



Any thoughts about my questions on if MQTT really should be able to be compiled 
out while leaving buttons functional (and the webserver possible) vs just 
leaving enough to do an OTA upgrade?

arendst:
Leave both mqtt ota and webserver upgrade please. Other than that can be compilled out.

khcnz:
> be_minimal is an 'attractive nusense', it looks like something that is useful,
> but isn't appropriate for anyone but experts to use. 

Yes it is attractively named... sounds like it complies a smaller/faster version - i fell into the trap when i first looked at the source - an alternative could be to change the name; perhaps "CONFIG_UPGRADE_ONLY_BUILD" or something like that.

I know you can't protect against every conceivable misreading of every #define but this one does stick out a little

ascillato:
Hi,

Is this PR still valid?

PR #232, opened 2017-03-19
make webserver port configurable

davidelang:
There was a request to make the webserver port a config option, it turned out to be trivial to do.


arendst:
Will include in next release

davidelang:
impelemented separately

Issue #227, opened 2017-03-18
MQTT Query for external switches?


dc75:
I'm currently running the latest version of Sonoff-Tasmota on a NodeMCU board with 2 external switches attached to GPIO4 & GPIO5.  The switches were configured through the web interface as Switch1 & Switch2.  Everything works fine & I receive an MQTT message when either switch changes state.  In my application the switches don't change state very often, so I was trying to find a way to query the state of each switch periodically via MQTT, just to make sure that I did not miss an MQTT message & end up with the software out of sync with the physical state of the switch.  I see how this can be done for GPIO assigned as a relay with the `cmnd/myesp/power1 status` command, but I can't seem to figure out how to accomplish this for GPIO assigned as a switch.

Perhaps I have overlooked something simple?  Any ideas on how to achieve this would be greatly appreciated!  Thanks!

ionciubotaru:
Look at: SwitchRetain and ButtonRetain in the [wiki](https://github.com/arendst/Sonoff-Tasmota/wiki/Commands)

dc75:
Thanks for the suggestion @ionciubotaru!  I've looked at both SwitchRetain & ButtonRetain, but I'm not sure that either of those would work for what I'm trying to achieve; I'm very new to this software & ESPs, so please feel free to correct me if I'm wrong!  

I'm using Node Red & I already retain the last switch state there, so it seems like SwitchRetain/ButtonRetain would be kind of redundant for my application.  What I'd really like to be able to do is to periodically query the GPIO that my switches are connected to.  My switches normally only change state a few times during a 24 hour period; since I only receive an MQTT message on a switch changing state, it can be many hours since I last received the state of the switch.  In this situation, I'd feel much more confident if I could periodically get a confirmation on the state of the switch.   :smiley:

ionciubotaru:
I understand your needs, but Tasmota was designed to be used with Sonoffs where the switch is always open, except the short moment when you press it. 
I suggest change the code and subscribe to a different additional topic and handle the received messages.
Tasmota is based on [knolleary MQTT library ](http://pubsubclient.knolleary.net/).
Take a look at [API Documentation](http://pubsubclient.knolleary.net/api.html)

davidelang:
take a look at switchmode, specifically follow and inverted follow

dc75:
[...]

PR #233, opened 2017-03-19
add status line to show button/switch status

davidelang:
Add a status line to show the state of buttons and switches (pretty meaningless with momentary switches, but with switchmode follow it's useful)

compile tested, but I have no devices with switches to function test it on.

see issue #227


davidelang:
implemented in a different way

PR #237, opened 2017-03-20
Add user.ino for user's function

ionciubotaru:


davidelang:
I'm pretty sure these functions won't do anything unless you put calls to them in sonoff.ino

PR #250, opened 2017-03-21
Resolved unhandled exception when posting commands to web console containing %s

khcnz:
Resolved unhandled exception when posting any command with a %s character to the web console.
Encoded web commands to allow for special chars (especially % when setting hostname or MQTTclient).
Enabled support for custom wifihost names with dynamic parameters.
Changed default wifi hostname to include last 6 of mac address to match mqtt default and allow simplification of formatting code.
Moved Wifi hostname to user_config.
Simplified and removed formatting code.

### DIFF
--- a/sonoff/settings.ino
+++ b/sonoff/settings.ino
@@ -702,13 +702,12 @@ void CFG_Delta()
       sysCfg.blinkcount = APP_BLINKCOUNT;
     }
     if (sysCfg.version < 0x03011000) {  // 3.1.16 - Add parameter
-      getClient(sysCfg.friendlyname[0], sysCfg.mqtt_client, sizeof(sysCfg.friendlyname[0]));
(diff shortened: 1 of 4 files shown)


arendst:
Great! I've been here half a year ago and decided to do it my way as you may have noticed.

Just a hint: Try to set the hostname to %s-%s and enjoy the exceptions...

khcnz:
OK well of course they are all just suggestions and i thought it could help with reducing & simplifying code - and these were issues i had run up against.

Agreed this could lead to other exceptions but at the time it wasn't worse than the old behaviour.

khcnz:
I *think* you also still need the url encoding in the javascript? i was having trouble with the characters sometimes getting truncated - i think the esp library must have some kind of state since it didn't always truncate. Also chrome developer tools shows the request is invalid if you have any non alphanumber char such as %

arendst:
If you mean the encodeUri and PSTR("%s") in webserver.ino then those were added in the latest release from yesterday and work indeed fine.

khcnz:
Yes sorry for some reason i thought you hadn't included the encode uri but you have i just missed it

PR #119, opened 2017-02-27
Added nodeMcu support, added IR control Toshiba HVAC

SirAd:
Hi,
I've added:
* nodeMCU hardware support
* extension for controlling Toshiba HVAC via IR.

To send IR to HVAC use command IRhvac with parameters (JSON)
IRhvac  {"HVAC_Mode": "<MODE>" , "HVAC_FanMode" : "<FAN_MODE>" , "HVAC_Temp" : <TEMP> , OFF : <0|1>}

Where 
- MODE is  HVAC_HOT, HVAC_COLD, HVAC_DRY, HVAC_AUTO (default)
- FAN_MODE is FAN_SPEED_1, FAN_SPEED_2, FAN_SPEED_3, FAN_SPEED_4, FAN_SPEED_5, FAN_SPEED_AUTO (default), FAN_SPEED_SILENT
- TEMP is temperature from 17 to 30 Celsius degrees
- OFF 1- turn off , 0-turn on (default)  
all parameters are optional, eg.:
IRhvac  {"HVAC_Mode": "HVAC_COLD" , "HVAC_FanMode" : "FAN_SPEED_AUTO" , "HVAC_Temp" : 19 , OFF : 0}
IRhvac  {"HVAC_Mode": "HVAC_HOT" , "HVAC_FanMode" : "FAN_SPEED_AUTO" , "HVAC_Temp" : 24 }
IRhvac  {"OFF" : 1 }

PR #257, opened 2017-03-21
Hvac support Mitsubishi and Toshiba added

SirAd:
Added support for Air Condition from Mitsubishi and Toshiba - extended from pull request #119


SirAd:
o send IR to HVAC use command IRhvac with parameters (JSON)
IRhvac {"HVAC_Mode": "<MODE>" , "HVAC_FanMode" : "<FAN_MODE>" , "HVAC_Temp" : <TEMP> , OFF : <0|1>, HVAC_Vendor: <Toshiba|Mitsubishi>}

Where

MODE is HVAC_HOT, HVAC_COLD, HVAC_DRY, HVAC_AUTO (default)
FAN_MODE is FAN_SPEED_1, FAN_SPEED_2, FAN_SPEED_3, FAN_SPEED_4, FAN_SPEED_5, FAN_SPEED_AUTO (default), FAN_SPEED_SILENT
TEMP is temperature from 17 to 30 Celsius degrees
OFF 1- turn off , 0-turn on (default)
HVAC_Vendor - optional  must be Toshiba (default) or Mitsubishi
all parameters are optional, eg.:
IRhvac {"HVAC_Mode": "HVAC_COLD" , "HVAC_FanMode" : "FAN_SPEED_AUTO" , "HVAC_Temp" : 19 , OFF : 0}
IRhvac {"HVAC_Mode": "HVAC_HOT" , "HVAC_FanMode" : "FAN_SPEED_AUTO" , "HVAC_Temp" : 24 , "HVAC_Vendor" : "Mitsubishi" }
IRhvac {"OFF" : 1 }

mihalski:
Toshiba works but Mitsubishi does not seem to. Going to battery power to get it closer to the unit to make sure it's not an IR signal issue.

Also, interestingly, my Mitsubishi remote goes down to 16 degrees.

mihalski:
I went as simple as IRhvac {"HVAC_Vendor" : "Mitsubishi", "OFF" : 1 } and got nothing at any distance from the AC unit. It is generating a signal, just not the right one.

I wonder if I could try capturing some IR signals for comparison :/

mihalski:
I'm TRYING to use a Vishay TSOP34838 infrared receiver with the sketch https://github.com/mdhiggins/ESP8266-HTTP-IR-Blaster but it seems to be picking up a LOT of random IR codes from the air (as well as legit, repeatable ones that I generate).

This happens when I power it at 5V. If I try 3.3V I get nothing.

I'll see if I can fish the codes I am after out of there.. or somehow otherwise quell the flood of false IR codes.

Hogesyx:
Mitsubishi HVAC IR has always been a pain, especially if you need to debug your sender code.
https://www.analysir.com/blog/2015/01/06/reverse-engineering-mitsubishi-ac-infrared-protocol/

mihalski:
It should work as this code is using a library that does work.

https://github.com/markszabo/IRremoteESP8266/blob/master/examples/TurnOnMitsubishiAC/TurnOnMitsubishiAC.ino

This example works and can be tweaked for different results.

SirAd:
Hi,
I've just added some debug. Just in case. Unfortunately I can't help because I do not have Mitsubishi AC unit. My Toshiba is working fine, but IR emitter must be put close to unit (I think is 3.3V power issue). Maybe it's the same problem with Mitsubishi?

mihalski:
Thanks.
I'm trying to work out how to capture and decode IR signals but running into some problems.

I have no resistors and am running at 5V for maximum reach (still seems to be in spec when measured with a multimeter).

Which Toshiba do you have? Does yours have additional controls for the vane that aren't exposed to us?

mihalski:
Amazing! It just worked...
```
20:10:55 CMND: IRhvac {"HVAC_Mode": "HVAC_COLD" , "HVAC_FanMode" : "FAN_SPEED_AUTO" , "HVAC_Temp" : 22 , "HVAC_Vendor" : "Mitsubishi" }
20:10:55 IRHVAC: Sent to Mitsubishi. Power: 1,  Mode: 24, Temp: 22C, Fan Speed: 0, Vane Mode: 7
20:10:55 MQTT: stat/master_bedroom/RESULT = {"IRHVAC":"Done"}
```
And the aircon came on. What the hell? Your code hasn't changed apart from the debugging. Could it have affected timing somehow?

Also:
```
20:13:08 CMND: IRhvac {"HVAC_Vendor" : "Mitsubishi", "OFF" : 1 }
20:13:08 IRHVAC: Sent to Mitsubishi. Power: 0,  Mode: 8, Temp: 16C, Fan Speed: 0, Vane Mode: 7
20:13:27 MQTT: stat/master_bedroom/RESULT = {"IRHVAC":"Done"}
```
Where are those settings coming from? Are you padding with values outside of the supported range? (though 16C should be a possible value, as it is on my remote).

SirAd:
> And the aircon came on. What the hell? Your code hasn't changed apart from the debugging. Could it have affected timing somehow?

No, just added debug. Maybe you did mistake with building previous version?

> Where are those settings coming from? Are you padding with values outside of the supported range? (though 16C should be a possible value, as it is on my remote).

If you miss some parameter I put default values (e.g. FAN_SPEED_AUTO is default fan option.
Temperature set must be between MITSUBISHI_AC_MIN_TEMP and MITSUBISHI_AC_MAX_TEMP defined in IRMitsubishiAC library. Missing value means 0 what is converted to MITSUBISHI_AC_MIN_TEMP = 16

mihalski:
I agree that it doesn't make sense, but if I had made a mistake with compiling the previous version then it shouldn't have worked for the Toshiba AC either.

I'll try the previous version again tonight after work to see if I still have the same problem. If I do then the only possible explanation would have to be something to do with the extra debugs changing the timing of the codes. It WAS sending something even when it wasn't working.

arendst:
Will be implemented in next release

arendst:
Will implement in next release

PR #258, opened 2017-03-22
Enable /RESULT to be customized via user_config.h

Hogesyx:
This MQTT subscribers that are only limited/hardcoded to a single topic to listen to both telemetry as well as result feedback since both are using similar JSON payload.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -462,7 +462,7 @@ void mqtt_publishPowerState(byte device)
 
   if ((device < 1) || (device > Maxdevice)) device = 1;
   snprintf_P(sdevice, sizeof(sdevice), PSTR("%d"), device);
-  snprintf_P(stopic, sizeof(stopic), PSTR("%s/%s/RESULT"), PUB_PREFIX, sysCfg.mqtt_topic);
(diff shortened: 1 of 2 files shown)


arendst:
This message is not telemetry but the result of a power on/off action and therefore is a RESULT response.

Hogesyx:
I know. Thus allowing it to be customized we can enable it to work like always using /RESULT or allow devices that can only monitor one topic to rename it to something else.

This might not be useful for you but it is extremely useful for people which need more customization for example people using home-assistant, as their default MQTT sensor module can only use one topic for both tele and trigger response.

It would really be nice if you approve this or allow us other method to rename topics via define.

arendst:
Some kind of implemented in next release. Use command MqttResponse to switch between RESULT or command (POWER)

Hogesyx:
Thanks for looking into it.

ascillato:
Hi,

Is this PR still valid?

Hogesyx:
Nope. This was meant for the old legacy version.

There was a feature that was added later to allow topic customization that
superseded this.

ascillato:
Ok,

Thanks a lot for your explanation. Very appreciated.

So, this can be closed?

Hogesyx:
Yes, thanks for the hard work doing spring cleaning.

PR #346, opened 2017-04-10
First step into Wiring an Adafruit Featherwing as build target

pimvanpelt:
Theo, you may recall I pinged you a while ago about adding a touch screen to Sonoff Touch (the device). I have made a little bit of progress in this, and wanted to see if you are OK with the direction.

Using #define USE_TOUCHSCREEN (set in [env:touch] in platformio.ini), I gate the inclusion of xdrv_touchscreen.ino and gate linking against the associated libraries (AdaFruit GFX/ILI9341/STMPE610). It creates an object 'TouchScreen' with a handle() method which is called in the main loop().

[env:sonoff] is unaffected by this change, compiles cleanly and runs on several of my Sonoff devices. I will be using [env:touch] going forward for this new target (which is also an ESP12e w/ 4M flash).

Currently, the mechanics of handle() do some simulated action, which will be replaced with functionality taken from sonoff.ino's internals. It registers touch (TOUCH_DOWN and TOUCH_UP) and currently only prints to serial, intended behavior is to wire this to MQTT message passing.

Could you let me know your thoughts on this, stylistically and any concerns you have? I'm not asking you to merge this for the time being, as I will need to get away from proof-of-concept mode and into a more productive UX. This adds 14.6K of code and 1.3K of memory.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -23,3 +23,14 @@ board_flash_mode = qio
 
 build_flags = -Wl,-Tesp8266.flash.1m64.ld -DMQTT_MAX_PACKET_SIZE=512
 lib_deps = PubSubClient, NeoPixelBus, IRremoteESP8266, ArduinoJSON
+
+[env:touch]
+platform = espressif8266
+framework = arduino
+board = esp12e
+upload_speed = 460800
+board_flash_mode = qio
+build_flags = -Wl,-Tesp8266.flash.4m1m.ld -DMQTT_MAX_PACKET_SIZE=512
+
+lib_deps = PubSubClient, NeoPixelBus, IRremoteESP8266, ArduinoJSON
+  Adafruit GFX Library, Adafruit ILI9341, Adafruit STMPE610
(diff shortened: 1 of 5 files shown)


pimvanpelt:
Wired in MQTT send/recv, RSSI, rtcTime and friendlyname[0].

pimvanpelt:
Wired in 'power' from sonoff.ino, drawing a green power icon if the relay is ON, darkgray if it is OFF.

arendst:
Looks promissing. If possible please also consider people using Arduino IDE instead of PlatformIO which involves some library management.

So in the end the touch panel will use the Sonoff Touch hardware except for the current touch switch? How do you connect the panel to the Sonoff Touch hardware?
Or do you plan to use a totally different esp8266 device?

pimvanpelt:
Closed - I have since moved to another micro controller OS (mongoose-os.com) and implemented the functionality (of Sonoff switches and the touch screen). Thanks for the consideration though and good luck with Tasmota!

PR #390, opened 2017-04-26
Make WS2182 control independent of relay

joecotton:
Allow 'POWER ON" and "POWER OFF" commands to work without turning on or off the WS2812 strip.


arendst:
PR rewritten and now supports all modules as an extra "virtual" relay. Will be released in next version (5.1.1).

homeseer666:
Is this feature enabled by default or do I need to enable it somewhere? I seem to have a similar issue where switching a physical relay on/off (HW622 board) affects a WS2812 on GPIO3.
Thanks.

PR #459, opened 2017-05-17
Implementation of counter sensors

maartendamen:
Hi Theo,

I have implemented code for the use of counter sensors. I am going to use this to measure the flow of my water meter, which is attached to my waterpump. It can also be used for a variety of other use cases, such as a electricity meter.

I hope you like it.

Regards,
Maarten

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -196,6 +196,8 @@ struct SYSCFG {
   // 5.0.4a
   char          mqtt_fulltopic[101];
 
+  unsigned long counters[4];
+
 } sysCfg;
 
 struct RTCMEM {
--- a/sonoff/settings.ino
+++ b/sonoff/settings.ino
@@ -452,6 +452,12 @@ void CFG_DefaultSet2()
 
   // 5.0.6
   sysCfg.mqtt_retry = MQTT_RETRY_SECS;
+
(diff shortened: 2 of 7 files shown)


arendst:
Nice implementation! I'll integrate soon but could take some weeks as I'm OoO in the near future ;-)

maartendamen:
Great, take your time.. I'll use my own binary in the meantime :-)

arendst:
How often do you expect these counters to increment? As you configured them in the sysCfg area it would mean that every counter update will also update/write to flash. If more than once in 10 seconds they have to be moved to RAM as it wears flash too much.

maartendamen:
In the case of my water flow meter this could be 460 pulses per 1 liter so quite a lot of changes. 
My thought was that it only will be saved when CFG_Save was been called. Am I wrong?

arendst:
Currently CFG_Save is called every second and if the hash value over the settings is changed between calls it will save to flash.

I propose to use the RTC area which saves the counters during a reboot and save the counters once a day in flash.

arendst:
Sooner than expected. Give it a go. See Wiki > Commands too

maartendamen:
Nice and quick one! I will check this implementation tonight.

PR #464, opened 2017-05-20
Changed HUE Device type to color supporting version

altelch:
Changed HUE Device Type to a color supporting version, the old type did just support brightness control. Had to add xy output to prevent Echo from handling the device correctly and then complain about disconnected bridge. The current Echo Firmware supports changing the color via voice control and now hopefully works with the emulated hue (not supported in germany yet).


arendst:
Will chck next week. Now visiting Naples

arendst:
Implemented so I guess this one can be closed?

altelch:
Yes, this can be closed. Thank you again,


     Heiko.


Am 11.06.2017 um 15:11 schrieb Theo Arends:

PR #497, opened 2017-06-03
added support for AM2320-Sensor

Staars:
Because it also uses I2C I added it to xsns_bmp.ino.
Code changes are rather small and I think it fits well to that section.
The substantial work was done by https://github.com/thakshak/AM2320 and I only had to refactor it a bit.

The first measurement will probably always give an error and I still think there is a problem with the wake function after some time (teleperiod). But the webserver and "cmnd/sonoff/status 10" works as expected.

### DIFF
--- a/sonoff/xsns_bmp.ino
+++ b/sonoff/xsns_bmp.ino
@@ -1,35 +1,44 @@
 /*
-  xsns_bmp.ino - BMP pressure, temperature and humidity sensor support for Sonoff-Tasmota
+ Copyright (c) 2017 Heiko Krupp and Theo Arends.  All rights reserved.
 
-  Copyright (C) 2017  Heiko Krupp and Theo Arends
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions are met:
 
-  This program is free software: you can redistribute it and/or modify
-  it under the terms of the GNU General Public License as published by
-  the Free Software Foundation, either version 3 of the License, or
-  (at your option) any later version.
+ - Redistributions of source code must retain the above copyright notice,
+   this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


Staars:
After adding the delay it works now on my test system without glitches.

arendst:
As the AM2320 is equivalent to the DHT22 I will not implement this code as it is a waste of precious code space. The AM2320 can be connected as either I2C device or as single wire as the DHT22 does. Configure your AM2320 as a DHT22 and you're good to go.

Regarding usability of the AM2320 I linked to this report before https://www.kandrsmith.org/RJS/Misc/Hygrometers/calib_many.html

Concluding:
AM2320 / AM2321
The devices I have tested appear similar to the DHT22 in overall performance, simply adding the convenience of I2C. Both my AM2320s exhibit the same intermittent fault of failing to return any data when humidity is in the range 50-70% RH. Oddly they appear to work reliably above and below that range. I never got to the bottom of whether this was something wrong in my set-up or an intrinsic sensor fault. Currently I am attributing it to bad luck in my getting a couple of faulty devices rather than a systematic fault with this model. Similarly the AM2321 intermittently fails to return a value above about 90% but works reliably at lower humidities. The AM2320 tested here shows quite a strong hysteresis though it is difficult to judge because of all the missing data points for NH4NO3 (60%).

In summary, they appeared similar to the DHT22 most of the time, but I had so many different intermittent problems I do not intend to use these devices in future experiments.

Staars:
Okay, no problem. Interesting article, BTW.

I know that the AM2320 supports 1-wire-protocol but IMHO it is a little bit arkward to set on power up and somehow for me it never worked in TASMOTA. (to be honest, I have to check this again ;)

Anyway, I understand your reason. 

Best regards,
Christian

PR #519, opened 2017-06-11
Rolling Shutters with Sonoff Dual

dino-de:
There were already some requests for a shutter enhancement here.
I use a Sonoff Dual (or Ch4) as switch for my rolling shutters.
My shutter motors have 3 connectors: N, Motor UP, Motor down and provide internal end switches.

The code here should use not more than 300 bytes of flash space and provides this functionnality:
Using the Wemo emulation, each ON/OFF command first shuts both relays off for 10 ms, then for a period of "pulsetime" relay 1 (ON command) or relay 2 (OFF command) switches on.
So I can tell my echo dot "Alexa Shutters On" and the shutters go up.
The motor can also be controlled with buttons at the internal pin header of the Dual: UP = button0 /
DOWN = button1. The top button (button2) should work as normal, i.e. toggle.
Any command during pulsetime stops the motor.

A new module type is defined in sonoff_template.h as "Shutter" (fairly the same as Sonoff Dual).
Set pulsetime a bit higher than the shutter rising time (in my case 40 sec. - in web console: "pulsetime 140"). PowerOnState should be 0 (console: "PowerOnState 0") .

Warning!
The code is not so much testet, I tried to avoid any side effects for the other moduls, but no one knows...

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -382,7 +382,7 @@ void setRelay(uint8_t rpower)
     power = (1 << Maxdevice) -1;
     rpower = power;
   }
-  if ((SONOFF_DUAL == sysCfg.module) || (CH4 == sysCfg.module)) {
+  if ((SONOFF_DUAL == sysCfg.module) || (SHUTTER == sysCfg.module) || (CH4 == sysCfg.module)) {  ///
     Serial.write(0xA0);
     Serial.write(0x04);
     Serial.write(rpower);
@@ -1623,7 +1623,7 @@ void do_cmnd_power(byte device, byte state)
     device = 1;
   }
   byte mask = 0x01 << (device -1);
-  pulse_timer[(device -1)&3] = 0;
+  if (SHUTTER != sysCfg.module) pulse_timer[(device -1)&3] = 0;  ///
   if (state <= 2) {
     if ((blink_mask & mask)) {
       blink_mask &= (0xFF ^ mask);  // Clear device mask
@@ -1639,11 +1639,27 @@ void do_cmnd_power(byte device, byte state)
     case 2: // Toggle
       power ^= mask;
     }
-    setRelay(power);
+      /// Sonoff Dual(or Ch4) as switch for rolling shutters
+      /// a new module type is defined in sonoff_template.h as Shutter (fairly the same as Sonoff Dual)
+      /// set pulsetime a bit higher than the shutter rising time (in my case 40 sec. - in web console: "pulsetime 140")
+      /// PowerOnState should be 0 (console: "PowerOnState 0")
+/// progmem 480516 Bytes (46%), data 43308 Bytes (52%)
+
+    if (SHUTTER == sysCfg.module) {
+      setRelay(0);            /// both relays off first
+      if (!pulse_timer[0]) {  /// any command during pulsetime stops the motors
+       delay(10);             /// wait 10 ms for relays shutoff - not shure this is needed
+       setRelay(2 - power);   /// ON -> relay 1, OFF -> relay 2
+       pulse_timer[0] = sysCfg.pulsetime[0];
+      } else
+        pulse_timer[0] = 0;    /// leave relays off
+    } else {
+      setRelay(power);
+      pulse_timer[(device -1)&3] = (power & mask) ? sysCfg.pulsetime[(device -1)&3] : 0;
+    }
 #ifdef USE_DOMOTICZ
     domoticz_updatePowerState(device);
 #endif  // USE_DOMOTICZ
-    pulse_timer[(device -1)&3] = (power & mask) ? sysCfg.pulsetime[(device -1)&3] : 0;
   }
   else if (3 == state) { // Blink
     if (!(blink_mask & mask)) {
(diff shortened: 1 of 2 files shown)


davidelang:
something is wrong with this file, it shows every line in sonoff.ino as changing (line endings possibly?)

that makes it very hard to review this.

I would be opposed to a shutter functionality that only works on the dual, and anything that only works with relay1 and relay2, this needs to be more flexible to be able to be used with any multi-relay module.

dino-de:
Sorry, I am not very experienced in github.
Loaded 2 files up: [sonoff.ino and sonoff_template.h](https://github.com/arendst/Sonoff-Tasmota/pull/519/files).
Then I changed sonoff.ino (deleted only 1 comment line) and must have made a mistake on uploading it again, so that the whole file was marked as changed.  So please ignore my second commit.
Maybe you have a tip how can I repair this.
This solution should work at least on the ch4 too. I bought the Duals for steering my shutters and that's
what they do for me now. I'm not happy with IThead's layout with the second processor and the serial protocol instead of using direct I/Os for buttons and relays, making it unneccessarily complicated.
If there is better hardware on the market, the code should be easly expandable.

davidelang:
At this oint, I think you will need to go back and re-generate the patch from scratch.

We need to have @arendst comment, but I think what you need to do is to create a new item type (shutter1up/shutter1down, shutter2up/shutter2down or similar) and have a way for them to be assigned to a relay rather than having to create a new module type for each device that will have this capability.

the way you are doing it means that you would need one new module type for the dual, one new module type for the ch4, one new module type for a sv that has a second relay attached, etc

dino-de:
Thanks for your comment.
As the title says, this patch is for the Sonoff Dual but the Ch4 should work without modifications, there is no need for a new module type.
Clearly, a shutter switch needs at least 2 relays. The only hardware I can test with 2 relays is the Dual.
These patches handle the dammed specific serial protocol of the IHeads for MY needs.
My approach is to steer my shutters with only ON/OFF commands, buttons and Alexa, no MQTT or Domotic server needed  (but possible), with common hardware.
I don't really know what do you mean with "create a new item type" in this context - maybe my fault - but if you can provide a general solution, I will agree...

davidelang:
the ch4 doesn't use the serial interface to the relays, so your module 
definition, which depends on that serial interface, won't work on the ch4. It 
would need a different definition, and your code tests for the module type 
"shutter" so the code won't work for anything else.

dino-de:
The original source code says:

void setRelay(uint8_t rpower)
{
  uint8_t state;

  if (4 == sysCfg.poweronstate) {  // All on and stay on
    power = (1 << Maxdevice) -1;
    rpower = power;
  }
  if ((SONOFF_DUAL == sysCfg.module) || (**CH4 == sysCfg.module)**) { 
    Serial.write(0xA0);
    Serial.write(0x04);
    Serial.write(rpower);
    Serial.write(0xA1);
    Serial.write('\n');
    Serial.flush();
  } 

But I will have a look to schematics...

arendst:
CH4 is NOT the latest Sonoff 4CH. It is a legacy 4 channel device controlled like a Sonoff dual by the serial interface. I think it is EOL. So this solution is usable for a Sonoff Dual only.

roondar:
I can confirm that works great.
image
Thanks both of you. 👍

PeggyFree:
Hello, what about using this 2ch item for future curtains projects? It looks more appropriate : https://www.banggood.com/SONOFF-2-Channel-AC-85V-250V-APP-Remote-Control-WIFI-Wireless-Switch-Socket-For-Smart-Home-p-1114958.html?rmmds=search

kabongsteve:
I missed this pull request, but have added my own "Isolate functionality that should achieve the same outcome as required, but is more flexible.

roondar:
And you will do a pull Request ?

kabongsteve:
Already done, and improved
Add Relay Interlock Functionality

roondar:
Nice.

tom211:
I have looked at the schematics from the dual. Isn´t it possible to get both relays on when using the header for the buttons because it is connected directly to the controller that controls the relays?

neographikal:
Also would like the shutter functionality, either on dual/4ch or on nodemcu controlling a bunch of relays

calcipher49:
Hi,
Have you any news about this request?
It should be nice to add dino-de mod and new Shutter module type in standard Tasmota code (using Sonoff Dual).
Also, a great addition should be intermediate position management for shutters.
IE Set to 0 = OFF, set to 100 = ON, and add  values between 0 and 100 setting appropriate relay ON during time = PulseTime * (New value - Old value)/100

Thanks

pviolero:
I have tested this modification with the release 5.1.6. the MQTT communication and the toggle are OK but the relays don't change their state. I'm using a sonoff dual R2 and not the dual.

Do you know how to modify the code to run on a dual R2 ?

flubber89:
any chance to get an update for the Dual R2 version?

getget1980:
Hello, same thing here, please update to Dual R2 (and integrate in Tasmota releases  :) )

kabongsteve:
I implemented something a little different

https://github.com/kabongsteve/Sonoff-Tasmota

https://community.home-assistant.io/t/sonoff-tasmota-extension-for-covers/44694/8

roondar:
Really impressive !! Do you think it's compatible with Domoticz and % level
of aperture ?
Do you use button 1 and 2  ?

PR #658, opened 2017-07-29
WEMOS SHT30 Shield

don-willingham:
I have successfully read the temperature and humidity from an [WEMOS SHT30 Shield](https://wiki.wemos.cc/products:d1_mini_shields:sht30_shield). I adapted their [library](https://github.com/wemos/WEMOS_SHT3x_Arduino_Library) to tasmota. It currently only supports 1 sensor with an address of 0x45. I really do not like the delay(500), so maybe the sht3x should be disabled by default. Although it works for me, it's still a bit beta. To avoid confusion I made sure all existing sht references were changed to sht1x. I also cleaned up some trailing white space.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1932,9 +1932,9 @@ void sensors_mqttPresent(char* svalue, uint16_t ssvalue, uint8_t* djson)
 #endif  // USE_DHT
 #ifdef USE_I2C
   if (i2c_flg) {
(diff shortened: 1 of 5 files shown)


don-willingham:
After merging with 5.5.1, the sensor isn't working. I have resolved the conflicts in my sht3x branch, and will continue to work in there.

arendst:
Released in version 5.10.0e

PR #660, opened 2017-07-29
Support for ADS1115 16bit 4-channel A/D converter

### LINKED ISSUE #338: I can provide extension for I2C ADS1115. Intrested?
I created a new file xsns_ads1115.ino linked on demand to support 16bit 4-channel A/D converter. Mr. Arend, interested in adding. Library used: Adafruit_ADS1X15-master

Code +1kb, Ram +968Byte.

syssi:
Fixes #338.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -22,7 +22,7 @@ framework = arduino
 board = esp01_1m
 board_flash_mode = dout
 build_flags = -Wl,-Tesp8266.flash.1m0.ld -DMQTT_MAX_PACKET_SIZE=512
-lib_deps = PubSubClient, NeoPixelBus, IRremoteESP8266, ArduinoJSON
+lib_deps = PubSubClient, NeoPixelBus, IRremoteESP8266, ArduinoJSON, i2cdevlib-ADS1115
 
 ; Sonoff minimal
 [env:sonoff-minimal] ; Placeholder to be configured
@@ -40,4 +40,4 @@ framework = arduino
 board = esp01_1m
 board_flash_mode = dout
 build_flags = -Wl,-Tesp8266.flash.1m0.ld -DMQTT_MAX_PACKET_SIZE=758 -DUSE_DS18x20 -DMESSZ=600
-lib_deps = PubSubClient, NeoPixelBus, IRremoteESP8266, ArduinoJSON, OneWire
+lib_deps = PubSubClient, NeoPixelBus, IRremoteESP8266, ArduinoJSON, OneWire, i2cdevlib-ADS1115
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1944,6 +1944,9 @@ void sensors_mqttPresent(char* svalue, uint16_t ssvalue, uint8_t* djson)
 #ifdef USE_BH1750
     bh1750_mqttPresent(svalue, ssvalue, djson);
 #endif  // USE_BH1750
+#ifdef USE_ADS1115
+    ads1115_mqttPresent(svalue, ssvalue, djson);
+#endif  // USE_ADS1115
   }
 #endif  // USE_I2C
   if (strstr_P(svalue, PSTR("Temperature"))) {
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -2035,6 +2035,9 @@ void every_second()
 #ifdef USE_BH1750
         bh1750_detect();
 #endif  // USE_BH1750
+#ifdef USE_ADS1115
+        ads1115_detect();
+#endif  // USE_ADS1115
       }
 #endif  // USE_I2C
     }
--- a/sonoff/user_config.h
+++ b/sonoff/user_config.h
@@ -158,6 +158,7 @@
 
 #define USE_I2C                                  // I2C using library wire (+10k code, 0.2k mem) - Disable by //
   #define USE_BH1750                             // Add I2C code for BH1750 sensor
+  #define USE_ADS1115                            // Add I2C code for ADS1115 16bit 4-channel A/D converter
   #define USE_BMP                                // Add I2C code for BMP/BME280 sensor
   #define USE_HTU                                // Add I2C code for HTU21/SI7013/SI7020/SI7021 sensor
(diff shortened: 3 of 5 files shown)


syssi:
Program size +18kB
Free memory -2kB

stefanbode:
I also use these devices, therefore a good idea to add. What I have found out, that also these devices have some large deviations, if only doing one read. Doing average is suitable, but I got some problems not able to solve. There seems to be some delay required. I used it to measure moisture with capacitive sensors.

syssi:
@stefanbode Does this approach fix the issue, too: https://github.com/arendst/Sonoff-Tasmota/issues/686

We don't need an average specially for the ads1115 anymore, right?

stefanbode:
I'm not sure, but I guess you should take a look. I assume there is no code reuse and you have to add the average also for the ADS1115. Anyhow, I tried this without great success. It must be tested anyway. On the A/D converter on the ESP the delay(1) (1 ms) this is fine and enough. On the ADS115 I did not get it working, but only invested less than 30 minutes.

syssi:
Alright. I will make some tests and report.

stefanbode:
Worked perfectly for me. Anyhow, I would suggest shortening the string "AnalogInputx". Even is this matches to the AnalogInput0 that comes from the ESP itself we create very long MQTT strings here. You should discuss this with arendt. For me it works but because also using some DS18B20 the message is really long and near its limit. Currently, I do not see the requirement to do an average, but further investigations required. Especially after wake up from deepsleep. I replaced my AD1115 implementation with yours. Great job and really like the I2C extensions.

stefanbode:
Ok, there are spikes and for sure some error values like 32.767. These errors have no clear reason when they appear. Also, negative values appear where they should not. I did run battery voltage measured over some hours. One idea could be filtering it. I just now used the 32 measures trick already used for AnalogInput0. 

Inserted/changed:
`  for (byte i = 0; i < 4; i++) {
    int16_t val;

    for (byte i = 0; i < 32; i++) {
      val += ads1115_getConversion(i);
      delay(1);
    }
    
    if (!dsxflg) {
      snprintf_P(svalue, ssvalue, PSTR("%s, \"%s\":{"), svalue, ads1115stype);
      *djson = 1;
      stemp1[0] = '\0';
    }
    dsxflg++;
    snprintf_P(svalue, ssvalue, PSTR("%s%s\"AnalogInput%d\":%d"), svalue, stemp1, i, val >> 5);
    strcpy(stemp1, ", ");
  }`

I did not change the webserver stuff because these values normally are not critical. It might be that I will add a filter because also the average can create unacceptable deviations because it is not an error in the near of the original value (Like AnalogInput0). It is an error sometimes FAR FAR away. Calculation efforts will increase. I check how the average will behave over night.

stefanbode:
I had a lot of investigations and tests done yesterday to find out where the wrong values are coming from. It looks like the WLAN and the different open sockets might cause the problem. An open CONSOLE on the Webbrowser cases more errors than a closed one. Syslog is less problematic. Also the MQTT I use causes trouble from time to time. Now I investigated a brunch of code to overcome the issue, or at least do not report wrong values in the MQTT stream. The code is more than lenghtly and changed several times to get all the situations where wrong data is generated. This is my current state that delivers stable values for at least one hour (teleperiod 15). And depending on the different acitivities sometimes the device reboot with BLOCKED LOOP error, but currently did not report ONE wrong value to MQTT.

```
void ads1115_mqttPresent(char* svalue, uint16_t ssvalue, uint8_t* djson)
{
  if (!ads1115type) {
    return;
  }
  char log[LOGSZ];
  char stemp1[10];
  byte dsxflg = 0;

  for (byte i = 0; i < 4; i++) {
    int32_t val, temp_val, max_val, min_val, count0;
    byte valid_measure = 1;
    while (valid_measure < 5) {
      val = 0;
      count0 = 0;
      for (byte j = 0; j < 32; j++) {
        temp_val = ads1115_getConversion(i);
        yield();
        val += temp_val;
        if (temp_val == 0) {
          count0 +=1;
        }
        if (j == 0) {
          max_val = temp_val;
          min_val = temp_val;
        } else {
          max_val = max(max_val, temp_val);
          min_val = min(min_val, temp_val);
        }
      }
      if ( max_val - min_val < 100 && 0 < max_val - min_val) {
        valid_measure = 10;
        snprintf_P(log, sizeof(log), PSTR("ADS1115: Probing %d converged. max: %d min: %d, avg %d, deviation %d"), i, max_val, min_val, val >> 5, max_val - min_val);
        addLog(LOG_LEVEL_DEBUG, log);
      } else {
        valid_measure += 1;
        snprintf_P(log, sizeof(log), PSTR("ADS1115: Probing %d not converged. max: %d min: %d, zero counts %d"), i,max_val, min_val,count0);
        addLog(LOG_LEVEL_ERROR, log);
      }
    }

    if (valid_measure == 10) {
      if (!dsxflg  ) {
        snprintf_P(svalue, ssvalue, PSTR("%s, \"%s\":{"), svalue, ads1115stype);
        *djson = 1;
        stemp1[0] = '\0';
      }
      dsxflg++;
      snprintf_P(svalue, ssvalue, PSTR("%s%s\"AnalogInput%d\":%d"), svalue, stemp1, i, val >> 5);
      strcpy(stemp1, ", ");
    }
  }
  *djson = 1;
  if (dsxflg) {
    snprintf_P(svalue, ssvalue, PSTR("%s}"), svalue);
  }
}
```

stefanbode:
I rolled back anything. Looks like any try to improve just make it worth.

syssi:
Which hardware and pins do you use?

stefanbode:
You can buy one of these on AliExpress: https://www.aliexpress.com/wholesale?catId=0&initiative_id=AS_20170824223853&SearchText=ads1115
Connecting SCL and SDA to GPIO 4 and 5 works well. They have all required resistors and capacitors already on board. As far as I can see, the ADS1115 do not like unconnected pins and then requesting them. Just to be aware.
Just if you want to run it on battery there is some additional effort required to avoid the ADS1115 connected to power during deepsleep.

syssi:
Oh no. ;-) I'm asking because of your spurious interrupt issues. I suspected a i2c bus at unusual pins.

PR #703, opened 2017-08-08
Increase minimum packet size to 475 in #error

don-willingham:
Version 5.5.0 increased MESSZ by 8, but not the #error message.


arendst:
Thnx. I love these fixes as they take no code space ;-)

Will be implemented in next release.

PR #719, opened 2017-08-10
Add Relay Interlock functionality

kabongsteve:
Hi theo,

I'm planning to use the multi-relay board as a blinds controller.  
Part of the requirements of this, and other uses, is to make sure that you cannot turn on multiple relays at the same time.  Thus the turning on or one relay, should force off any other relay.

So I added a command called Isolate, allowing you to set the isolation for each relay.
The value of Isolate is the relay(s) that should be turned off when this relay is turned on.  It currently uses the bit field version of relays, so the Relays values are 1, 2, 4, 8 (so that multiple can be used).

Perhaps I should support a text string of 1 and 0 of O and F and parse it?

Thus to turn off all other relays when Relay 1 is turned on, you would use Isolate1 14
Or to simply switch between relays 1 and 2, you would use two Isolate commands
Isolate1 2
Isolate2 1

I added a 250ms delay between the Offs, and subsequent On, to make sure there was some separation.

I haven't tested with my blinds as yet, but may need to add an "IsolationDelay" setting to allow users to set the time delay between the offs and subsequent on, to make sure the controlled devices can respond appropriately.

Steve


arendst:
Delays in code are deadly for responsiveness.

I suggest to get yourself a Sonoff 4Ch Pro and it's all done in semi-hardware while still having the Tasmota features.

kabongsteve:
Truish..
But Pro is almost twice the price and I already bought 2 4CH.
I need "inching mode" but closer to 30 seconds, it only supports 4.
And inching and lockout together, as well as independently settable for any/all the relays, not just a 1 to all.

I'll recode to remove the delay call, and use a PulseTimer equivalent for the Isolation Delay

Maybe even rename to Interlock instead of Isolation, so its more consistent

kabongsteve:
I've refactored command to be called Interlock.
Added an InterlockTime, using time range as pulseTime
removed the delay call, and supported the InterlockTime correctly.
changed command setting, so you pass relay numbers in simple string
Thus Interlock1 23  means for relay1, turn off 2 and 3

arendst:
I will implement global interlock in version 5.5.2k and up.

It does not use your timers but still enables the use of pulsetime.

kabongsteve:
That's a real shame.

Your changes mean I won't be able to use your version of isolation, as it works across all 4 relays.

I have a single 4CH running two sets of blinds, and can run each independently, including the extra inbetween delay for safety.

neographikal:
It would be nice to be able to assign pairs of GPIO pins that will control one shutter (in an array maybe?). This would make it more flexible than just a custom configuration for the interlock functionality on the dual. The dual/4ch would have a standard setting for this.

I've got one nodemcu that controls 5 shutters, so 10 GPIO's ;)

meingraham:
I know this is closed and had meant to comment long ago... I'd prefer a configurable grouping:

- All channels
- 2 & 2
- 1 & 3

ascillato:
See open issue #824

PR #759, opened 2017-08-19
Allow for increased precision of power measurements

PeteBa:
I have done some analysis of the power precision using sonoff-tasmota and believe there is greater than 1 watt precision achievable with the hlw chip (at least in the low to mid range). See graph below. This change proposes extending the VoltRes command into the Power domain. In addition, this would provide consistency between the two measurement functionality.

huafan resolution

There are two main changes:
1) move to a floating point calculation to avoid integer style "floor" rounding when calculating wattage in hlw_readEnergy. This could result in a more accurate baseline reading
2) allow for a single decimal of precision as analysis shows that the chip - at least with Huafan Switch shunt resistor and voltage divider - can provide this

I would agree that at very low current levels (approaching sub-30mA) then there is too much instability and cut-out to pretend accuracy but for above, say, 2W then there is precision to be had.

Appreciate any feedback.

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -41,7 +41,7 @@ typedef union {                            // Restricted by MISRA-C Rule 18.4 bu
     uint32_t spare16 : 1;
     uint32_t spare17 : 1;
     uint32_t spare18 : 1;
-    uint32_t spare19 : 1;
+    uint32_t wattage_resolution : 1;
     uint32_t voltage_resolution : 1;
     uint32_t emulation : 2;
     uint32_t energy_resolution : 3;
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1152,6 +1152,12 @@ void mqttDataCb(char* topic, byte* data, unsigned int data_len)
       }
       snprintf_P(svalue, sizeof(svalue), PSTR("{\"PressRes\":%d}"), sysCfg.flag.pressure_resolution);
     }
+    else if (!strcmp_P(type,PSTR("WATTRES"))) {
+      if ((payload >= 0) && (payload <= 1)) {
+        sysCfg.flag.wattage_resolution = payload;
+      }
+      snprintf_P(svalue, sizeof(svalue), PSTR("{\"WattRes\":%d}"), sysCfg.flag.wattage_resolution);
+    }
     else if (!strcmp_P(type,PSTR("VOLTRES"))) {
       if ((payload >= 0) && (payload <= 1)) {
         sysCfg.flag.voltage_resolution = payload;
--- a/sonoff/xsns_hlw8012.ino
+++ b/sonoff/xsns_hlw8012.ino
@@ -173,7 +173,7 @@ void hlw_savestate()
   sysCfg.hlw_kWhtotal = rtcMem.hlw_kWhtotal;
 }
 
-void hlw_readEnergy(byte option, float &et, float &ed, uint16_t &e, uint16_t &w, float &u, float &i, float &c)
+void hlw_readEnergy(byte option, float &et, float &ed, uint16_t &e, float &w, float &u, float &i, float &c)
 {
 /* option 0 = do not calculate period energy usage
  * option 1 = calculate period energy usage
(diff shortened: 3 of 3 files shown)


arendst:
I'll test it and hope to see some positive results too as I also hate the low power instabilities. Nice integration though 👍

arendst:
The HLW8012 provides power, current and voltage values as a pulse length. The lower the power, the longer the pulse length. For a Power values of 4 Watt it returns a pulse length of almost 3 seconds. I probe the power signal for up to 3 seconds before deciding what value it has.

To better read power values below 4W you'll have to wait longer for the pulse length to end.

Would you be so kind to test that if you make the below changes to xsn_hlw8012.ino at line 82 does it provide better values?

```
  if (hlw_cf_plen > 5000000) {
    hlw_cf_plen = 0;  // Just powered on
  }
  hlw_cf_timer = 20;  // Support down to 3W which takes about 4 seconds
```

PeteBa:
Hey, so I managed to get a side-by-side of both the Sonoff POW and Huafan switch using a bunch of available devices as loads with the following comparison:

comparison

I then modified the firmware as suggested but max'ed the cf_timer cut-out so I could step through a much lower range. I also switch to my programable load and took a bit more care to re-calibrate both devices using the hlwpset command. I stepped down in 1/2 watt increments from 4 watts.

comparison low power

My view would be:
- both devices can easily support a single decimal precision (and just seen that you have taken it onboard thanks !)
- the shunt resistor/voltage divider settings make a big difference at the low end with the sonoff being useful down to 2.5w. and the Huafan getting some great stability down to about 1 watt (this is as low as I can go with my usb load)
- unsurprisingly I found that my usb load gave far more stable readings than when using standard consumer devices as loads so mileage may vary when it comes to value at this low a level
- probably the big question is whether extending the "wait" time is worth it in order to get these low readings. This delay becomes a little annoying when switching a device off takes 10 seconds to register as zero watts!

On the last point, maybe, using the voltage reading as a proxy for zero watts might work i.e. if voltage is reading zero then assume that power is also zero even though we may still be waiting for the "cf_timer" cutoff ?

arendst:
Thnx Pete, will look into a way to extent the time for low loads. The problem is that I'm currently using the power load to decide to set the voltage to 0 as voltage is always present on the HLW.

That's a catch 22 then ;-)

khcnz:
Wouldn't it be useful to show/broadcast the voltage even when no load? The voltage is still there even if no current is being consumed.

PeteBa:
Theo, just seen the recent commits to remove the cf_timer logic. Not sure if this is still work-in-progress but I think the change may have a problem but wanted to check my thinking first.

I had thought the decision-tree for the Sonoff Pow was to report conditions on the load-side of the relay as follows:
1) Sonoff POW Relay OFF => Power, voltage and current report as zero
2) Sonoff POW Relay ON
             |--> 2a) Load ON => Power, voltage and current report based on load consumption from hlw reading
             |--> 2b) Load OFF => Voltage report based on hlw reading but power and current report zero

Trying out the latest development branch, I am seeing odd behaviour in scenario 2b), for example,
if the sonoff relay is on and the load is switched from on to off (e.g. a lamp via a bulb side switch) then the sonoff will continue to report a non-zero power and current. Looking at the code, this occurs because the hlw chip never generates a zero power/current report and therefor, without the cf_timer timeout, the sonoff continues to  publish the last registered reading

- When switching off the lamp at the bulb, the sonoff generally got a fairly low final reading from the hlw, less than a watts, but on a few of occasions it was up around 10-15 watts continuing to be published ongoing.

- This also manifest with the current report as I was seeing around 30mA being published despite the lamp being off at the bulb. Not sure whether the current is real (i.e. powering the sonoff itself ?) or a zero error. In the previous code this wouldn't have happened as the logic was to zero current when power was zero. 

Also, a bit to khcnz's point above, if there is no load attached to the sonoff and you switch on the relay then you will get a zero voltage reading being published. I would agree that this should be the hlw chip reading as the voltage will be present on the load side of the relay regardless of whether current/power is being consumed.

I will take a closer look at the code and see if I can come up with any suggestions to help.

arendst:
I'll revert the latest changes but will extent the time to measure low power values.

PeteBa:
Closing this PR as the ability to increase precision of power readings (WattRes) has been provided. Opened new PR #796 to progress logic for determining when load is considered off.

PR #796, opened 2017-08-27
Added cutoff limit and adjusted power/current/voltage publishing logic

PeteBa:
Theo, as per #759 , I have had a go at adjusting the logic for publishing power/voltage/current as per:

1) Sonoff POW Relay OFF => Power, voltage and current report as zero
2) Sonoff POW Relay ON
        |--> 2a) Load ON => Power, voltage and current report based on load consumption from hlw reading
        |--> 2b) Load OFF => Voltage report based on hlw reading but power and current report zero

In addition, I added a "cutoff" timer to proxy for the load being switched-off and a zero power/current reading. If the timeout is reached (10 seconds) before receiving a hlw power interrupt then assume that the load has been switched off and zero power and current readings.

I will close #759 as that was primarily about increasing power reading precision which has now been delivered - thank you - and use this as "cutoff/publishing logic"

### DIFF
--- a/sonoff/xsns_hlw8012.ino
+++ b/sonoff/xsns_hlw8012.ino
@@ -63,6 +63,7 @@ unsigned long hlw_Ecntr;
 unsigned long hlw_EDcntr;
 unsigned long hlw_kWhtoday;
 uint32_t hlw_lasttime;
+boolean hlw_load_off;
 
 unsigned long hlw_cf1u_pcntmax;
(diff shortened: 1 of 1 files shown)


arendst:
Implemented in master release 5.7.0

haniham:
You serch for [SetOption21](https://tasmota.github.io/docs/Commands/#setoption21)

Issue #698, opened 2017-08-06
Fixed: HUE App wants to update emulated HUE devices


chinds185a:
So I successfully flashed this firmware onto a 4CH sonoff  today. I have turned on the hue emulation. I want to control these devices with Alex and ultimately Apples home kit.

Alex does not discover the devices so I guess I need to add them in the HUE app first? 

Going to the HUE app it find the new bridge and the 4 named devices. However I cannot save the light/add them to a room etc as the HUE app is continuously trying to update the firmware on the new emulated hue bridge and obviously fails.

How can I get each ch discover-able by Alexa? Using wemo emulation works but again only for a single Chanel.

arendst:
Update emulated HUE devices as designed. Don't know how to fix. Probably have to tell the App I have a way to new version....

Alexa should work without any Hue App.

Courty40:
I've just tried the same thing with a 4ch and version 5.5.2 and I'm seeing the same behaviour.
Alexa see ch1 of the 4ch device in Wemo mode but doesn't detect the Sonoff at all in Hue mode. Tried rebooting/power off etc
I had this working a few versions ago on a 2ch Sonoff so I do t believe it's the Alexa

Courty40:
And I can confirm the hue app finds the Sonoff and tries to upgrade it even though the Alexa device search doesn't
Something changed in the Hue discovery protocol maybe ?

arendst:
My Hue code changed between version 5.2.4 and 5.3.0 and might be causing you trouble. I notice the same behaviour on my Hue App regarding the firmware update. I just ignore it.

I do not have Alexa so I can't test. If others have problems with it too please let me know.

Courty40:
Just remoted into one I know works with Hueand that's a BN-SZ LED and its running 5.3.0

Courty40:
More info, Alexa does poll the unit and the unit is responding but Alexa doesn't accept the Sonoff

23:44:31 HTTP: Root
23:44:33 HTTP: Console
23:44:46 Wifi: Checking connection...
23:44:46 Wifi: Connected
23:44:48 UPnP: HUE 3 response packets sent to 192.168.1.16:50000
23:44:48 UPnP: HUE 3 response packets sent to 192.168.1.16:50000

chinds185a:
Still having problems with this. Alexa just doesn’t discover the hue devices. 

Is it worth downgrading the firmware used?

altelch:
Hi,

[...]

PR #854, opened 2017-09-09
Fix for Alexa detection of HUE Emulation and fix Update Request in HUE App

altelch:
#698 is fixed with this.


altelch:
Echo/Alexa requires "basic" and not "Basic". After this simple change detection worked flawless. The rest was required to get the HUE App satisfied...

PR #912, opened 2017-09-23
Add in user-configurable generic ESP8266 device

mzealey:
Use-case: supporting misc non-Sonoff devices without needing code-changes, and
in order to easily create custom devices based on ESP8266 type processors.


davidelang:
the wemos device already is fully user configurable.
k

arendst:
As David stated the WeMos D1 mini device does this and provides the ADC input too ;-)

Issue #920, opened 2017-09-25
Sonoff pow not able tu use temp sensor


RemCom:
In de current software versions (5.8) when using a Sonoff pow it is not possible to configure a temp sensor on GPIO1 or GPIO3.

I believe this is still possible. The schematic doesn’t suggest that these ports are used in the energy measuring hardware.

Is it possible to make this available in the software?

arendst:
See why not supported  https://github.com/arendst/Sonoff-Tasmota/wiki/Sonoff-Pow

RemCom:
There is nothing on that page where you suggest that it isn’t possible.
The only thing that you mention is that it isn’t a smart idea to connect your pc/laptop to it because you have a change to make a ground loop. But still, just like in the normal sonoffs. The 3.3v is still 3.3v and the ground is always the ground and the serial wires are not used.
If these voltages are doing strange stuff the processor should also fry itself if it would be any different. The only reason why it is a bad idea to connect something to the serial ports is a pc with witch your connecting 2 grounds from a different power supply to each other. That can cause a surge to the hardware because the 2 ground levels are trying to get to the same level potentially frying any hardware in the process.
Al already have a sonoff pow with a local hack running power measurements and having a temp sensor for a few weeks.

RemCom:
Sorry for the lack of enters in the text. The got lost somewhere typing the text on my phone

davidelang:
actually, the 3.3v is 123v and the ground is 120v

that's why connecting to these pins is not supported. If you hook up an 
opto-isolator you could safely use those pins, but if you can do that, then you 
can modify the sources to support it.

RemCom:
didn’t noticed that the ground was a 120 volts. But also without the opto-coplers you can safely use it to hookup a sensor. Just put some shrink sleeve around the pins (which is always a good thing to do). Then its safe to use without a opto copler. Remark that I only talk about connecting temp sensors.

A risk that you still can have is that your temp sensor has a metal case which is connected to the “ground” and having 120 volt on it. That’s not safe. But with a sensor like a AM2302 you have little to no risk.

The problem of keeping my own source for this that I can’t easily update without doing some extra work.

Is it an idea to make this option available but only after you clicked a box that your aware of the danger that your taking? Possibly with a link to the wiki or something like that.

mormor1971:
[...]

PR #921, opened 2017-09-25
Add gpio1 and gpio3 back to sonoff pow.

remcom:
see issue #920. No reason or hw issue why these ports should be not available.

Reasons that are given are only a risk when connecting it to a pc or other device with a own power supply


davidelang:
On Mon, 25 Sep 2017, Remco wrote:

> see issue #920. No reason or hw issue why these ports should be not available.
>
> Reasons that are given are only a risk when connecting it to a pc or other device with a own power supply
> You can view, comment on, or merge this pull request online at:

not true, the reason is that ground is not ground. It's a danger to put those 
pins out to things that consider it safe to have ground exposed for people to 
touch.

PR #946, opened 2017-10-01
Fix #945 ('Set Level' should turn on the light)

### LINKED ISSUE #945: Light not turned on when dim level is set
According to https://www.domoticz.com/wiki/Domoticz_API/JSON_URL%27s#Set_a_dimmable_light_to_a_certain_level, the light should be turned on when dim level is set.
However, there seems to be a bug on this line in domoticz_mqttData():
`if (sfl_flg && (sysCfg.led_dimmer[i] == nvalue)) {`
If dim level is unchanged, light will not be turned on.

Maybe the above line should be changed to:
`if (sfl_flg && (sysCfg.led_dimmer[i] == nvalue) && ((power >> i) &1)) {`

emontnemery:


arendst:
Works if you only use Domoticz to control the device.
Fails as soon as you use the device webpage and want to power off the device; Domoticz will send power on the device as a result of power off message send by device.

The way this works is as follows:
1 - Webpage power button pressed to power on device
2 - Device is turned on and two messages are send to Domoticz; one with Dimmer info, the other with power info.
3 - Domoticz returns the messages as commands for both Dimmer setting and Power setting.
4 - As the Domoticz dimmer value equals the device dimmer setting nothing happens.
5 - As the Domoticz power value equals the device power setting nothing happens either.

Domoticz is a bit of a MQTT message generator resulting in this unwanted behavior. The original solution did work for both the webpage AND Domoticz as long as you do not expect to turn on the light by sending the same dimmer value; sending a power on does work though.

I've been trying some solutions but do not want to implement some kind of mutex where a message send by the device needs to be validated by a return from Domoticz as it may well be that a MQTT message from Domoticz is never received resulting in a miss-communication on next interrogation.

So I'm afraid I have to leave it as it is unless you come up with a very clever idea ;-)

emontnemery:
Here's a new attempt:
Send dimlevel together with the on/off status message. One potential concern is that Domoticz seems to ignore dimlevel sent with an "off" status message, I'm not sure if that will/might have side effects?
(I did mess up a bit when rebasing the pull request, sorry about that)

PR #953, opened 2017-10-02
Support for ESP8266_HVIO 

tom211:
I have tested this board with tasmota and it works fine. There are two 230 V outputs and two 230 V inputs for common wall switches.
All information could be found here:
[https://luani.de/projekte/esp8266-hvio/](https://luani.de/projekte/esp8266-hvio/)

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -160,6 +160,7 @@ enum module_t {
   WITTY,
   YUNSHAN,
   MAGICHOME,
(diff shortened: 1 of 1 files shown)


davidelang:
Please update the wiki with this board.

arendst:
In pre-release 5.8.0e

meikelmoeller:
Problems with Alexa support, HUE emulation does not work. Items are found, but they don't respond.

Issue #955, opened 2017-10-02
PWM level at startup


emontnemery:
I'm using the Sonoff BN-SZ configuration.
Values from sysCfg.power and sysCfg.led_dimmer are shown in UI and sent to MQTT, however PWM level itself is not initiated.

Edit:
Some more details.
With sysCfg.poweronstate = 0:
UI matches system state after a cold boot.
UI does not match system state after a restart since resetInfo.reason is neither REASON_DEFAULT_RST nor REASON_EXT_SYS_RST.
Even though GPIOs might be untouched, PWM is probably stopped and should be reinitiated?

With sysCfg.poweronstate = 3, but sysCfg.flag.savestate = 0:
UI does not match system state if there are some saved values in sysCfg.power and sysCfg.led_dimmer

arendst:
As you rightfully suggested setting sl_any to 1 should handle the first situation.

The last situation will be implemented as that all saved power will be set to Off once SaveState (SetOption0) is set to 0 resolving the situation of saved value miscompare. Dimmer will still be saved as it is always the current dimmer value (Although power is Off dimmer value can still be set and will be used as the last dimmer value when powered on).

Both changes will be in next (pre)release.

PR #960, opened 2017-10-03
Add support for inverted PWM

emontnemery:
Add support for inverted PWM control, for example when P-channel MOSFET is used to drive LED.

Note:
Default state of PWM needs to be implemented.
What works for me for dimmer control is to simply set sl_any to 1 instead of 0 at the end of sl_init, but maybe this has some side effect?
See also #955

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -285,6 +285,7 @@ mytmplt my_module;                    // Active copy of GPIOs
 uint8_t pin[GPIO_MAX];                // Possible pin configurations
 uint8_t rel_inverted[4] = { 0 };      // Relay inverted flag (1 = (0 = On, 1 = Off))
 uint8_t led_inverted[4] = { 0 };      // LED inverted flag (1 = (0 = On, 1 = Off))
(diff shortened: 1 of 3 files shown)


arendst:
sl_any should work fine this way.

I support your commit too but have to think about the location of the inverted options.

arendst:
Do you have a real use-case for this other than there might be P-Channel MOSFET being used...

emontnemery:
Thanks for merging!

However, with the changed order of upins_t, this logic no longer works:
```
else if ((mpin >= GPIO_PWM1_INV) && (mpin <= GPIO_PWM5_INV)) {
  pwm_inverted[mpin - GPIO_PWM1_INV] = 1;
  mpin -= 5;
}
```
Maybe a more robust solution would be to change to:
```
else if ((mpin >= GPIO_PWM1_INV) && (mpin <= GPIO_PWM5_INV)) {
  pwm_inverted[mpin - GPIO_PWM1_INV] = 1;
  mpin -= GPIO_PWM1_INV - GPIO_PWM1;
}
```

PR #976, opened 2017-10-06
Witty Cloud color command support

KPK75:
RGB PWMs changed to enable color command.


arendst:
In pre-release 5.8.0h

PR #1023, opened 2017-10-16
Support connecting to MQTT brokers without username/password

mortenmathiasen:
By example a default Mosquitto broker installation will setup no requirements for username and password. Actually, the default Mosquitto installation will reject connections that tries to apply username and password as they will be unknown to the broker. This commit allows connecting to such brokers as username and password will not be applied if they are configured as empty strings.


davidelang:
I've always run my systems with no password on mqtt, if that's not working any 
longer, when did it break (I admit it's been a little bit since I've updated by 
devices)

mortenmathiasen:
I don’t know when/if Mosquitto has changed its connection rules. The error occurs for brokers that are strict about checking if username/password is correct, even if empty username/password is allowed by the broker. This check makes sense if different users have different permissions.

arendst:
Adjusted version in next (pre)release.

mortenmathiasen:
I see that you have only committed changes only to Sonoff.ino. But this is not sufficient to connect to a mqtt broker with empty mqtt username and password. Default mqtt username and password will apply from userconfig.h which is not changed:

#ifdef USE_MQTT_TLS
  #define MQTT_HOST            "m20.cloudmqtt.com"  // [MqttHost]
  #define MQTT_FINGERPRINT     "A5 02 FF 13 99 9F 8B 39 8E F1 83 4F 11 23 65 0B 32 36 FC 07"  // [MqttFingerprint]
  #define MQTT_PORT            20123                // [MqttPort] MQTT TLS port
  **#define MQTT_USER            "cloudmqttuser"      // [MqttUser] Mandatory user
  #define MQTT_PASS            "cloudmqttpassword"  // [MqttPassword] Mandatory password**
#else
  #define MQTT_HOST            "domus1"          // [MqttHost]
  #define MQTT_PORT            1883              // [MqttPort] MQTT port (10123 on CloudMQTT)
  **#define MQTT_USER            "DVES_USER"       // [MqttUser] Optional user
  #define MQTT_PASS            "DVES_PASS"       // [MqttPassword] Optional password**
#endif

Default MQTT_USER and MQTT_PASS in userconfig.h need to be set to empty default values.

arendst:
No need to change user_config.h as you can reset both values with commands ``MqttUser 0`` and ``MqttPassword 0``.

arendst:
Implemented in master release v5.9.0

PR #1045, opened 2017-10-21
Support for KMC 70011 Smart Plug

zerog2k:
adds support for KMC 70011 Smart Plug
https://www.amazon.com/KMC-Timing-Monitoring-Network-125V-240V/dp/B06XRX2GTQ


arendst:
In latest (pre)release

PR #1054, opened 2017-10-22
fix some german words

pgollor:
The main change is that `kWh` is not current it is energy.


arendst:
In pre-release 5.8.0n

pgollor:
Okay thanks. But in future it would be nice to merge the commit, because if there is a question anyone knows which person had done the changes.

PR #1233, opened 2017-11-22
DTH sensor: Distinguish "signal already there" from "timeout"

casualoak:
Ported from adafruit/DHT-sensor-library
see https://github.com/adafruit/DHT-sensor-library/pull/82


arendst:
Thnx.

There is more to it as the uint32_t data types will need to be changed to int32_t in order to represent -1.

I'll implement in next (pre)release.

casualoak:
Thanks!

PR #1270, opened 2017-11-28
translate on off status on web root page

reloxx13:
translated:
grafik

grafik


arendst:
No need to implement this as:

- It's fixed in code where it should be fixed in the language file
- There is a command (StateText) to configure On/Off/Toggle and Hold messages. See wiki.

reloxx13:
ahh u r right, did not see this command.

i thought this is the default response from the modul which cannot be changed, thats why i replaced it with the on/off var from the translation file :)

thanks for the hint 👍

PR #1271, opened 2017-11-28
some corrections for the german translation

reloxx13:
i made some corrections for the german translation.

is it ok to change the date format in the beginning?


arendst:
Merged with Development today

PR #1358, opened 2017-12-14
Minor Typo in HTML

reloxx13:


reloxx13:
Heya, i would like to add, that style="width... does also work.

in my line was a missing = , thats why i corrected it.

br
nils

arendst:
I know it worked but this way I gain some code bytes back...

reloxx13:
ahh okay, good to know thank you :)

drbeat:
width attribute on table is deprecated in html5
so best practice would be
`style="width:100%"`

arendst:
Thought so... Will revert the width change in next pre-release.

reloxx13:
hey,

u missed the unit here:

grafik

arendst:
I thought it would default to px but after reading this https://stackoverflow.com/questions/11274904/fallback-for-css-attributes-without-unit you're right so I'll add px to it.

To my surprise it indeed makes a difference and it looks like I envisioned so thanx for the comment!

reloxx13:
you also have everything as % and not as px, i dont know why you made px here instead of 60% and 40% Oo

arendst:
It's px because I want the input field of fixed length. Currently looking at the domoticz part as there it is still wrong too. Might change to % sometime...

PR #1449, opened 2017-12-27
Italian language added

gtortone:
Hi,
in this PR I added support for italian language, but I skip to translate words
used in JSON (MQTT) messages to avoid that changing language can break
HA configuration;

in principle these words (e.g. temperature, humidity, etc.) should be used only in
one language (e.g. English) but translated in Sonoff-Tasmota web interface; a possible
solution could be to define a separate label for this purpose...


arendst:
Agree. I will make different defines for the JSON messages in due time. Until then I will add your translation to the list.

arendst:
See #1449

gtortone:
Hi,
I added new translations keeping JSON label in English, but I realized that some labels 
are still used in JSON message (not in telemetry) and they are not separated by web interface labels.

e.g. D_ABORTED, D_ACTIVE, D_APPENDED, D_BLINK, D_BLINKOFF....

arendst:
You're right regarding D_ABORTED etc.

Considering the current POLL result I guess most defines will stay in English and only the non-JSON ones will be translated.

gtortone:
I realized that "Logging Parameters" web page displays the third option as "Syslog level" but the word
"Syslog" is obtained in webserver.ino (line 856) using D_SYS and "log level"... but in italian this does
not work fine because "log level" (English) translate as "livello di log" (Italian) and concatenation of
D_SYS with "livello di log" is "Syslivello di log" that is a nonsense...

I suppose this is related to Syslog verbosity and in this case D_SYS must be always in English and
its value could be fixed as "Syslog"... if this is the case the line 

```
page.replace(F("{b0"), F(D_SYS));
```

can be replaced with:

```
page.replace(F("{b0"), F(D_SYS " "));
```

arendst:
I'll change D_SYS into D_SYS_LOG_LEVEL, D_WEB into D_WEB_LOG_LEVEL and D_SERIAL into D_SERIAL_LOG_LEVEL an update handling in webserver accordingly. This should solve future translations too at the cost of some code space.

PR #1465, opened 2017-12-29
Feature/Fix Information Build Date

reloxx13:
Build Date is used from `__DATE__` .
This macro is always in the english locale, meaning the month is Dec. 
In Germany it is Dez(ember).

This fixes the Information page:
unbenannt

  
  
Fixed:
pp1


arendst:
Nice one!

Fixed in next (pre)release.

PR #1496, opened 2018-01-03
fix degree symbol

reloxx13:
&deg; did not get exchange in web interface

wemos

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -119,7 +119,7 @@ static const char kMonthNames[] = D_MONTH3LIST;
 
 // webserver.ino
 #ifdef USE_WEBSERVER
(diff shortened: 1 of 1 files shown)


arendst:
Strange. My browsers (Edge, Firefox and Chrome) all show

image

drbeat:
this should be a bug in his browser and not on tasmota side
or even a bug on the 5.10 version of tasmota

should only happens if his browser messes up html parsing

domdomegg:
@reloxx13 

What browser and version of that browser are you using? If you view source, what is the raw HTML as it might have been escaped weirdly - possibly specific to German translation?

reloxx13:
just tested this again with the latest 6.1.1 without the pullrequest changes and the degree char is correct now.

grafik

PR #1503, opened 2018-01-04
Change default GPIO assigment for h801 boards

wouterbaake:
I started using an h801 board and found out that the standard GPIO assignment is a bit strange, as it requires the user to figure out the that FETs for W1 and W2 (as marked on the box) manually need to be assigned in the module configuration. I fixed this in the template and also opened up the possibility of reusing the TX/RX pins.

Assigned PWM4 and PWM5 to GPIO14 and GPIO04 respectively. These are not user configurable on the board as they are hooked up to FETs.

Made GPIO2 and GPIO3 user assignable as is the standard for the TX and RX pins on sonoff boards.


emontnemery:
+1, I agree the update in this PR makes sense!

Discoverability is not so good for these boards..
Maybe it would make sense, in a further update, to let user choose between a few "templates" for custom LED dimmer type device, for example:
- RGB (SetOption15 will be set to 1)
- RGBW (SetOption15 will be set to 1, assigns GPIO14 to PWM4)
- RGBWC (SetOption15 will be set to 1, assigns GPIO14 to PWM4, GPIO04 to PWM5)
- 5 channel dimmer (needs some work, purpose is to allow 5x dimmer control)
- 5 channel PWM (SetOption15 will be set to 0)

namadori:
+1 for this. Also, on the board there is also J3 exposing GPIO0. Do you think it is safe to map that as user configurable?

namadori:
@arendst Any news on merging this? It seems an easy change, I'm waiting for it to add temperature sensors to my two H801.
Thanks 😃

emontnemery:
@namadori No need to wait, set module type to "generic", reboot, and you can then set it up just like the H801 but add the sensors too. Check sonoff_template.h to see how the H801 is mapped.

arendst:
Question as I do not own H801 hardware:

With the changes you made regarding GPIO_TXD to GPIO_USER are you still able to make a serial connection to the H801? I doubt it as the device uses a different GPIO for Serial TxD than all other devices and the software adjusts for it.

So let me know if Serial still works when both TXD and RXD are changed to USER.

renne:
Is GPIO0 mapped as GPIO_USER now?

arendst:
Changed GPIO00 to from GPIO_KEY1 to  GPIO_USER

PR #1518, opened 2018-01-05
Make module configuration web page dynamic

emontnemery:
Make module configuration web page dynamic (show user GPIO for selected module)
Allow setting of Option 15 from module configuration web page
Allow setting module and GPIO at the same time

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -236,6 +236,7 @@ typedef struct MYIO {
 typedef struct MYTMPLT {
   char         name[15];
   myio         gp;
+  uint8_t      option15;
 } mytmplt;
 
 const uint8_t kNiceList[MAXMODULE] PROGMEM = {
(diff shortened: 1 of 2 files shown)


emontnemery:
@arendst Do you think this is too big / unnecessary?
A simpler option to make module setting page less confusing might be to split module configuration page in two: 
1. Page for changing the module, with warning message GPIO settings will be reset
2. One page for setting GPIO + Option 15
Or perhaps keep it as one page and just add the warning message about GPIO + add possibility to set Option 15

arendst:
I've tries to use this PR but my screen stayed blank. I also noted a growth in web page size which is already to high so abandoned it for now.

In essence it's something I always wanted using the display block etc. so if you can get it working with the latest development built I will have a look at it again.

emontnemery:
@arendst OK, I rebased to latest development and verified it works in Chrome and Edge.
Let me know what you think, what could also be done is:
- Either remove the "nicelist", (it's not used, the module list is instead sorted by js) or remove the js sorting
- Clean up HandleModuleConfiguration it became hard to read
- Fix logic to add helpful explanation for each pin (I think the explanation for wemos / generic makes more sense than the default explanation. I think default explanation only applies to standard sonoff devices?)

eazuka:
+1

ascillato2:
Thanks for sharing your ideas. By now, This PR is not going to be merged due to now exists a module configuration in actual Tasmota. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

PR #1561, opened 2018-01-10
Add french translation

OlivierFrancais:
Add french translation.


arendst:
In next (pre)release

calcipher49:
Hi, 
French language is ok, but (in 5.11.1b) still missing from platformio.ini
Need to be added manually before compiling

OlivierFrancais:
I've updated the translation file.
I've added missing lines, correct one typo and improve some translations.

calcipher49:
I have issues with french version due to use of comma as decimal separator. For example this string is not correct json:

> {"Temperature":12,9}
Better to keep dot as decimal separator

arendst:
@calcipher49 fixed in next pre-release

OlivierFrancais:
Sorry I didn't check the usage of this separator.
For me this separator as to be used only for direct display to user. Of course, in JSON, the decimal separator is only the dot. But, in good french, the decimal separator is a comma (and space for thousands separator).

arendst:
@OlivierFrancais No need to check as I am suppose to use the dot for JSON and the language option for other stuff. Currently it will be using the dot overall as I need to rewrite all non JSON values at the cost of code space. So currently the language option will not be used.

PR #1572, opened 2018-01-11
Allow Cross-Origin (removed by ESP Update)

reloxx13:
Default Cross Origin was allowed before 2.4.0

Now its not in the headers anymore (check here: http://www.esp8266.com/viewtopic.php?f=6&t=16679 )

Im working on a Webinterface (https://github.com/reloxx13/SonWEB) to manage all sonoffs in network, i cannot executes commands without this.

coud break other communications too

> 
> Failed to load http://192.168.10.53/cm?cmnd=Power1: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://192.168.10.180' is therefore not allowed access.
> 
> 
> Failed to load http://192.168.10.53/cm?cmnd=Status%202: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://192.168.10.180' is therefore not allowed access.


arendst:
In next (pre)release

PR #1582, opened 2018-01-12
Added template for raw esp8266 modules.

Damme:
Added template for raw esp8266 modules exposing all gpio.


davidelang:
how is this better than the wemos module?

Damme:
How could I have missed that one? *facepalm*

davidelang:
I think it would be great if we could rename 'wemos' to 'generic'

Damme:
I can agree on that :)

PR #1633, opened 2018-01-18
Added a Power Window state

jonlidgard:
I'm using a Sonoff-Pow to monitor the state of a TV. Even if Sonoff power state is on the TV could be switched off at the unit or put into standby. I wanted a way to monitor this state & report it back via MQTT. So I've developed some code to output the window that the the current power drawn is in.
There are PowerWindowLow & PowerWindowHigh commands to set these windows in Watts. (Default 5W & 30W). There is a +/-2W hysteresis in the code around these values.
0 <= Current Power < PowerWindowLow-2  = OFF
PowerWindowLow+2 < Current Power < PowerWindowHigh-2 = STANDBY
PowerWindowHigh+2 < CurrentPower = ON.
Setting PowerWindowLow to 0 results in just OFF or ON states.
PowerWindow is output in the tele/SENSOR message & on change in the stat/RESULT message.
Also displayed on the web page.
I hope you think it is a worthwhile extension and will consider incorporating into the next release.
I've added some basic translations for 'Standby' but not 'Power Window'
Jon Lidgard 18/01/18


arendst:
Investigating...

kintrupf:
Could you please explain what exacly PowerHigh and PowerLow are supposed to do? The list of supported MQTT commands does not say much about them, only that they are some kind of threshold value.
I found out that I need to set them to something other than 0 to get reports about changed power levels with the ENERGY topic, so I set PowerHigh to 1 to get a report whenever my washing machine uses more than 1 watt. Throughout the day I get dozens of reports whenever the machine uses something between 3 and 20 watts for a few seconds in standby. However when I start the machine it draws > 2000 watts and nothing is reported?

ascillato2:
Thanks for sharing your ideas. By now, This PR is not going to be merged due to already exists commands for achieve this. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

Issue #1634, opened 2018-01-18
MQTT_MAX_PACKET_SIZE error when compiling


ascillato:
What about to modify _sonoff.ino_ with the following code so as there is no need to go and modify other libraries?

**sonoff.ino, Line 41:**
```
// Max message size calculated by PubSubClient is (MQTT_MAX_PACKET_SIZE < 5 + 2 + strlen(topic) + plength)
#if (MQTT_MAX_PACKET_SIZE -TOPSZ -7) < MIN_MESSZ  // If the max message size is too small, throw an error at compile time. See PubSubClient.cpp line 359
  #error "MQTT_MAX_PACKET_SIZE is too small in libraries/PubSubClient/src/PubSubClient.h, increase it to at least 1000"
#endif
```
change to:
```
// Max message size calculated by PubSubClient is (MQTT_MAX_PACKET_SIZE < 5 + 2 + strlen(topic) + plength)
#if (MQTT_MAX_PACKET_SIZE -TOPSZ -7) < MIN_MESSZ  // If the max message size is too small, throw a warning at compile time. See PubSubClient.cpp line 359
  #warning "MQTT_MAX_PACKET_SIZE is too small in libraries/PubSubClient/src/PubSubClient.h, increasing it to 1000"
  #undef MQTT_MAX_PACKET_SIZE
  #define MQTT_MAX_PACKET_SIZE 1000
#endif
```

@arendst Testing Tasmota v5.11.1c. It is working very good!!!!

arendst:
Did you test it with Arduino Ide too? I did a year ago and didn't work. Pls try too.

ascillato:
Yes, in Arduino IDE works:

![image](https://user-images.githubusercontent.com/35405447/35120006-802fa3c4-fc75-11e7-8cfe-b79e85a20b00.png)

It works just fine.

arendst:
Pls provide the output of command status 6

ascillato:
@arendst Where I can find the output of command status 6 ?

davidelang:
you can send the command via mqtt, serial, or the web console

ascillato:
Thanks @davidelang 

The output of Status 6 on the Console is:

![image](https://user-images.githubusercontent.com/35405447/35196553-5893a6be-feb2-11e7-8b06-c96dc9028138.png)

![image](https://user-images.githubusercontent.com/35405447/35196554-5dac63ca-feb2-11e7-957a-4721b39bfbff.png)

[...]

PR #1657, opened 2018-01-20
Not edit library PubSubClient.

Sergey-SRG:
That would not edit library PubSubClient for each project.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -40,7 +40,7 @@
 
 // Max message size calculated by PubSubClient is (MQTT_MAX_PACKET_SIZE < 5 + 2 + strlen(topic) + plength)
 #if (MQTT_MAX_PACKET_SIZE -TOPSZ -7) < MIN_MESSZ  // If the max message size is too small, throw an error at compile time. See PubSubClient.cpp line 359
-  #error "MQTT_MAX_PACKET_SIZE is too small in libraries/PubSubClient/src/PubSubClient.h, increase it to at least 1000"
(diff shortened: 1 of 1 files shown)


arendst:
Serge, tried it a year ago. Might work with platformio but pls verify it works  with Arduino ide as that always failed.

Sergey-SRG:
Works perfectly with Arduino ide.

ascillato:
This is exactly as #1634.

The code works on Arduino IDE

Sergey-SRG:
A more practical solution, it is better to accept it #1634

Sergey-SRG:
#1634

ascillato (approved):
As in #1634 and #1657, this approach is good so as not to modify another libraries.

If you want, can be added also a warning to the user like:

```
// Max message size calculated by PubSubClient is (MQTT_MAX_PACKET_SIZE < 5 + 2 + strlen(topic) + plength)
#if (MQTT_MAX_PACKET_SIZE -TOPSZ -7) < MIN_MESSZ  // If the max message size is too small, throw a warning at compile time. See PubSubClient.cpp line 359
  #warning "MQTT_MAX_PACKET_SIZE is too small in libraries/PubSubClient/src/PubSubClient.h, increasing it to 1000"
  #undef MQTT_MAX_PACKET_SIZE
  #define MQTT_MAX_PACKET_SIZE 1000
#endif
```

ascillato:
There is also a small typo error on a comment on this file (_sonoff.ino_) Line 28:

#include "sonoff.h"                         // **Enumaration** used in user_config.h

and should be

#include "sonoff.h"                         // **Enumeration** used in user_config.h

PR #1683, opened 2018-01-23
Use c-string methods in xplg_wemohue.ino to save memory

emontnemery:
Use c-string methods in xplg_wemohue.ino (-1k flash)
Move sensor and extension jump tableto flash (-70 byte RAM)

Note: I have verified the changes xplg_wemohue.ino only by accessing the API through browser (valid JSON, seems fully aligned with before change). Let me know if this functionality can be tested with Domoticz.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -186,10 +186,6 @@ uint8_t light_type = 0;                     // Light types
 
 boolean mdns_begun = false;
 
-uint8_t xdrv_present = 0;                   // Number of drivers found
-boolean (*xdrv_func_ptr[XDRV_MAX])(byte);   // Driver Function Pointers
-uint8_t xsns_present = 0;                   // Number of External Sensors found
-boolean (*xsns_func_ptr[XSNS_MAX])(byte);   // External Sensor Function Pointers for simple implementation of sensors
 char my_version[33];                        // Composed version string
 char my_hostname[33];                       // Composed Wifi hostname
 char mqtt_client[33];                        // Composed MQTT Clientname
--- a/sonoff/xdrv_01_light.ino
+++ b/sonoff/xdrv_01_light.ino
@@ -891,18 +891,18 @@ void LightHsbToRgb()
 
 /********************************************************************************************/
 
-void LightReplaceHsb(String *response)
+void LightGetHsbFixedPoint(uint16_t *hue, uint8_t *sat, uint8_t *bri)
 {
   if (light_subtype > LST_COLDWARM) {
     LightRgbToHsb();
-    response->replace("{h}", String((uint16_t)(65535.0f * light_hue)));
-    response->replace("{s}", String((uint8_t)(254.0f * light_saturation)));
-    response->replace("{b}", String((uint8_t)(254.0f * light_brightness)));
+    *hue = (uint16_t)(65535.0f * light_hue);
+    *sat = (uint8_t)(254.0f * light_saturation);
+    *bri = (uint8_t)(254.0f * light_brightness);
   } else {
-    response->replace("{h}", "0");
-    response->replace("{s}", "0");
-//    response->replace("{b}", String((uint8_t)(2.54f * (float)Settings.light_dimmer)));
-    response->replace("{b}", String((uint8_t)(0.01f * (float)Settings.light_dimmer)));
+    *hue = 0;
+    *sat = 0;
+//    *sat = (uint8_t)(2.54f * (float)Settings.light_dimmer);
+    *sat = (uint8_t)(0.01f * (float)Settings.light_dimmer);
   }
 }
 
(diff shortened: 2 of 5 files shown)


arendst:
I'm trying to get the grasp of your function pointer changes... At least they save a lot of (future) space.

The rest will take time to implement as I want to investigate the consequences too.

Thnx for you many input.

emontnemery:
cool. let me know if I can help with the testing!

arendst:
@emontnemery  What do you think (or is common knowledge) is the preferred way of concatenating a string?

I see you use a construction like:
```
int tmpiter = 0;
tmpiter += snprintf_P(mqtt_data+tmpiter, sizeof(mqtt_data)-tmpiter, "Start text");
tmpiter += snprintf_P(mqtt_data+tmpiter, sizeof(mqtt_data)-tmpiter, " Additional text");
```
as I use:
```
snprintf_P(mqtt_data, sizeof(mqtt_data), "Start text");
snprintf_P(mqtt_data, sizeof(mqtt_data), "%s Additional text", mqtt_data);
```

I see the advantage of mine in that it doesn't use an extra variable (useful as used in my string concatenation between sensors).

Just asking, not judging.

emontnemery:
I didn't give it much thought..
Your version is slightly slower, because the end of the string must be found for each operation, but at the same time much less error prone.
My version is faster, but also error prone. In this case I forget to bounds check the return value from snprintf so my version is quite broken.
Your version is much better in this case, let me know if I should update the PR.

arendst:
No need to update PR. I worked it already.

arendst:
This won't work as the char string will become too long (>893 chars) for more than one device. We'll have to stick to Strings here.

emontnemery:
Oops! But instead of using mqtt_data, a buffer can be allocated on the stack, that's what String has to do anyway. 
```
char tmp[1024];
snprintf_P(tmp, sizeof(tmp), "Start text");
snprintf_P(tmp, sizeof(tmp), "%s Additional text", tmp);```
By the way, snprintf behaviour is undefined when output is same as output buffer, but maybe it works when you use it for concatenation like this?

**Comment by emontnemery** [comment]
@arendst I'll rework the changes to use local variable instead of global variable. Should I do within this PR or close this and open a new one?

**Comment by arendst** [comment]
You may do it here.

**Comment by emontnemery** [comment]
OK, now fixed:
- Works with worst case response length (4 devices, each friendly name 32 characters + NUL)
- Return value from snprintf handled, and response string is correctly truncated

Let me know if I should rebase the branch since there has been changes to `sonoff/xplg_wemohue.ino`

I do:
```
char* wp = cresponse;
char* end = &cresponse[RESPONSE_SIZE];
if (wp < end) wp += snprintf_P(wp, end-wp, "Start text");
if (wp < end) wp += snprintf_P(wp, end-wp, " Additional text");
```
Instead of:
```
snprintf_P(cresponse, sizeof(cresponse), "Start text");
snprintf_P(cresponse, sizeof(cresponse), "%s Additional text", mqtt_data);
```
Because the second method has undefined behavior (http://pubs.opengroup.org/onlinepubs/9699919799/functions/sprintf.html):

> If copying takes place between objects that overlap as a result of a call to sprintf() or snprintf(), the results are undefined.

**Comment by emontnemery** [comment]
I just now realized this style of code is used throughout the project:
```
snprintf_P(cresponse, sizeof(cresponse), "Start text");
snprintf_P(cresponse, sizeof(cresponse), "%s Additional text", cresponse);
```
Maybe it's fine, as long as one is **very** careful to only use it for concatenation?
Should I rewrite the PR to use that style instead if the explicit concatenation with a write pointer?
A suggestion from stack overflow is to write a wrapper for snprintf to do concatenation in safe way without copies of code to keep track of a write pointer and out-of-bounds checks everywhere, that could also be an alternative.

**Comment by arendst** [comment]
I prefer the current use concatenation as used in the project. 

What's the overhead of the Stack Overflow wrapper?

**Comment by emontnemery** [comment]
OK, I'll update the PR to use the sprintf-style concatenation instead!

This is the proposal from SO, I'll test and see if it makes any impact on code size.
(It needs to be updated to make sure bufSize - len is positive though as size_t is unsigned)
```
size_t 
snprintfcat(
    char* buf,
    size_t bufSize,
    char const* fmt,
    ...)
{
    size_t result;
    va_list args;
    size_t len = strnlen( buf, bufSize);

    va_start( args, fmt);
    result = vsnprintf( buf + len, bufSize - len, fmt, args);
    va_end( args);

    return result + len;
}
```

arendst:
This may be a problem with snprintf_P using Progmem data. I suggest to stay using the projects snprintfs

emontnemery:
There is a progmem version of vsnprint: vsnprintf_P.
But anyway, I'll revert to just using snprintf for concatenation since it seems to work fine.

emontnemery:
I just rebased this patch, and it turns out code size is now **lower** when using C++ string methods!
@arendst has there been some changes in the Arduino core to explain this?

arendst:
Don't know.

Issue #1630, opened 2018-01-18
How to fix Home Assistant (Hass) discovery python exceptions


arendst:
Release 5.11.1c provides Hass Mqtt discovery for both switch and light. It works using the parameters in file xdrv_06_home_assistant.ino and I'm impressed with the ease of adding Tasmota devices to Hass this way.

I have two issues:

1) When I look at Hass logging there are several exceptions caused by parameter errors in Color and/or Brightness depending on the type of light (ColdWarm, RGB, RGBCW). While discovery seems to work these errors should not happen. Anyone with knowledge about Hass (and Python) might provide some info here for me to fix.  

2) Hass allows for effect selection (Tasmota Scheme) but I'm unable to get it working. It has something to do with Python String Lists and how Hass discovery wants to receive the Effect list in the JSON message. Again, if you know how to use it please provide the info here or add a PR.

emontnemery:
Related to MQTT discovery: Maybe default MQTT topic should be something more unique than "sonoff"?

arendst:
To address a device you need a unique topic. Users normally come up with bedroom, kitchen etc. I just use a name like sonoff and a numer. The fallback topic is unique anyway.

As the friendlyname will be th Hass name they will probably change that one to bedroom light.

So no need to make things more unique.

emontnemery:
What I'm after is to flash a device, and have it automatically discovered without too many manual steps.
This doesn't work if all devices use topic "sonoff".
How does the "fallback topic" work?

I agree the "Friendly name" does not need to be unique for the reason you mention.

davidelang:
you either need to use the fallback topic, or set the topic to something unique 
for your system.

when you are configuring things one at a time, having a known topic that the new 
device starts out with is great.

when you are mass configuring them, you need to use the fallback topics (and 
know what they will be) to configure them.

personally, I bring them up one at a time after flashing them so that I know for 
sure that the one I just powered up is the one I am working with. I've had too 
many cases over the years of 'autodetection' not identifying the devices 
properly, and even a rare case of "oops, that's not the device I though was 
there" takes so much time to resolve that I'm much happier with the slightly 
more manual process.

emontnemery:
Fallback topic will not be used by tasmota for sending tele information or the hass-style discovery, right?
[...]

PR #1684, opened 2018-01-23
Support unique MQTT topic (configured topic suffixed with chip ID)

emontnemery:
As discussed in #1630. I think it's useful, and it requires only very small code changes.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -192,7 +192,8 @@ uint8_t xsns_present = 0;                   // Number of External Sensors found
 boolean (*xsns_func_ptr[XSNS_MAX])(byte);   // External Sensor Function Pointers for simple implementation of sensors
 char my_version[33];                        // Composed version string
 char my_hostname[33];                       // Composed Wifi hostname
-char mqtt_client[33];                        // Composed MQTT Clientname
+char mqtt_client[33];                       // Composed MQTT Clientname
+char mqtt_topic[33];                        // Composed MQTT topic
 char serial_in_buffer[INPUT_BUFFER_SIZE + 2]; // Receive buffer
 char mqtt_data[MESSZ];                      // MQTT publish buffer and web page ajax buffer
 char log_data[LOGSZ];                       // Logging
@@ -207,6 +208,7 @@ void GetMqttClient(char* output, const char* input, byte size)
   uint8_t digits = 0;
 
   if (strstr(input, "%")) {
+    // Sanitize printf-style format string "input" by allowing at most one format specifier, and replacing it with type 'X'
     strlcpy(output, input, size);
     token = strtok(output, "%");
     if (strstr(input, "%") == input) {
@@ -419,7 +421,7 @@ void MqttPublishPrefixTopic_P(uint8_t prefix, const char* subtopic, boolean reta
     romram[i] = toupper(romram[i]);
   }
   prefix &= 3;
-  GetTopic_P(stopic, prefix, Settings.mqtt_topic, romram);
+  GetTopic_P(stopic, prefix, mqtt_topic, romram);
   MqttPublish(stopic, retained);
 }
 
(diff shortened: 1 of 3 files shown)


davidelang:
Why is the fallback topic not good enough?

emontnemery:
Because it does not work for telemetry and hass discovery.

davidelang:
it can be used for a scripted method to change the main topic.

hass discovery is a reasonable issue, I don't know what it needs.

SGF-Lon:
I would be in favour of this going ahead & being implemented.  Are you suffixing it with the 4 character decimal "characters" or the 6 digit Hex characters?  I am currently using the 4 character decimal "characters" and would like to stay with that naming convention.  Will that be possible with this mod?

emontnemery:
@SGF-Lon I'm using the 6 digit hex characters.
The issue with the 4 digit decimal number (or in reality, 0..8192, since it's a larger number truncated to 13 bits) is that there is a 50% chance for a collision already with 100 devices (birthday paradox). For hostname that doesn't really matter since dhcp-server will handle collisions, but for the mqtt topic there is no arbiter.
With the 6 digit hexadecimal you must reach around 5000 devices to have such high collision risk (and maybe there is no collision risk anyway since the hex number is simply the SoC's serial number).
Having said this, i could update the PR to support both cases?

SGF-Lon:
@emontnemery Good points.  I probably won't have 100 devices within
my home (!) - but somebody might.  So, your approach is best.  If you
can support both by some simple usage of the editing characters (i.e
%04d versus %06X) - or something like similar - that would be
fantastic.  Otherwise, I guess I will end up converting to 6 digit
suffixes.  Thanks.

On Thu, 25 Jan 2018, at 19:28, emontnemery wrote:
> @SGF-Lon[1] I'm using the 6 digit hex characters.
>  The issue with the 4 digit decimal number (or in reality, 0..8192,
>  since it's a larger number truncated to 13 bits) is that there is a
>  50% chance for a collision already with 100 devices (birthday
>  paradox). For hostname that doesn't really matter since dhcp-server
>  will handle collisions, but for the mqtt topic there is no arbiter.>  With the 6 digit hexadecimal you must reach around 5000 devices to
>  have such high collision risk (and maybe there is no collision risk
>  anyway since the hex number is simply the SoC's serial number).>  Having said this, i could update the PR to support both cases?


> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/sgf-lon
  2. https://github.com/arendst/Sonoff-Tasmota/pull/1684#issuecomment-360573519

emontnemery:
@SGF-Lon OK, I updated as you suggest.

emontnemery:
I don't get why there are linker errors due to `strcspn` and `strspn` in the travis build :(

drbeat:
maybe missing <string.h> ?

arendst:
I've managed to find support for strcspn and strspn but version 2.3.0 printf is not compatible either as can bee seen here:
```
topic
15:01:11 CMD: topic
15:01:11 RSL: Group 0, Index 1, Command TOPIC, Data 
15:01:11 MQT: stat/%.*s/RESULT = {"Topic":"wemos6"}
mqttclient
15:02:21 CMD: mqttclient
15:02:21 RSL: Group 0, Index 1, Command MQTTCLIENT, Data 
15:02:21 MQT: stat/%.*s/RESULT = {"MqttClient":"DVES_%06X"}
```
Where is says in 2.4.0
```
topic
15:07:13 CMD: topic
15:07:13 RSL: Group 0, Index 1, Command TOPIC, Data 
15:07:13 MQT: stat/wemos6/RESULT = {"Topic":"wemos6"}
mqttclient
15:10:12 CMD: mqttclient
15:10:12 RSL: Group 0, Index 1, Command MQTTCLIENT, Data 
15:10:12 MQT: stat/wemos6/RESULT = {"MqttClient":"DVES_%06X"}
```
Considering the amount of issues with 2.4.0 I still want support for 2.3.0. 

So without the correct snprintf_P support this will not be integrated in Tasmota.

arendst:
Will be in next pre-release together with #1871

arendst:
It's in 5.12.0b

PR #1685, opened 2018-01-23
Make hass discovery a compile-time option

emontnemery:


arendst:
In next pre-release as default disabled

PR #1705, opened 2018-01-25
Add helper function to avoid duplicated calls to WebServer->arg()

emontnemery:
This saves around 4k code, presumably because duplication of code to handle std::String construction and destruction is avoided.
Note, not possible to do:
```
static const char* getarg(const char*) {
  return WebServer->arg(arg).c_str();
}
```
Because the String returned by WebServer->arg will be destroyed when leaving the helper function.

### DIFF
--- a/sonoff/webserver.ino
+++ b/sonoff/webserver.ino
@@ -314,6 +314,13 @@ uint8_t upload_error = 0;
 uint8_t upload_file_type;
 uint8_t upload_progress_dot_count;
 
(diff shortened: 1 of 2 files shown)


danphamus304:
https://github.com/arendst/Sonoff-Tasmota/commit/39ddce5256524ed01a2b5e148263b9018eb7cafc

arendst:
In next pre-release

PR #1719, opened 2018-01-26
Updated for B1 light: Use Option 21 to unlink power

bensuffolk:
Setting Option21 = 1 will stop the POWER being automatically turned on when a dimmer, colour or CT command is sent.

This is useful when using the wakeup light as you can set the desired target colour or CT without the lighting coming on before issuing the wakeup command.


arendst:
Will do.

arendst:
In next pre-release

PR #1722, opened 2018-01-27
Corrections to Spanish Language File (es-AR.h)

ascillato:
Typo corrections and better choice of words for context.


arendst:
In next pre-release

Issue #1721, opened 2018-01-27
translate error dutch airquality


Stephan296:
At the dutch webinterface of the Sonoff SC shows at airquality: Lucht kwalitiet instead of Lucht kwaliteit

arendst:
Fixed in 5.11.1g just released

PR #1723, opened 2018-01-27
Language error in Dutch (airquality) #1721

ascillato:
At the dutch webinterface of the Sonoff SC shows at airquality: Lucht kwalitiet instead of Lucht kwaliteit. (#1721)


arendst:
In next pre-release

PR #1726, opened 2018-01-27
Language Support for Title (ES, IT, FR)

ascillato:
Added language support on the web page title for spanish, french and italian.

Tested Ok.

if you use spanish, french or italian the title will be displayed [MODULE] [NAME]
if you use others the title will be displayed [NAME] [MODULE]


arendst:
Doesn't work for platformio so I had to use the solution I came up with.

ascillato:
Ok, I understand. So, this is like a workaround until Platform.io has support to **#if** ?

So, what about this name change:

`#ifdef LANGUAGE_MODULE_NAME`

for

`#ifdef LATIN_LANGUAGES`

and also _LATIN_LANGUAGES_ should be on _USER_CONFIG.H_ under _localization_.

Closing.

Issue #1587, opened 2018-01-13
Security issue ? wireless information not protected


matsube11:
Sonoff-Tasmota: 5.11.1 20180107
Sonoff wifi

Hi,

I realized today while I was rebooting my wireless access point that my Sonoff device once disconnected was showing the wifi parameters page without requiring any password. Even though I setup a web admin password, password is only requested if changes are to be made on the page.

So it's trivial to get the wifi SSID and password information from someone using a sonoff tasmota, just disconnect it from wifi, connect to it and that's it.

Am I missing something here ???

Thanks

davidelang:
change your wifi mode so that it doesn't present an AP if it fails to connect to 
one.

matsube11:
Thanks, changed successfully using WifiConfig 4

I'm new to Tasmota, this is amazing piece of software but this default behavior is a critical security breach imho.
Wonder how many people have their wifi crendentials open for anyone to access due to that :/

ascillato:
Hi,

I test this issue and this allows you to change the Wifi configuration so you can take yours neighbour's sonoff and connect it to your network. In that initial page you only can see the SSID but not the password.

After that, you can not control it. The Sonoff ask you for the user and password that you previously configure.

So, this is just an anoying issue that can take your sonoff from your network.

I consider that a possible solution is to automatically change the WifiConfig variable to WIFI RECONNECT after the sonoff connect to a network. Then you should be able to configure the WifiConfig on the Webserver Menus.

I can take a look on doing this if this solution is considered ok.

davidelang:
and what if someone wants their device set this way? how to they avoid your "I 
know what's best for you" feature?

ascillato:
The logic I follow was:

1- On the first time you power up the sonoff, it shows as AP so you can set up your WIFI with your phone
2- Sonoff connects to your Wifi
3- Now you can set up the features of your sonoff over your wifi.

From step 2 to 3, Sonoff can change the
    WIFI_CONFIG_TOOL: WIFI_MANAGER
to
    WIFI_CONFIG_TOOL: WIFI_RETRY

Because now your Sonoff is attached to your Network. Each time it get disconnected from your router, will reconnect to it. But, if you want that each time the sonoff get disconnected from your wifi the sonoff shows itself as AP,, you should be able to configure that from the EXTRA CONFIG menu of the webserver.

What do you think?

[...]

PR #1727, opened 2018-01-27
Solved Security Issue (#1587) on webserver.ino

ascillato:
Solved issue for Tasmota 5.11.1g that sends to web browser the MQTT password. #1587 can be closed

tested ok using Arduino IDE and NodeMCU


ascillato:
@arendst 

What do you think of this little change, so the MQTT password is not sent to the browser, increasing security?

PR #1730, opened 2018-01-28
Fixed size log buffer

emontnemery:
Instead of array of dynamically sized strings, allocate fixed chunk of memory for web log.
Main purpose is to have less dynamic memory usage.
Instead of a fixed number of retained log messages, as many log messages as can fit in the chunk of memory are kept (so more messages if they are short, and fewer if they are long).

Also saves some memory, because the array of strings has an overhead of at least 20 byte / entry for the String objects.

The implementation is very simplistic and not very efficient, but there is not a lot of logging going on so should be OK.


arendst:
Testing. Looks promising! Now even command ``cfgdump 0 30`` keeps some useful lines in the buffer.

emontnemery:
@arendst I turned on some more warnings (-Wall) when debugging, and I was anyway modifying around there :)

arendst:
In next pre-release

PR #1774, opened 2018-02-03
Fixed a couple of bugs and many warnings

projectkk2glider:
that were found with `-Wextra` compiler flags


arendst:
In next pre-release

PR #1836, opened 2018-02-09
Fix pointer on button CSS

domdomegg:
Tiniest change, was just really bugging me that when you hover over the buttons the cursor icon doesn't change - to me it makes it feel less snappy for some reason.

Having gone through all the menus I couldn't find somewhere where changing the cursor hover behaviour could cause any issue.


arendst:
In next releases

PR #1871, opened 2018-02-11
Ability to insert chip ID in update URL and MQTT client name

Leseratte10:
Tasmota supports the placeholder "%06X" for the MQTT client name which inserts the chip ID in hexadecimal - unfortunately, it does not support other placeholders. I have added "%d" for the chip ID in decimal, and "%04d" for the short chip ID (same as used in the hostname). This is because I identify my sonoff devices using the 4-digit hostname number and also have that in the MQTT client name. 

In addition, I added support for placeholders to the OTA update URL. If the update URL contains "%d" (or "%04d"), that will be replaced by the short chip ID (chipID & 0x1fff). If it contains "%u", that will be replaced by the (long) chip ID (both in decimal).

This means, one could now compile one firmware, flash that to multiple Sonoffs, and they'd have different MQTT client names including the ID (in different variants, not just 6-digit hex), and, if the OTA url in the firmware is set correctly (containing a placeholder), the server which you load the update from knows which device requested the update - great for deploying a new testing firmware to just one device.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -215,11 +215,19 @@ void GetMqttClient(char* output, const char* input, byte size)
       token = strtok(NULL, "");
     }
     if (token != NULL) {
-      digits = atoi(token);
-      if (digits) {
-        snprintf_P(output, size, PSTR("%s%c0%dX"), output, '%', digits);
-        snprintf_P(output, size, output, ESP.getChipId());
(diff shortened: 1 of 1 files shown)


arendst:
Had that a year ago in client id and then people start using strings where integers were expected and they received exceptions everywhere. So I had to revert to a very strict way of doing things.

Looking at your implementation it might wotk though. Investigating.

PR #1884, opened 2018-02-12
Add RfKeyX 5 command which shows the learnt RF code

ErikAndren:
Small change that lets the user show currently learned rf keys by adding an additional RfKeyX argument.


arendst:
In next pre-release

PR #1927, opened 2018-02-15
handle preflight requests

WoHinDu:
Currently, I'm writing a little web programme to make the batch updating from multiple tasmota devices easier. 

My preferred method to update the devices is to send the firmware directly to the device via a POST request. That is basically the same what the original built-in updater is doing. The problem is that a browser sends before the actual POST request a so-called preflight request (an OPTIONS request). The browser asks the web server whether the request is allowed or not. Currently, the preflight requests are not handled by tasmota and therefore the response is an error 404. 

This pull request implements the correct handling of preflight requests for the firmware update page.

### DIFF
--- a/sonoff/webserver.ino
+++ b/sonoff/webserver.ino
@@ -347,6 +347,7 @@ void StartWebserver(int type, IPAddress ipweb)
       WebServer->on("/up", HandleUpgradeFirmware);
       WebServer->on("/u1", HandleUpgradeFirmwareStart);  // OTA
       WebServer->on("/u2", HTTP_POST, HandleUploadDone, HandleUploadLoop);
+      WebServer->on("/u2", HTTP_OPTIONS, HandlePreflightRequest);
       WebServer->on("/cm", HandleHttpCommand);
       WebServer->on("/cs", HandleConsole);
       WebServer->on("/ax", HandleAjaxConsoleRefresh);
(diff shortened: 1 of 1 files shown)


ascillato:
Hi

There is already a Web based application for uploading firmware to several sonoff at the same time. 

https://github.com/reloxx13/SonWEB

WoHinDu:
Hi,

I know SonWeb. That's great work. The problem is that the Tasmota devices need either direct access to the internet or you need to set up a web server on your local machine. My devices are not allowed to connect directly to the internet due to security concerns. Setting up a web server is not trivial and for a lot of people too much effort for just updating a few devices. Of course, that is not a problem of SonWeb but in general of the update mechanism. 

In my opinion, the easiest way to send the firmware directly to the devices is to handle the preflight requests. That are just a few lines of code and I can't think of any problems.

ascillato:
Hi, I understand. So, just few lines on _webserver.ino_ and https://github.com/WoHinDu/Tasmota-Firmware-Updater can be used, right?

That is a nice python solution. I can think that this approach is also very useful for an android app to update Tasmota Firmwares directly.

Like your approach as another update option. :+1: 

Hope to be included in Tasmota.

BTW, Installing SonWEB on windows is just unziping the provided ready to use **SonWEB-XAMP-Portable-2018-01-20.zip** on **c:\XAMPP** and run once **setup_xampp.bat** and then run **xampp-control.exe**
Then on a web browser just type **localhost** as address and you have it.

ascillato (approved):

ascillato:
Hi @WoHinDu,

I have tested your code, works very good. Also I tested that did not interfere with SonWEB, neither with local web update.

So, hope that @arendst include this feature on Tasmota.

Thanks

arendst:
Will be in next pre-release

WoHinDu:
@ascillato @arendst Thanks a lot for reviewing this code so fast!

Issue #1842, opened 2018-02-09
Runtime wrong


reloxx13:
Hello, its me agian x.x

The new runtime as seconds would be great, but the time is so much wrong counted, its useless :(

From 24h real uptime only 7-10h are counted.

30mins real are ~7mins in Tasmota.

within 10sec only 3-4sec are counted.

arendst:
That's probably because you use sleep. I don't use sleep and am very happy. I never understood why anyone wanted sleep. It's your choice.

reloxx13:
heya,

since you are form netherland, ill post this in german. tell me if you dont understand this:


> Welche auswirkungen hat der in einer Festen Installation, habe aktuell 9 x Tasmota Sonoffs mit Iobroker und 3 x Echo Dot am laufen, und es werden noch ein paar mehr werden, an der hälfte sind aktuell über koppelrelais noch Taster Schaltungen dran, um konventionell das Licht einzuschalten, Frage wäre, wie schnell reagieren die Geräte bzw. weckt Alexa die Geräte über den iobroker aus dem Sleep auf?
> 
> In meiner Zukunftsplaung sind ca. 30 Tasmota Geräte geplant und der Stromverbrauch bzw. die Leistung ist ja auch ein Kostenfaktor, somit hätte ich dann 30 x 1,1 Watt = 33 Watt x 24 Std x 365 Tage = ca. 289 KWh im Jahr Stromverbrauch was bei ca. 0,26 Cent die KWh kosten von ca. 75 Euro im Jahr wären , im Sleepmodus wären es ja bei 30 x 0,4 Watt = 12 Watt x 24 Std x 365 Tage = ca. 105 KWh im Jahr Stromverbrauch was bei ca. 0,26 Cent die KWh kosten von ca. 27 Euro im Jahr wären!
> 
> Denke das sollte man nicht vernachlässigen und wenn durch den Sleepmodus keine Nachteile in der Performance entstehen wäre dies nicht nur bei Akkugestützen Systemen Interessant!

reloxx13:
in short:
without sleep: 1,1watt = ~ 289KWh/year = ~75€
with sleep: 0,4watt = 105KWh/year = 27€

but yep, I thought it will be the sleep and/or other loop lag.

reloxx13:
how about saving the start date and additional subtract it from current time.

but only the saved starttime would be enough for other platforms to get the calculate the real uptime themselfes.

arendst:
[...]

PR #1938, opened 2018-02-16
Feature/remember startup datetime

reloxx13:
According to  issue #1842 i added code for saving the startup datetime as utc of the device after the first successfull ntp response.

available in StatusPRM.StartupDateTimeUtc


reloxx13:
oh well, i need to fix my git x)

PR #1940, opened 2018-02-17
Add Check for lwIP version

ascillato:
Add Check for lwIP version

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -35,6 +35,7 @@
 #include "i18n.h"                           // Language support configured by user_config.h
 #include "sonoff_template.h"                // Hardware configuration
 #include "sonoff_post.h"                    // Configuration overrides for all previous includes
+#include "lwip/init.h"
 
 // Libraries
 #include <PubSubClient.h>                   // MQTT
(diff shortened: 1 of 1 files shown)


reaper7:
I am using a esp8266 arduino git version, in this case I do not get any error 
because _core_version.h_ contains **ARDUINO_ESP8266_GIT_VER**
maybe for users like me it would be good to add something similar

arendst:
No need as the problems are solved in the git version where I suppose you talk about the stage version.

ascillato:
I tested Tasmota with the new **lwIP 2.4.1 release** and it works with _no problem_ :+1: . I use _lwIP v2 lower memory_ and the firmware is a bit smaller.

@arendst have you tested it?

arendst:
Thnx for testing. Will do soon too.

arendst:
Just provided version 2.4.1 boards.txt and platform.txt specific updates for tasmota.

KPK75:
I did some tests with 2.4.1 and regardless of lwip it just hungs after after some time ( under minute ). Maybe this is the cause : [https://github.com/esp8266/Arduino/issues/4497](https://github.com/esp8266/Arduino/issues/4497)

ascillato:
Interesting. I have my device working for hours without problem. Which
tasmota version did you use and with which options?

So as to try to reproduce that bug.

Also how did you make the update? Serial / OTA?
I used the serial with the arduino option to erase all the flash (just to
try it).

Did you build the firmware? Arduino / platform.io?

KPK75:
ESP-01 devices. Latest dev version. First OTA then serial. All the same. No bugs. Just hungs - stops working. Do you use webserver ?

ascillato:
Yes. With webserver. But on a nodemcu

KPK75:
After downgrade to 2.4.0 (lwip 1.4) it is working fine.

KPK75:
Found it! 2.4.1 and Sleep >0 == hung.
[https://github.com/esp8266/Arduino/issues/4485](https://github.com/esp8266/Arduino/issues/4485)

ascillato:
Great. I was not using sleep. Thats why. Very good job. Thanks

PR #1998, opened 2018-02-23
Add OTA "push" method

soif:
This PR implements the ability to receive OTA updates. This allows the "Two Steps Update" method as successfully implemented by ESPEasy or Espurna ([learn more](https://github.com/xoseperez/espurna/wiki/TwoStepUpdates)). 

When using this method : 
- (1) one can first upload a very small special Uploader Firmare that retains the wifi/network settings (I've successfully compiled one [there](https://github.com/soif/EspBuddy/blob/master/firmwares/TasmotaUploader.OTA-0x20161209.bin), with sources [here](https://github.com/soif/EspBuddy/tree/master/sketches/TasmotaUploader.OTA))
- (2) then Upload a bigger firmware  (up to near 700k). :-)

These 2 steps can be performed from the web interface (very simple for most users) OR directly in platformio or espota.py that are able to make direct OTA updates. 

I've even build a nice [EspBuddy](https://github.com/soif/EspBuddy) wrapper script that make use of this to allow (among many others features) to  update OTA one (or more in batch) ESP firmware in a single simple command.... 👍 

BTW As I've added this as a _#ifdef USE_ARDUINO_OTA_, it should not add a single octet of memory is not activated, but when used it paves the way to be able to *easely* release larger firmwares ...

HTH

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -65,14 +65,11 @@
 #endif  // USE_SPI
 
 
-
 #ifdef USE_ARDUINO_OTA
 #include <ArduinoOTA.h>
 #include <ESP8266mDNS.h>
 bool ArduinoOTAtriggered=false;
 #endif
-
-
 
 // Structs
 #include "settings.h"
@@ -2817,6 +2814,7 @@ void setup()
   RtcInit();
   XsnsCall(FUNC_INIT);
 
+
 #ifdef USE_ARDUINO_OTA
     ArduinoOTAInit();
 #endif
@@ -2865,12 +2863,10 @@ void loop()
 
 
 
-
-
 #ifdef USE_ARDUINO_OTA
 /********************************************************************************************\
   Allow updating via the Arduino OTA-protocol. (this allows you to upload directly from platformio)
-  \*********************************************************************************************/
+\*********************************************************************************************/
 
 void ArduinoOTAInit(){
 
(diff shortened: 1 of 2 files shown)


reloxx13:
it should work already with this: https://github.com/arendst/Sonoff-Tasmota/pull/1927 ???

soif:
This is not the same method at all ! 

#1927 just fixes the ability to send the POST request directly to the webserver, 
This is fine but you are still limited to the near 502k flash limit, and you can NOT use platformio nor espota.py to do a direct upload (really convenient when developing).

My approach directly [inspired](https://github.com/xoseperez/espurna/wiki/TwoStepUpdates) by Espurna :
- use the standard OTA method supported by platformio, espota.py, Arduino IDE, and certainly many others tools.
- Does not eat any memory at all if not activated by the define (so it doesn't hurt at all if regular users don't activate it).
- Eat 11.5k of precious flash memory **BUT in return** allow you to push the limit for new firmware to about **720k** instead of the actual 502k : _This is a fresh new 218k more flash memory available to use_ :-)

BTW The intermediate firmware could certainly also contains the #1927 patch to also accept direct POST, if needed.

Wouldn't you like to get half more flash memory to add new features in Sonoff-Tasmota  ? :-p :-)

arendst:
The flash memory size has already been solved by providing the minimal version to be loaded first as documented here https://github.com/arendst/Sonoff-Tasmota/wiki/What's-New with regard to Ota-Magic.

Using the Tasmota provided platformio and arduino scripts AND a local webserver allow for two-step upgrades initiated from the devices either using the group topic for mass update or using a specific topic to update only one device.

The goal of tasmota is not having to edit user_config AT ALL so enabling extra functionality for Arduino OTA is an extra step using, as I understand, additional 11.5k 

For now I see no real advantage but I'll look in to it.

soif:
I saw the OtaMagic thing and yes this is sort of the same process : upload a small firmware, then a second larger.

The only small issue with this method is that it NEEDS an external webserver to automate the process  : not a problem for me personally (I'm a sysadmin), but really not easy for regular users. 

On the other hand, take a look at my [EspBuddy](https://github.com/soif/EspBuddy) and experiment how easy it is to update **dozen of ESP babies**, with different firmware (Espurna, EspEasy, Tosmota), while keeping backup of all your firmwares or copies of each ESP settings, in one central place, with just ONE command ! 


>The goal of tasmota is not having to edit user_config AT ALL

This why *user_config_override* exists, no?
I personally have edited A LOT of settings to fit my particular needs (SSID+Key, own syslog server, fixed IPs + Class B network, own time server...).

>so enabling extra functionality for Arduino OTA is an extra step 

_You might certainly "#define USE_ARDUINO_OTA" as default if you like the feature_ 👍 

Anyway Thanks for you time and this great firmware, 🍻

soif:
any hope to get this merged ?

arendst:
Give it a try...

soif:
Seems to work fine! 
Thanks for including it 🍻  
and amazed that you've succeeded to reduce it to 4.5k only 😮 

-----

_BTW I don't know if it is related to this commit (I guess no) but defining a SYS_LOG_HOST now results in an endless reboot loop. I've opened a separate issue #2227 about that._

PR #2005, opened 2018-02-23
Create pl-PL-d.h

RTurala:
Localization for Polish with diacritics


arendst:
Can this replace the current pl-PL.h file as I think one language file should be enough?

RTurala:
In my opinion it can replace original file. I haven't noticed increased memory usage with this translation.

PR #2091, opened 2018-03-06
Display hostname in header. Closes #1006

### LINKED ISSUE #1006: Feature request:  Hostname on startup webpage
I would like to see the Hostname on the first webpage seen when you connect to the device.

Like This:

main_menu_-_2017-10-10_17 51 07

jonatanolofsson:
When accessing host by ip, it is unclear which host you're at. This PR aims to add the hostname to the head of the page. See also issue #1006

### DIFF
--- a/sonoff/webserver.ino
+++ b/sonoff/webserver.ino
@@ -99,9 +99,9 @@ const char HTTP_HEAD_STYLE[] PROGMEM =
   "<div style='text-align:center;color:red;'><h3>" D_MINIMAL_FIRMWARE_PLEASE_UPGRADE "</h3></div>"
 #endif
 #ifdef LANGUAGE_MODULE_NAME
-  "<div style='text-align:center;'><h3>" D_MODULE " {ha</h3><h2>{h}</h2></div>";
+  "<div style='text-align:center;'><h3>" D_MODULE " {ha</h3><h2>{h}</h2><h3>{hn.local ({ip)</h3></div>";
 #else
-  "<div style='text-align:center;'><h3>{ha " D_MODULE "</h3><h2>{h}</h2></div>";
+  "<div style='text-align:center;'><h3>{ha " D_MODULE "</h3><h2>{h}</h2><h3>{hn.local ({ip)</h3></div>";
 #endif
 const char HTTP_SCRIPT_CONSOL[] PROGMEM =
   "var sn=0;"                    // Scroll position
@@ -442,6 +442,8 @@ void ShowPage(String &page, bool auth)
   }
 
   page.replace(F("{ha"), my_module.name);
+  page.replace(F("{hn"), my_hostname);
+  page.replace(F("{ip"), WiFi.localIP().toString());
   page.replace(F("{h}"), Settings.friendlyname[0]);
   if (HTTP_MANAGER == webserver_state) {
     if (WifiConfigCounter()) {


sonoff/webserver.ino
```diff
@@ -841,7 +842,6 @@ void HandleWifi(boolean scan)
   }
 
   page += FPSTR(HTTP_FORM_WIFI);
-  page.replace(F("{h1"), Settings.hostname);
```
ascillato: Should be there

sonoff/webserver.ino
```diff
@@ -442,6 +442,7 @@ void ShowPage(String &page, bool auth)
   }
 
   page.replace(F("{ha"), my_module.name);
+  page.replace(F("{h1"), Settings.hostname);
```
ascillato: should be change to
`page.replace(F("{h1"), my_hostname);`

ascillato (changes requested):
As explained on #1006, using `Settings.hostname` does not work as intended. Should be changed to `my_hostname`

ascillato:
Also, if you like, can be added the IP address and all the new line can be like:

_**SOnOff_123.local (192.168.1.101)**_

All this info can be reached also on Information Menu.

jonatanolofsson:
Thank you @ascillato ,
I updated the PR accordingly. Good idea with the ip, I added that as well.

sonoff/webserver.ino
```diff
@@ -841,7 +843,6 @@ void HandleWifi(boolean scan)
   }
 
   page += FPSTR(HTTP_FORM_WIFI);
-  page.replace(F("{h1"), Settings.hostname);
   page.replace(F("{s1"), Settings.sta_ssid[0]);
```
ascillato: It is missing the line:

`page.replace(F("{h1"), Settings.hostname);`

Should be there because `Settings.hostname` is different than `my_hostname`.

sonoff/webserver.ino
```diff
@@ -99,9 +99,9 @@ const char HTTP_HEAD_STYLE[] PROGMEM =
   "<div style='text-align:center;color:red;'><h3>" D_MINIMAL_FIRMWARE_PLEASE_UPGRADE "</h3></div>"
 #endif
 #ifdef LANGUAGE_MODULE_NAME
-  "<div style='text-align:center;'><h3>" D_MODULE " {ha</h3><h2>{h}</h2></div>";
+  "<div style='text-align:center;'><h3>" D_MODULE " {ha</h3><h2>{h}</h2><h3>{h1.local ({ip)</h3></div>";
```
ascillato: **{h1** should be **{h4** so as not making conflict with WIFI Settings Menu

sonoff/webserver.ino
```diff
@@ -99,9 +99,9 @@ const char HTTP_HEAD_STYLE[] PROGMEM =
...
 #else
-  "<div style='text-align:center;'><h3>{ha " D_MODULE "</h3><h2>{h}</h2></div>";
+  "<div style='text-align:center;'><h3>{ha " D_MODULE "</h3><h2>{h}</h2><h3>{h1.local ({ip)</h3></div>";
```
ascillato: **{h1** should be **{h4** so as not making conflict with WIFI Settings Menu

sonoff/webserver.ino
```diff
@@ -442,6 +442,8 @@ void ShowPage(String &page, bool auth)
   }
 
   page.replace(F("{ha"), my_module.name);
+  page.replace(F("{h1"), my_hostname);
```
ascillato: **{h1** should be **{h4** so as not making conflict with WIFI Settings Menu

ascillato (changes requested):

sonoff/webserver.ino
jonatanolofsson: Ah, right, I misunderstood, my bad

ascillato:
Tested the @jonatanolofsson approach with the review changes and work.

image

ascillato (approved):
Tested and works very nice :+1:

jonatanolofsson:
Awesome! Thanks a lot for the testing!

arendst:
Won't implement as it clutters the display with redundant information to be found on the information page. At the time you see this page you already found the correct name/ip address.

jonatanolofsson:
I agree. In my case I often connect by IP, but it's difficult to be sure I
got the right one (of my 35 devices). This is such a simple fix, I find it
silly it's being rejected tbh

Geitde:
This may work when you have a single switch or whatever. When you have a dual, quad, octo switch the logic would be to use the host name as it represents the entire device, what the webpage represents.

Currently it shows the name of  one of the devices switches, which is in worst case only 1/8 of the device. So instead of reading "KitchenControl" on the webpage you get "PowerSocket5" or something. Then you need to remember which socket that is and in which room. Even if it is named like a kitchen device this is kind of stupid, when there is a customizable name describing the entire device by name.

It like "We take the ford" compared to "We take car 2". Which sentence would you prefer as more logic?

Frogmore42:
@meingraham I am not sure if your comment was directed at mine or something else.  But, I am confused.  This PR is about adding the hostname to the UI.  I agree that using names is much nicer.  I had never tried using multiple relays before.  I now see that the friendly name is really used by Hue/Belkin emulation.  I don't use that (as I use HA-Bridge for control from Alexa).  I believe it is confusing to use the friendly name for the first relay for part of the hostname and also display it at the top of every page when you use multiple relays.  My guess is that Tasmota was originally designed with a single relay in mind.

I don't find the information about the type of device I have it configured as is at all important.  The name of it is important.  It seems Friendly Name 1 is the first part of the hostname, so displaying Friendly Name 1 does make some sense.  However, there are multiple names and that can be confusing.  On my network, I have to use the IP address to get to the web UI.  Once I am there, if I really care I can see the ip address in the address bar of my browser, so I don't need or want to see it in big type on the UI.  I do want to see the topic name I have used, since that is how I usually refer to it from MQTT.  I would really like a way to set the name once and have it be used for both topic and displayed on the UI.

With multiple relays/switches it gets more complicated.  Perhaps each relay/switch/button pair should have a different topic name, but that would be confusing for STATE of the device and for SENSOR readings.  What topic would they use.

ascillato2:
Thanks for sharing your ideas. By now, This PR is not going to be merged due to be outside the actual scope of Tasmota. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

PR #2130, opened 2018-03-10
Module Title in a Multi-language way for web pages

ascillato:
With this PR, it is proposed to change the way that the title of the module is expressed on the web pages.

At this moment is: 

  **[module type] module**

So, as example, for Generic was **Generic Module** but in Spanish (_Generic Módulo_), Italian (_Generic Modulo_) and French (_Generic Module_), the words are in wrong order.

So, the idea is to be:

   **Module: [module type]**

So, for example, Generic module title should be: **Module: Generic**


ascillato:
No, because that PR was rejected due to platform.io compatibility and the approach that is in the actual version is not automatic. So, thats why this other approach of using the same module title structure for all languages in the way of **Module: Generic**, **Module: Sonoff Basic**, etc

In that way, is correct for all languages.

ascillato:
What do you think @arendst ?

Is Ok for you this approach for the title to be **Module: [module_name]** ?

This should be ok for all languages

PR #2190, opened 2018-03-18
MQTT - Serial bridge

danielztolnai:
Software serial - MQTT bridge implemented as an xdrv. Useful when interfacing with devices having only a serial port for communication like for example projectors and older AV receivers.

Quick info and usage:
 - Configure pins "SerBr Rx" and "SerBr Tx"
 - Set the baud rate through cmnd/sonoff/SBrBaud (Can be anything from 1200 to 115200)
 - Send data through cmnd/sonoff/serialbr
 - Received data is published to stat/sonoff/SERIALBR
 - Delimiter functionality can be enabled by sending the ASCII code of the delimiter to cmnd/sonoff/SBrDelim
 - Delimiter functionality can be disabled by sending 0xff to cmnd/sonoff/SBrDelim
 - Added support for whitespace characters. It is possible to transmit for example a newline character by putting its escaped version ("\n") in the MQTT message

### DIFF
--- a/lib/TasmotaSerial-1.1.0/src/TasmotaSerial.cpp
+++ b/lib/TasmotaSerial-1.1.0/src/TasmotaSerial.cpp
@@ -111,6 +111,7 @@ bool TasmotaSerial::isValidGPIOpin(int pin)
 bool TasmotaSerial::begin(long speed) {
   // Use getCycleCount() loop to get as exact timing as possible
   m_bit_time = ESP.getCpuFreqMHz() *1000000 /speed;
+  m_high_speed = speed > 9600;
   return m_valid && (speed <= TM_SERIAL_BAUDRATE);
 }
 
@@ -149,7 +150,7 @@ int TasmotaSerial::available()
 }
 
 #ifdef TM_SERIAL_USE_IRAM
-#define TM_SERIAL_WAIT { while (ESP.getCycleCount()-start < wait) optimistic_yield(1); wait += m_bit_time; }  // Watchdog timeouts
+#define TM_SERIAL_WAIT { while (ESP.getCycleCount()-start < wait) if (!m_high_speed) optimistic_yield(1); wait += m_bit_time; } // Watchdog timeouts
 #else
 #define TM_SERIAL_WAIT { while (ESP.getCycleCount()-start < wait); wait += m_bit_time; }
 #endif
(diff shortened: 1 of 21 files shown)


arendst:
You're 8 hours too late. I have just released this functionality. ;-)

danielztolnai:
@arendst Your implementation is using the hardware serial. In a software serial implementation you can choose pins, and you will not have the ESP's reset messages on the serial line. Correct me if I've missed something or said something wrong.

arendst:
You are right but what currently uses tasmotaserial I see exceptions randomly so I choose the hardware option this time. I'll look into your implementation too.

danielztolnai:
@arendst I've been testing this (using 38400 bps) for around 48 hours now and it seems to work (for now). I hope it will work for you as well. :-) Above the advantages I have mentioned before, this also has the comfort of working on devices that use the hardware serial for something else, like the older Sonoff Dual.

arendst:
Convinced! I will add to next pre-release. Well done.

arendst:
Implemented.

danielztolnai:
@arendst Please excuse my bluntness, but the implementation in my pull request is still superior to yours.

1. You've disabled sending raw binary data. MQTT itself supports this, and it works well in my implementation.
2. You've also disabled using any binary byte as the delimiter.
3. You've wrapped the data received from the serial port in JSON. I understand that all Tasmota telemetry and status data uses JSON, but in the case of a serial port, it brings more problems to the table than advantages.
4. The code that you use for MQTT input preprocessing removes all whitespace characters using the isspace() function but you have disabled the "\s" escaped space character in the Unescape() function. This makes it impossible to send a space character through the serial port.
5. My implementation has the ability to disable unneeded parts of the code to save flash and memory space.

**Points 1., 2. and 3. make it impossible to use your implementation as a transparent serial bridge**, which is required in multiple serial port use cases. (Where you have to transfer raw binary data)

I am not pushing you to use my implementation as it is, but please fix these issues (1. to 4.) as having a transparent serial bridge has more opportunities than having a text only one. (Also, the transparent implementation has no drawbacks compared to the text only one.)

aidanmountford:
Thanks guys.

I was literally looking at writing this last week ...  I will run it up over the weekend and see how I go...   

A

PR #2202, opened 2018-03-20
Added support for Zengee WF017 LED Strip controller

aidanmountford:
Added Support for Zenggee ZJ-WF017-A (ESP12S))

https://www.ebay.com/p/Smartphone-Android-IOS-WiFi-Music-Controller-for-RGB-5050-3528-LED-Strip-Light/534446632?_trksid=p2047675.l2644

This pull request adds patch to force it to be recognised as RGB light by default.


arendst:
I'll add Zengge and will also change default pwm_control (SetOption15) from 0 to 1 for newly flashed devices providing a better initial PWM experience.

Now that command Channel is available the need for SetOption15 0 is also a lot less.

PR #2211, opened 2018-03-21
Feature/hsb features

aidanmountford:
This completes the channel handling in v5.12h by:

   *  Returning the channel information in the status
   *  Returning the channel information (in status and result) as an integer rather than a string.


arendst:
Did it differently using a channel array shortening the string considerably  ;-)

Thnx for updates.

aidanmountford:
No worries.

I am sure I will cook up some more as time progresses :) 

Sorry it took two shots to get it right (no idea why I didnt put the status in there in the first place)

A

PR #2228, opened 2018-03-22
Really ignore user_config_override.h and make it optionnal

soif:
while _user_config_override.h_ is correctly in .git_ignore, it also exists in the repo. 
This lead this file to be overwritten when cleanly switching git branches, this is thus far from ideal.

This PR : 
- makes _user_config_override.h_ completely optional with the USE_CONFIG_OVER flag. 
- remove the _user_config_override.h_  file to fix the git issue
- provide some more examples in the sample config
- shows a compiler warning to confirm that the user_config_override.h is included
 
I've splitted all changes in individual commits to make it easy for you to pick/revert thoses that you whish or not.

HTH


arendst:
I still can't find a more user friendly option then the current one. 

Your solution is fine for people knowing more about compile options and I wonder if it will ever work in Arduino IDE where, as far as I know, the user has no way to provide compile options.

soif:
The "more user friendy" feature is that the _user_config_override_ file is NOT overwritten when switching branch, which is a real user annoying pain  😶 

Following you comments in this file, you don't really see the need for users to change this file, so this certainly reserves the use of this file to people a lit bit more expert than the average basic user, and who certainly use better IDE than the Arduino dumb one (ie the widely used [PlatformIO](https://platformio.org/)). So adding ONE compile flag (or more for customs needs) is really easy for them.

For the remaining "basic" users, using Arduino IDE, there are STILL two methods available:
1) Add the compilation flag inside the [Arduino IDE config file](https://arduino.stackexchange.com/questions/4729/add-d-option-via-boards-txt) just ONCE, which is not the smartest way to do it, but which should work flawlessly. (I may add a detailed comment in the file about it , if you like)
2) Simply remove the #ifndef / #endif statement, which lead to the exact same situation as today ==> they will need to change this again, if they git pull, but at least all their custom settings will NOT be erased each time !...

Or we could also simply use  the proposed _user_config_override-sample.h_ , and let the `#include user_config_override.h` as it is currently (without `#ifdef USE_CONFIG_OVER` ) , then tell the user  wanting to compil (from the README.md), that they HAVE TO first rename _user_config_override-sample.h_ to _user_config_override.h_ , once.

_The goal is still the same : make the user settings survive to furure git updates !_
 
In an ideal world there should be a compiler macro like `#IfFileExists <file>, #include <file>`, this PR is just the usual workaround, widely used in many many projects.

What you think ?

Issue #1249, opened 2017-11-25
Support of Sonoff Dual R2 variant


vanyid:
Hello Tasmona community,

I have purchased 3 x Sonoff Dual modules. But those are different from any other mentioned module. (or just I not found anything on the web, maybe my fault) Does anyone have the same board?

The label on the board: **Sonoff Dual R2**. Ver 1.0 , 2017-7-20 (maybe the manufacturing date)
Inside the case there is only one chip a **ESP8285**.
The relays and the LEDs are directly connected to the ESP.

I will open a pull request according to the new module as soon as I am able to check in the necessary modifications to my fork repo.

Here is the pinning of the ESP:
This variant is more simpler then the previous one. (and perhaps cheaper for iTead)

PIN | Function
-- | --
GIO0 | BUTTON 0 (inside header)
GIO5 | L2 RELAY
GIO9 | BUTTON 1 (inside header)
GIO10 | SWITCH (on the case)
GIO12 | L1 RELAY
GIO13 | BLUE LED (inverted)

Some pictures about the board:

![img_e5188 2](https://user-images.githubusercontent.com/658247/33235044-ea436562-d230-11e7-8f9c-b7913067e698.JPG)

![img_e5187](https://user-images.githubusercontent.com/658247/33234871-c21a5bd4-d22d-11e7-8862-fbba9add7c99.JPG)

![img_e5189](https://user-images.githubusercontent.com/658247/33234878-df3a5426-d22d-11e7-894f-36c2ab53d897.JPG)

![img_e5190](https://user-images.githubusercontent.com/658247/33234880-ec1065d2-d22d-11e7-8700-c4901098b5cd.JPG)

davidelang:
since you have found what's connected to what, you should be able to use the 
wemos module type and define the various GPIOs

vanyid:
As far as I see no it will not work 100%. Because the Button1 and the case switch are connected to the GPIO9 and 10, but in the "WeMos D1 mini" the GPIO 6 - 11 are not defined as GPIO_USER.

And otherwise in my opinion if this is an official Sonoff device then is should be supported by default.
I do not know when will iTead update the official wiki page and provide the schematic for this variant?

arendst:
The R2 will be included in Tasmota as a seperate device.

For now I use your findings regardiong GPIO and add it to the next (pre)release.

ThomDietrich:
I hate to be that guy to always push the documentation...
[...]

PR #2258, opened 2018-03-25
Enable user mode for GPIO2 and GPIO5 on the Sonoff Dual

danielztolnai:
GPIO 2 and 5 on the Sonoff Dual are unused, so the pads can be routed out and used for senor connection.

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -308,10 +308,10 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
   { "Sonoff Dual",     // Sonoff Dual (ESP8266)
      0,
      GPIO_TXD,         // GPIO01 Relay control
-     0,
+     GPIO_USER,        // GPIO02 Optional sensor
      GPIO_RXD,         // GPIO03 Relay control
      GPIO_USER,        // GPIO04 Optional sensor
-     0,
+     GPIO_USER,        // GPIO05 Optional sensor
      0, 0, 0, 0, 0, 0, // Flash connection
      0,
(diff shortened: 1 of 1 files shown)


arendst:
Are these GPIOS easily access-able to the regular user?

danielztolnai:
They require the same amount of soldering skills as the currently available user mode GPIOs as they are all available on the pads near the ESP8266. While creating the image below, I've realized that 3 other GPIOs can be accessed the same way.

sonoff_dual_gpios
Note1: GPIO15 has a 10k pull-down resistor on the board, the others are not connected to anything at all.

localhost61:
And what about the Dual R2? Do you know for them? I'm willing to add a DS18B20.

danielztolnai:
@localhost61, personally I do not own a Dual R2, but based on the photos in #1249 the following GPIOs should be available: GPIO2, GPIO4, GPIO14 and GPIO16. They are on the same pin number as on the ESP8266 (see the annotated photo above) and the same amount of soldering skills are required to connect a sensor to them.

localhost61:
You are write, those pins or not connected and 6,7,8 & 11 neither.
But after thinking, the R2 version don't have the constraints of the R1, it seems that nothing prevent to use RX or TX on the header, provided that one remove the sensor(s) for serial programming, which is even not necessary in the case of OTA update.

digiblur:
Happened to see this one referenced in my Dual R2 research.  According to previous threads here the device has a ESP8285.  I found one site documenting the GPIO pins and it states GPIO0, 2, and 15 will drive start up states and 6, 7, 8 and 11 are unsable.  Only leaves with a few other free pins available on the chip itself.

https://hackaday.io/project/28220-one-tube-nixie-clock/log/70865-esp8285-gpio

PR #2294, opened 2018-03-30
add german timer translation

reloxx13:
grafik


Jason2866:
Gerät isnt a good choice for translation of timer in my humble opinion.
There is only one device (Gerät) which is switched at different times.
Maybe to long but better translation is Schaltzeitpunkt or simple let it not translated.
Timer is a well known word in German even for non English speaker

arendst:
Perhaps I better change "device" to "Output".

arendst:
But now I changed it to "Ausgang"

Is that "nice" German or should it stay as "Output"?

Jason2866:
Ausgang is ok!  It is a philosopic debatte

PR #2307, opened 2018-03-31
Sensirion SGP30 Support

gemu2015:
support for SGP30 environmental sensor

### DIFF
--- a/lib/Adafruit_SGP30-master/.github/ISSUE_TEMPLATE.md
+++ b/lib/Adafruit_SGP30-master/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,46 @@
+Thank you for opening an issue on an Adafruit Arduino library repository.  To
+improve the speed of resolution please review the following guidelines and
+common troubleshooting steps below before creating the issue:
+
+- **Do not use GitHub issues for troubleshooting projects and issues.**  Instead use
+  the forums at http://forums.adafruit.com to ask questions and troubleshoot why
+  something isn't working as expected.  In many cases the problem is a common issue
+  that you will more quickly receive help from the forum community.  GitHub issues
+  are meant for known defects in the code.  If you don't know if there is a defect
+  in the code then start with troubleshooting on the forum first.
+
+- **If following a tutorial or guide be sure you didn't miss a step.** Carefully
+  check all of the steps and commands to run have been followed.  Consult the
+  forum if you're unsure or have questions about steps in a guide/tutorial.
+
+- **For Arduino projects check these very common issues to ensure they don't apply**:
+
+  - For uploading sketches or communicating with the board make sure you're using
+    a **USB data cable** and **not** a **USB charge-only cable**.  It is sometimes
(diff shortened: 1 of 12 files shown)


arendst:
Give it a try and let me know if it works as I do not own this sensor.

gemu2015:
it works a couple of weeks already

the sensor is from eckstein on ebay  =>https://www.ebay.de/itm/282823690099 <https://www.ebay.de/itm/282823690099>

arendst:
It might work with your code but I did a rewrite using the sensors handbook regarding stability.

Pls have go at my implementation and let me know if it works too.

gemu2015:
after OTA the module stays offline, i must disassemble device and flash via serial.  

i do not have the time at the moment, i will report at least tomorrow

gemu2015:
ok, it works.

merging your new version updated my user_config_overwrite and i lost my ap access (why is that, it is on gitignore ??)

thank you!

gemu2015:
you should name it eCO2 since it is not the same as CO2

arendst:
Yeah that gitignore thing is an item I'm working on in another PR. It seems I will have to provide a user_config_override_sample file instead of the ready made user_config_override.h

Glad it works. What's the difference between CO2 and eCO2?

gemu2015:
From the SGP30 datasheet => equivalent CO2 => eCO2
1 CO2eq output is based on a hydrogen measurement. The SGP is not suited for applications where real CO2 detection is required.
Von meinem iPad gesendet

gemu2015:
A Major accurracy improvement would be to add a Humidity value to the SGP30 calculation In case you have a Humidity Sensor Running in parallel 

Von meinem iPad gesendet

arendst:
Looked at the humidity part but calculating absolute humidity seems to be a challenge in it self. The sensirion handbook formula is way of compared to a google absolute humidity table.

The formula from here https://onlineconversion.vbulletin.net/forum/main-forums/convert-and-calculate/322-relative-humidity-absolute-humidity/page2?t=567 provides another different value.

For now I stop developing humidity support.

If you want to get humidity and temperature from an attached sensor I used the following code within the Sgp30Show Json routine succesfully:
```
      double sgp30_tm = 0;
      double sgp30_rh = 0;

      char *temperature = strstr(mqtt_data, D_JSON_TEMPERATURE);  // ..,"SI7021-14":{"Temperature":22.0,"Humidity":42.8},"TempUnit":"C"}
      if (temperature) sgp30_tm = strtod(temperature +strlen(D_JSON_TEMPERATURE) +2, NULL);
      char *humidity = strstr(mqtt_data, D_JSON_HUMIDITY);  // ..,"SI7021-14":{"Temperature":22.0,"Humidity":42.8},"TempUnit":"C"}
      if (humidity) sgp30_rh = strtod(humidity +strlen(D_JSON_HUMIDITY) +2, NULL);
```

gemu2015:
Yes you are right, the special absolute humidity is a complicated formula and probably not worth the effort

the difference between CO2 and eCO2 is however important and we should not use the same name

PR #2317, opened 2018-04-01
Sunrise timer

gemu2015:
add the ability to select sunrise and sunset calculated times to a timer.

#define USE_SUNRISE

added a mode variable with

=> given time, sunrise  or sunset

also added longitude and latitude for time calculation 

// nor working yet =>
however it does not update the time on selecting the options
only after reentering the timer setup

remark => used 2 bits of currently unused mday TIMER field


gemu2015:
Ok , the code is huge (about 20k)

but i had my own timer implementation for about a couple of weeks and needing no external sensor is nice. a light sensor would shut down also during a dark storm during the day

i use it for roller shutters sonoff t1/2 gang and used it for christmas lightning with an s20 

you have to enter longitude and latitude once to get the correct result but is works perfect.

arendst:
I'm looking into it. 

Pity it takes at least 14k for the sunrise/sundown mesurement but compared to Dooticz it is only 2 minutes off.

I'll add it as an option but I still have to find out what happens at TimerEverySecond ...

arendst:
Ah, that makes sense.

Now I'm looking at the Zeitzone parameter set to 2 in dusktilldawn. I wonder if UTC time would make a better source with Zeitzone set to 0.

gemu2015:
with Timezone set to 99 this 2  hours worked for me, but UTC might be better ?

arendst:
It worked for you as you (and I) live in timezone 2. Others live on differen timezone so I have to take that into account. Will study.

reaper7:
only for sure...

> Add support for Sunrise and sunset tools (**+16k**)

+16k ?

gemu2015:
15,6k at present mainly because of angle functions and double math

arendst:
This is how it looks like:

image

gemu2015:
Well done !!!

arendst:
@gemu2015 couldn't do it without you.

gemu2015:
What about my oldest mod ?  SML reader ?

you only need a phototransistor and a 1k resistor to monitor your houses current consumption.

works with most current meters according to https://www.volkszaehler.org <https://www.volkszaehler.org/> most meters output these values without intervention

some need commands to emmit the values this would require a ir diode for sending 

This was my 1. mod, i have put it into a sonoff 4 channel into the meter box

it decodes SML binary  or OBIS asci

 
by the way a comment to SGP30 :

i examined again the original Sensirion docus and driver examples and now there are some conclusions:

the sensor by itself calculates a continuous baseline calibration which takes at least 1 hour to become stable (at 1 second sample rate)
this means that the sensor results are valid after at least one hour
this baseline could be stored in non volatile memory if one wants to shorten this restart period. Sensirion suggest to store the baseline at 1 hour intervals (30 seconds make no sense either)

in this case the values stored in NVM could be written to the baseline registers at a restart of the module. so the values would be valid directly at startup.

i do not think this is a very important feature and this baseline thing can be omitted totally.

probably one could count the first hour after a restart and mark the result with a "!" mark to sign that they are not yet valid.

PR #2372, opened 2018-04-05
German Translation for Timer

reloxx13:
Translated new
- Sunset
- Sunrise

Adjusted
- Subnetmask
- subscribe to


Leseratte10:
"Subnetmask"? In networking context, I have never seen "Subnetmask" being used as german translation for "subnet mask". They all use "Subnetzmaske". And "subscribe zu" as in "subscribe zu topic" sounds weird, too. 
I'd recommend changing "Subnetmask" to "Subnetzmaske" and "subscribe to" to "Abonniere" with another set of quotation marks around the topic itself so it reads Abonniere "topic" instead of Abonniere topic.

PR #2377, opened 2018-04-06
Sml

gemu2015:
my last PR for a while

This is the cheapest SML (OBIS) reader for electronic house energy meters. You simply need a phototransistor (e.g. TEKT5400S)   and a 1k Resistor to convert any sonoff or other esp8266 device into a  smart meter reader.
this works with meters that emit values without a request (most meters do)
it reads either binary (SML) or ASCI obis values.


gemu2015:
SML is indeed dedicated to mainly german manufacturers, however OBIS is international (also used in the Netherlands Smart Meters)

there are some differences eg baurate ist 115kb in the netherlands, 9600 in most german meters. there are also differences in bit coding 7+p or 8

the phototransistor is already provided in smart meters in the netherlands, so you can connect directly to the serial port (https://www.netbeheernederland.nl/_upload/Files/Slimme_meter_15_a727fce1f1.pdf <https://www.netbeheernederland.nl/_upload/Files/Slimme_meter_15_a727fce1f1.pdf>)

the term smart meter is also special and means readable by the electric company and or connected to the WEB, in germany we do not have many of them

but we have a lot of electronic meters that are not smart but emit their data through IR Photodiodes.

normally people connect with a read head (about 50 to 100 Euros) to a serial port via usb and typically a raspberry pi.

the esp8266 tasmota solution is much cheaper here

i do know nothing about other countries regarding meters though.

reloxx13:
No intention to add it anymore?

Some German Users use this already from the Fork:
http://forum.creationx.de/forum/index.php?thread/1095-d0-zähler-sml-auslesen-mit-tasmota/

PR #2378, opened 2018-04-06
Sunrise offset

dasMopo:
After learning about @gemu2015 's sunrise timer pulled in from #2317 and and instantly loving it , 
I wondered why there was no functionality to add an offset. 

So I did implement it myself.

It exploits the Timer.time field which in sunrise mode is otherwise unused.
This way, the sunrise / sundawn time can be positively or negatively be offset by up to 11h59m.
If a given offset would cause an over- or underflow, the actual days to be considered for firing the
timed event are shifted accordingly.

[However, only the selected (sunrise/sundawn) time + offset and originally choosen day(s) are 
currently displayed]

### DIFF
--- a/sonoff/xdrv_09_timers.ino
+++ b/sonoff/xdrv_09_timers.ino
@@ -464,7 +464,7 @@ const char HTTP_TIMER_SCRIPT[] PROGMEM =
     "if((m==1)||(m==2)){"                                         // Sunrise or sunset is set
       "p=pt[ct]&0x7FF;"                                           // Load stored time for offset calculation
       "q=Math.floor(p/60);"                                       // Parse hours
-      "if(q>12){q-=12;qs('#odr').selectedIndex=1;}"               // Negative offset
+      "if(q>=12){q-=12;qs('#odr').selectedIndex=1;}"               // Negative offset
         "else{qs('#odr').selectedIndex=0;}"
       "if(q<10){q='0'+q;}qs('#oho').value=q;"                     // Set offset hours
       "q=p%60;if(q<10){q='0'+q;}qs('#omi').value=q;"              // Set offset minutes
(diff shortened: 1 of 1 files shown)


gemu2015:
i was also thinking of an offset

however there is also an additional possibility

in fact we do not have ONE sunrise but more definitions

civil twilight
nautical twilight
and astronomical twilight

it is just the angle definition in the time equation formula.

  h (D) = -0.8333 normaler SA & SU-Gang
  h (D) = -6.0 civile Dämmerung
  h (D) = -12.0 nautische Dämmerung
  h (D) = -18.0 astronomische Dämmerung

one could make these also selectable

eazuka:
+1 👍

localhost61:
@ajayapala you push it pretty far and it seems a funny request. What is your usage purpose? I'm thinking of reducing backlite when working on screen display or watching TV.

localhost61:
@ajaypala you push it pretty far and it seems a funny request. What is your usage purpose? I'm thinking of reducing backlite when working on screen display or watching TV.

ajaypala:
@localhost61 Where I live, having the lights turn on at sunset is too early. Civil twilight is probably closer to when lights are needed. Anyway, it is just an enhancement, not really a necessity or major feature. I'll just use the offset feature. The difference in time from sunset to the twilight events changes throughout the year but I think the maximum change is 10 minutes for where I live, so no big deal. It will be more of an issue closer to the poles.

gemu2015:
constant offsets have the disadvantage of not being constant over the year especially if you get closer to the poles

defining the angle would assure a constant brightness for on off times 

why not as a compromise defining the DAWM TYPE in the config file 

e.g.

#define DAWN_NORMAL -0.8333
#define DAWN_CIVIL -6.0
#define DAWN_NAUTIC -12.0
#define DAWN_ASTRONOMIC -18.0

#define DAWN_ANGLE DAWN_CIVIL


in DuskTillDawn

double h =DAWN_ANGLE*RAD;

arendst:
@gemu2015 Will do

PR #2443, opened 2018-04-14
Use digitalWrite for PWM values over the threshold

nhorvath:
The threshold that capped PWM at 0xFC caused my MOSFET to incinerate itself because it wasn't fully open (100 Ohms @ 3A = very hot) even when dimmer set to 100. Using digitalWrite once above the threshold will avoid the PWM issues it was added to fix and also allow the FET to be fully ON when the dimmer is set to 100.


arendst:
To continue on @Frogmore42 what would happen if you keep the dimmer on 90% for say an hour? In that case the FET will definitively not be fully open but should still be able to provide 90% duty cycle power.

I always wondered how these Arilux controllers manage to supply over 3A at 12V using no heatsink at all. They function great though.

nhorvath:
The FET i'm using is a logic level one. Here's the datasheet: https://www.fairchildsemi.com/datasheets/FQ/FQD13N10L.pdf

I'm using a nodeMCU instead of an actual sonoff (I use your awesome project on my sonoff basics and wanted to try converting my RF controlled LEDs to wifi using it).
Running 12V LED strips that total about 3A. Turns out the FET i thought I blew up just got so hot it melted the solder on the leads and the smoke was just it burning the ABS case on the node it was touching, yet it somehow still tests good and appears to work (I replaced it to be safe).

Anyway when I check the resistance with no load while the PWM is at 252 it's about 90-100 Ohms. I didn't actually measure the current with it on that level of PWM so it's probably less than 3A but the strips are pretty bright still so it has to be a decent amount of current. using digital write above 252 drops it to only 8 Ohms including some connectors and wiring. When under load it looks like the FET was dropping about 1V across it. I'm not an expert in electronics by any means so my testing methodology could be totally wrong... but it seems to produce a lot less heat after this change.

Still dumps a lot more heat than the RF controller I replaced though... datasheet for the FET that controller appears to use is here: 
http://www.aosmd.com/pdfs/datasheet/AO4410.pdf

If you understand why that could be I would appreciate the insight.


Back to the change though... I didn't look back through the history of this section of code but the comment seems to indicate that your PWM library didn't like values above 0xFC and caused flicker. This change would bypass the PWM library when the user calls for it to be fully on instead of capping the PWM duty cycle at 98.8% on like it currently does. It still uses PWM below that threshold.

Frogmore42:
R(DS)(On) is an important value for your usage as it determines how much power the FET needs to dissipate (when turned on).  Yours lists it at 180mOhm (Max), but that is with V(GS) 10V.  The other one lists:  RDS(ON) < 5.5mΩ (VGS = 10V) RDS(ON) < 6.2mΩ (VGS = 4.5V)

That is a BIG difference.  Look at Figure 1 of your.  With V(GS) (gate to source voltage) of 3V (which is about what you will get out of an esp8266 shows V(DS) of almost 10V (ie a 10V drop across the FET).  The same chart on the other shows less than a 0.25V for almost any gate voltage and up to 20A.

You measured 1V drop across the FET, assuming 3A, that is 3W.  Your FET is rated for an absolute maximum of 2.5W, but this assuming an ambient temp of 25C and 1 square inch of 2oz copper as a heat sink.

The other FET would probably have less than 0.25V drop which would (at 3A) mean it would need to dissipate less than 1W (probably somewhere between 0.5 and 0.8W).  This is much easier to achieve, possibly without a heat sink at all.  If we look at based on R(DS)(On), assuming 10mOhm (at 3ishV, though it should be less than this), I^2R says the power dissipated is 9*0.010=0.1W.  At that power level it probably wouldn't even get hot without a heat sink.

The bottom line is that the FET is not appropriate for your application.

nhorvath:
Thank you for the explanation. I see now the big difference in Rds on. Yes
the other FET doesn't even have a sink and it didn't even get hot inside
the heatshrink it was packaged in (definitely warm though). This runs my
fish tank lights so I actually just ran a little flexible tubing sucking
water from one of the pumps and watercooled the FET for now which is
holding it at about 65C (poor thermal transfer between FET and water due to
the tubing, but good enough to keep it cooler). I will look into ordering
one with a lower Rds on.

I still think the code change is an improvement though as when someone puts
the dimmer value at 100 they would expect 100% duty cycle.

Frogmore42:
I think your code change is good.  I just didn't see how not having it was causing the issue you were seeing.

nhorvath:
I agree it's really a problem with the FET i'm using. It's strange that my
multimeter (fluke 179) sees the duty cycle as resistance though. There's no
way it's correct now that I think about it... when I set it at 90% it shows
2.5k, 80% 5k and so on, but there's no way that much current could flow
through 5k.

Jason2866:
Try this, it works in my setup to drive the led stripe
https://www.aliexpress.com/wholesale?catId=0&initiative_id=SB_20180416084622&SearchText=LED+RGB+Amplifier+DC12-24V

nhorvath:
Thanks, but ny strips are 12V not 24V.

Jason2866:
Amplifier for 12-24Volt!

nhorvath:
@Frogmore42 Do you see any reason this FET wouldn't be a suitable replacement? https://www.fairchildsemi.com/datasheets/FD/FDP8896.pdf

I'm looking for a through hole component otherwise I'd just order some of the ones the RF controller used. Thanks for all your help.

Frogmore42:
Yeah, that looks like it would work fine. V(GS)(TH) the voltage it starts turning on is low.  Figure 8 shows a low V(DS) (voltage drop) for current under 10A at the low V(GS) you will get from an esp8266.  Mouser has them for $1.10 each which is a decent price.

nhorvath:
Just following up for completeness' sake in case google leads someone here. Replacement FET came today and you were right they don't even get hot without a heatsink.
I wound up getting a Chinese clone of the FDP8896: 10 for $4 delivered... can't beat that. https://www.aliexpress.com/item/Free-Shipping-10Pcs-FDP8896-FDP5800-HY3003P-SUP85N03-TO-220-92A-30V-Power-MOSFET-Transistor/32851546281.html?spm=a2g0s.9042311.0.0.KvHC2Y

ascillato:
@nhorvath 

From your last comment, it means that this PR is not longer needed? We can close it? Or there is something else about this?

ascillato2:
Thanks for sharing your ideas. By now, This PR is not going to be merged due to be outside the actual scope of Tasmota. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

Issue #1964, opened 2018-02-18
 Ultrasonic distance measurement 


nosaj66au:
Hi Guys,
    Any plans for ultrasonic distance measurement sensors? I have about 4 application I could use it for now if it ran on Tasmota. I'm currently playing with this one at the moment "Jsn-sr04t Ultrasonic Module Distance Measuring Transducer Sensor Waterproof" code is very small and simple.
Jason

MimbaMonkeyHouse:
I'd also be interested in this for various water tanks!

Oxyandy:
Water tanks ? I am using float switches.. the so far have been working great

MimbaMonkeyHouse:
Thanks for the tip, but I've not seen floating switches adequate to check water level in 3m deep - 8000l tanks. I was looking at these for swimming pool water level where amplitude is not as wide as in rain collection tanks...

nosaj66au:
Hi all, Not that i wanted too, but I have now have loaded ESPEasy onto a Wemos Mini and the used the waterproof Ultrasonic module no need for volt shiffers the module runs fine on 3.3v. Bingo!!! Even have it running on Solar and 3400mA battery.  
Jason

MimbaMonkeyHouse:
Cheers for that nosaj66au. I may try that then! What solar panels are you using?

nosaj66au:
Hi MimbaMonkeyHouse, 
https://www.bigdiscount.com.au/solar-panel-monocrystalline-cells-3w-5w-8w-9v.html
the 8w version most likly overkill a I have enable sleep mode which draws 4.9mA and jumps to 120mA to measure and send the MQTT message.

MimbaMonkeyHouse:
Nice

Issue #113, opened 2017-02-26
SR04 Ultrasonic sensor support?


Killobite:
First off, fantastic firmware!  I have installed this firmware on more than 20 Sonoffs and am really appreciating the capabilities and attention to detail!

I do find that I would really like to add the use of a SR04 distance sensor for one of my implementation and am wondering if there may be any plan or ideas on how to do so?

Thanks!

arendst:
The SR04 needs 5V. Sonoff can only supply 3V3. Do you expect the SR04 to work reliably on 3V3?

AlexTransit:
Sonoff basic - use 2 power
5v - via trancformater (power relay)
3.3v - afrer stabiliser

Killobite:
I will actually be using it with a Wemos D Mini and can take the 5V right off the VCC pin.  I have a mock up working with 3.9.21 and just threw the sonar code in the void loop and sent the output to serial.  I started to dig into an "easy" way to publish the distance into MQTT but that's when I thought I would reach out and see if anybody has any plan/code on it. I had to put a little millis delay in to get it to read correctly and only report out if the distant changes by more than 10 inches but it works reliably, I just have to figure out hot to make it pretty for the web interface and MQTT output.

My use case is a garage door control with open/closed (distance) indicator and a sump pump monitor/override with water level (distance) logging.

Here is my simple loop code in case you are interested:
if (millis() >= lMillis + 250) {

      lMillis = millis();
      sDistNew = Ultrasonic.distanceRead(INC);
    
      if (sDistNew > sDist + 10 ){
        sDist = sDistNew;
        Serial.print("Distance in Inches: ");
        // Pass INC as a parameter to get the distance in inches
        Serial.println(sDist);
        //delay(1000);
      }
      else{
          if (sDistNew < sDist - 10 ){
          sDist = sDistNew;
          Serial.print("Distance in Inches: ");
          // Pass INC as a parameter to get the distance in inches
          Serial.println(sDist);
          //delay(1000);
      }
     }
}

f0m3:
Thats a cool idea. Id like to use the sonoff dual for my garage door control but id like to know if it is open or closed too...

hamwong:
Wish SR04/05 will support in sonoff, power should not be the reason make SR04 not add as SR501 also use 5V

quadflight:
Great idea. Did you get it to work?

BernardLaurenzo:
The sonoff dev board has 5v 
[...]

PR #2444, opened 2018-04-14
SR04 ultrasonic sensor support

nferreyra:
This PR adds a new driver for the SR04 ultrasonic device family. It was tested using a JSN-SR04T-2.0 reading the water level of my basement sump. I'm also planning to use it with my garage door controller to check if the gate is open or closed and when closed to see if the car is parked.
I'm adding it here because I've seen some interest on it, namely with #113 and #1964.
Feel free to adapt it to your coding style and if you need me to change something just shout.

Thanks for your work @arendst!


ruimarinho:
@arendst would you be open if I replace this implementation with https://playground.arduino.cc/Code/NewPing which adds support for SR04, SRF05, SRF06, DYP-ME007, Parallax PING and **RCW-0001** (3V3)?

It also has a built-in digital filter which I've found to be absolutely needed for error correction on these sensors.

PR #2450, opened 2018-04-15
Home-Assistant: Discovery and Force Light Setting

anduchs:
Since I'm totally new to Tasmota and this is my first code, I'm posting this as RFC, what I did correctly and what not ?

I guess I need to add the other languages, however I do not speak them. How shall I proceed here ? Add the english version to the other .h files as well ?

Also how do I proceed about the version in the Settings.h file and the version-updates in the Settings.ino file ?

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -253,6 +253,9 @@
 #define D_CMND_SENSORRETAIN "SensorRetain"
 #define D_CMND_PUBLISH "Publish"
 
+#define D_CMND_HASS_DISCOVERY "HassDiscovery"
+#define D_CMND_HASS_FORCELIGHT "HassForceLight"
+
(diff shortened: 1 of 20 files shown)


arendst:
From what I see you want to add a HAss configuration menu to the webserver and another option for a fixed light.

Looks fine keeping in mind that any webpage needs to be as small as possible to keep in memory before it is send to the user.

Just add your language defines as english to the other languages and let local people translate it over time.

Changes in settings.h (and settings ino) normally will need a version update to become auto-implemented. I suggest you make it functional with your version bump and if I might implement I'll incorporate it in the current development branch.

anduchs:
@arendst Yes that's correct. Now I have english versions added for all the other languages.

arendst:
Implemented in v6.0.0b just relesed

PR #2486, opened 2018-04-19
Hard Thermostat With Web Controls

IceStuff:
This is the code for the hard thermostat allowing control either heating or cooling on device for Sonoff TH.  Control can be via webpage or MQTT.


reloxx13 (changes requested):
You should add any new language keys at least in english in all other i18n files, if not, compiling will fail for other languages.

arendst:
Apart from the user interface is it's functionality not basically the same as a rule like:
```
{"Rule":"ON","Once":"ON","Rules":"on ds18b20#temperature>24 do power off endon on ds18b20#temperature<23 do power on endon"}
```
With variations in warming/cooling or other sensor type. Once the rule is configured the unit will perform like a thermostat standalone too.

sebfromme:
Guys, pls could I kindly ask you how to include IceStuffs code into my build or where to add arendsts rule? I'm highly interested in getting exactly this function.
Thx Seb

ascillato2:

gabrielklein:
I think this feature was great! What is missing to include it in the main code?

(With some define to avoid "overloading" the code)

PR #2487, opened 2018-04-19
Update Readme with requirement of 2.4.0 of esp8266

ascillato:
Update requirement of version 2.4.0 or superior of the esp8266 board libraries


Jason2866:
I dont find this a good idea that 2.4.0 is neccersary in general.2.4.0 has bugs that 2.3.0 doesnt have. There are lot of users with promlems and questions based on using 2.4.0 !
If i understand it right 2.4.0 is only neccersary for KNX. I find it a better idea to check at compile
time and give a warning that for KNX 2.4.0 is needed.

ascillato:
Hi, I'm downgrading the esp-knx-ip library to use 2.3.0 so KNX will run on 2.3.0

I will make the PR soon

ascillato:
The fix will be just:

Delete ESPAsyncUDP library 
Update ESP-KNX-IP library

No other change will be needed. And everything under ESP8266 v2.3.0

ascillato:
#2492

PR #2505, opened 2018-04-20
Don't display passwords on console / in "stat". Fixes #2504.

### LINKED ISSUE #2504: Presenting password exposure through RESULT on. Mqtt.
I just saw a very simplified video how to hack Tasmoto on YouTube. Most of it bullshit, if you secure your mosquito server like explained in the Wiki. Anyhow I would vote for disabled password transmition on serial and Mqtt for the ssid password , mqtthost password, fingerprint, Mqttuser password.

If the user do not remember, his problem. Passwords should not transmitted into cleartext by whatever method. All modern software system do not net you access a password in cleartext once after entered.

Leseratte10:
Related to issue #2504: Don't show passwords in MQTT result when called without parameter. 

Even though MQTT should only be reachable in the internal network (and not to an attacker), it can't hurt to hide passwords.


reloxx13:
https://github.com/arendst/Sonoff-Tasmota/pull/1727

https://github.com/arendst/Sonoff-Tasmota/pull/1601

PR #2515, opened 2018-04-21
Add support for mapping ADC input to PWM output with new USE_ADC_PWM

CapnBry:
A new ADC driver reads from the ADC input once per second, if the new value is more than 1% changed from the previous value, a fast read session begins, and the ADC is sampled every 50ms for 10s. The 1% change requirement allows Pwm# to be set in software without the ADC resetting the software-set value on each read.

The input ADC value is mapped to a PWM output (0-PwmRange) using the ADC low/high range set using the command AdcPwm target,lo,hi where target is the PWM output (1-MAX_PWM) and lo and hi are the ADC values to LERP between. Examples (assumes PwmRange 1023)

```
# Set a 1-to-1 mapping of the ADC input value onto Pwm2
AdcPwm 2,0,1023
# Disable mapping ADC value to any Pwm
AdcPwm 0
# ADC input values of 0-100 are mapped to Pwm1 (0-PwmRange)
AdcPwm 1,0,100
# Negative values are ok, a 10 ADC value becomes 0 Pwm3 output
AdcPwm 3,-10,1013
# Inverted ranges are ok, regardless of Pwm4's inverted-ness
AdcPwm 4,1000,-50
```

A typical use would be to attach a potentiometer to the ADC input (properly scaled down to the ADC input max voltage) and have it drive a PWM dimmer. I wasn't sure if this should drive a LIGHT or PWM driver but the light driver is pretty complicated so I could imagine way too many ways a user might want to directly change a value (overall brightness, a single RGB channel, color temperature, etc) so I decided to make it a simple PWM which might work more generally for more cases.

NOTE: I intentionally do not store the Pwm output value in Settings to persist between reboots. The first ADC forced read 50ms after startup will set the intensity. 

Requires USE_ADC_VCC not be defined.

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -370,6 +370,9 @@
 #define D_CMND_LATITUDE "Latitude"
 #define D_CMND_LONGITUDE "Longitude"
 
+// Commands xdrv_12_adc.ino
+#define D_CMND_ADCPWM "AdcPwm"
+
 /********************************************************************************************/
 
 #ifndef MY_LANGUAGE
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -258,7 +258,11 @@ struct SYSCFG {
   byte          knx_GA_param[MAX_KNX_GA];  // 6E2  Type of Input (relay changed, button pressed, sensor read <-teleperiod)
   byte          knx_CB_param[MAX_KNX_CB];  // 6EC  Type of Output (set relay, toggle relay, reply sensor value)
 
-  byte          free_6f6[266];             // 6F6
+  uint8_t       adc_pwm_target;            // 6F6
+  int16_t       adc_pwm_map_lo;            // 6F7
+  int16_t       adc_pwm_map_hi;            // 6F9
+
+  byte          free_6fb[261];             // 6FB
 
   char          rules[MAX_RULE_SIZE];      // 800 uses 512 bytes in v5.12.0m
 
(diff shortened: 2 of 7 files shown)


don-willingham:
I may try to leverage this to drive the brightness of a strip of WS2812B leds.

CapnBry:
Hi Theo! I did look briefly at trying to do this with rules but there were a few reasons why I did not
* Lack of math support means there wouldn't be any way to adjust the LERP. My ADC only reads from 4-999 and I wanted to map that to 100%-10% PWM (inverted because my potentiometer is plugged in backwards, but retaining the invertedness of the PWM itself). I feel like you'd still want to be able to configure the mapping even if the `ANALOG#A0` returned 0-100
* analogRead() is pretty slow so I wouldn't want to call it too frequently on idle, which is why I went for the "once per second until the value changes then a short period of faster measurements". Switching fully to a 1Hz read all the time + a rule would be too slow for feedback to the user, but 20Hz (50ms) wastes too much time if the user isn't actively changing the analog value.
* I didn't want to write to flash. Because the analog value can change a lot even with just one interaction (the user grabbing the knob and adjusting it for 5 seconds), the amount of flash writes could be quite high to store the intermediate values which are discarded 50ms later. This can be mitigated by increasing  the Settings commit delay obviously.

I'd be happy to make changes or contribute to a better more flexible solution depending on what you come up with!

ascillato2:

PR #2555, opened 2018-04-27
some bug fixes & typo

vladspbru:
fixes after static analyzer and compile warns


arendst:
Thnx for finding the timer errors.

The Ds18x20 items are as designed although I agree it doesn't look nice. The endless loop will never occur as at some time the OneWire Search will fail.

vladspbru:
But OneWire Search will never fail on new/china chip with CHIPID not in the list

PR #2614, opened 2018-05-03
Add W1_PARASITE_POWER optimization for DS18X20

zuckschwerdt:
There are additional requirements when using DS18x20 type sensors in parasite power mode:
- keep the data line high for the conversion time
- provide enough power on the data line for the conversion.

The first is accommodated by bookkeeping `w1_power_until` millis until the bus is usable again.

The latter (missing a strong pull-up) by triggering conversion on each sensor round-robin instead of all at once. To that `Ds18x20Convert()` is called with `FUNC_EVERY_SECOND` only and `Ds18x20Read()` is blocked if a conversion is still running.

I tested this with 5 sensors on a 25m bus with a 1k5 pull-up for about 2mA (worst case at 3V) usable power (the sensor can sink up to 4mA, the ESP up to 20mA) with perfectly stable results.

The whole optimization is optional with the `W1_PARASITE_POWER` define.

### DIFF
--- a/sonoff/user_config.h
+++ b/sonoff/user_config.h
@@ -255,6 +255,7 @@
 // -- One wire sensors ----------------------------
                                                  // WARNING: Select none for default one DS18B20 sensor or enable one of the following two options for multiple sensors
 //#define USE_DS18x20                              // Optional for more than one DS18x20 sensors with id sort, single scan and read retry (+1k3 code)
+//#define W1_PARASITE_POWER                        // If using USE_DS18x20 then optimize for parasite powered sensors
 //#define USE_DS18x20_LEGACY                       // Optional for more than one DS18x20 sensors with dynamic scan using library OneWire (+1k5 code)
 
 // -- I2C sensors ---------------------------------
--- a/sonoff/xsns_05_ds18x20.ino
+++ b/sonoff/xsns_05_ds18x20.ino
@@ -43,6 +43,10 @@ uint8_t ds18x20_index[DS18X20_MAX_SENSORS] = { 0 };
 uint8_t ds18x20_sensors = 0;
 uint8_t ds18x20_pin = 0;
 char ds18x20_types[9];
+#ifdef W1_PARASITE_POWER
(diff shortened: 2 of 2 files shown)


zuckschwerdt:
I hope this helps others to use TASMOTA even in weird cases. Let me know if the code style is lacking or I missed some corner case.

arendst:
Nice.

But can it be done without the delays. 750mS delay is way too long to wait for just one sensor.If there are more sensors the delays will be in the seconds...

zuckschwerdt:
There is no regular active delay. Only if you request the FUNC_WEB_APPEND (an ajax update) at a bad time the answer will be delayed until the W1 bus is ready. The FUNC_JSON_APPEND (via MqttShowSensor) is called before the FUNC_EVERY_SECOND so there can never be a delay to teleperiod reports.

Are you concerned about a web request possibly being delayed? If so, we could just report cached values, i.e. read each value before the conversion request (measurements won't be any older than they are now (less than N+1 seconds for N connected sensors). That would trade `w1_power_until` for an array of cached temperature values.

ascillato2:

zuckschwerdt:
Sorry for not keeping this updated. I still maintain this patch and did rebase ;) I have now force pushed the current state and I guess I'll clean it up and then ping you to re-open this.

zuckschwerdt:
Somehow the updated commit doesn't show up in this PR. Should I open a new PR? The changes on current `development` are zuckschwerdt/Sonoff-Tasmota@0584e73

ascillato:
But wait.

ascillato:
you have a delay command inside your code. Please, change that in order not to make tasmota stuck

ascillato:
Sorry, drivers with delay are not going to be merged due to make conflicts with other functions of Tasmota

zuckschwerdt:
Thanks. Ok, I'll add back the code to ignore one update/read cycle if there is a write going. It's actually important not to cut the power supply while an eeprom write is under way.

PR #2618, opened 2018-05-03
Update it-IT.h

fragolinux:
corrected accented vowels

### DIFF
--- a/sonoff/language/it-IT.h
+++ b/sonoff/language/it-IT.h
@@ -51,7 +51,7 @@
 
 // Common
 #define D_ADMIN "Admin"
-#define D_AIR_QUALITY "Qualita' dell'aria"
+#define D_AIR_QUALITY "Qualità dell'aria"
 #define D_AP "AP"                    // Access Point
 #define D_AS "come"
 #define D_AUTO "AUTO"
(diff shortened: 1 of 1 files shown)


arendst:
While you are here could you also please translate some English words/sentences introduced between version 5.12.0 and 5.13.0.

Especially around lines 378 - 396.

arendst:
Thnx for the update.

Add translations is an issue. What I normally do is add the English words to it and leave the message "Updated until …" in the header as is.

The translator is then expected to go over latest commits to see if there are any changes made to the language file. Boring I know but that's the only way I can think of other than sending emails to ever increasing number of translators who are changing just as easy...

fragolinux:
no, i think that's fine... the important thing is that you add the new strings to the translation file, in english of course, then as you say it's up to the translator to translate the new ones, without having to go to the original english and redo the compare line by line...
some of them are "tricky" to be translated, by the way... let's say the ones you just asked me: i don't know what a knx network is, so i can't check if they're all correct... the worst are single words, as we don't know in which context they're used and can be misinterpreted... in these cases i suggest you to add a comment to the end of the line with a sample phrase to best identify the word...

PR #2694, opened 2018-05-11
Eastron SDM120 energy meter - new sensor

gtortone:
Hi,
I just finished integration of Eastron SDM120 energy meter (http://www.eastrongroup.com/productsview/14.html);

please skip my VS Code setup files included in this PR...

Regards,
Gennaro


gtortone:
sdm120-home
sdm120-config

altelch:
Hi,

how do you connect the SDM to the ESP? Do you use a MAX3485 converter as 
the SDM is a RS485 Modbus device?

Greetings,


       Heiko.


Am 11.05.2018 um 11:06 schrieb Gennaro Tortone:

gtortone:
Hi @altelch I use this RS485<->TTL adapter

rs485-ttl

on ESP side I connect TX, RX, GND, +5V (ESP power the adapter)
from meters I connect A, B, GND

arendst:
Thnx Genarro!! Even my latest serial changes are incorporated!

Will release soon.

ionciubotaru:
Great improvement, just add a page for this device in the wiki and, if possible, support for 3 phase model
SDM630-Modbus v 2 100A

gtortone:
Thanks a lot ! A PR for SDM630 is already in place (https://github.com/arendst/Sonoff-Tasmota/pull/2735); for the wiki page I will work soon on it !

rflx18:
hi, my rs485 converter is the one that  follow, is it compatible?
do we really need gnd to the sdm?
018-max485

rflx18:
im trying to get info with this but no sucess, the SDM is a 230 witch i believe is not the issue.
any clues?

gtortone:
Hi,
the adapter you have is not compatible because it converts RS485 to RS232... then you need to add a converter from RS232 to UART-TTL;

this adapter:
image

is "all-in-one" RS485 to UART-TTL

rflx18:
thanks for the reply, but in all sites its says rs485 to ttl, i just dont understand the DE and RE that says drive enable to send and receive data.

gtortone:
yes, also Amazon says RS485 to TTL but it is wrong... they mount MAX785 IC that is a RS485 to RS422 (like RS232 but differential)

rflx18:
can you please share a link of one i can purchase, darn another month waiting :(

rflx18:
another strange thing, i tried to ground the RE to make it low to be alwaysing receiving, ant the tasmota send the command to retrieve power but says unknow commnad!

gtortone:
take into account that RS422 has +6V/-6V levels (https://en.wikipedia.org/wiki/RS-422) instead UART-TTL has 0V-3.3V... I suggest you to disconnect this adapter because there is a risk to damage RX/TX pins of ESP

gtortone:
This is the adapter I bought on Amazon (Italy):

https://www.amazon.it/WINGONEER-alladattatore-seriale-convertitore-livello/dp/B06XHH6B6R

rflx18:
hi, i have made it to work with the SDM library, there is a Dere pin that we can trigger to enable the usage of max485, now the reall question is how can we include that in tasmota, would be awesome to make it work the same way.

#define DERE_PIN                            4                           //digital pin for control MAX485 DE/RE lines (connect DE & /RE together to this pin)

gtortone:
Hi @rflx18,
I did not use this serial implementation https://github.com/reaper7/SDM_Energy_Meter but TasmotaSerial developed by @arendst.

the pin you are reporting is available on RS485/RS422 adapter while there is not such pin in RS384/UART-TTL... btw this has to be included in TasmotaSerial library

rflx18:
sorry gtortone, this is out of my league :), is it possible to include that pin for tasmota porposes also?
im finding much more dificult to get a rs485 than a max485

gtortone:
I don't know TasmotaSerial library in depth... but take into account that following your request you need an additional pin (TX, RX, DERE) that on Sonoff is not always a simple thing...

rflx18:
i know that, im using a nodemcu or a wemos, and the goal of the tasmota is to use more than 3 gpoio :)
i will w8 for good news, thanx

gajotnt:
Is there any trick to make the SDM send information via the RS485 to TTL?
I have everything connected and i get no information.
Only the send indicator of the adapter blinks

localhost61:
@gtortone, just to make things clear, I own the same adapter as @rflx18 and the chip is **really** a Max485. The bad news is that it's not designed to work under 3.3V. So it's OK for a Arduino Pro Mini but not a Sonoff Dual R2. :-/

gtortone:
@localhost61 infact you need this adapter:

https://www.amazon.it/WINGONEER-alladattatore-seriale-convertitore-livello/dp/B06XHH6B6R

that works with 3.3V/5V voltage levels:

https://ae01.alicdn.com/kf/HTB1PDHWIVXXXXcGaXXXq6xXFXXXv/SCM-TTL-to-RS485-Converter-485-to-UART-Serial-Port-Automatic-Flow-Control-Module-High-Quality.jpg

localhost61:
Yes @gtortone , it's already in the pipe. ;-)

localhost61:
@gtortone, your adapter is very interesting because it matches pin for pin the J2 header on a Sonoff Dual R2. Unfortunately I couldn't succeed to make it work "out of the box" on SDM120 and SDM230.
Even disconnected from the modbus, but fitted on the Dual R2 connector, the RX LED blinks every second, even if I invert RX pin and TX pin through the module configuration webpage.
Then guess I'll have to take the slow way, initiate at first my Eastron modules and study the ModBus protocol. :sight:

gtortone:
@localhost61 this is my configuration of SDM120 for Sonoff Basic: 
image

I suggest you to double check speed of SDM module and speed inside Sonoff-Tasmota configuration file user_config.h (SDM120_SPEED or SDM630_SPEED)

localhost61:
Thanks for the tips Gennaro! I did the same, SDM-TX on Serial In and SDM-RX on Serial Out, it's logical !
I had already thrown an eye on the SDM driver code and seen you didn't implement any new command to manage them from the console, I'll have to get a windows app and learn how to write proper settings in the 4X registers and check if everything is in order, 32 bit float values and CRC make it difficult to do that manually :-/ . I already have the USB to RS485 adapter for the job (and for monitoring too.)
Currently, my SDM120 both default to node 01 at 2400,8E1 and SDM230 to node 01 too at 9600,8E1 (this one can be setup on device manually use both buttons whereas the SDM120 needs a modbus master for that).
I'm aware I have to address them separately but my goal is to mount them as a bus and I may dedicate a WeMos D1 pro for them.

localhost61:
Hi,
Unfortunately, it seems my adapter is broken, the SD3486 doesn't work properly.
- When the adapter is only powered and other pins not connected, the pullups/pulldown make that the bus should be HiZ. It,'s not the case here, the 4k7 pullup on A+ doesn't pull the value to Vcc but 2.24V
- When the adapter is running on the Dual R2 and an other master is connected on the bus, with no slave and polling the bus each second, then the RxD Led should blink on the adapter, not only the TxD driven by Tasmota.
- For those interested, here is a schematic I've reverse designed from the adapter:
adapter rs485 3v3

localhost61:
Shame on me! The adapter pinout doesn't match the Dual R2. RX and TX are twisted.
It explains why RxD Led blinked and not the TxD one. Once crossed it works ^^.

It' seems that there is only one working option that is to map Tasmota SDM120Tx and SDM120Rx on the physical TxD and RxD pins, other GPIO pins can be selected but the SDM data don't appear on the main webpage.

gtortone:
@localhost61 I'm happy you solve your communication problem !
regarding pin selection this is strange because SDM120 code uses TasmotaSerial library that can be mapped on every GPIO pin...

localhost61:
> SDM120 code uses TasmotaSerial library

@gtortone maybe not when compiled with core 2.3.0 which defaults now.

localhost61:
@gtortone 
I just downloaded latest build  6.2.1.3, compiled with old core 2.3.0 and uploaded OTA to this Dual R2.
It acts the same, same home page and 75SDM120Tx can be exchanged with 76SDM120Rx in the module configuration with no difference.

I would like to see if I can change the code in order to add a command to change the modbus slave target ID in order to monitor my three devices, that should be rather straightful.

arendst:
@localhost61 Xavier, just got myself a SDM230 and seeing how you hooked it up confirms my connection.

I see you do not use the Gnd of the RS485 interface. Is it not needed? I disconnected mine too and it seems to work just fine indeed.

I connected only a simple load of 15W so that's probably an issue but I see the Export Reactive value incrementing together with the Import Active and Total counters. I cannot explain this as I'm definitly not supplying power.

localhost61:
I guess that Gnd may be used for long distance ModBus for the Shield, not used with a twisted pair
About exporting reactive energy I don't think there is an issue with the driver here, as it only reports the registers content.
It seems weird when you monitor a simple load on the grid but it simply means that your *load* is mostly capacitive and then the meter consider that you *export* reactive power (sin(φ)<0)), with an inductive load like a drill machine or an old fashion neon with ballast it should increase the *import* reactive power counter. 
With switching power supplies, when the current is not a pure sinusoid, there is a reactive part too but I don't know how to tell if it will be inductive or capacitive.

localhost61:
I just tested with a vacuum cleaner. It's power factor is optimized at full power, but at minimal power, it's only 0.5 and I could verify that as an inductive load it increase *import* reactive energy counter. The *export* energy counter remains unchanged. I currently have a "capital" of 44kVAr in the *export* energy counter because my SDM230 was loaded 24/7 by 2 SDM120 that are seen as very highly capacitive loads.
---
I notice right know that there is a confusion in my translation, but maybe in the english version too, you tell me...
Power times a duration equates to an energy, all the `D_POWERUSAGE` strings are not power but energy counters. In French we don't say Power supply but Alimentation, meaning feeding so for us power is always in Watt (or maybe horse power, it depends ;-) ) and never in Wh :-(

localhost61:
I was wrong,  `D_POWERUSAGE` strings are truly power values. The issue is only in regards to SDM220 specific IMPORT/EXPORT strings.

PR #2713, opened 2018-05-12
Create templates for Bugs, Requests and Troubleshooting

ascillato:
As an attempt to help on the issues resolutions, a clasification is needed also providing some steps for the descriptions required.


ascillato:
This is how it looks when you press the NEW ISSUE Button

image

Giving to the user, templates to be used as a guide

PR #2717, opened 2018-05-13
Enable user mode for GPIO2, GPIO5 and GPIO16 on the Sonoff RF/Basic 

brainnex:
Hello,
not a big deal but if GPIO4 is in the List why not make it complete and put all usable GPIOs in the list.

Related to:
https://github.com/arendst/Sonoff-Tasmota/issues/2650
https://github.com/arendst/Sonoff-Tasmota/pull/2258

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -256,10 +256,10 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
   { "Sonoff Basic",    // Sonoff Basic (ESP8266)
      GPIO_KEY1,        // GPIO00 Button
      GPIO_USER,        // GPIO01 Serial RXD and Optional sensor
-     0,                // GPIO02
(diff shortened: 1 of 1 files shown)


reloxx13:
As mentioned before, use generic module.

I think this will only confuse "normal" users, cuz their are no other (prepeared) pins on the board and its neccessery to solder it on the ESP directly.

Also, this would make module types senseless, cuz every Module with an ESP got more than the displayed GPIOs on the webui.

Dispaying GPIO4 is rather an old part which was added before generic module type or cuz its the most used GPIO for self-soldering.

You could also use the RX and TX pins if you need more than the GPIO14.

brainnex:
OK. No Problem. I used already successfully your suggest. Thanks.

reloxx13:
dude, theo got the last word on it not me, im just reviewing, argumenting and sharing my thoughts about it :)

arendst:
Agree with @reloxx13 so closing this one

PR #2726, opened 2018-05-14
KNX: Add control for all Tasmota Outputs.

ascillato:
Add control by KNX for all Tasmota Outputs
Before this, only Relays were able to be controlled by KNX. Now all Outputs can be controlled using KNX.

### DIFF
--- a/sonoff/xdrv_11_knx.ino
+++ b/sonoff/xdrv_11_knx.ino
@@ -108,14 +108,14 @@ device_parameters_t device_param[] = {
 
 // device parameters (information that can be sent)
 const char * device_param_ga[] = {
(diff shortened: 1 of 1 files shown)


ascillato:
I found that enabling KNX with last release v5.13.1b, I had to disable more sensors to free memory, in order to solve resets of error: exception 28.

jeylites:
This sounds great! Just a though to consider if it makes it make sense. 

It will be nice to have two version of KNX for the sake of space. It’s not always one will need a full blown KNX with additional sensors. 

On my setup all is needed is Relay. There’s not much one can do with a T1 switch by adding more sensors. 

So perhaps we could look at KNX and KNX extended with sensors for a though. What do you think?

PR #2736, opened 2018-05-15
Update Bulgarian translations to v5.13.1c

znanev:
- Fix some bad translations
- Do not use Cyrillic in Si units:
    - Si units use only Latin or Greek letters. Writing SI units with Cyrillic letters (like it used to be according to BDS) is no longer allowed.


arendst:
Thnx.

znanev:
Changes merged as part of 12ffefe08177f29e088331a9b7166e1ecafc2ed6

PR #2737, opened 2018-05-15
Ignore Visual Studio Code local settings

znanev:
Ignore local Visual Studio Code settings files.

It doesn't hurt having more files / folders in .gitignore file. Rather not having them is irritating for users that use Visual Studio Code, because files from .vscode folder will appear every time as uncommitted changes in git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@
 .gcc-flags.json
 sonoff/user_config_override.h
 build
+
+## Visual Studio Code specific ######
+.vscode


gtortone:
Sorry,
my fault !

I did not realize it...

znanev:
No problems at all!

I was just wondering if it was a deliberate removal, because it is quite common that people use different IDE / tools and it is hard to please everyone anyway :)

PR #2742, opened 2018-05-16
 Add KNX Communication Enhancement

ascillato:
Added option to enable KNX Communication Enhancement. If activated, will make the device send 3 times the same telegram. This drastically reduce the telegram lost, improving communications issues.


ascillato:
Screenshot of the KNX Configuration menu with the new option

image

PR #2838, opened 2018-05-26
Adds counter inc and dec using command

IVI053:
Enables the use commands like `counterX 10+`, `counterX 1+` or `counterX 1-`

I was in need to increment a counter using rules. Seemed like there is no proper way of doing this easily so I changed a few lines of code to do this.


arendst:
Did it my way:

- Set value: ``counter1 33``
- Increment value: ``counter1 +11``
- Decrement value: ``counter1 -12``

PR #2885, opened 2018-06-01
Increase Switch to max. 8

Jason2866:
Increase Switch to 8


Jason2866:
For users who wants to connect (for example on a 4CH) another switch or
on a WeMos just switches to control "devices" in hardware without the use of the HA Console.

arendst:
Correctly implemented in v6.0.0b

PR #2955, opened 2018-06-10
Add Arithmetic commands for VAR

ascillato:
Hi @arendst,

For your consideration to see if this new commands can be included in Tasmota.

New Arithmetic commands to be used with `VARs`:

`ADD1` to `ADD5`: Add a value to `VARx`
Syntax: `ADDx value`
Usage: `ADD1 15`
Result: `VAR1 = VAR1 + 15`

`SUB1 `to `SUB5`: Substract a value from `VARx`
Syntax: `SUBx value`
Usage: `SUB1 15`
Result: `VAR1 = VAR1 - 15`

`MULT1 `to `MULT5`: Multiply a value to `VARx`
Syntax: `MULTx value`
Usage: `MULT1 15`
Result: `VAR1 = VAR1 * 15`

`SCALE1 `to `SCALE5`: Scale a value from a low and high limit to another low and high limit and store it in `VARx` (directly equivalent to MAP arduino command)

Syntax: `SCALEx value, fromLow, fromHigh, toLow, toHigh`

where,

_value_: the number to scale
_fromLow_: the lower bound of the value’s current range
_fromHigh_: the upper bound of the value’s current range
_toLow_: the lower bound of the value’s target range
_toHigh_: the upper bound of the value’s target range

_(omitted values are taken as zero)_

Usage: `SCALE1 15, 0, 100, 0, 1000`
Result: `VAR1 = 150`

This code is very little (just 496 bytes) and it expands the rules posibilities of Tasmota.

With this PR, issues where a sensor value is only sent when it changes a delta, can be solved applying these new commands as the following rule example:

```
rule
on AM2102#temperature>%var1% do backlog var1 %value%; publish stat/sonoff/temp %var1%; var2 %var1%; add1 2; sub2 2 endon
on AM2102#temperature<%var2% do backlog var2 %value%; publish stat/sonoff/temp %var2%; var1 %var2%; add1 2; sub2 2 endon
```

### DIFF
--- a/sonoff/xdrv_10_rules.ino
+++ b/sonoff/xdrv_10_rules.ino
@@ -74,11 +74,15 @@
 #define D_CMND_EVENT "Event"
 #define D_CMND_VAR "Var"
 #define D_CMND_MEM "Mem"
+#define D_CMND_ADD "Add"
+#define D_CMND_SUB "Sub"
+#define D_CMND_MULT "Mult"
+#define D_CMND_SCALE "Scale"
 
 #define D_JSON_INITIATED "Initiated"
 
-enum RulesCommands { CMND_RULE, CMND_RULETIMER, CMND_EVENT, CMND_VAR, CMND_MEM };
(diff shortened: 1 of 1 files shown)


ascillato:
Added to the wiki (commands and rules)

PR #3008, opened 2018-06-16
rf receive HEX DEC setting

sammy2360:
hi arendst,
created a setting in setoption to switch between HEX and DEC data in the json object when an RF signal is recieved on the sonoff bridge.

i have used setoption 28 as it was labeled as spare.
i have tested this and all seems to be working as it should.

in sonoff.ino i have changed the CMND_SETOPTION index from 26 to 28 which may enable the setting for knx_enable_enhancement. 
this may have to be looked at. maybe by adding ||(index == 28) instead.

thanks for all your great work.
sam

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -35,7 +35,7 @@ env_default = sonoff
 
 [common] ; ************************************************************
 ; *** Esp8266 core for Arduino version 2.3.0
-platform = espressif8266@1.5.0
+;platform = espressif8266@1.5.0
 ; *** Esp8266 core for Arduino version 2.4.0
 ;platform = espressif8266@1.6.0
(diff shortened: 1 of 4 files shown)


ascillato (changes requested):
Hi,

Can you modify your PR so as to leave platformio.ini file without changes?

Thanks

sammy2360:
yes ill download a fresh copy and delete the modified one. apologies.

sammy2360:
ok that is the platform.ini back to the original state. it is still listed in the pr but with no changes. will that disappear? please forgive me im new to github.

arendst:
Just implemented. Thnx.

PR #3041, opened 2018-06-22
update of zh-CN language

killadm:


ascillato:
Sorry.

Your file contains >>>>>>

May be, a fastest way could be that you go (with your web browser) to the file in this repo and modify it directly. You can copy and paste your file from your previous PR. Github will give you then, the option to make a pull request just for that modification.

Thanks

PR #3044, opened 2018-06-22
add TMP112 support

cputoaster:
Seems to work for me with a TMP112AIDRLR, there is an offset of a few °C, should only be 0.5 °C. Don't know why.

### DIFF
--- a/sonoff/user_config.h
+++ b/sonoff/user_config.h
@@ -264,6 +264,7 @@
   #define USE_HTU                                // Add I2C code for HTU21/SI7013/SI7020/SI7021 sensor (+1k5 code)
   #define USE_BMP                                // Add I2C code for BMP085/BMP180/BMP280/BME280 sensor (+4k code)
 //    #define USE_BME680                           // Add additional support for BME680 sensor using Adafruit Sensor and BME680 libraries (+6k code)
+//    #define USE_TMP112                            // Add I2C code for TMP112 temperature sensor
   #define USE_SGP30                              // Add I2C code for SGP30 sensor (+1k1 code)
   #define USE_BH1750                             // Add I2C code for BH1750 sensor (+0k5 code)
 //  #define USE_VEML6070                           // Add I2C code for VEML6070 sensor (+0k5 code)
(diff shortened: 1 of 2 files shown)


cputoaster:
Actually turns out it was just a bad test setup, with heat radiation from the flasher / power supply. Works fine.

andrethomas:
You need to make provision for negative temperatures - see datasheet http://www.ti.com/lit/ds/symlink/tmp112.pdf - Refer to table 2 and its descriptive text below the table.

It appears to work similar to the LM75AD sensor with regard to negatives so you could borrow that method of conversion used in the LM75AD code.

Also, you need to add support for F - This is also demonstrated in the LM75AD driver using the M75ADShow function and the LM75ADGetTemp where it uses the configured temperature unit to perform the conversion automatically to either C or F depending on user config.

See https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/xsns_26_lm75ad.ino to see what I mean - perhaps a bit of restructuring and cleanup would be a good idea... just my 2 cents.

andrethomas:
@arendst I see that - tried finding some just for testing purposes... can only find chips.

Also, I2C address conflicts with LM75AD.

cputoaster:
FYI, I was using this for testing: http://wiki.aprbrother.com/en/ESP_Soil_Moisture_Sensor.html

I think this could help, but have to see when I have time: https://github.com/ControlEverythingCommunity/TMP112/blob/master/C/TMP112.c

cputoaster:
How would you go about the address conflict? Just never compile with both of them? Or some elaborate detection?

andrethomas:
How would you go about the address conflict? Just never compile with both of them? Or some elaborate detection?

The simplest would be to check if it was defined and then undefine it with a warning that it has been disabled. Its unlikely that someone would use both sensors at the same time but I am not sure how the LM75AD will respond to your detection code for the TMP112 if they are connected on the same I2C bus.

Either way, safe is to do some exception handling on #ifdef statements like what can be seen in https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/xsns_27_apds9960.ino with specific reference to this piece in the code

```
#if defined(USE_SHT) || defined(USE_VEML6070)
  #warning **** Turned off conflicting drivers SHT and VEML6070 ****
  #ifdef USE_SHT
  #undef USE_SHT          // SHT-Driver blocks gesture sensor
  #endif
  #ifdef USE_VEML6070
  #undef USE_VEML6070     // address conflict on the I2C-bus
  #endif
#endif
```

PR #3102, opened 2018-06-29
Add support for I2C MCP23008 GPIO for 5V tollerant GPIO INPUT ONLY

andrethomas:
Adding support for MCP23008 I2C GPIO expander, but limited to INPUT ONLY and will only transmit mqtt based on logging interval configuration... so momentary changes are not detected at this stage.

The main idea is to have expanded inputs which are 5V tolerant.

So yeah, it basically works like a sensor insofar that it reports the state of the 8 gpio input states as and when the chip is polled.


andrethomas:
Actually, let me add interrupt support to publish as

`stat/sonoff/RESULT = {"Time":"2018-06-30T17:49:56","MCP23008":{"D0":0, "D1":1, "D2":0, "D3":0, "D4":0, "D5":0, "D6":0, "D7":0 }}`

To be enabled using USE_MCP23008_INT

Also, going to add support to enable internal pullup resistors using define USE_MCP23008_PULLUP

Once complete, I'll try commit this again

PR #3107, opened 2018-06-30
Add support for I2C MCP23008/MCP23017 for GP INPUT

andrethomas:
Adding support for I2C MCP23008 GPIO chip - ONLY INPUT IS SUPPORTED

It is disabled by default in user_config.h

If enabled will provide mqtt telemetry result as follows

`tele/sonoff/SENSOR = {"Time":"2018-06-30T18:31:49","MCP23008":{"D0":0, "D1":0, "D2":0, "D3":0, "D4":0, "D5":0, "D6":0, "D7":0 }}`

Internal pull-up weak resistors can be enabled by uncommenting USE_MCP23008_PULLUP which means all inputs will return 1 unless pulled down to ground externally.

The information above will only be reported to mqtt frequency based on the Telemetry period configured for the device - I believe that defaults to 300-second intervals but may be changed in the device configuration after flashing.

Immediate notification on GP input state changes from wither high to low, or low to high can be received by enabling USE_MCP23008_INT in the configuration file which will enable interrupt functionality of the MCP23008 chip - This is polled every 50 milliseconds and will report as follows if and when any of the input pins change state via mqtt messages

`stat/sonoff/RESULT = {"Time":"2018-06-30T18:33:00","MCP23008":{"D0":1, "D1":0, "D2":0, "D3":0, "D4":0, "D5":0, "D6":0, "D7":0 }}
`


andrethomas:
I must add that the intention is not to extend GP OUTPUT with this chip but rather to provide additional inputs and more specifically inputs which are 5V tolerant.

andrethomas:
Resulting display on web interface:

image

Resulting information on MQTT and Interrupt behaviour:

image

andrethomas:
@arendst  Is it possible for you to add 'uint8_t mcp23008_config[8];' to Settings (struct SYSCFG) if you accept this commit? - I am unclear on the process needed to take to make an amendment to setting structure without messing up upgrades for other people :)

Maybe useful to be able to configure pull-up and interrupt behaviour per pin and store in flash settings.

Thanks!

andrethomas:
@arendst 

Perhaps to make easier, define a union struct

```
typedef union {
  uint8_t data;
  struct {
    uint8_t enable : 1;    // Enable INPUT
    uint8_t pullup : 1;    // Enable internal weak pull-up resistor
    uint8_t inten : 1;     // Enable Interrupt on PIN
    uint8_t intmode : 1;   // Change on STATE or match COMPARATOR
    uint8_t intcomp : 1;   // Interrupt COMPARATOR
    uint8_t b5 : 1;
    uint8_t b6 : 1;
    uint8_t b7 : 1;
  };
} MCP23008_Config;

```

And then add

`MCP23008_Config      mcp23008_config[8];`

to the SYSCFG struct, then I can easier access the bits required.

?

renne:
@arendst @andrethomas 
Finally more inputs for Sonoff devices! :1st_place_medal: 

Please keep the [MCP23017](https://www.microchip.com/wwwproducts/en/MCP23017) (16 GPIOs) in mind when merging the code.

andrethomas:
@renne if @arendst agrees and is willing to make a change to configuration struct I can add support for MCP23017 but then I think we need to do it as one driver so that would require modification to my request on the configuration struct - perhaps

```
typedef union {
  uint8_t data;
  struct {
    uint8_t enable : 1;    // Enable INPUT
    uint8_t pullup : 1;    // Enable internal weak pull-up resistor
    uint8_t inten : 1;     // Enable Interrupt on PIN
    uint8_t intmode : 1;   // Change on STATE or match COMPARATOR
    uint8_t intcomp : 1;   // Interrupt COMPARATOR
    uint8_t b5 : 1;
    uint8_t b6 : 1;
    uint8_t b7 : 1;
  };
} mcp230xx_config;
```

and then in SYSCFG

`MCP230xx_Config mcp230xx_config[16];`

This way we can handle configuration of either 8 input or 16 input in the same configuration space.

I don't think its necessary to support 8 input and 16 input independently so the driver will work with only one of the two on the I2C bus. I do not see a need to go beyond either a single 8 input or single 16 input at this point.

Jason2866:
Is it possible -with a not a to big afford- that you adapt your driver for the PCF8574?
This chip is easy and cheap to get on a pcb (used for I2C to LCD)  
https://www.aliexpress.com/item/1pcs-Serial-Board-Module-Port-IIC-I2C-TWI-SPI-Interface-Module-for-Arduino-1602-LCD-Display/32404454240.html?spm=a2g0s.13010208.99999999.301.384e3c003HGBsk

andrethomas:
```
Is it possible -with a not to big afford- that you adapt your driver for the PCF8574?
This chip is easy and cheap to get on a pcb (used for I2C to LCD)
```

Although it sounds like a cool idea my main goal at this point is to just expand input for ESP8266 based devices. The idea basically started in using the mcp23008 as a "sensor" and not a device as such.

I think there are others who have tried what you are suggesting before and for some reason, @arendst does not appear to want to explicitly support those type of devices (as an expansion of OUTPUT) in the sonoff firmware. You should definitely take a peek at his display branch - I think I saw some I2C displays and other display support being developed there by @arendst over here 

https://github.com/arendst/Sonoff-Tasmota/tree/development_display

So simple answer is no, I do not plan for it, but if there is demand and blessing from the developers then I'm willing to write the code I guess.

I'm also not sure I have any PCF8574 devices in stock but I will take a peek sometime during the week.

Jason2866:
@andrethomas 
Thank you for your answer. The PCF8574 not as "Output" device for a LCD.
Same use case as you intention for the MCP23008. Just to expand Inputs (5 Volt tolerant)
If i understand the datasheet of the PCF8574 correct it can be used for Input or Output

I use the branch display from @arendst an it works nice :-)

andrethomas:
```
Thank you for your answer. The PCF8574 not as "Output" device for a LCD.
Same use case as your intention for the MCP23008. Just to expand Inputs (5 Volt tolerant)
If i understand the datasheet of the PCF8574 correct it can be used for Input or Output

```
Thanks for the clarification... Sure, could be added to work the same as I plan to use mcp23008 - Will need to get some chips and do some testing - first want to finish mcp23008 integration but will keep the pcf chip in mind.

Thanks

renne:
@stefanbode supports the PCF8574 as output in his [Tasmota-mod](https://github.com/stefanbode/Sonoff-Tasmota/wiki/PCF8574-I2C-DIO-extension). Maybe he is interested in adding input support.

andrethomas:
Yeah, he is planning on adding

andrethomas:
**note to self - do not reply using a phone because I might accidentally close the request :)**

I think one of the main drawbacks of the PCF chip is that it does not appear to support interrupt polling via I2C so I can add it but it will not cover all the functionality provided by the MCP variant.

I'm in the process of adding the web UI configuration for the MCP230XX chips and I need to still connect a mcp23017 to see if I can programmatically determine the difference as to which one was connected and act/configure accordingly.

Jason2866:
For the MCP23017 i found versions with a PCB not only chips as for the MCP23008!
I will order
https://www.aliexpress.com/item/MCP23017-Serial-Interface-Module-IIC-I2C-SPI-MCP23S17-Bidirectional-16-Bit-I-O-Expander-Pins-10Mhz/32830799627.html?spm=2114.search0104.3.1.b5e365a9ImjQpL&ws_ab_test=searchweb0_0,searchweb201602_3_10320_10152_10151_10065_10321_10344_10068_10342_10547_10343_10322_10340_10548_10341_10696_10084_10083_10618_10304_10307_10820_10821_10302_10843_10059_100031_10319_10103_524_10624_10623_10622_10621_10620,searchweb201603_30,ppcSwitch_5&algo_expid=7f0ce58d-edb6-4e95-bae2-3aa94034301f-0&algo_pvid=7f0ce58d-edb6-4e95-bae2-3aa94034301f&priceBeautifyAB=0

andrethomas:
@Jason2866 looks like that one will work. I got detection of MCP23008 vs MCP23017 working... actually quite easy

```
    if (I2cValidRead16(&buffer, mcp23008_address, MCP23008_IODIR)) { // Can we get a valid response? If so, then chip must be valid for this address
      if (buffer == 0xFF00) {
        mcp23008_type=1; // We have a MCP23008
      } else {
        if (buffer == 0xFFFF) {
          mcp23008_type=2; // We have a MCP23017
        }
      }
```

The MCP23008 returns 0's on the second byte for IO direction so we can distinguish between the two during detection.

I see the description and the photo of the link do not match - I have not checked if the MCP23Sxx (with SPI support) will still respond to I2C but it would seem logical that it would.

Needless to say there needs to be some changes to the variable names etc to mcp230xx as opposed to being strictly mcp23008 (yes, I am rather specific in those type of conventions) hehe

andrethomas:
Progress for today

image

andrethomas:
@arendst  @renne @Jason2866 

I think I've completed the initial work - probably some room for opto but its weighing in on +4k3 flash & +100 bytes of ram.

I think I've tested every possible combination of use.

Jason2866:
I hope the chinese sends my MCP230017 fast. I cant wait to test and build my master switch panel.
Thank you @andrethomas 👍😃

andrethomas:
@Jason2866 Usually my luck is its on some slow tugboat of sort... my results on chinese orders range from 2 weeks to 3 months. For this code I just hooked the chips up on breadboard to a ESP8266 12F

The MCP only needs power,ground, + to enable and 3 wires to ground to set the I2C address, and of course the SCL and SDA wires... was quite an easy build and seems to work reliably.

I'm too scared to ask what your "master switch panel" will be controlling... 👍😃

Jason2866:
Control even more as planed with #2885 😎

andrethomas:
Sounds like something I need to do also... i hate walking into my entertainment room only to find my phone to turn stuff on.

Nice thing is the interrupts are sending on individual state changes now... I need to look at how switches work - no clue... never looked into it on the esp modules
```

21:50:05 MQT: stat/sonoff/RESULT = {"Time":"2018-07-05T21:50:05","MCP230XX_INT":{"Pin":"D1", "State":0}}
21:50:11 MQT: stat/sonoff/RESULT = {"Time":"2018-07-05T21:50:11","MCP230XX_INT":{"Pin":"D1", "State":1}}
```

andrethomas:
@arendst 

I had to add `#define D_CONFIGURE_MCP230XX "Configure MCP230xx"` to en-GB.h

I'm not sure what the process is to add them to other language files, or if it is needed or not - please advise.

Thanks

arendst:
@andrethomas regarding switches:

- currently only support eight and that shouldn't be a problem.
- I implement virtual switches overlaying the hardwired GPIOs by some changes to SwitchHandler.

I was working on the TM1638 driver but by lack of hardware I had to stall it as the current implementation doesn't seem to work. The virtual switch handler should work though so I advise you take a look at lines 201 to 211 of xsns_28_tm1638.ino how to use the eight virtual switches with the first eigth inputs of your device. Shouldn't be too hard.

How to you plan to use the received state changes of your 16 inputs? By using rules or by a HA tool?

arendst:
I'll update the other languages when needed to keep this PR as small as possible regarding file count.

andrethomas:
> I was working on the TM1638 driver but by lack of hardware I had to stall it as the current implementation doesn't seem to work. The virtual switch handler should work though so I advise you take a look at lines 201 to 211 of xsns_28_tm1638.ino how to use the eight virtual switches with the first eigth inputs of your device. Shouldn't be too hard.
> 
> How to you plan to use the received state changes of your 16 inputs? By using rules or by a HA tool?

For now I'm receiving on node-red via mqtt, was just curious about switches and what they do so will take a peek and see if it is of interest.

Thanks!

andrethomas:
@arendst The TM1638 looks really cool and interesting - added 3 to my next cart from banggood - will help you with debugging when they get here.

Jason2866:
Hi together,
for me it is ok to get the status of the inputs via mqtt! I will do the "magic" in OpenHab
Bit it would be nice if virtual switches in Tasmota would be realized....
Thanks for your work!

renne:
If you need up to 48 cheap buttons on a ESP8266 have a look at #3142 ;-)

andrethomas:
@arendst Question - I keep my fork up to date with

```
git clone https://github.com/andrethomas/Sonoff-Tasmota.git
git remote add upstream https://github.com/arendst/Sonoff-Tasmota.git
git fetch upstream
git pull upstream development
git push
```

It does pull your new commits and pushes them back to my fork which is all perfectly in order but why do I see these inside this PR?

image

andrethomas:
@Jason2866 I'm not going to explore adding the PCF chip to this due to lack of interrupt access via I2C on the PCF

@arendst Anything else you want me to add/change to this?

Jason2866:
@andrethomas 
Thanks for taking a look! Since you have included MCP23017 (where a DIY PCB board is availiable)
this is the better choice. The PCB was the reason for me asking for supporting the PCF.
Thank you once again!

andrethomas:
@Jason2866 No sweat, you're welcome... had to write it for me anyway so may as well share it 👍

PR #3216, opened 2018-07-13
Add Joywell 5ch power strip

fnordsh:
Add support for Joywell 5ch power strip.
See: https://www.amazon.de/Steckdosenleiste-Steckdose-App-Steuerung-Bedienung-Joywell/dp/B079C1HPSK

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -201,6 +201,7 @@ enum SupportedModules {
   SONOFF_POW_R2,
   SONOFF_IFAN02,
   BLITZWOLF_BWSHP2,
+  JOYWELL_5CH,
   MAXMODULE };
 
 /********************************************************************************************/
@@ -243,6 +244,7 @@ const uint8_t kNiceList[MAXMODULE] PROGMEM = {
   SONOFF_BN,
   SONOFF_IFAN02,
   SONOFF_BRIDGE,
+  JOYWELL_5CH,
   CH1,
   CH4,
   MOTOR,
@@ -884,6 +886,21 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
      GPIO_HLW_CF1,     // GPIO14 BL0937 or HJL-01 CF1 voltage / current
      GPIO_REL1,        // GPIO15 Relay (0 = Off, 1 = On)
      0, 0
+  },
+  { "Joywell 5ch",     // Joywell power strip (ESP8285)
+     0,                // GPIO00 (D0)
+     GPIO_LED1_INV,    // GPIO01 (D7)
+     0,                // GPIO02 (D2)
+     GPIO_KEY1,        // GPIO03 (D8)
+     GPIO_REL2,        // GPIO04 (D4)
+     GPIO_REL1,        // GPIO05 (D5)
+     0, 0, 0, 0, 0, 0, // Flash connection
+     GPIO_REL3,        // GPIO12 (D12)
+     GPIO_REL4,        // GPIO13 (D13)
+     GPIO_REL5,        // GPIO14 (D14)
+     0,                // GPIO15 (D15)
+     0,                // GPIO16 (D16)
+     0                 // ADC0 Analog input (A0)
   }
 };
 


gertmenke:
Just to clarify: The article description suggests this thing has 4 channels, but in fact there are 5 relays, one for each socket and one for the included usb power supply.

Jason2866:
Could you post some photos? Where are Rx, Tx, GPIO 0, Gnd?
How to disassemble? Something "special" to know?
THX!

gertmenke:
I've posted some details in the sonoff-users group a while ago:
https://groups.google.com/forum/#!topic/sonoffusers/edE3DMJccs8

But I'll try and add some more details and/or better photos here when I get back home this afternoon/evening.

gertmenke:
Okay, here are more details:

It's helpful to have a triangle-screwdriver:
dsc_0424a
These are the same screws we all love from the Teckin/Blitzwolf sockets.
You *can* use a flat or tri-wing screwdriver, if you are careful, but I'd recommend getting a matching bit, as they are not expensive.

The plastic hooks are easily loosened; I like to use a guitar pick, but a flat screwdriver works just as well.

dsc_0282
On the left is the USB power supply, on the right is the control board.

dsc_0291
Remove the plastic clip that holds the mains cable in place.
The control board is held by 3 Phillips screws.

dsc_0292

Here's the documentation for the TYWE2S module:
https://fccid.io/2ANDL-TYWE2S/User-Manual/Users-Manual-3596121

As far as I can see, GPIO0 can only be reached via a test point on the back, so you'll need to desolder the module.

dsc_0300a

After flashing Tasmota (and preferably making sure that it connects to your WLAN correctly), solder the module back in. When putting the back cover back on, make sure the socket contacts are still in the correct place. (When closed, the back cover holds them in place.)

Well, that's it!

I really like these power strips; even though flashing the firmware is a bit of work at first, it's still faster than flashing 4 or 5 single sockets, and they have the best switchable-sockets-per-euro ratio that I've found so far. Also, they look good on my workbench. :-)

If you have any further questions, just ask!

Jason2866:
Thank you for your detailed how to!
Just my 2 cents, the build quality / design is not the best...

gertmenke:
@Jason2866 
>  Just my 2 cents, the build quality / design is not the best...

I agree, it looks a bit cheap, and the button feels a bit fragile.
But I'm sufficiently confident that it won't burn my house down. :-)

gertmenke:
@Jason2866 
Out of curiosity: What are your main concerns, and how would you improve it?
I think I'd have added an air gap between the mains and the 3V parts of the control board, and the socket contacts could probably be fixated better. But at least they've included a fuse.

Jason2866:
The things you mentioned. Low Voltage to Ac voltage "spaces" wiring and cable management.
Finally soldered connections in power strip, that isnt the best idea. After VDE this isnt allowed....
Better is crimpend or Punktgeschweißt (i dont know the english word)
The Power rails if the got warm maybe get deformed and this makes bad contact 
and than it got hotter....

gertmenke:
The power strip has indeed only the CE symbol, no VDE. (I think the latter is not mandatory in the EU, but it's still interesting to know.)

gertmenke:
It'd be interesting to know about other available options. There are some other power strips on Amazon that look like they're probably ESP8266-based, but this is the only one with 4 sockets that I know of.

gertmenke:
You're probably right about the lottery thing. In my opinion, the original Sonoff S20 have a much better internal design than some of the "no-name" clones, or the Teckin/Blitzwolf ones, for example. (But the latter have power measurement and look at least okay-ish to me.)

Jason2866:
Yes, i have original! Sonoffs too. Build quality is for China good.
I bougt a few days ago a gosund powermeter socket. It is not as bad...
But i never would connect or switch devices which needs much current unattended for a longer time.

gertmenke:
Hi, just to make sure I'm not doin anything wrong here: Is there any kind of protocol I should adhere to, to get my PR merged? Or should I just wait more patiently? :slightly_smiling_face:

ascillato:
Please, be patient :+1:

gertmenke:
Okay! :slightly_smiling_face:

arendst:
Why are you so eager to add this device to Tasmota?
 
The proposed changes can be implemented by using the Generic device easily...

gertmenke:
Of course that's possible - which is true for most of the over 40 devices already in there, I suppose.

It just would be nice not having to look up the correct settings for 7 different GPIOs every time I set up one of these devices. In fact, I do have this patch in my custom builds, but every time you add a device upstream, I get conflicting device id numbers, which can lead to problems...

One time I had to reflash a power strip because the ids changed and the device would not boot properly with the new settings (the same id now being used to indicate another device). Maybe it expected some sensor that was not present? I don't know. It just wouldn't connect to my WLAN any more and I couldn't make it go into setup mode either. Since these things need to be soldered apart for flashing (see above) that was rather annoying.

Now, these are by no means unsolvable problems, but it would still be nice if I wouldn't need to worry about it any more. And, I imagine, others would also benefit from the added convenience of not having to figure out the GPIO settings themselves (or at least not having to google them every time).

I take it you are reluctant to merge this because you don't want to clutter the list of devices too much?

arendst:
It's your last sentence and the fact how many people are using these devices.

Feel free to initiate a poll to find out how many users are using what device.

gertmenke:
I see. Thank you for the feedback.

ascillato2:
Thanks for sharing your ideas. By now, This PR is not going to be merged in order to mantain the device list in the firmware as short as possible. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

fnordsh:
This is actually the same device as the "MANZUKO smart power strip" which was merged in #4590 on Dec. 11, 2018. I'm not sure what made you change your mind, but happy to see it got integrated after all.

PR #3220, opened 2018-07-14
Added Generic ESP8285 module template based on the Generic module

uberflyx:
Added generic ESP8285 module template based on the Generic module except with GPIO 9 and 10 also exposed for GPIO_USER. 

This is useful for ESP8285 based boards such as the Sonoff 4CH when the buttons are used differently (eg. broken out to switches) as the generic module type doesn't allow access to these GPIOs.

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -174,6 +174,7 @@ enum SupportedModules {
   EXS_RELAY,
   WION,
   WEMOS,
+  ESP8285,
   SONOFF_DEV,
   H801,
   SONOFF_SC,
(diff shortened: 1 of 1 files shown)


Jason2866:
Which module has a 8285 and isnt supported?
The 4ch is supported and GPIO 9 and 10 are used in this module
Not supporting GPIO 9 and 10 was a design decision from Theo because more problems in support will result if supporting. 
Not experienced user will try to use 9 and 10 because the dont know which chip the have and the 
might even know that there are different ESP chips out there....
Profis can do whatever the want like you😃👍

uberflyx:
I suppose you are right.. The pins aren't exposed on the M2 module either.
Seems that it's only the 4ch that uses them for the buttons! I guess it's a
bad idea.

ascillato:
Hi

I also agree with Jason and Theo. This will bring more problems than solutions to new users. But I also want to thank you for sharing. This will help anyone, that already knowing that is using this esp type, look for expose those pins.

Issue #2116, opened 2018-03-09
Request: Send / Receive RAW IR data for any device


cyberelectronics:
Hi, 
Using IRrecvDumpV2 example sketch from IRremoteESP8266 library, recommended by David Conran (Thanks!), I was able to find the RAW codes for my pellet stove IR remote control and with IRMQTTServer example sketch I played back succesfuly the RAW data to the pellet stove.

The question is it possible to implement this RAW data Receive/Transmit function over MQTT in Tasmota (no need for decode, just receiving/sending RAW data info over MQTT)?

**Here is the RAW data:**
`Extraflame Ecologica Idro Pellet Stove - IR commands

**Button 1+2 = Power ON / OFF**
Encoding  : UNKNOWN
Code      : 9EB58962 (11 bits)
Library   : v2.3.3

Raw Timing[21]:
   +  6744, -  2462,    +  3396, -  1624,    +   884, -   788,    +   912, -  1610, 
   +  1722, -   814,    +   858, -  1628,    +   884, -   788,    +   884, -  1626, 
   +  1716, -   794,    +  1722, -  1626,    +  1726

uint16_t rawData[21] = {6744, 2462,  3396, 1624,  884, 788,  912, 1610,  1722, 814,  858, 1628,  884, 788,  884, 1626,  1716, 794,  1722, 1626,  1726};  // UNKNOWN 9EB58962

**Button1 = Set Temperature Higher**
Encoding  : UNKNOWN
Code      : 4B3AB9D1 (12 bits)
Library   : v2.3.3

Raw Timing[23]:
   +  6742, -  2464,    +  3420, -  1626,    +   858, -   790,    +   904, -  1642, 
   +  1694, -   814,    +  1696, -   842,    +   832, -   814,    +   860, -  1652, 
   +  1694, -   814,    +   858, -   814,    +   858, -   814,    +  1696

uint16_t rawData[23] = {6742, 2464,  3420, 1626,  858, 790,  904, 1642,  1694, 814,  1696, 842,  832, 814,  860, 1652,  1694, 814,  858, 814,  858, 814,  1696};  // UNKNOWN 4B3AB9D1

**Button2 = Set Power Higher**
Encoding  : UNKNOWN
Code      : 2B890461 (10 bits)
Library   : v2.3.3

Raw Timing[19]:
   +  6796, -  2462,    +  3394, -  1626,    +   910, -   762,    +   886, -  1634, 
[...]

PR #3227, opened 2018-07-15
Added MIDEA/KOMECO IRHVAC protocol

lvta0909:
Added support to MIDEA/KOMECO HVAC protocol.
Its somewhat similar to Mitsubishi hvac protocol.

More information about the protocol / source code reference:
http://veillard.com/embedded/midea.html
https://github.com/sheinz/esp-midea-ir/blob/master/midea-ir.c

### DIFF
--- a/sonoff/xdrv_05_irremote.ino
+++ b/sonoff/xdrv_05_irremote.ino
@@ -46,6 +46,21 @@ IRMitsubishiAC *mitsubir = NULL;
 
 const char kFanSpeedOptions[] = "A12345S";
 const char kHvacModeOptions[] = "HDCA";
+
+// HVAC MIDEA / KOMECO
+// References
+// http://veillard.com/embedded/midea.html
+// https://github.com/sheinz/esp-midea-ir/blob/master/midea-ir.c
+
+#define HVAC_MIDEA_HDR_MARK 4420 // 8T high
+#define HVAC_MIDEA_HDR_SPACE 4420 // 8T low
+#define HVAC_MIDEA_BIT_MARK 553 // 1T
+#define HVAC_MIDEA_ONE_SPACE 1660 // 3T low
+#define HVAC_MIDEA_ZERO_SPACE 553 // 1T high
+#define HVAC_MIDEA_RPT_MARK 553 // 1T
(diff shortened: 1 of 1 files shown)


ascillato2:
Hi,

Thanks a lot for sharing your ideas. 

As Tasmota can't handle ALL possible IR protocols, due to flash size restrictions, we should be better implementing a RAW IR send/receive as requested at #2116.

By now, This PR is not going to be merged due to be outside the actual scope of Tasmota. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

Zuikkis:
I'd like to disagree on this decision.

Tasmota already has many "unnecessary" extensions, like tons of different sensors which are not useful for most people. But it's great to have them in case you need them! They can be enabled and disabled in the config when building, when disabled they don't take up space.

IR_HVAC is default disabled in my_user_config.h, so adding Midea does not take any space.

RAW IR send/receive does not really help for air conditioners, because generally they always have huge IR data packets which cannot be received (or sent?) unless you recompile Tasmota with increased buffer sizes, as explained by many users in #2116..

I have this Wifi IR blaster:
https://github.com/arendst/Sonoff-Tasmota/wiki/YTF-Wifi-Smart-IR---Tuya-Universal-Smart-IR-Hub---TYWE3S-module

Recent Tasmota supports it. With such device it would be great to have a Tasmota firmware that supports as wide range of IR devices as possible, but really is no need to have sensor support etc..

So just let users build what they want. :)

I modified the Midea commit to fit current development branch, and it seems to work nicely. Thanks @lvta0909 !

no-Legend:
@Zuikkis 
I own two mobile hvac's from media (in Germany the company sell as comfee or koenic)
I tried to use the raw command, but I was not able to get the device controlled.
Raw was not working in any kind I tried it.

Is there a way to integrate the midea code in the latest master?

arendst:
Working on it now.

arendst:
Done

no-Legend:
@arendst 
I tested now the protocol for midea. Works good for my midea (comfee,koenic) hvac.

PR #3274, opened 2018-07-20
Added support for fujitsu irhvac

taxx:
Added support for Fujitsu IRHVAC.
Tested with fujitsu units having "AR-RY4" remotes.

### DIFF
--- a/lib/IRremoteESP8266-2.2.1.02/src/IRremoteESP8266.h
+++ b/lib/IRremoteESP8266-2.2.1.02/src/IRremoteESP8266.h
@@ -196,7 +196,7 @@
 #define SEND_MITSUBISHI_AC   true
 
 #define DECODE_FUJITSU_AC    false  // Not written.
-#define SEND_FUJITSU_AC      false
+#define SEND_FUJITSU_AC      true
 
 #define DECODE_DAIKIN        false  // Not finished.
 #define SEND_DAIKIN          false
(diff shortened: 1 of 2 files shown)


ascillato2:
Hi,

Thanks a lot for sharing your ideas. 

As Tasmota can't handle ALL possible IR protocols, due to flash size restrictions, we should be better implementing a RAW IR send/receive as requested at #2116.

By now, This PR is not going to be merged due to be outside the actual scope of Tasmota. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

huexpub:
> Hi,
> 
> Thanks a lot for sharing your ideas.
> 
> As Tasmota can't handle ALL possible IR protocols, due to flash size restrictions, we should be better implementing a RAW IR send/receive as requested at #2116.
> 
> By now, This PR is not going to be merged due to be outside the actual scope of Tasmota. Please, if you have further questions or ideas, do not hesitate on contacting us in the Tasmota Discord Chat. Thanks.

bad news!! now IRremoteESP8266 lib add most new irhvac support， irhvac will be usefull。。。

ascillato:
Why bad news? Instead of supporting just a few protocols, with raw ir can be supported a wider range. All available protocols can't be implemented due to flash and memory restrictions. That is why, a raw IR will be better.

huexpub:
> Why bad news? Instead of supporting just a few protocols, with raw ir can be supported a wider range. All available protocols can't be implemented due to flash and memory restrictions. That is why, a raw IR will be better.

1.raw ir send also no userfull for some ac，such as fujitsu，my  ac to  power-on then raw recived  2  ir codes together , i use openmqttgateway and  esp8266irlib example  send power-on raw  data  for fujitsu It doesn't work. allso i test panasonic ac ，doesn't work too ,but other rc6 codes ,fan work!
2.raw data，how get ? maybe study ? you must get most raw data，and this data  can't  direct use by homeassistant climate and so on，

now i use fujitsu-tasmota and media-tasmota IRHVAC  work for homeassitant  with my custom_components Program very well, so i hope irhvac Integration into compilation for userfull man!

thans, Above personal view and usage experience

PR #3284, opened 2018-07-21
Bugfix MCP230xx driver

andrethomas:
1. Force IO direction register to INPUT if pin is enabled
2. Prevent sensor29 from being used on pins outside of valid pins for MCP230xx variant detected... i.e. 0-7 for MCP23008 and 0-17 for MCP23017


andrethomas:
Error - only enabled for interrupt - must move code :|

PR #3291, opened 2018-07-21
MCP23008/MCP23017 - Remove web config & extend sensor29 command

andrethomas:
1. Removed web configuration option as it uses too much ram in relation to benefit
2. Updated the config structure to use a single pinmode 3 bit value to determine the mode of an input pin which makes it easier to implement in the code
3. Extended the sensor29 command as follows:

`sensor29 reset` - Will reset all pins to input without internal pull-up resistors

`MQT: stat/sonoff/RESULT = {"Sensor29":{"D":99,"MODE":99,"PULL-UP":99}}`

Note the 99's are returned in all values due to the reset confirmation

`sensor29 pin,?` - Will send a response as follows showing the current configuration of a pin

`MQT: stat/sonoff/RESULT = {"Sensor29":{"D":0,"MODE":1,"PULL-UP":0}}`

Existing sensor29 commands for configuration persist as follows, to configure pin 0 as input with pull-up:

`sensor29 0,1,1`

`Response receive: MQT: stat/sonoff/RESULT = {"Sensor29":{"D":0,"MODE":1,"PULL-UP":1}}`

Configure pin 8 (Pin B0 on the chip) as input with interrupt on LOW with pull-up using the command
`sensor29 8,3,1`

And interrupt is reported when the pin goes low as follows

`MQT: stat/sonoff/RESULT = {"Time":"2018-07-21T17:12:08","MCP230XX_INT":{"Pin":"D8", "State":0}}`

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -155,11 +155,9 @@ typedef union {
 typedef union {
   uint8_t data;
   struct {
-    uint8_t enable : 1;                    // Enable INPUT
+    uint8_t pinmode : 3;                    // Enable INPUT
     uint8_t pullup : 1;                    // Enable internal weak pull-up resistor
-    uint8_t inten : 1;                     // Enable Interrupt on PIN
-    uint8_t intmode : 1;                   // Change on STATE or match COMPARATOR
-    uint8_t intcomp : 1;                   // Interrupt COMPARATOR
+    uint8_t b4 : 1;
     uint8_t b5 : 1;
     uint8_t b6 : 1;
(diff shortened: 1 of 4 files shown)


andrethomas:
@arendst D_CONFIGURE_MCP230XX may also now be removed from the language files.

andrethomas:
Wiki at https://github.com/arendst/Sonoff-Tasmota/wiki/MCP23008-MCP23017 updated to remove web config sections and add additional information for the sensor29 command

PR #3309, opened 2018-07-23
ccs811

gemu2015:
completely new try from arends/development


arendst:
Great!

moneybag:
i cant find the ccs811 sensor in the sensor list of tasmota. 
what happens to the wake pin?

PR #3313, opened 2018-07-23
Fix crash via cmd "Status 0"

ckesc:
If you execute `Status 0` you'll get a crash `Soft WDT reset`.
It can be caused by Home Assistant monitoring too

This is because of too long operation in `PublishStatus` funciton (>50ms)
Now we use `yield()` to let Wifi module process its data.

See more at
http://arduino-esp8266.readthedocs.io/en/latest/reference.html?highlight=yield#timing-and-delays


ascillato:
Hi,

I'm not having that crash.

Do you have that issue in the default version?

ascillato:
Have you checked the https://github.com/arendst/Sonoff-Tasmota/issues/3289 ?

There were some binaries with problems due to travis issue.

If you use the recently published binaries by Theo, those are ok.

Or build it by yourself

ckesc:
Yes, I checked.
It seems that my issue is reproducable on Sonoff POW2. 

Bacause: 
- for this device default Baudrate is set to 4800.
- this device have much information to print in response to `Status 0`

All of this causes long execution of `MqttPublishPrefixTopic_P`

Printing one status doesn't take too much time. But if execute all statuses at once - you get long operation, that takes more than 50 ms.
That's why the software watchdog resets the device.

arendst:
Cannot reproduce.

Pls provide output of ``status 0`` as I suspect you are using syslog...

arendst:
Pls try the change just released.

ckesc:
Partially fixed.

There is no crashes now.
But it seems that the device is still doing too much work when generating `Status 0` report.
If I open Tasma Admin (which polls `Status 0` every 5 seconds) and start ping, I got gaps: 2-3 packet loss (see https://gist.github.com/ckesc/7c970b1c57b53f8d7f29070af545ffd4)

And if I change Baudrate to 115200, I get good ping: no packet loss, 1-26 ms

I looks like this bug appeared at 6.x.x versions. 
Maybe you changed something, that made message sending heavier?
Is there is a way to profile perfomance?

ascillato:
> Is there is a way to profile perfomance?

Yes, you can recompile with Debug ON for CPU usage. Please, see the file _xdrv_99_debug.ino_ and enable the key `//#define USE_DEBUG_DRIVER`

ascillato:
Please, provide also your `STATUS 0` output and your findings. Thanks

arendst:
A status 0 every 5 seconds is a disaster waiting to happen...

I know TasmoAdmin uses the HTTP interface to get information but as status 0 results in many MQTT messages (and probably syslog and serial if enables) too it's a major impact on Tasmota performance. If you enable serial AND a low baudrate exceptions are expected as the output just cannot be delivered in time.

You'll have to decide if you want serial output anyway once your device is in prodiction.

As the main problem is solved by introducing a yield I hereby close this PR.

PR #3321, opened 2018-07-24
Opto'd MCP230xx_Show() shaving off another 92 bytes of flash

andrethomas:
Opto'd MCP230xx_Show() shaving off another 92 bytes of flash

### DIFF
--- a/sonoff/xsns_29_mcp230xx.ino
+++ b/sonoff/xsns_29_mcp230xx.ino
@@ -178,16 +178,16 @@ void MCP230xx_Show(boolean json)
 {
   if (mcp230xx_type) {
     if (json) {
-      if (mcp230xx_type == 1) {
+      if (mcp230xx_type > 0) { // we have at least 8 pins
         uint8_t gpio = MCP230xx_readGPIO(0);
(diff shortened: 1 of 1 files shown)


andrethomas:
Compilation differential says codebase = 1900 bytes :)

andrethomas:
@arendst 

Any idea why I get different sizes on Arduino vs platformio ?

On arduino I get a difference of 1900 bytes whereas on platformio I get a difference of 1856 bytes ?

andrethomas:
I see the same when conditionally including or excluding USE_MCP230xx_OUTPUT from branch mcp230xx_output

On platformio it needs +416 bytes and on arduino it needs +440 bytes

Perhaps is something missing on compiler optimization on the arduino core?

arendst:
I've seen patformio optimization too progressive resulting in the Travis failures earlier this week where variables were not on 16 and 32 byte boundaries. That might be the case too between the Arduino and the platformio compiler.

I've seen these differences since at least two years so I don't worry too much.

Do you compile on Linux or Windows? I ask because Travis uses Linux and there I noticed the byte boundary issue.

andrethomas:
@arendst I generally use Windows since most of my other stuff is running on that but am comfortable with linux also... reason I started toying with platformio is because I want to make something like https://nodemcu-build.com/ for the Tasmota firmware which will pull latest code and build a firmware for a user based on specific selections such as which sensor libs to include etc...

PR #3345, opened 2018-07-27
Support send / receive of binary serial data

emontnemery:
Add command serialsend4 which supports sending and receiving raw binary data, including NUL character.
Sending of data is done with binary data in the mqtt message

Received data is sent as hex-strings, example:
`tasmota/sonoff_0FAC39/tele/RESULT:{"SerialReceived":"403f10003c0304430000948583120d"}`
Another option could be to send the received data in a non JSON-endoded message.

The changes can be included also in the softserial bridge in xdrv_08_serial_bridge.ino, please let me know and I'll update the PR with sames changes there.


ascillato:
:+1:

PR #3365, opened 2018-07-29
Add wildcard '+' Character between 'Rule' and 'On' to concatenate new rule with existing rules

JackNJack:
Please add a feature on Rules that able us to concatenate new rule to existing rule without copy and pasting the old rules to update the Rule. Like when I add '+' character between `Rule` and `on`, the new rule will be added to the existing rule in the memory.

Example:
Existing Rule .           :    `on System#Boot do Backlog PowerOnState 1; RuleTime1 15 endon`
At CLI if I enter          :    `Rule + on Rules#Timer=1 do PowerOnState 3 endon`
The Resulting Rule is:     `on System#Boot do Backlog PowerOnState 1; RuleTime1 15 endon on Rules#Timer=1 do PowerOnState 3 endon`


JackNJack:
Thank you guys! Hooray!

PR #3370, opened 2018-07-29
Make interrupt output more rule friendly

andrethomas:
Update code to provide interrupt message in the format that is easier to interpret

`MQT: stat/sonoff/RESULT = {"Time":"2018-07-29T19:28:57","MCP230XX_INT":{"D0":1}}`

### DIFF
--- a/sonoff/xsns_29_mcp230xx.ino
+++ b/sonoff/xsns_29_mcp230xx.ino
@@ -161,7 +161,7 @@ bool MCP230xx_CheckForInterrupt(void) {
                 }
                 if (report_int) {
                   snprintf_P(mqtt_data, sizeof(mqtt_data), PSTR("{\"" D_JSON_TIME "\":\"%s\""), GetDateAndTime(DT_LOCAL).c_str());
-                  snprintf_P(mqtt_data, sizeof(mqtt_data), PSTR("%s,\"MCP230XX_INT\":{\"Pin\":\"D%i\", \"State\":%i}"), mqtt_data, intp+(mcp230xx_port*8), ((mcp230xx_intcap >> intp) & 0x01));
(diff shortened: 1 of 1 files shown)


andrethomas:
@arendst The interrupt responses do not activate rules like

`rule on MCP230XX_INT#D0=1 do power on,ON endon on MCP230XX_INT#D0=0 do power off,OFF endon`

Any thoughts on how I should modify the interrupt routine so that it may be picked up by rules?

Thx!

arendst:
There is no command ``power on,ON`` ?!?!?

Rules only function on JSON messages like {"Level1":{"Level2a":10,"Level2b":9}} as ``on Level1#Level2a=10 do something endon``.

As there can be many triggers keeping the device too busy I have to schedule them. So every 100 mSec one possible trigger is polled and processed by the rule service. Depending on the amount of sensors enabled this soon takes up 2 seconds before the same trigger is send to the rule service again.

Triggers need to be provided as JSON message and the trigger scheduler just round-robin visits the individual sensors FUNC_JSON_APPEND and sends the JSON result to the rule processor.

With this info you might understand that interrupts will never be services by rules when they appear but only when the scheduler is ready to ask for it.

To work around this I use another approach in the tm1638 driver where I send a single changed bit to the switch handler to be serviced immediately as one of max eight switches using their switchtopic/switchmode settings. These can then also be a trigger to rules. Come to think about this you might be able to trigger the rule immediately by a construct I use for these switches too:
```
    snprintf_P(mqtt_data, sizeof(mqtt_data), PSTR("{\"Switch%d\":{\"State\":%d}}"), device, state);
    result = XdrvRulesProcess();
```

Keep in mind that calling the XdrvRulesProcess function too often will definitively cause your device to hang/exception or otherwise do things you do not want.

andrethomas:
> There is no command power on,ON ?!?!?

heh, no there isn't... I think I was tired when I typed :)

I'll toy with your workaround but do not think it will be implementable due to the fact that the user may cause excessive amounts of XdrvRulesProcess()'s due to floating pins with int enabled and polled every 50 milliseconds.

Thanks for the info's, appreciated.

PR #3408, opened 2018-08-04
Fix detection of LM75AD

andrethomas:
Prevent false detection of LM75AD on same addresses as other devices such as the ADS1115 by polling a specific register that has a known value valid only for the LM75AD

### DIFF
--- a/sonoff/xsns_26_lm75ad.ino
+++ b/sonoff/xsns_26_lm75ad.ino
@@ -48,17 +48,19 @@ uint8_t lm75ad_addresses[] = { LM75AD_ADDRESS1, LM75AD_ADDRESS2, LM75AD_ADDRESS3
 
 void LM75ADDetect()
 {
-  uint8_t buffer;
+  uint16_t buffer;
 
   if (lm75ad_type) { return; }
 
   for (byte i = 0; i < sizeof(lm75ad_addresses); i++) {
     lm75ad_address = lm75ad_addresses[i];
-    if (I2cValidRead8(&buffer, lm75ad_address, LM75_CONF_REGISTER)) {
-      lm75ad_type = 1;
-      snprintf_P(log_data, sizeof(log_data), S_LOG_I2C_FOUND_AT, "LM75AD", lm75ad_address);
-      AddLog(LOG_LEVEL_DEBUG);
-      break;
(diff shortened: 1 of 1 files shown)


arendst (approved):

andrethomas:
@arendst Thanks for always merging my stuff so quickly - much appreciated.

I'm not sure who developed the ADS1115 driver but it needs to have something similar to what I did for the LM75AD as it also gets detected if an LM75AD is present and the allsensors code is used... I do not currently have one but I've got one in my shopping card from banggood on the next order from there so will look at it then but if you have one and have time perhaps check if there's a specific register that has a unique value which can verify that the i2c address is indeed used by an ADS1115.

It looks like Config Register would = 0x8583 on power-up/reset so perhaps performing a 16bit read like I did in the LM75AD and comparing it to that of contents 0x8583 to confirm the validity of chip on an address.

I think these kind of scenarios are going to become increasingly useful to have multiple sensor types on the same address ranges whilst Tasmota can correctly detect which device is on which address.

I would make the code change but have no way of testing it at the moment.

andrethomas:
Something like this

```
void Ads1115Detect()
{
  if (ads1115_type) {
    return;
  }

  uint16_t buffer;  
  for (byte i = 0; i < sizeof(ads1115_addresses); i++) {
    ads1115_address = ads1115_addresses[i];
    if (I2cValidRead16(&buffer, ads1115_address, ADS1115_REG_POINTER_CONFIG)) {
      if (buffer == 0x8583) {
        Ads1115StartComparator(i, ADS1115_REG_CONFIG_MODE_CONTIN);
        ads1115_type = 1;
        snprintf_P(log_data, sizeof(log_data), S_LOG_I2C_FOUND_AT, "ADS1115", ads1115_address);
        AddLog(LOG_LEVEL_DEBUG);
        break;
      }
    }
  }
}

```

arendst:
I'm currently on the Moselradweg so can't do much Tasmota wise :-)

Will be back in 8 days (long river LOL).

andrethomas:
Enjoy :-) 👍

Jason2866:
@arendst 
Have a nice ride. If your tour goes through germany there are some nice places too 😃

PR #3440, opened 2018-08-09
Improvement request Sonoff-Tasmota with Cover mode support 4ch (2 groups = 2 ch for cover + 2 ch for lights)

ctripodi:
Hi all
I wanted to use a 4CH lx-wifi-03o switch to control 1 cover (2 channels) and 2 lights with the others 2 channels (4CH  into 2 groups one for lights and 2 channels for the cover motor), The solution needs to prevent, that the up/down motors can be activated the same time.
@arendst  Could be possible to add this feature to the standard version?
Thanks.
Some like this:
a169490ed0d20c2cb80323eaee5e39a88050c05f


arendst:
A PR without useful code is not the means to request new functionality.

PR #3463, opened 2018-08-12
display support for SSD1306 and Waveshare e-paper

gemu2015:
i always wanted small local displays at least with some sensors. but also an overview display with more space.
i inspected the tasmota display branch but found to much overhead there.
so i created my own display driver for SSD1306 and waveshare e-paper

see info and syntax below.

 
[Tasmota displays.txt](https://github.com/arendst/Sonoff-Tasmota/files/2280874/Tasmota.displays.txt)
oled
e-paper


andrethomas:
I think you meant to propose this PR to the display branch? No way it will be included in development branch... way too many changes.

gemu2015:
i only added about 10 lines in 2 files!  i merged several times with the main branch, github doesnt distinguish this 

i still think it would be nice to have display support in the main branch. this could be turned off by default. 
and it only adds about 10 k of code if used.
and by using software 3 line  spi you are not longer restricted to the hardware  SPI lines of the cpu mostly not accessible.

sensor support in the current display branch is obsolete using rules is much more flexible.

andrethomas:
Maybe your fork was just out of date... sync it and then push your changes - see here how to update your fork:

https://gist.github.com/CristinaSolana/1885435

You'll need to specify the branch so

git remote add upstream -t development
and
git pull upstream development

then push back to your fork and do a git push
add your changes
push again
then make PR

PR #3468, opened 2018-08-12
Remove unnecesary flash writes

habek:
If SettingsSave was called with rotate>0, flash was always writed, even if location is set to eeprom flash slot

### DIFF
--- a/sonoff/settings.ino
+++ b/sonoff/settings.ino
@@ -225,20 +225,20 @@ void SettingsSave(byte rotate)
  * stop_flash_rotate 1 = Allow only eeprom flash slot use (SetOption12 1)
  */
 #ifndef BE_MINIMAL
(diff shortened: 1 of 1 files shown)


arendst:
The non-zero SettingsSave rotate option is **incidentally** used to force a save of the settings disregarding the crc check.

The regular SaveSettings is performed with option 0 either for a page or for the fixed EEPROM location. and thus repects the crc check.

In case of a force save from one of the pages to the EEPROM area the crc of the EEPROM area is unknown so a save should always take place.

habek:
I agree with You in generally, but I have a problem with unstable WiFi. When sonoff loss WiFi connection, it is rebooting. In my case, after a problematic night I have thousands of new writes. Besides this I had to remove bootcount incrementation to save flash life.

PR #3478, opened 2018-08-13
update mcp230xx to allow inverted output

MadDoct:
Created pinmode 6 to allow setting inverted output, so it can easily be used with active low relay boards... Also changerd some of the output to make it more readable...

### DIFF
--- a/sonoff/xsns_29_mcp230xx.ino
+++ b/sonoff/xsns_29_mcp230xx.ino
@@ -1,5 +1,5 @@
 /*
-  xsns_29_mcp230xx.ino - Support for I2C MCP23008/MCP23017 GPIO Expander (INPUT ONLY!)
+  xsns_29_mcp230xx.ino - Support for I2C MCP23008/MCP23017 GPIO Expander (INPUT AND OUTPUT)
 
   Copyright (C) 2018  Andre Thomas and Theo Arends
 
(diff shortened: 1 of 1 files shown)


MadDoct:
Why did travis ci fail?

andrethomas:
Not sure what travis was trying to do. It seems to pass building though.

andrethomas:
@MadDoct Question

Do you know if the rules cares about the -'s you added to make it easier for humans to read?

I ask because there's one value which could still be changed

snprintf_P(mqtt_data,sizeof(mqtt_data), PSTR("%s\"OUTD%i\":%s,"),mqtt_data,pinx,stt);
TO
snprintf_P(mqtt_data,sizeof(mqtt_data), PSTR("%s\"OUT-D%i\":%s,"),mqtt_data,pinx,stt);

Of course that would mean something like

on MCP230_OUT#D0 do xxx endon (its current form)

Should work as 

on MCP230_OUT#D-0 do xxx endon

Makes it easier to read, provided the rules code doesnt care about the "-" in the variable name

I'd test it but have no way of doing that from where I am now

MadDoct:
Well, unfortunately I had problems with the "-" in home assistant, so have
changer from using "-" to "_" in my version of the code.. I'll do a pull
request today and will also change it in the output you referenced...
Did you try my implementation of pinmode 6? What did you think of the
coding? (I'm in no way related to IT... I work in health services... ;) LOL)

A terça, 14/08/2018, 14:52, andrethomas <notifications@github.com> escreveu:

PR #3495, opened 2018-08-15
Add option for PlatformIO 2.4.2/1.8.0

reloxx13:
2.4.2 support for platformIO
https://github.com/platformio/platform-espressif8266/releases/tag/v1.8.0


digiblur:
I tried the staging option over the weekend and noticed on two separate esp8266's they will stop responding to any wifi activity after being online for 15 seconds or so.  I tried the new option in this commit today and compiled the firmware and I get the same thing where it only lasts for a short bit.  If I change back to 2.3.0 and recompile then everything works as expected of course.  Are any of you seeing this?

digiblur:
Rude much?

PR #3509, opened 2018-08-17
MCP23008/MCP23017 Enhancements

andrethomas:
```
- Updated telemetry output formatting to use _'s instead of -'s 
  to please HA software
- Added interrupt mode via sensor29 command;
  sensor29 pin,pinmode,pullup,intmode(optional)
  intmode 0 = Immediate report telemetry and cause event
  intmode 1 = Immediate cause event only
  intmode 2 = Immediate report telemetry only
  intmode 3 = Both immediate telemetry and immediate event is disabled
  * This required change in settings.h in Mcp230xxCfg
- OUTPUT telemetry sent when tele_period == 0 ONLY for OUTPUT pins and
  only if PIN is enabled for OUTPUT
- Make OUTPUT pins adhere to Settings.flag.interlock
- Added pinmode 6 for inverted output function
```


andrethomas:
@arendst Is there a callback for something faster than FUNC_EVERY_50_MSECOND ?

I want to report the number of millis() since last time interrupt occurred and I want to implement some form of interrupt priority configuration in magnitudes of 10ms if possible.

The application in my case is time critical because I'm using interrupts to measure the time difference between impulses received from an energy meter :)

andrethomas:
Currently getting this out on interrupt

`MQT: stat/sonoff/RESULT = {"Time":"2018-08-19T00:55:32","MCP230XX_INT":{"D0":0,"MS":325}}`

Math on HA translates to an active reading of 6,923 Watt :)

I guess that's good enough

andrethomas:
```
300       250 ms since last interrupt
3.33        4 impulses per second
200       240 impulses per minute
12000   14400 impulses per hour
7.5         9 kWh

```
Thats a variance of 1.5kWh estimate on time difference if limited to 50ms callback

10ms callback would give me a resolution of 258W
5ms callback would give me a resolution of 127W

PR #3529, opened 2018-08-20
Enable AsyncUDP for KNX (only valid for esplib v2.4.2+)

ascillato:
For KNX:

If using esp library v2.3.0, do not use this PR
If using esp library v2.4.2, this PR can be used for better/faster KNX communication.

This PR just enable the library used for KNX communication (esp-knx-ip) to use the AsyncUDP instead of WifiUDP, making the KNX communication faster. If you enable this, platformio will not compile under v2.3.0. It only works from v2.4.1 and up.

### DIFF
--- a/lib/esp-knx-ip-0.5.1/esp-knx-ip.h
+++ b/lib/esp-knx-ip-0.5.1/esp-knx-ip.h
@@ -7,12 +7,12 @@
 #ifndef ESP_KNX_IP_H
 #define ESP_KNX_IP_H
 
-//#define USE_ASYNC_UDP     // UDP WIFI Library Selection for Multicast
+#define USE_ASYNC_UDP     // UDP WIFI Library Selection for Multicast
                           //   If commented out, the esp-knx-ip library will use WIFI_UDP Library that is compatible with ESP8266 Library Version 2.3.0 and up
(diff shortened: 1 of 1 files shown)


arendst:
Hi Adrian,

can't this be done using a test for the correct core version like:
```
#include <core_version.h>                   // Arduino_Esp8266 version information (ARDUINO_ESP8266_RELEASE and ARDUINO_ESP8266_RELEASE_2_3_0)
#ifdef ARDUINO_ESP8266_RELEASE_2_3_0

#endif
```

ascillato:
Hi,

The idea of using the `#ifdef` is very good, but the main problem is that Platformio will not compile if you use the esplib v2.3.0 and the line 58 in _esp-knx-ip.h_ (`#include <ESPAsyncUDP.h>`) is not commented out. There is an `#ifdef` before that line, but platformio precompile all the libraries before linking (no matter if you are going to use them or not), so, when doing that, Platformio can not find functions of the AsyncUDP Library that are present since the esplib v2.4.1 and up.

That is why, if Tasmota is going to continue the compatibility with esplib v2.3.0, this PR has NOT to be merged. If Tasmota is going to use esplib 2.4.2 and up, this PR can be merged.

ascillato:
As sleep command is not working in esplib v2.4.2, this PR needs not to be merged in order to allow Tasmota be compiled under v2.3.0.

PR #3542, opened 2018-08-21
MCP23008/MCP23017 - update interlocking output to include the pair information

andrethomas:
Update interlocking output to include the pair information


andrethomas:
Ah, I see @MadDoct already made PR directy - perfect, will close those one.

PR #3544, opened 2018-08-21
MCP230xx - Make space for more config options

andrethomas:
Updated mcp230xx_config from 8bit to 16bit to support additional
configuration options to be added.

`uint8_t int_report_defer : 4;`
Number of interrupts to ignore until reporting (default 0, max 15)
Will cause driver telemetry and/or event output until required number
of interrupts have occurred.

`uint8_t int_count_sec : 1;`
0 = Not, 1 = Report number of interrupts since last FUNC_EVERY_SECOND callback 
if more than 0 occured (overrides int_report_mode if enabled)

`uint8_t int_count_min : 1;`
0 = Not, 1 = Report number of interrupts since last FUNC_EVERY_SECOND*60 callback
if more than 0 occured (overrides int_report_mode if enabled)

In the two cases above the plan is to implement a counter like configuration so
FUNC_EVERY_50_MSECOND will still run but will not report immediate telemetry or event
but report the number of interrupts which occurred either in the last 20 FUNC_EVERY_SECOND (1 second)
callbacks, or in the last 60 FUNC_EVERY_SECOND callbacks (1 minute)

Also added uint8_t mcp230xx_int_prio which will allow future configuration of
interrupt priority. Currently, interrupts are polled using the FUNC_EVERY_50_MSECOND
callback - Functionality will be added to implement multiples of FUNC_EVERY_50_MSECOND
in low power applications where the user may require interrupt polling only once per second
for example, then interrupt priority would be set to 20 (20*50ms = 1 second)

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -154,16 +154,16 @@ typedef union {
 typedef union {
   uint16_t data;
   struct {
-    uint8_t pinmode : 3;                   // Pin mode (1 through 6)
-    uint8_t pullup : 1;                    // Enable internal weak pull-up resistor
-    uint8_t saved_state : 1;               // Save output state, if used.
-    uint8_t int_report_mode : 2;           // Interrupt reporting mode 0 = immediate telemetry & event, 1 = immediate event only, 2 = immediate telemetry only
-    uint8_t int_report_defer : 4;          // Number of interrupts to ignore until reporting (default 0, max 15)
-    uint8_t int_count_sec : 1;             // 0 = Not, 1 = Report number of interrupts since last FUNC_EVERY_SECOND callback if more than 0 occured (overrides int_report_mode if enabled)
-    uint8_t int_count_min : 1;             // 0 = Not, 1 = Report number of interrupts since last FUNC_EVERY_SECOND*60 callback if more than 0 occured (overrides int_report_mode if enabled)
-    uint8_t spare13 : 1;
-    uint8_t spare14 : 1;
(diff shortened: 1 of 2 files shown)


arendst:
pls change uint8_t to uint16_t in the settings union.

andrethomas:
Well spotted, Thanks!

ascillato:
```
snprintf_P(log_data, sizeof(log_data), PSTR("your information"));
AddLog(LOG_LEVEL_INFO);
```

andrethomas:
Thx

Check line 1506 on current codebase, should be
`if ((pin[GPIO_SWT1 +i] < 99) || (pin[GPIO_SWT1_NP +i] < 99)) {`

I think you forgot the extra () :)

andrethomas:
In sonoff.ino - forgot to say which file :)

andrethomas:
`if ((pin[GPIO_SWT1 +i] < 99) || (pin[GPIO_SWT1_NP +i] < 99)) {`

andrethomas:
@arendst  or
`if (pin[GPIO_SWT1 +i] < 99 || pin[GPIO_SWT1_NP +i] < 99)`
Better, I think.

arendst:
I love brackets so this is what it is going to be:
```
if ((pin[GPIO_SWT1 +i] < 99) || (pin[GPIO_SWT1_NP +i] < 99)) {
```
Thnx for alerting. Could have been a nasty bug...

andrethomas:
:))

Issue #3314, opened 2018-07-24
MQTT lost connection problem happen after Wifi router restart.


wongnam:
My sonoff basic, sonoff touch, RF bridge are facing issue with MQTT not reconnect after i restart my wifi router.

All my Sonoff devices are using Tasmota version 6.1.1b, core 2.30, MQTT_ESPMQTTARDUINO, self compile by VSC.

symptom:

- Sonoff devices can reconnect to wifi router.
- Sonoff devices webserver can use normally
- Sonoff devices are no response to MQTT server on my Pi 3 B+. (no mqtt reconnection happen)
- restart Sonoff devices-> all will back to normal function as expect.
- Important note:  above issue will happen if i am using option WifiConfig 5

10:23:04 CMD: status 0
10:23:04 MQT: stat/mypc/STATUS = {"Status":{"Module":1,"FriendlyName":["My PC"],"Topic":"mypc","ButtonTopic":"0","Power":0,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"ButtonRetain":0,"PowerRetain":0}}
10:23:04 MQT: stat/mypc/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"http://sonoff.maddox.co.uk/tasmota/sonoff.bin","RestartReason":"Software/System restart","Uptime":"5T15:17:53","StartupUTC":"2018-07-18T12:05:11","Sleep":0,"BootCount":4,"SaveCount":9,"SaveAddress":"FB000"}}
10:23:04 MQT: stat/mypc/STATUS2 = {"StatusFWR":{"Version":"6.1.1b","BuildDateTime":"2018-07-18T17:42:33","Boot":31,"Core":"2_3_0","SDK":"1.5.3(aec24ac9)"}}
10:23:04 MQT: stat/mypc/STATUS3 = {"StatusLOG":{"SerialLog":2,"WebLog":2,"SysLog":0,"LogHost":"","LogPort":514,"SSId":["Multimedia Wi-Fi Network","Wong"],"TelePeriod":300,"SetOption":["00008009","55A18000"]}}
10:23:04 MQT: stat/mypc/STATUS4 = {"StatusMEM":{"ProgramSize":417,"Free":584,"Heap":17,"ProgramFlashSize":1024,"FlashSize":1024,"FlashMode":3,"Features":["00000809","0F083380","0C000000","00000096","00000000"]}}
[...]

PR #3558, opened 2018-08-23
Fix MQTT Reconnection Bug

ascillato:
If you restart your wifi router, and you have **Tasmota_MQTT Library** with `wificonfig 5`, Tasmota don't realize of this _MQTT disconnection_.

This solves issue #3314

### DIFF
--- a/sonoff/xdrv_01_mqtt.ino
+++ b/sonoff/xdrv_01_mqtt.ino
@@ -109,7 +109,6 @@ TasmotaMqtt MqttClient;
 
 bool MqttIsConnected()
 {
-  //return mqtt_connected;
   return MqttClient.Connected();
 }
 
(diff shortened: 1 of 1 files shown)


ascillato:
After several tests of this Tasmota_MQTT Library I found it very reliable. Please, consider adding this library as default. Thanks.

uberflyx:
Thank you so much! This bug has been destroying my life for ages! ;)

meingraham:
Trying to confirm that the newest development branches now default to using PubsClient again, resolve this issue as well. That, as of now, the recommendation is, if we're compiling our own firmware images, to switch back to using PubsClient over TASMOTA_MQTT because it addresses this problem and also has a smaller memory footprint. Is this an accurate statement?

ascillato:
yes, thanks :+1:

PR #3565, opened 2018-08-23
MCP230xx - Initial commit for additional sensor commands

andrethomas:
MCP230xx - Initial commit for additional sensor command settings

sensor29 reset was reworked

Added
```
sensor29 intpri,x // x = 0 to 20 to set multitudes of 50ms callbacks to skip before interrupt is called
```

The following is added for setting only but does not yet have impact - will add code for logging info as the functionality of these inhibit/alter/disable other settings.

```
sensor29 intdef,pin,x // defferred interrupt for pin (only setting implemented - function will be pushed later
sensor29 intsec,pin,x // 0=disable(default), 1=enable - only setting code - function will be pushed later
sensor29 intmin,pin,x // 0=disable(default), 1=enable - only setting code - function will be pushed later
```

### DIFF
--- a/sonoff/xsns_29_mcp230xx.ino
+++ b/sonoff/xsns_29_mcp230xx.ino
@@ -57,6 +57,10 @@ uint8_t mcp230xx_address;
 uint8_t mcp230xx_addresses[] = { MCP230xx_ADDRESS1, MCP230xx_ADDRESS2, MCP230xx_ADDRESS3, MCP230xx_ADDRESS4, MCP230xx_ADDRESS5, MCP230xx_ADDRESS6, MCP230xx_ADDRESS7, MCP230xx_ADDRESS8 };
 uint8_t mcp230xx_pincount = 0;
 uint8_t mcp230xx_int_en = 0;
+uint8_t mcp230xx_int_prio_counter = 0;
+uint8_t mcp230xx_int_report_defer_counter[16] = {0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0};
+uint16_t mcp230xx_int_count_sec[16] = {0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0};
+uint16_t mcp230xx_int_count_min[16] = {0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0};
 
 unsigned long int_millis[16]; // To keep track of millis() since last interrupt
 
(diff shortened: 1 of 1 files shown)


andrethomas:
On second thought, I guess I should rework most of the mqtt responses to settings to just provide logging output.. will it use less flash ?

arendst:
It only uses less flash if the same text is configured with PROGMEM

ascillato:
Using the String() function a lot, uses a lot of memory and can result in an exception reboot. Try to reduce it as much as possible.

andrethomas:
@ascillato Yeah, thats what I am afraid of... let me close this then and rework using the existing subStr function which I think you originally wrote and then moved to the support.ino - so it's already there in the code, may as well use it. It may mean I need to rethink how the sensor29 command parameters work but in the interest of saving ram and possibly some flash it may be worth it?

andrethomas:
Closing PR to re-work MCP230xx_Command() to move away from String objects.

ascillato2:
The main problem with string() is that is not fixed the amount of memory that it uses. So, it allocate memory (in the subroutine you called) and until that subroutine is ended, that RAM is used. So, any other command or another string instance, can lead to an exception reset. You will not see that in the report of platformio after compilation because all this problem is done on the fly while it is being executed in the ESP8266.

So, at first you may think that using char arrays and subStr() was a lost of time, but now you have avoided A LOT of future issues.

Glad you change it.

andrethomas:
Yeah, it seems a lot cleaner and I took a slightly different approach to command parameter detection, like:
```
  if (!strcmp(sub_string,"RESET"))  sensor29_cmnd = MCP230xx_CMND_RESET;
  if (!strcmp(sub_string,"SET"))    sensor29_cmnd = MCP230xx_CMND_SET;
  if (!strcmp(sub_string,"INTPRI")) sensor29_cmnd = MCP230xx_CMND_INTPRI;
  if (!strcmp(sub_string,"INTDEF")) sensor29_cmnd = MCP230xx_CMND_INTDEF;
  if (!strcmp(sub_string,"INTSEC")) sensor29_cmnd = MCP230xx_CMND_INTSEC;
  if (!strcmp(sub_string,"INTMIN")) sensor29_cmnd = MCP230xx_CMND_INTMIN;
  if (!strcmp(sub_string,"POWER"))  sensor29_cmnd = MCP230xx_CMND_POWER;
```

Then I can switch() through to the command parameter and process from there.

Only downside is that some commands will change on output functionality.

`sensor29 pin,ON/OFF/T`
becomes
`sensor29 POWER,pin,ON/OFF/T`

But thats probably also for the better.

andrethomas:
Oh, and setting a pin config from
`sensor29 pin,pinmode,pullup,intmode`
to
`sensor29 set,pin,pinmode,pullup,intmode(optional)`

andrethomas:
@ascillato Maybe its because its late, but any idea why this kind of method segfaults?

```
const char MCP230XX_RESPONSE_SET[] PROGMEM = "{\"Sensor29_D%i\":{\"INT%s\":\"%s\"}}";

const char _txt_off[] PROGMEM = "OFF";
const char _txt_on[] PROGMEM = "ON";
const char _txt_toggle[] PROGMEM = "TOGGLE";

const char* const txtonoff[] PROGMEM = {
  _txt_off,
  _txt_on,
  _txt_toggle
};

snprintf_P(mqtt_data, sizeof(mqtt_data), MCP230XX_RESPONSE_SET,pin,"MIN",txtonoff[Settings.mcp230xx_config[pin].int_count_min]);
```

It works if I use something like this:

```
const char* OnOffTxt(uint8_t onoff) {
  switch (onoff) {
    case 0:
      return "OFF";
      break;
    case 1:
      return "ON";
      break;
    case 2:
      return "TOGGLE";
      break;
  }
}

```

ascillato:
I don't know why that doesn't work. I just use as Theo has done in the drivers.

arendst:
You cannot use progmem for parameter data like you try with textonoff.

Forget textonoff and use my routine for user defined on/off/toggle/hold messages. Look for examples in sonoff.ino regarding SetOption result message.

andrethomas:
Thx

andrethomas:
In settings.h we have the Settings structure, which on the current codebase = 3584 bytes in size (obviously the linker script is aligning the bytes to 4-byte words)

In current codebase we have:
```
Mcp230xxCfg   mcp230xx_config[16];       // 6F6
uint8_t       mcp230xx_int_prio;         // 716

byte free_717[183];                      // 717
```
Needing an additional uint16_t for the interrupt timer I need to add it as

```
Mcp230xxCfg   mcp230xx_config[16];       // 6F6
uint8_t       mcp230xx_int_prio;         // 716
uint16_t      mcp230xx_int_counter;      // 717
  
byte          free_720[180];             // 720
```
Note that free_ has decreased by 3 bytes - I have to do this otherwise the linker script causes Settings to increase to 3588 because the last 8 bits of the newly defined uint16_t flows into the next 4 bytes aligned word.

Is there specific way in which this is handled for people contributing or is it just a case of having to manually check it when changes are made so that the size does not change because as we know if the size changes the device will reset to default CFG on next boot after flashing the new firmware with a larger sized Settings structure?

arendst:
People have to have knowledge of compiler optimization and variable boundaries.

A uint8_t, byte or char (8 bits) can be placed anywhere.
A uint16_t or int16_t (16 bits) has to be placed on an address at 16-bit boundary
A long or int (32 bits) has to be placed on an address at 32-bit boundary

So in your case the address has to be 718.

So for the Settings structure this means:
```
byte       free_717[1];            // 717
uint16_t   mcp230xx_int_timer;     // 718

byte       free_71a[180]           // 71A
```

andrethomas:
Cool. that works - so there was a 2 byte boundry issue somewhere further up or down - basically the free[180] should always be divisible by 4, unless the alignment is further down in the Settings struct.

andrethomas:
Duh, no wait. I understand what you are saying now :)

arendst:
Nope.

The alignment error was you placing a uint16_t next to mcp230xx_int_prio. The unit16_t could not be placed next to mcp230xx_int_prio as that address was not a 16-bit boundary. The compiler selects the next possible boundery and adds a free byte.

andrethomas:
Realised a second before you posted :)

PR #3579, opened 2018-08-24
MCP230xx - For the love of brackets :))

andrethomas:
Special commit for you @arendst


MadDoct:
Why did you do this?! The braces introduced bugs because they are in the wrong location in several places...

arendst:
Should work fine now.

Bracketing is an art ;-)

andrethomas:
Hi, sorry was offline until now.

I did momentarily have a wtf moment when I changed this particular one
`if ((interlock && pinmo) == Settings.mcp230xx_config[pin+pinadd].pinmode)`
and thought I would test it before doing the PR but it somehow escaped my mind.

Apologies for the hassle and thanks for fixing it up nicely... see you made a few nice cosmetic changes also :)

Thanks again!

PR #3597, opened 2018-08-26
MCP230xx - Resolve compiler warnings

andrethomas:
IntModeTxt()
Force return "" if unknown value was sent as parameter

ConvertNumTxt()
Force return "" if an unknown value was sent as a parameter

MCP230xx_CheckForInterrupt()
change return from bool to void since the function does not return anything

MCP230xx_SetOutPin()
resolve compiler warnings when pin shifts are done with values outside of the parenthesis


andrethomas:
And yes, tested interlock

0 or 1
2 or 3
4 or 5
6 or 7

PR #3630, opened 2018-08-29
Non-blocking wait()-function added.

renne:
@arendst @ascillato2 @andrethomas 
This commit adds a wait()-function to wait for a fixed number of milliseconds without blocking. Blocking operation is optional (second parameter "blocking" = "true").

!!! Blocking operations should be avoided wherever possible. !!!

### DIFF
--- a/sonoff/support.ino
+++ b/sonoff/support.ino
@@ -717,6 +717,29 @@ void ShowSource(int source)
   }
 }
 
+/**
+ * Delay function.
+ *
+ * !!! Avoid blocking wherever possible !!!
+ * 
+ * @param   milliseconds  Milliseconds to wait.
+ * @param   blocking      Blocking if "true", non-blocking if "false" (default).
+ * 
+ * @return                Always "true" to allow concatenating of logical operators.
+ */
+bool wait(uint32_t milliseconds, bool blocking = false) {
+  if (blocking) {
+    delay(milliseconds);
+  } else {
+    uint32_t endpoint = millis() + milliseconds;
+    while(millis() < endpoint) {
+      yield();
+    }
+    return true;
+  } // if
+} // wait
+
+
 /*********************************************************************************************\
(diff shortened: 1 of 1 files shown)


renne:
@arendst @ascillato2

Is there any interest in such a wait()-function?

I need my fork for sensor development.

andrethomas:
I recommend that you do this inside your 50MS callback of the driver itself. I doubt it will have any use outside of your driver... also doubt that we want people to start using a function that blocks other functions from running - remember Tasmota is not a RTOS - just my 2c ;)

arendst:
I agree with @andrethomas

renne:
@arendst @andrethomas
`delay()` blocks other tasks. The while loop with `yield()` does the same like `delay()` but does **not block** other tasks. The `block_other_tasks` parameter is false by default and meant only in case it is really necessary to block other tasks (atomic operations, etc.).

arendst:
In addition to @andrethomas I find this code in core_esp8366_wiring.c used by Tasmota providing the code for delay:

```
void delay_end(void* arg) {
    (void) arg;
    esp_schedule();
}

void delay(unsigned long ms) {
    if(ms) {
        os_timer_setfn(&delay_timer, (os_timer_func_t*) &delay_end, 0);
        os_timer_arm(&delay_timer, ms, ONCE);
    } else {
        esp_schedule();
    }
    esp_yield();
    if(ms) {
        os_timer_disarm(&delay_timer);
    }
}
```

andrethomas:
@renne What @arendst found above confirms, esp8266 core we're using is an RTOS in itself and what we see as our main loop is just another "task" in the RTOS.

renne:
The Arduino API documentation listed `delay()` as blocking. Obviously the Expressif guys did a better job. It's good to know `delay()` is non-blocking and should not be used in atomic operations.

PR #3665, opened 2018-09-01
Update README.md

andrethomas:
Updated readme

The development branch is checked hourly for changes and if new commits were made it will be rebuilt on the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ If you like **Sonoff-Tasmota**, give it a star, or fork it and contribute!
 
 See [RELEASENOTES.md](https://github.com/arendst/Sonoff-Tasmota/blob/development/RELEASENOTES.md) for release information and [sonoff/_changelog.ino](https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/_changelog.ino) for detailed change information.
 
-The compiled development versions from current codebase are built around 6AM GMT+2 everyday and posted at http://thehackbox.org/tasmota/ (this web address can be used for OTA too). Remember that these are based on the current development codebase and it is not recommended to flash devices in hard to reach places with this firmware.
+The development codebase is checked hourly for changes and if new commits have been merged and compile successfuly they will be posted at http://thehackbox.org/tasmota/ (this web address can be used for OTA too). It is important to note that these are based on the current development codebase and it is not recommended to flash it to devices used in production or which are hard to reach in the event that you need to manually flash the device if OTA failed. The last compiled commit number is also posted on the same page along with the current build status (if a firmware rebuild is in progress).
 
 ### Disclaimer
 :warning: **DANGER OF ELECTROCUTION** :warning:


ascillato:
Cool!

andrethomas:
I was bored... sometimes when bored cool things may happen.

It works! :)

image

arendst:
@andrethomas That looks very good!

But...

Would it be possible to rename the 1m0 files to their default names? The current naming makes OTA-Magic unusable. OTA-Magic uses the minimal version if the requested image does not fit in free space. The mechanism used is as follows:
- It copies OTAUrl and from the back of the OTAUrl it searches the name for a dash (-) and if present it removes the text after the dash.
- It then inserts the text **minimal** and tries to load the minimal version.
- Once the minimal version is loaded it will use the saved OTAUrl to load the requested image.

As you renamed the files using the postfix 1m0 this mechanism will fail.

My suggestions are one of the below where the first probably is most useful:
- remove the 1m0 from the 1M images but leave the 4m0 in place as they will never need OTAMagic
- place the binaries in a separate folder for either 1m0 and 4m0
- do nothing but tell users that OTAMagic will fail as long as the 1m0 versions have a 1m0 postfix.

andrethomas:
@arendst 

> remove the 1m0 from the 1M images but leave the 4m0 in place as they will never need OTAMagic

That sounds like the best plan overall, so I've made the code changes and its currently building the last commit so you can confirm in 10mins or so if the file naming is correct :)

andrethomas:
```
xplg_wemohue.ino: In function 'void HandleUpnpEvent()':
xplg_wemohue:392:5: error: expected ';' before '}' token
     }
     ^
```

There is a bug in the current codebase :|

arendst:
You're right. I failed to compile. Will update.

arendst:
Fixed.

andrethomas:
Forcing rebuild :)

arendst:
Great work did regarding the provision for development releases.

andrethomas:
I hope it will reduce the number of issues opened by people who cannot install a dev environment but want something from the development codebase :)

arendst:
Did you manage to alignment issue?

andrethomas:
The alignment issue seems to be related to platformio and not Linux - compiling with Arduino on Linux the file size is still different to that on Windows but I saw at least one confirm of someone failing with platformio on exact same release where the binaries from the development binaries compiled by Arduino on Linux seem to work - so not totally ruled out but looks like it may be a step in the right direction

andrethomas:
Build complete - please check if filenames conform to requirements?

arendst:
Great. Looks OK.

andrethomas:
Updated table output so that it is more clear to understand which binaries to use

image

arendst:
Or make two table....

andrethomas:
You mean one for 4M's and one for 1M's ?

arendst:
Yep.

andrethomas:
Makes sense... the sorting was confusing me also

andrethomas:
Done

arendst:
Checked. OK!

PR #3704, opened 2018-09-05
add wifi network info to the default display screen

yhfudev:
This PR is for display the WiFi network information such as IP address, MAC, SSID, host name etc to the default display screen to check if the device is connected properly after re-connection.


arendst:
As the ip address and RSSI values are not yet present after a restart I have to make them optional. Will also change the code to implement it for all display types. It will show starting with displaymod 2 and up.

To see the IP address and RSSI values just execute the displaymode command again after wifi is connected.

Issue #3700, opened 2018-09-05
fatal exception on setting KNX PA


jpadie:
**Describe the bug**
when setting a public address the sonoff crashes and reboots with a fatal exception 28.
core 2.3.0
tasmota v 6.2.0.2

_Also, make sure these boxes are checked [x] before submitting your issue - Thank you!_
- [x] _Searched the problem in issues and in the wiki_
- [x] _Hardware used_ :  sonoff basic
- [x] _Provide the output of command_``status 0`` :
```
{"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"***","RestartReason":"Exception","Uptime":"0T00:01:46","StartupUTC":"2018-09-05T08:43:02","Sleep":0,"BootCount":2,"SaveCount":5,"SaveAddress":"F7000"}}
{"StatusFWR":{"Version":"6.2.0.2","BuildDateTime":"2018-09-05T10:15:59","Boot":31,"Core":"2_3_0","SDK":"1.5.3(aec24ac9)"}}
{"StatusLOG":{"SerialLog":2,"WebLog":2,"SysLog":0,"LogHost":"","LogPort":514,"SSId":["Livebox-A217","Gaia-Rabastens-Free"],"TelePeriod":300,"SetOption":["02008009","55818000","00000000"]}}
{"StatusMEM":{"ProgramSize":420,"Free":580,"Heap":15,"ProgramFlashSize":1024,"FlashSize":1024,"FlashMode":3,"Features":["00000809","1D003080","000003A1","21021796","00000000"]}}
{"StatusNET":{"Hostname":"bedswitch-2077","IPAddress":"192.168.1.25","Gateway":"192.168.1.1","Subnetmask":"255.255.255.0","DNSServer":"192.168.1.1","Mac":"60:01:94:AD:88:1D","Webserver":2,"WifiConfig":0}}
{"StatusMQT":{"MqttHost":"mqtt.no3.co.uk","MqttPort":8883,"MqttClientMask":"DVES_%06X","MqttClient":"DVES_AD881D","MqttUser":"***","MqttType":3,"MAX_PACKET_SIZE":1000,"KEEPALIVE":15}}
[...]

PR #3710, opened 2018-09-05
Fix Exception Issue

ascillato:
Solves Bug #3700

### DIFF
--- a/sonoff/support.ino
+++ b/sonoff/support.ino
@@ -143,7 +143,7 @@ char* subStr(char* dest, char* str, const char *delim, int index)
   int i;
 
   // Since strtok consumes the first arg, make a copy
-  strlcpy(dest, str, strlen(str));
+  strncpy(dest, str, strlen(str));
   for (i = 1, act = dest; i <= index; i++, act = NULL) {
     sub = strtok_r(act, delim, &ptr);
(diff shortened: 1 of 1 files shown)


andrethomas:
@ascillato 

I still can't successfully extract the last parameter without a trailing delimiter:

```
command: '0,?'
paramcount: 2
param #1 '0'
param #2 '?@'
```

Change line
`strncpy(dest, str, strlen(str));`
to
`strncpy(dest, str, strlen(str)+1);`

Fixes it to:

```
command: '0,?'
paramcount: 2
param #2 '0'
param #2 '?'
```

ascillato:
Yes, it stills need +1 in length before calling `substr()`

In _xdrv11_knx.ino_ file, this was done before calling the function, like for example line 1159: `char sub_string[XdrvMailbox.data_len +1];`

What it is not needed anymore is the extra delimiter.

With the small change that this PR does, it reverts a commit that modify substr() and result in an exception for KNX driver.

andrethomas:
Yeah I know that, but for last parameter strtok_r(act, delim, &ptr); doesnt pull the string terminator, hence you need to do strlen(str)+1 in subStr() so that the \n is copied also.

Issue #3628, opened 2018-08-29
Tasmota needs sound !!


gemu2015:
i created an MP3 driver for this extremely cheap MP3 Player => DFPlayer mini (<2 Euros, 3 watt audio)
3 commands:

MP3Play track     track =>1-N
MP3Volume volume  volume => 0-100
MP3Stop

you simply place any mp3 files on a micro sd card and can then play it
i use it for sensor alarms when e.g. air quality is too bad
and i want to use it for extending my haus alarm siren
you could even think of an alarm clock with different alarm sounds or even musik.

it consumes nearly no resources, needs only one pin (software uart transmit)

attached is the (tiny) driver file. simply add it to the project (currently uses fixed GPIO 16 as transmit)

[xdrv_98_mp3.ino.zip](https://github.com/arendst/Sonoff-Tasmota/files/2332006/xdrv_98_mp3.ino.zip)

arendst:
Ordered one. Now where do I get the music from....

mike2nl:
gemu2015 has optimized his mp3 driver 18 hours ago:
https://github.com/gemu2015/Sonoff-Tasmota/blob/display-drivers/sonoff/xdrv_98_mp3.ino

He has added some defines, changed the void for mp3 play and changed the beloning code for the defines.

Here the file: 
[xdrv_98_mp3.zip](https://github.com/arendst/Sonoff-Tasmota/files/2334782/xdrv_98_mp3.zip)

mike2nl:
BTW:
- we need a 3.3K .. 10K resistor when we will work with RX and TX in both lines.
- in single mode is a 10K better before the RX pin from the player.
- or use a level shifter because the player won't play with 3.3VDC. Min is around 4.3VDC.
- direct connect can be risky (info from internet).

mike2nl:
gemu2015:
strange because when i use 3,3 VDC i get these rare effect that the player stops after a second
and the blue LED goes off. I can be possible that my 3.3VDC regulator, see photo please, is
not 100% ok but it works with other 3.3VDC hardware. So then my player must be damaged.

The 3.3V is stable until 800mA that means that something must be wrong. Will see in it what's
going wrong. 
And yes with a 3.3VDC power supply you need no resistor (data sheet).

Thank's for your feedback.

andrethomas:
> The 3.3V is stable until 800mA

That's 2.64 watts of power... in theory a 8 ohm speaker should not use more than 413mA @ 3.3V

[...]

PR #3723, opened 2018-09-07
Added DFPlayer Mini MP3 Player

mike2nl:
Based on the issue Tasmota needs sound !! #3628 from gemu2015 i have added all 
needed changings including the xdrv_91_mp3.ino driver. Renamed it because there 
is a lot of stuff in the range around 96, 97 and 98.  The GPIO is now selectable and
no longer fixed to GPIO 16 as in the original driver xdrv_98_mp3.ino. Formated the 
code a little bit for better and faster reading.

Right now only three commands are prossible:
e.g. MP3Play 001 means first track
e.g. MP3Volume 60 set volume to 60% = (60 * 30 / 100)
e.g. MP3Stop yeah stops the playing

Link to the issue:
-> https://github.com/arendst/Sonoff-Tasmota/issues/3628

### DIFF
--- a/sonoff/language/bg-BG.h
+++ b/sonoff/language/bg-BG.h
@@ -456,6 +456,7 @@
 #define D_SENSOR_I2C_SCL  "I2C SCL"
 #define D_SENSOR_I2C_SDA  "I2C SDA"
 #define D_SENSOR_WS2812   "WS2812"
+#define D_SENSOR_DFR562   "MP3 Player"
 #define D_SENSOR_IRSEND   "IRsend"
 #define D_SENSOR_SWITCH   "Ключ"   // Suffix "1"
 #define D_SENSOR_BUTTON   "Бутон"   // Suffix "1"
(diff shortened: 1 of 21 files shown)


crxporter:
Lately I've been struggling to choose between ESPeasy and Tasmota for my wemos and sonoff boards. This issue is the final one that I have keeping one of my chips on ESPeasy - the MP3 player.

There are two things- first I'd love to see the ability to send music to TWO DFPlayers on one wemos board. I have this all soldered up and running on ESP easy (but for many reasons I'd like to be back with tasmota, mostly because no ESP easy ever stays connected to wifi for more than 2 or 3 days). To get it running I duplicated the plugin and renamed everything so it would show up as a second player allowing me to choose a second pin. I'm sure I could do this in tasmota but I don't know how.

Second- the 3 commands provided are almost enough- the way I understand the player is you can have any one of the following:
-255 audio files in the root folder (play up to 255 files)
-3000 (I think) MP3 files in a folder called MP3
-255 audio files each in folders 1-100

I would be very interested in the second of those options (3000 MP3 files). I'll be anxiously watching future release notes for more development on this!

PR #3784, opened 2018-09-12
Let syslog to be active when emulation is active too

ascillato:
solves issue https://github.com/arendst/Sonoff-Tasmota/issues/2109


mrohner:
What's the solution now?

PR #3818, opened 2018-09-16
MCP230xx - Change address selection from auto to user defined

andrethomas:
Some I2C liquid crystal displays fall within the same range registered for the MCP23008/MCP23017 so this PR removes the automatic address detection and allows the user to specify the actual address in the user_config.h file (or other methods they use)

With this, the user can specify the I2C address (most development boards seem to use 0x20) and avoid unwanted data to be written to 0x27 which is a common address for PCF based I2C LCD modules.


andrethomas:
@Jason2866 Wiki updated also.

MadDoct:
@andrethomas, this means that you can only have one mcp and not several... Having the option to have several would be ideal...

andrethomas:
@MadDoct I believe 16 additional GPIO is sufficient for now - adding more instances will significantly increase RAM demand so even if this was done it would not be merged to this development branch.

Also, I have other plans for the other I2C addresses provided by the MCP chip ;)

andrethomas:
Also, if you look at the cpu time used by the check for interrupt routine you will see that it will be tight to make multiple devices viable within the 50ms interrupt polling ability.

PR #3852, opened 2018-09-20
LM75AD - Force function parameters to be C and C++ compliant

andrethomas:
So a bit of a theory.

I was thinking about what we experienced here https://github.com/arendst/Sonoff-Tasmota/issues/3678#issuecomment-419727653 wherein removing the LM75AD driver was changing the behaviour of the code.

In C it is perfectly fine to have a function like

void LM75ADDetect()
{
}

and the C compiler will treat it that this function does not have a parameter.

In C++ this is not acceptable for the case of no parameters because it is intended to allow you to pass any number of unknown variables or types.

My theory is that because we use C and C++ code mixed in the project that perhaps the platformio compiler recipe is not correct insofar that it may be trying to link functions like void LM75ADDetect() with a pointer to a potential parameter being cast to the function - of course there is no handling inside the function for this as would be required in C++ so it leads to a pointer being created that may be pointing to absolutely nothing and could lead to unexpected behaviour.

To explicitly define a function that has no parameters in C++ one has to do

void LM75ADDetect(void)

This, of course, is also valid and accepted by a C compiler and is treated the same as the compiler would treat void LM75ADDetect() during compilation/linking.

So with this theory in mind, I propose these changes to the LM75AD driver and see if someone can reproduce the problems from issue 3678 after compiling on platformio and not getting a reboot loop or other problems.

Either way, it cannot break anything and its probably better to use (void) in all cases anyway.

If this does solve the problem it proves that the recipe imported by the platformio esp8266 core needs to be looked into to make sure that C and C++ are segregated correctly and linked appropriately.

Just a shot in the dark, but worth the try I think.

### DIFF
--- a/sonoff/xsns_26_lm75ad.ino
+++ b/sonoff/xsns_26_lm75ad.ino
@@ -46,7 +46,7 @@ uint8_t lm75ad_type = 0;
 uint8_t lm75ad_address;
 uint8_t lm75ad_addresses[] = { LM75AD_ADDRESS1, LM75AD_ADDRESS2, LM75AD_ADDRESS3, LM75AD_ADDRESS4, LM75AD_ADDRESS5, LM75AD_ADDRESS6, LM75AD_ADDRESS7, LM75AD_ADDRESS8 };
 
-void LM75ADDetect()
+void LM75ADDetect(void)
 {
   if (lm75ad_type) { return; }
 
@@ -64,7 +64,7 @@ void LM75ADDetect()
   }
 }
 
-float LM75ADGetTemp() {
+float LM75ADGetTemp(void) {
   int16_t sign = 1;
 
   uint16_t t = I2cRead16(lm75ad_address, LM75_TEMP_REGISTER);
--- a/sonoff/xsns_29_mcp230xx.ino
+++ b/sonoff/xsns_29_mcp230xx.ino
@@ -181,7 +181,7 @@ void MCP230xx_ApplySettings(void) {
   MCP230xx_CheckForIntCounter(); // update register on whether or not we should be counting interrupts
 }
 
-void MCP230xx_Detect()
+void MCP230xx_Detect(void)
 {
   if (mcp230xx_type) {
     return;


andrethomas:
Updated the mcp detect also, just for fun.

andrethomas:
I was hoping for rubber underpants ;)

andrethomas:
@arendst 

Some interesting results. I did the same platformio compilation with the exact same copy of the source with only one difference - the one has (void) and the other is () in cases where functions do not have parameters.

image

[test.zip](https://github.com/arendst/Sonoff-Tasmota/files/2409164/test.zip)

The files are exactly the same size in bytes, but the content is not mapped the same.

andrethomas:
Interesting to note the offset in alignment

image

aquaticus:
A better way to analyse such problems is to generate linker map than just compare binaries.
Add the following options to gcc:
`-Xlinker -Map=program.map`
and compare for 2 compilation variants.

andrethomas:
> -Xlinker -Map=program.map

Tried those, but platformio.ini does not seem to pass them. I guess it would need to be done in the build.py file, not sure.

andrethomas:
Seems to accept -Wl,-Tesp8266.flash.1m0.ld -Wl,-Map,output.map

andrethomas:
@aquaticus 
Recompiled both variants and included their .bin and .map files
[test.zip](https://github.com/arendst/Sonoff-Tasmota/files/2409198/test.zip)

andrethomas:
image

andrethomas:
So function() vs function(void) does make a difference... question is if the difference is good or not.

arendst:
Nice research!

When I check Adafruit cpp they mostly use void as it should. Looking in the Arduino core though it's more like a lottery; some functions have void, other have not.

So when we change Tasmota to using void it's still not clear if it's useful as most core is 50/50.

aquaticus:
There are differences in binaries/map because `foo()` and `foo(void)` are different functions. So this is not surprising.

The question is why changing signature of the function (or disable LM75AD module) make WiFi working.
If you are using `foo()` without any tricks there should be no issues and the program should work the same with `void` or without it. You could face the problem if you're getting more parameters from the stack than it was actually passed (there is no way to know number of args passed).
For example, you declare function as `void foo()` and call it as `foo(1,2)` which is perfectly fine for C and then your assembler implementation of that function takes 3 params from stack instead of 2. That is obviously bad.

I am suggesting that this change actually only moved the problem somewhere else. The issue was just hidden by changing function signature.
Easy prove for that would be disabling any of other modules and see if it helps. If yes, the problem is not related to `void`, if not -- we still won't be sure.

The only valid declaration of parameter-less function in C is using `void` as argument. For C++ it does not matter.

andrethomas:
Yes, it could be moved somewhere else because of this.

In fact, the binaries worked fine on the previous release with LM75AD enabled so it's not directly related to the LM75AD driver.

Most of the adafruit libraries are C++ based so in those cases having functions without parameters as () would be fine for C++.

If we can get a device which malfunctions on the original 6.2.1 compiled with platformio to fail again with different variations of including/excluding drivers then maybe it could be proved or disproved.

I think it may also be worth exploring what is inside

./platforms/espressif8266@1.5.0/platform.py
./platforms/espressif8266@1.5.0/builder/main.py

as compared to the platform.txt and boards.txt we have for Arduino which appears to compile and link correctly with the same xtensa binaries (compiler, linker etc)

Maybe its just there that we need to make sure that the recipe for platformio's imported esp8266 core is using the same compiler flags/directives/parameters as we have for arduino.

andrethomas:
So here is almost identical (except for bytes that show build time/date) compilation with (void) vs () under Arduino - So something is in my platform.txt (For Arduino) that is not in ./platforms/espressif8266@1.5.0/builder/main.py - most likely C compliance related parameters - it causes the output to be the same for the arduino compiling irrespective whether () or (void)

[arduino.zip](https://github.com/arendst/Sonoff-Tasmota/files/2410306/arduino.zip)

andrethomas:
Another interesting change.

If you clone the source repo and go into the directory and do:
platformio run
It downloads and installs the 1.5.0 core, as expected and compiles.

```
PlatformManager: Installing espressif8266 @ 1.5.0
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
espressif8266 @ 1.5.0 has been successfully installed!
The platform 'espressif8266@1.5.0' has been successfully installed!
The rest of packages will be installed automatically depending on your build environment.
PackageManager: Installing toolchain-xtensa @ ~1.40802.0
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
PackageManager: Installing tool-esptool @ ~1.409.0
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
PackageManager: Installing framework-arduinoespressif8266 @ ~1.20300.1
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
CorePackageManager: Installing tool-scons @ ~2.20501.4
Downloading  [####################################]  100%
Unpacking  [####################################]  100%
tool-scons @ 2.20501.4 has been successfully installed!
```

After compiling it pops up the following innocent looking suggestion

```
There are the new updates for platforms (espressif8266)
Please update them via `platformio platform update` command.

If you want to manually check for the new versions without updating, please use `platformio platform update --only-check` command.
```

Doing this removes the original build and link recipes and then uses a new one from a new folder going forward.

Very interesting part is the change in the LINKFLAGS defined prior to the update and after the update

Before
```
    LINKFLAGS=[
        "-Os",
        "-nostdlib",
        "-Wl,--no-check-sections",
        "-u", "call_user_start",
        "-Wl,-static",
        "-Wl,--gc-sections"
    ],

```
After
```
    LINKFLAGS=[
        "-Os",
        "-nostdlib",
        "-Wl,--no-check-sections",
        "-Wl,-static",
        "-Wl,--gc-sections",
        "-u", "call_user_start",
        "-u", "_printf_float",
        "-u", "_scanf_float"
    ],
```

So something has changed on the esp8266 core imported by platformio and this may not be the only change!

andrethomas:
So here goes another test.

Compiled binary prior to the innocent "platformio platform update" suggestion and a binary for the same codebase after the suggested "platformio platform update"

image

I think we can confirm now that the change from a core folder to what seems like a framework method has changed the way platformio is creating the binaries for Tasmota - This also explains why everything worked fine until support for 2.4.2 was added (even though I used 2.3.0 in all test scenario's above)

Two binaries of before and after attached for interest sake.
[after_upgrade.zip](https://github.com/arendst/Sonoff-Tasmota/files/2412137/after_upgrade.zip)
[before_upgrade.zip](https://github.com/arendst/Sonoff-Tasmota/files/2412138/before_upgrade.zip)

aquaticus:
@andrethomas Do you know what are the versions of the tools that produce working binary even without `void` change?

andrethomas:
@aquaticus 
Yes, in both cases exactly the same:
version 4.8.2 (crosstool-NG 1.20.0) on the sane Linux machine.

andrethomas:
Seems the same, except for the symlinks
Left side is platformio's binaries it downloaded, right side is arduino's binaries downloaded
image

andrethomas:
When compiling with Arduino, the output is different from platformio... one can see different parameters passed to gcc and g++

See attached - I'm not sure yet how to add the additional ones to platformio but it looks like it can be done in the platformio.ini file but does not seem to allow specific for gcc and g++

[piolog.zip](https://github.com/arendst/Sonoff-Tasmota/files/2412630/piolog.zip)
[arduinobuild.zip](https://github.com/arendst/Sonoff-Tasmota/files/2412631/arduinobuild.zip)

andrethomas:
```
PIO
xtensa-lx106-elf-g++ -o .pioenvs/sonoff/firmware.elf -Os -nostdlib -Wl,--no-check-sections -u call_user_start -Wl,-static -Wl,--gc-sections -Wl,-wrap,system_restart_local -Wl,-wrap,register_chipv6_phy -Wl,-Tesp8266.flash.1m0.ld .pioenvs/sonoff/src/core_esp8266_timer.c.o .pioenvs/sonoff/src/core_esp8266_wiring_digital.c.o .pioenvs/sonoff/src/core_esp8266_wiring_pwm.c.o .pioenvs/sonoff/src/sonoff.ino.cpp.o -L/home/tasmota/.platformio/packages/framework-arduinoespressif8266@1.20300.1/tools/sdk/lib -L/home/tasmota/.platformio/platforms/espressif8266@1.5.0/ldscripts -L.pioenvs/sonoff -Wl,--start-group .pioenvs/sonoff/lib6c9/libWire.a .pioenvs/sonoff/libc4f/libSPI.a .pioenvs/sonoff/lib488/libAdafruit-GFX-Library-1.2.9.a .pioenvs/sonoff/lib53d/libAdafruit_ILI9341-1.2.0-Tasmota-1.0.a .pioenvs/sonoff/lib4e5/libIRremoteESP8266-2.2.1.02.a .pioenvs/sonoff/lib6a0/libArduinoJson-5.11.2.a .pioenvs/sonoff/lib08f/libesp-epaper-29-ws-20171230-gemu-1.0.a .pioenvs/sonoff/lib3ee/libOneWire-2.3.3.06.a .pioenvs/sonoff/libac5/libESP8266WiFi.a .pioenvs/sonoff/lib8f6/libESP8266WebServer.a .pioenvs/sonoff/lib235/libTasmotaSerial-2.0.0.a .pioenvs/sonoff/lib9e5/libAdafruit_LED_Backpack-1.1.6.a .pioenvs/sonoff/libfb7/libTicker.a .pioenvs/sonoff/libba7/libBME680_driver-bme680_v3.5.9.a .pioenvs/sonoff/lib941/libPubSubClient-2.6.09.a .pioenvs/sonoff/libade/libMutichannel_Gas_Sensor.a .pioenvs/sonoff/lib0d5/libNeoPixelBus-2.2.9.a .pioenvs/sonoff/lib4a4/libI2Cdevlib-Core.a .pioenvs/sonoff/libcce/libI2Cdevlib-ADS1115.a .pioenvs/sonoff/libe34/libESPAsyncUDP-master.a .pioenvs/sonoff/lib01f/libEEPROM.a .pioenvs/sonoff/lib401/libesp-knx-ip-0.5.1.a .pioenvs/sonoff/lib6b8/libESP8266mDNS.a .pioenvs/sonoff/lib46a/libLiquidCrystal_I2C-1.1.3.a .pioenvs/sonoff/libc7c/libC2Programmer-1.0.0.a .pioenvs/sonoff/libf52/libTasmotaMqtt-1.1.1.a .pioenvs/sonoff/libfaf/libESP8266HTTPClient.a .pioenvs/sonoff/libb54/libAdafruit_SSD1306-1.1.2.a .pioenvs/sonoff/libc8d/libESP8266httpUpdate.a .pioenvs/sonoff/libbaa/libArduinoOTA.a .pioenvs/sonoff/lib582/libJoba_Tsl2561-2.0.7.a .pioenvs/sonoff/lib7c7/libAdafruit_SGP30-1.0.0.13.a .pioenvs/sonoff/libbab/libAdafruit_CCS811-1.0.0.14.a .pioenvs/sonoff/lib4bd/libDNSServer.a .pioenvs/sonoff/libcc0/libMPU6050.a .pioenvs/sonoff/libad8/libesp-mqtt-arduino-1.0.1.02.1.a .pioenvs/sonoff/libFrameworkArduinoVariant.a .pioenvs/sonoff/libFrameworkArduino.a -lmesh -lwpa2 -lsmartconfig -lpp -lmain -lwpa -llwip_gcc -lnet80211 -lwps -lcrypto -lphy -lhal -laxtls -lgcc -lm -lstdc++ -Wl,--end-group

Arduino
xtensa-lx106-elf-gcc -g -Os -nostdlib -Wl,--no-check-sections -u call_user_start -Wl,-static -L/theo/arduino/portable/packages/esp8266/hardware/esp8266/2.3.0/tools/sdk/lib -L/theo/arduino/portable/packages/esp8266/hardware/esp8266/2.3.0/tools/sdk/ld -Teagle.flash.1m0.ld -Wl,--gc-sections -Wl,-wrap,system_restart_local -Wl,-wrap,register_chipv6_phy -o ../../../../build/sonoff.ino.elf -Wl,--start-group /theo/build/sketch/core_esp8266_timer.c.o /theo/build/sketch/core_esp8266_wiring_digital.c.o /theo/build/sketch/core_esp8266_wiring_pwm.c.o /theo/build/sketch/sonoff.ino.cpp.o /theo/build/libraries/esp-knx-ip-0.5.1/esp-knx-ip-config.cpp.o /theo/build/libraries/esp-knx-ip-0.5.1/esp-knx-ip-conversion.cpp.o /theo/build/libraries/esp-knx-ip-0.5.1/esp-knx-ip-send.cpp.o /theo/build/libraries/esp-knx-ip-0.5.1/esp-knx-ip-webserver.cpp.o /theo/build/libraries/esp-knx-ip-0.5.1/esp-knx-ip.cpp.o /theo/build/libraries/EEPROM/EEPROM.cpp.o /theo/build/libraries/ESP8266WiFi/ESP8266WiFi.cpp.o /theo/build/libraries/ESP8266WiFi/ESP8266WiFiAP.cpp.o /theo/build/libraries/ESP8266WiFi/ESP8266WiFiGeneric.cpp.o /theo/build/libraries/ESP8266WiFi/ESP8266WiFiMulti.cpp.o /theo/build/libraries/ESP8266WiFi/ESP8266WiFiSTA.cpp.o /theo/build/libraries/ESP8266WiFi/ESP8266WiFiScan.cpp.o /theo/build/libraries/ESP8266WiFi/WiFiClient.cpp.o /theo/build/libraries/ESP8266WiFi/WiFiClientSecure.cpp.o /theo/build/libraries/ESP8266WiFi/WiFiServer.cpp.o /theo/build/libraries/ESP8266WiFi/WiFiUdp.cpp.o /theo/build/libraries/ESP8266WebServer/ESP8266WebServer.cpp.o /theo/build/libraries/ESP8266WebServer/Parsing.cpp.o /theo/build/libraries/Ticker/Ticker.cpp.o /theo/build/libraries/ESP8266HTTPClient/ESP8266HTTPClient.cpp.o /theo/build/libraries/ESP8266httpUpdate/ESP8266httpUpdate.cpp.o /theo/build/libraries/DNSServer/DNSServer.cpp.o /theo/build/libraries/ESP8266mDNS/ESP8266mDNS.cpp.o /theo/build/libraries/Wire/Wire.cpp.o /theo/build/libraries/PubSubClient-2.6.09/PubSubClient.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/IRrecv.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/IRsend.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/IRtimer.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/IRutils.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Aiwa.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Argo.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Coolix.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Daikin.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Denon.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Dish.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Fujitsu.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_GlobalCache.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Gree.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_JVC.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Kelvinator.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_LG.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Mitsubishi.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_NEC.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Nikai.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Panasonic.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Pronto.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_RC5_RC6.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_RCMM.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Samsung.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Sanyo.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Sharp.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Sherwood.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Sony.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Trotec.cpp.o /theo/build/libraries/IRremoteESP8266-2.2.1.02/ir_Whynter.cpp.o /theo/build/libraries/C2Programmer-1.0.0/c2.cpp.o /theo/build/libraries/C2Programmer-1.0.0/ihx.cpp.o /theo/build/libraries/TasmotaSerial-2.0.0/TasmotaSerial.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/HsbColor.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/HslColor.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/HtmlColor.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/HtmlColorNameStrings.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/HtmlColorNames.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/HtmlColorShortNames.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/NeoEsp8266UartMethod.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/NeoGamma.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/NeoPixelAnimator.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/NeoPixelAvr.c.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/NeoPixelEsp.c.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/RgbColor.cpp.o /theo/build/libraries/NeoPixelBus-2.2.9/internal/RgbwColor.cpp.o /theo/build/libraries/SPI/SPI.cpp.o /theo/build/libraries/Adafruit_SGP30-1.0.0.13/Adafruit_SGP30.cpp.o ../../../../build/arduino.ar -lm -lgcc -lhal -lphy -lpp -lnet80211 -lwpa -lcrypto -lmain -lwps -laxtls -lsmartconfig -lmesh -lwpa2 -llwip_gcc -lstdc++ -Wl,--end-group -L../../../../build
```

Interesting that platformio uses g++ and Arduino uses gcc to do this last part.

aquaticus:
This is linker call, so it really does not matter if it's gcc or g++. Linker is the same in both cases.

PR #3866, opened 2018-09-21
PCA9685 I2C Hardware PWM - Experimental initial commit

andrethomas:
With reference to issue https://github.com/arendst/Sonoff-Tasmota/issues/3830

This adds support for the PCA9685 hardware PWM board which adds 16 x 12bit PWM capable outputs via I2C.

The device resets to 50hz (commonly used by servos) but can be changed with commands as indicated below:

From console, commands:
```
driver15 pwmf,x // Where X is PWM frequency from 24Hz to 1526Hz
driver15 pwm,pin,pwmvalue // Where pin = 0 to 15 and pwmvalue = 0 to 4096
driver15 pwm,pin,ON // Turns pin ON
driver15 pwm,pin,OFF // Turns pin OFF

```
pwmvalue of 0 will turn the pin off completely whereas 4096 will turn the pin on completely.

Settings are not perpetual (yet, perhaps if there is a demand for this we can add it) so will revert to default on reset/power-on.


andrethomas:
```
driver15 pwmf,1000 // 1khz
driver15 pwm,0,2048 // enable 50% duty cycle

```
image

;)

PR #3905, opened 2018-09-26
'decode-config.py' version bugfixes & json output enhancements

curzon01:
- add new arguments '--json-indent' and '--json-compact'
- fix poblem with invalid string chars when having garbadge in flash
- fix errors using configs for Tasmota < v6.2.1
- fix setting definition for v5.13.1/v5.14.0


ascillato:
Cool!

PR #3908, opened 2018-09-26
Fix logic for forced light discovery

emontnemery:
In home-assistant, a switch can be turned on/off and a light can be turned on/off but may also have additional capabilities such as brightness, color etc.

Home-assistant can handle simple relays announced as light, but it can't handle a light announced as a switch.
Fix handling of setoption30 to reflect this.

Suggested wiki update:

| Command | Payload | Description |
| --- | --- | --- |
| SetOption30 | 0 / off | Automatic announcement as 'switch' for relays and 'light' for PWM in Home-Assistant MQTT discovery message. |
| SetOption30 | 1 / on | Enforce Home-Assistant discovery as 'light' platform also for relays. |


ascillato:
Done.

Thanks for share :+1:

PR #3921, opened 2018-09-28
Udated the driver veml6070

mike2nl:
Lots of things changed for a better code, so far men can say that.
Looks more professional.
json output added and changed. Bug from version 1.0.0.1 is out
web server output changed for less code. Better to handle now
Added FUNC_EVERY_SECOND so as advised, 

Theo please have a look on:
void Veml6070EverySecond(void)
{
  // all = 10..15[ms]
  if (11 == (uptime %100)) {

All tested, plus timing. Timings are added in [ms]. 
So far the driver works much better and faster at start ;-).

In i18n.h i have added that #define D_JSON_UV_INDEX_TEXT "UvIndexText"
for the json and web server output.


arendst:
Well done. Nice to see small timings.

I think I have to get me a sensor too ;-)

andrethomas:
@arendst Get 2 - that way you can know which side of the house is the safest to exit from ;)

PR #3931, opened 2018-09-29
v1.5.0011: 'decode-config.py' fixes & enhancements

curzon01:
- add configuration data header to json and binary outputs
--add template @v and @f for output filename
- add output of crc values on crc error
- add '--exit-on-error-only' arg make it possible to process partly valid data
- replaced '--sort <word>' by '--unsort'
- changed data size and data crc error into warnings
- changed output always in json format
- removed arg '--format' (obsolete)
- removed collections module (obsolete)

v1.5.0010: 'decode-config.py' add more detailed outputs

- add bit structure for struct TimeRule, Timer, Mcp230xxCfg
- add program return code desc
- add error handling for additonal python modules
- changed 'Settings' definition to handle array of struct
- changed field conversion string to formular evaluation for values
- fix and enhance output values

### DIFF
--- a/tools/decode-config.py
+++ b/tools/decode-config.py
@@ -19,23 +19,25 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
 Requirements:
      - Python
      - pip install json pycurl urllib2 configargparse
 
+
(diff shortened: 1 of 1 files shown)


curzon01:
Wrong source

PR #3947, opened 2018-09-30
Update function for pca9685 driver.

labastov:
Update function for pca9685 driver.

* rename driver to sensor (xdrv_15 to xsns_33)
* added display information on webgui (function FUNC_WEB_APPEND) (+0k112 code).
* added send jonson information (FUNC_JSON_APPEND)
* added key #defined PCA9685_DEFAULT_FREQ.

Could you insert files to project.
* xsns_33_pca9685.ino
* user_config.h (udated)


andrethomas:
Hi

I don't think we want to import additional libraries for this driver when code can be added directly to the driver for specific functionality required - i.e. we want to keep dead code (code that is not being used) outside of the driver. This is why it was done in the simplistic way that it was... to keep its footprint as small as possible.

Lastly, this is a driver and not a sensor so changing the naming convention is incorrect in my opinion.

With 53 files changed I do not see any specific benefits documented in the description of your PR.

What could be useful is:
#define PCA9685_DEFAULT_FREQ

I am tentative on whether or not we want web appends as the webui is mainly designed for configuration purposes and not necessarily intended for dashboard purposes - for monitoring during implementation of some purpose you want it may always be observed from the console during setup.

The telemetry data may be useful to some users but it would simply be reporting what a user or HA system has already sent to the driver so again in the name of saving flash and ram may or may not be critical to this specific driver.

arendst:
In addition the proposed PR is littered with local code so this won't be merged.

labastov:
Sorry for my English.
1. The Driver  xdrv_interface.ino has not support FUNC_JSON_APPEND and i need rename to sensor.
2. Sent information over mqtt need for home automatization. (I use OH2). Only sent packet for change parametet is not enough.
2. Webgui have higt intormation for easy answer and I made it as option enble over key USE_PCA9685_DISPLAYOUTPUT.

labastov:
I could not split the branches and delete the extra files

andrethomas:
I will add define #define USE_PCA9685_FREQ option for you and enable telemetry json output. I think that will satisfy your needs?

Web UI information is a no-go area for me - it is not intended for diagnostic/driver control purposes and should be limited to sonoff device support and at most sensor support.

labastov:
need merge only two files 
xsns_33_pca9685.ino    - driver
user_config.h    - coments for keys.

labastov:
by default webgui - disable, but key USE_PCA9685_DISPLAYOUTPUT  ley any user  enable webgui information. 
This is your project. You can make any resolution, but  key USE_PCA9685_DISPLAYOUTPUT  made your project more functionaly.

mike2nl:
About the PR from labastov:
- It is not an idea to have a folder/directory named Playground or NoSupportAddOns?
- To put in there tested and running stuff but with no support!
- The user who has written that driver must give the support.
-- If he won't do that no transfer to that directory.
- The suer has to hold the driver uptodate based on changings from tasmota, so not, delete it.
- Or the driver was not used in 12 month delete it.

andrethomas:
@labastov 

Perhaps for your information, in addition to outputting the telemetry on telemetry period configured as mentioned before PR https://github.com/arendst/Sonoff-Tasmota/pull/3954 also adds a new command
`driver15 status`
Which causes an immediate response with the same JSON provided by telemetry so it is useful for configuration validation.

If you need more features (other than the web ui which was ruled out) please feel free to open a feature request issue.

Thanks for your ideas and contribution so far, much appreciated.

PR #3949, opened 2018-09-30
PCA9685 - Add #define USE_PCA9685_FREQ

andrethomas:
Add support for user to set default frequency to be used for PWM in user_config.h


andrethomas:
Hang tight, adding telemetry.

andrethomas:
@arendst Thanks, am done with these changes for now - I will not be implementing web ui status because the information is now presented in telemetry json.

arendst:
Indeed refrain from implementing any web gui for this.

When doing the telemetry in here it won't be possible for rules to act on it. This can only be done as sensor using the FUNC_JSON_APPEND call. It also won't show up in status 8 or 10. This makes it a bit of a messy driver...

Anyway. I'll merge and see what comes out of it.

andrethomas:
@arendst  Understood.

What are the implications of adding FUNC_JSON_APPEND support to the driver interface - I'm not particularly keen on calling this a xsns because it is not a sensor - but maybe this does not matter?

andrethomas:
Also doubt that the output would serve any purpose than to update values in a HA system - Unlikely to use rules based on a specific pin's PWM value, for example.

arendst:
Adding FUNC_JSON_APPEND to drivers will increment the overall rule scheduling which I'm not fond of.

In the latter case just leave it as it is.

PR #3954, opened 2018-10-01
PCA9685 - bugfix for telemetry + status command

andrethomas:
Fix bug for updating pca9685_pin_pwm_value[pin] so that telemetry data is valid, otherwise all pin values were 0.

Also, added new command
`driver15 status`

Which causes the same JSON created for telemetry to be returned immediately.


andrethomas:
Thx Theo

I asked on Discord, but I don't think you go there often but I asked if you would like to add andrethomas2 for access to help Adrian manage/close issues - not sure if you saw it or if you don't agree ;)

PR #3960, opened 2018-10-02
Create Hebrew Language File

rt400:
Hi ,
i Create new language file base on english file..
hope you can add it .


rt400:
thanks..
also need to make change in user_config.h.
and there is away also to make firmware bin file in hebrew language ?

PR #3993, opened 2018-10-06
Update cs-CZ.h

synekvl:
I have tried to make a test edit as per your description and I updated the line 31 correcting Tasmota version... I hope everything is correct without any harm to your excellent work.

Cheers


arendst:
Congratulations!

Issue #2871, opened 2018-05-31
Real Time Clock


Lawrencezarb:
From my understanding the Timer function relies on the Sonoff unit have access to a NTP server.

If communication is lost to the NTP or the Wifi I assume that the Timers will cease to function as the Sonoff will not know what time it is.

Again I assume the Sonoff has no inbuilt clock to hold the time once it gets its Time from the NTP.

If this is correct would it be possible to add a Real Time Clock board to the Sonoff as used in a Raspberry pi?

ascillato:
> From my understanding the Timer function relies on the Sonoff unit have access to a NTP server.

Yes

> If communication is lost to the NTP or the Wifi I assume that the Timers will cease to function as the Sonoff will not know what time it is.
> Again I assume the Sonoff has no inbuilt clock to hold the time once it gets its Time from the NTP.

If the wifi or internet is lost **after** Tasmota sync the time from internet, the timers will continue working. The ESP8266 will continue counting the time. It has the internal hardware to do that. The problem is that the Clock in the ESP8266 is not backed-up with a battery, so with a reset or a power loss, it will forget the time.
If Tasmota boots and can not reach a Time server, the timers will never trigger. Because, Tasmota don't know the time at startup.

> If this is correct would it be possible to add a Real Time Clock board to the Sonoff as used in a Raspberry pi?

Raspberry Pi don't have a builtin Real Time Clock either. It has the same problem that the ESP8266. Can keep track of the time but it is not battery backed-up when losing power. You can add a RTC to a Raspberry as explained in this [link](https://learn.adafruit.com/adding-a-real-time-clock-to-raspberry-pi/overview)

About the RTC on ESP8266 there is an article in this [link](http://www.instructables.com/id/TESTED-Timekeeping-on-ESP8266-Arduino-Uno-WITHOUT-/) that explains the ESP8266 precision for the Time.

And at this moment Tasmota does not have support for an external RTC hardware for reading and saving the time backed by a battery and to be used when there is no Wifi or Internet Connection.

Lawrencezarb:
Thanks for the information and clarificaton. With regards to my Raspberry Pi RTC, I was referring the an external RTC module that can be added to the Pi. 

It would be interesting to see this implemented.

trogper:
You could set up a NTP server on your Raspi, so the ESP would get it's time from there.

fritserasmus:
> 

[...]

PR #4029, opened 2018-10-10
Add support for DS3231 Extrnal RTC with I2c interface

guyelg:
Tested with nodeMcu 1.0.
sensor number 33

### DIFF
--- a/sonoff/user_config.h
+++ b/sonoff/user_config.h
@@ -322,6 +322,8 @@
       #define MTX_ADDRESS6     0x76              // [DisplayAddress6] I2C address of sixth 8x8 matrix module
       #define MTX_ADDRESS7     0x00              // [DisplayAddress7] I2C address of seventh 8x8 matrix module
       #define MTX_ADDRESS8     0x00              // [DisplayAddress8] I2C address of eigth 8x8 matrix module
+  #define USE_DS3231                          // Enable use DS3231 external RTC , usefall when you don't have avaliable WIFI. see docs in the source file (+1k2 code)
+//  #define USE_RTC_ADDR                      //you can change the addrsss of the DS3231 RTC, default is 0x68, not mandatory fieled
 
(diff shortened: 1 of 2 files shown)


sonoff/user_config.h
```diff
@@ -322,6 +322,8 @@
       #define MTX_ADDRESS6     0x76              // [DisplayAddress6] I2C address of sixth 8x8 matrix module
       #define MTX_ADDRESS7     0x00              // [DisplayAddress7] I2C address of seventh 8x8 matrix module
       #define MTX_ADDRESS8     0x00              // [DisplayAddress8] I2C address of eigth 8x8 matrix module
+  #define USE_DS3231                          // Enable use DS3231 external RTC , usefall when you don't have avaliable WIFI. see docs in the source file (+1k2 code)
```
ascillato: This need to be commented. By default it is not being used by sonoff devices.

sonoff/user_config.h
```diff
@@ -322,6 +322,8 @@
...
+//  #define USE_RTC_ADDR                      //you can change the addrsss of the DS3231 RTC, default is 0x68, not mandatory fieled
```
ascillato: Please, add at least the default address to be left as an example for address syntax. If left alone can be confused with an option instead that a number required.

ascillato (changes requested):

sonoff/user_config.h
andrethomas: Another option is to combine the two

`#define USE_DS3231   0x68`

ascillato:
Hi,

By default this driver needs to be disabled. Please, update that for user_config.h

ascillato:
This solves #2871

ascillato (approved):

andrethomas (approved):

ascillato:
@guyelg

Thanks :+1:

PR #4031, opened 2018-10-11
Fix compiler error on minimal firmware

ascillato:


arendst:
I'm out but I think it's better to move WaitForReset in webserver to above ifdef be_minimal.

andrethomas (approved):

arendst:
Just fixed it another way.

andrethomas:
Yeah, seems logical ;) Thx

PR #4062, opened 2018-10-14
Add Reset 4 command to reset defaults but keep wifi config

andrethomas:
This PR solves https://github.com/arendst/Sonoff-Tasmota/issues/4061 by adding command

reset 4

Which resets the device to defaults in user_config settings, but retains what was configured for ssid1,ssid2,password1 and password2.


ascillato (approved):

ascillato:
Good idea reset 4 command.

ascillato:
@jziolkowski

Yes. This PR implements only the command reset 4.

The web GUI remains the same so as to keep flash and memory as low as possible.

andrethomas:
@jziolkowski Yes, as indicated only command - remember you can still issue this command from the web console also, so there's no need to have any additional features added to the web interface.

PR #4080, opened 2018-10-16
fix RF bridge raw receive

Portisch:
fix receive len and fix the 0x55 uart sync

### DIFF
--- a/sonoff/xdrv_06_snfbridge.ino
+++ b/sonoff/xdrv_06_snfbridge.ino
@@ -301,7 +301,7 @@ void SonoffBridgeReceived()
 boolean SonoffBridgeSerialInput()
 {
   // iTead Rf Universal Transceiver Module Serial Protocol Version 1.0 (20170420)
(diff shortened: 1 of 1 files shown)


arendst:
Thx

No offend but what is the difference between:

```
if (serial_in_byte_counter == 3) { 
        if ((0xA6 == serial_in_buffer[1]) || (0xAB == serial_in_buffer[1])) {  // AA A6 06 023908010155 55 - 06 is receive_len 
```
and
```
if (serial_in_byte_counter == 3) { 
        if ((0xA6 == serial_in_buffer[serial_in_byte_counter - 2]) || (0xAB == serial_in_buffer[serial_in_byte_counter - 2])) {  // AA A6 06 023908010155 55 - 06 is receive_len 
```
other than that it costs more code space?

Portisch:
should also work but serial_in_byte_counter is counter. so the buffer should be handled with this buffer.
Just a "nice" coding issue.

PR #4095, opened 2018-10-17
Add support for different variants of TUYA-dimmer

chaosmaster:
There are many Dimmers using the TUYA serial-protocol, they use different configurations for addressing the dimmer, as well as reset-functions (Button-presses) etc.

This adds support for other Dimmers using the TUYA protocol as mentioned in #469 

This patch adds a new option to select the dimmer ID used in the serial communication.
It can either be set at compile-time via:
`#define TUYA_DIMMER_ID <id>`

Or at runtime via:
`TuyaDimmerId <id>`

The ID defaults to **3** to not break stuff.

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -246,6 +246,7 @@
 #define D_CMND_SERIALDELIMITER "SerialDelimiter"
 #define D_CMND_BAUDRATE "Baudrate"
 #define D_LOG_SOME_SETTINGS_RESET "Some settings have been reset"
+#define D_CMND_TUYA_DIMMER_ID "TuyaDimmerId"
 
 // Commands xdrv_01_mqtt.ino
 #define D_CMND_MQTTHOST "MqttHost"
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -334,6 +334,7 @@ struct SYSCFG {
   uint16_t      web_refresh;               // 7CC
   char          mems[MAX_RULE_MEMS][10];   // 7CE
   char          rules[MAX_RULE_SETS][MAX_RULE_SIZE]; // 800 uses 512 bytes in v5.12.0m, 3 x 512 bytes in v5.14.0b
+  uint8_t       tuya_dimmer_id;
                                            // E00 - FFF free locations
 } Settings;
 
--- a/sonoff/settings.ino
+++ b/sonoff/settings.ino
@@ -419,6 +419,11 @@ void SettingsDefaultSet2()
   // Module
 //  Settings.flag.interlock = 0;
   Settings.module = MODULE;
+#ifdef TUYA_DIMMER_ID
+  Settings.tuya_dimmer_id = TUYA_DIMMER_ID;
+#else
+  Settings.tuya_dimmer_id = 3;
+#endif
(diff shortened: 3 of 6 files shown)


ascillato (approved):

arendst:
Considering the amount of "patching" I'm preparing a separate driver file for this hardware based on yours and digiblur code.

It will also use command SetOption34 to control the dimmer id.

So hold on for a few days to prepare a driver file with the code. Once released I think it's easier for you to make it work and I have the chance to disable the whole thing easily.

chaosmaster:
@arendst 
That sounds good, I was actually considering doing this before digiblurs code got merged.
May I ask, why not use a named option instead of SetOption34?
Also would be great to keep an option to set it at compile-time.
Thanks for all your great work!

arendst:
As you noticed the named option impacts several areas of the code. The SetOption framework is already in place and doesn't require newly reserved settings space as it is already reserved.

Hold on for the update today...

chaosmaster:
Great, thanks for explaining, I'll stay tuned :)

arendst:
Give it a go.

BTW Why do you want to change wificonfig to 2 using dedicated code? It's a no-go.

chaosmaster:
I'll check it out.
On the tuya-dimmers the reset-signal is controlled by the MCU.
Depending on the variant, it is either sent as a serial-packet or it pulls up GPIO_0.
It isn't a separate button though and gets activated using a specific procedure on the MCU.
(On the one i have, turn off the dimmer and press "-" for +-15 seconds)
So we can't have a "hit once for smartconfig, twice for AP" kind of thing.
That's why AP is hardcoded.
I think the original software actually switches modes on each 15-sek button hold.
What would you suggest to do here?

digiblur:
Wow.  I had a few minutes to look over the changes. Great work!  I like the modular approach with the driver as it definitely lessens the impact of things as we move forward and only make this better.  I'll do some tests this evening (US time here).

arendst:
What happens if you do not configure GPIO0 as a button as did digiblur?

In that case no button press is detected. If you still need a kind of button configure it as a switch.

chaosmaster:
> Give it a go.
> 
> BTW Why do you want to change wificonfig to 2 using dedicated code? It's a no-go.



> What happens if you do not configure GPIO0 as a button as did digiblur?
> 
> In that case no button press is detected. If you still need a kind of button configure it as a switch.

If it's not configured then the I can't reset the wifi-configuration.
It needs to detect GPIO_0 in order to reset wifi.

digiblur:
Just an FYI. On my variant, GPIO0 isn't even soldered to anything off of the little ESP8266 Tuya module.  The ON/OFF button on the device including the touch sensors on the faceplate to go up and down are handled by the dimming MCU.  The ESP8266 is basically just the WiFi radio of the device and done all over TX/RX pins.  I also had to watch for the 5 second hold (defined in the dimming MCU) when it would send over the WiFi reset packet to push Tasmota into the wifi manager mode.  On my variant, the switch would still work with all it's functions without the ESP8266 code it just of course wouldn't accept and/or send back it's states via MQTT.  

Serial protocol:
https://docs.tuya.com/en/mcu/mcu-protocol.html

chaosmaster:
@arendst  I have created another pull-request for the button-handling: #4099
Please consider adding this, otherwise there is no way to reset wifi-configuration on this device.

PR #4099, opened 2018-10-18
Add reset-button-handler vor tuya-dimmer

chaosmaster:
Some of the TUYA-Dimmers use GPIO_0 to tell the ESP that wifi-configuration should be reset.
The Procedure to trigger this is as follows:
- Turn of the dimmer
- Hold Arrow-down-button for +-15 Seconds
- The MCU will pull down GPIO_0
This code mimics 4 short presses as described here:
https://github.com/arendst/Sonoff-Tasmota/wiki/Button-usage

Without this there is no way to trigger wifi-configuration.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1723,7 +1723,20 @@ void ButtonHandler()
             ExecuteCommandPower(button_index +1, POWER_TOGGLE, SRC_BUTTON);  // Execute Toggle command internally
           }
         }
-      } else {
+      }
+#ifdef USE_TUYA_DIMMER
+      else if (TUYA_DIMMER == Settings.module) {
+        if ((PRESSED == button) && (NOT_PRESSED == lastbutton[button_index])) {
+          snprintf_P(log_data, sizeof(log_data), PSTR(D_LOG_APPLICATION D_BUTTON "%d " D_LEVEL_10), button_index +1);
+          AddLog(LOG_LEVEL_DEBUG);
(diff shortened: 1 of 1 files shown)


arendst:
OK, but...

Why do you need this? There are devices with no button at all. Once installed and configured using wifimanager there is no real need to use a button to reset anything.

The latest release (except sonoff-classic) defaults to wifimanager anyway so no need to select it. A reset using the button is also not a daily task. Might be handy during software development but certainly should not be needed once in production.

Convince me ;-)

chaosmaster:
> Convince me ;-)
I'll try :P

I'v had situations where the device was misconfigured and didn't start wificonfig on it's own.
And since the device has this feature in it's firmware it would be great to make use of it.
This way I can just start wifi-config using the procedure described in the devices manual.
It doesn't necessarily need to be hardcoded to use WIFI_MANAGER.
Instead the button-handler could just start whatever is set as WIFI_CONFIG_NO_SSID.
Or ideally rotate modes, i.e.:
- start with WIFI_MANAGER
- if another button-press is detected while in WIFI_MANAGER start SMARTCONFIG
- while in SMARTCONFIG start WPS
- while in WPS go back to STA
etc

It would be really nice to have the feature when needed.
The device is mounted in the wall and not easily accessible.
Additionally flashing it requires soldering directly to the ESP-module and since there isn't much room in the housing the wires should probably be removed after flashing.

chaosmaster:
Thank you :)

PR #4110, opened 2018-10-19
Obi socket user pins

chaosmaster:
Note I have used LED4 instead of LED2, so LED2 stays available for user-defined GPIOS

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -1107,7 +1107,7 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
      GPIO_LED1,        // GPIO04 Blue LED
      GPIO_REL1,        // GPIO05 (Relay OFF, but used as Relay Switch)
      0, 0, 0, 0, 0, 0, // Flash connection
-     GPIO_LED4,        // GPIO12 (Relay ON, but set to LOW, so we can switch with GPIO05)
+     GPIO_LED2,        // GPIO12 (Relay ON, but set to LOW, so we can switch with GPIO05)
      GPIO_USER,        // GPIO13
      GPIO_KEY1,        // GPIO14 Button
      0,
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -1112,7 +1112,7 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
      GPIO_KEY1,        // GPIO14 Button
      0,
      GPIO_USER,        // GPIO16
-     GPIO_ADC0         // Analog input
+     GPIO_USER         // Analog input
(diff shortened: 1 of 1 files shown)


ascillato:
> Note I have used LED4 instead of LED2, so LED2 stays available for user-defined GPIOS

why?

Led2 should be the logic one. why using led1 and led4?

This will confused people using this template.

sonoff/sonoff_template.h
```diff
@@ -1102,14 +1102,17 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
...
+     GPIO_USER,        // GPIO00
+     0,0,0,
+     GPIO_LED1,        // GPIO04 Blue LED
+     GPIO_REL1,        // GPIO05 (Relay OFF, but used as Relay Switch)
      0, 0, 0, 0, 0, 0, // Flash connection
-     GPIO_LED2,        // GPIO12
-     0,                // GPIO13
-     GPIO_KEY1,        // GPIO14 switch button
-     0, 0, 0
+     GPIO_LED4,        // GPIO12 (Relay ON, but set to LOW, so we can switch with GPIO05)
```
ascillato: Sorry, stills not understand why this?

ascillato (changes requested):

sonoff/sonoff_template.h
chaosmaster: Because it isn't really used as LED on this device (as mentioned in the comment)
It's just set to LED so it stays LOW.
and this way LED2 can be used with the user-configurable GPIOs

chaosmaster: I thought it might be more confusing, when adding an LED, that you can't choose LED2 even though there isn't an LED2

ascillato: Sorry, not convinced.

I think that the actual template is easier to understand.

If you want to control the relay led, may be you can set that GPIO as another GPIO_USER

ascillato: and let the user to select what to do with that.

What do you think?

chaosmaster: you can't controll the relay-led.
GPIO_12 must be set to LOW so the relay can be switched with GPIO_5
There is only one controllable LED which is at GPIO_4 (Blue led)
But I can change back to LED2 if you think that is better.

chaosmaster:
changed back LED4 to LED2

ascillato (approved):

ascillato:
I'm sorry, the analog pin should be GPIO_ADC0 not GPIO_USER. My bad. Sorry

Made the PR https://github.com/arendst/Sonoff-Tasmota/pull/4111 to fix that. Thanks

PR #4170, opened 2018-10-26
Change OTA URL to http://thehackbox.org

Jason2866:
Due to the fact of more and more user issues related to old outdated version on
http://sonoff.maddox.co.uk/tasmota/sonoff.bin

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -79,7 +79,7 @@
 #define WEB_LOG_LEVEL          LOG_LEVEL_INFO    // [WebLog] (LOG_LEVEL_NONE, LOG_LEVEL_ERROR, LOG_LEVEL_INFO, LOG_LEVEL_DEBUG, LOG_LEVEL_DEBUG_MORE)
 
 // -- Ota -----------------------------------------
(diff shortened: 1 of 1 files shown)


arendst:
Are these the release versions or the development versions?

Jason2866:
Development Version

arendst:
So it's better I should make a new release. Been thinking about it but was reluctant considering the compile issues.

@andrethomas do we now have a correct compile procedure which also satisfies the Basic R2?

If so:
- Pls hand me the (updated) compile script
- I also suggest to make a release folder here too so all the bins are in one basket ;-)

jobhh:
Is there a specific reason why the unencrypted version of http://thehackbox.org/tasmota/release/sonoff.bin is used? Is https not supported in the firmware?

It seems like a security flaw for OTA updates. The website does have a proper Let's Encrypt certificate, so https seems like a valid option.

ascillato:
The OTA feature provided by the Arduino Core, does not support https due to memory constraints in the ESP8266 chip.

There is not enough RAM for that.

jobhh:
Cool, thanks for the answer :)

PR #4173, opened 2018-10-26
Small updates to make the system working locally

mrBussy:
# COM port selection
After compiling I was not able to upload the firmware. Investigating learned that the COM settings where incorrect. I added a README entry to indicate that a developer has to change it to their own setup.

# Custom setting files
When creating multiple devices it is convinient to create multiple configuration files. Added the option to ignore these files using the following structure: user_config_override.*.h

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .clang_complete
 .gcc-flags.json
 sonoff/user_config_override.h
+sonoff/user_config_override.*.h
 build
 
 ## Visual Studio Code specific ######
@@ -15,3 +16,4 @@ build
 .vscode/.browse.c_cpp.db*
 .vscode/c_cpp_properties.json
 .vscode/launch.json
+firmware.map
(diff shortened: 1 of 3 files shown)


ascillato:
Hi, sorry but your changes make mandatory the use of user_config_override.h

That is not the case for everyone. 

Better should be to remain the actual config as it is.

ascillato:
Also, this breaks the actual automatic compilation system for the precompiled bins from development branch.

ascillato2:
Closing as this PR makes mandatory the use of user_config_override.h and also breaks the automatic compiling system.

Anyway, thanks for sharing your ideas and we hope you continue doing that.

mrBussy:
You are right. I accedently uploaded a config where the DUSE_CONFIG_OVERRIDE is set. Shall I make a pull request with this issue fixed? Or are you not interested in the changes at all?

ascillato:
It is better to let the user choose which COM port to use and if they want or not to use the USER_CONFIG_OVERRIDE.

PR #4179, opened 2018-10-27
decode-config.py v2.0.0000: add full backup/restore capability

curzon01:
- add full backup/restore capability
- add Tasmota .dmp format to be able to upload previously save Tasmota files
- add help files decode-config.md
- add Tasmota *.dmp format for backup
- new args:
	--restore-file, --extension, --no-extension,
	--full-help, --verbose
- changed args:
	--exit-on-error-only	to --ignore-warnings
	--output-file		to --backup-file
	--output-file-format	to --backup-type
	--hide-pw			to --json-hide-pw
	--unhide-pw		to --json-unhide-pw
- removed (hidden) args:
	--sort, --unsort, --raw-values, --no-raw-values
	-raw-keys, -no-raw-keys
	defaults are no raw keys/values and sort


curzon01:
wrong version in code

PR #4195, opened 2018-10-28
[Feature] support multiple bme sensors

reloxx13:
relates   #1049 #1251 #2261 #2707 #4195

This PR adds abilitiy to connect 2 I2C BMP Sensors at the same time on the same bus 🚌 

grafik

grafik

NOTE:
arendst said in some tickets that this will eat to much flash.
I dont know how to get the more used flash now :D

For BME280 to enable the 2nd address:
vt

PR #4196, opened 2018-10-28
Multiple BMP fix model detection

reloxx13:
fix model detection
related pr #4195

`GetTextIndexed(bmp2x_sensors[bmp2x_count].bmp_name, sizeof(bmp2x_sensors[bmp2x_count].bmp_name), bmp2x_sensors[bmp2x_count].bmp_model, kBmpTypes);`

use `bmp2x_sensors[bmp2x_count].bmp_model` and not `i`

### DIFF
--- a/sonoff/xsns_09_bmp2x.ino
+++ b/sonoff/xsns_09_bmp2x.ino
@@ -1,5 +1,5 @@
 /*
-  xsns_09_bmp.ino - BMP pressure, temperature, humidity and gas sensor support for Sonoff-Tasmota
+  xsns_09_bmp2x.ino - BMP pressure, temperature, humidity and gas sensor support for Sonoff-Tasmota
 
   Copyright (C) 2018  Heiko Krupp and Theo Arends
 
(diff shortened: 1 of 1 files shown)


mike2nl:
@reloxx13 
Question:
i saw the both ino files for the bmp sensors on your repository and asked myself
what that means because the code is different. 

So my question is why two sensor driver files?
- if this for the BMP2xx sensor type or it is for two sensors of the same type?

I ask too because:
- i have to add the new ino file to the excel file sheet i made with all supported sensors
for tasmota. It is in the wiki for a week now under:
https://github.com/arendst/Sonoff-Tasmota/wiki/Sensors-Supported-by-Tasmota

I hope you or arendst can give me a hint about that. Thank You in any case.

reloxx13:
heya,

there are two files now like for the SHT driver, one which supports only one driver and one for max 2 bmp. it should be independent if you use 2 diffrent bmp's or 2 bmp's of the same type, as long as the addresses are diffrent.

maybe you should just merge sht and sht3x and bmp and bmp2x each in one row, cuz both do the same, and the x's drivers just support multiple sensors.

arendst:
It was a very good decision from @reloxx13 to make a seperate driver file. It provides an immediate fix for people wanting 2 sensors while giving  me time to evaluate and someday make one driver out of it..

mike2nl:
@reloxx13 @arendst 
Thank You for your answers...
So i can add that to the excel sheet and some text info too.
- Theo:
..while giving me time to evaluate and someday make one driver out of it..
I was on it to see whats the code is doing and testing to merge, ha, ha.
See who is faster again. So much fun today!

PR #4199, opened 2018-10-29
tuya: reset wifi status LED

chaosmaster:
Some of the tuya-modules don't allow direct control of the status-led via GPIO.
Instead a serial packet indicating wifi-status must be sent to the MCU.
To keep the status-led from blinking annoyingly, just always turn it off for now.
Might implement proper indication of wifi-status at a later time.

See #469 for further details.


qingz2004:
The WiFi LED shows wrong state after one or two days. I have to reset the device to get it corrected. Is it ok to set it once a while instead of only before the ack is recevied? Or can I manually set the LED so I can have it set it in a rule once a while.

PR #4233, opened 2018-10-31
Reduce CPU usage at boot time

ascillato:
If many drivers and sensors have been enabled, there is a peak of CPU usage at boot time that it is translated as a peak in power consumption. This address to a wifi connection issue in poor power regulated devices like Sonoff Basic R2. This fix reduces cpu usage making the wifi connection to establish. This don't degrate the actual performance due it is used only at boot time.

Also, it is noticed a faster wifi connection in other devices too.

This POC cames from an idea discussion with @andrethomas of delaying drivers at startup. Thanks for that! :+1:

### DIFF
--- a/sonoff/xdrv_interface.ino
+++ b/sonoff/xdrv_interface.ino
@@ -249,6 +249,9 @@ boolean XdrvCall(byte Function)
   boolean result = false;
 
   for (byte x = 0; x < xdrv_present; x++) {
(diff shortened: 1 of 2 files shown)


ascillato:
@arendst wait a little to merge it. Will add a configurable #define FUNC_CALLBACK_DELAY

ascillato:
will do later. let's try like this

andrethomas (approved):

reloxx13:
maybe this will also help to reduce power/cpu on boot

grafik

tasmota requested ntp time three times at boot at the same second

PR #4239, opened 2018-11-01
add v6.3.0.2 config changes

curzon01:


TobiasBg:
`Setting_6_3_0.update({` needs to be `Setting_6_3_0_2.update({`!

curzon01:
update on next release - it currently doesn't matter because on 6.3.0 the update area is unused so nothing happens.

PR #4240, opened 2018-11-01
Changes to driver boot delay

andrethomas:
Add driver boot delay while wifi is not connected to

xdsp_interface.ino
and
xnrg_interface.ino

And make it configurabe in my_user_config.h using

`#define DRIVER_BOOT_DELAY 1`

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -269,6 +269,10 @@
 // -- Internal Analog input -----------------------
 #define USE_ADC_VCC                              // Display Vcc in Power status. Disable for use as Analog input on selected devices
 
+// -- BOOT Delay for drivers ----------------------
+
+#define DRIVER_BOOT_DELAY       1                // Number of milliseconds to retard driver cycles during boot-up time to reduce overall CPU load whilst Wifi is connecting
+
 // -- One wire sensors ----------------------------
                                                  // WARNING: Select none for default one DS18B20 sensor or enable one of the following two options for multiple sensors
 //#define USE_DS18x20_LEGACY                       // Optional for more than one DS18x20 sensors with dynamic scan using library OneWire (+1k5 code)
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -269,10 +269,6 @@
 // -- Internal Analog input -----------------------
 #define USE_ADC_VCC                              // Display Vcc in Power status. Disable for use as Analog input on selected devices
 
-// -- BOOT Delay for drivers ----------------------
-
-#define DRIVER_BOOT_DELAY       1                // Number of milliseconds to retard driver cycles during boot-up time to reduce overall CPU load whilst Wifi is connecting
-
 // -- One wire sensors ----------------------------
                                                  // WARNING: Select none for default one DS18B20 sensor or enable one of the following two options for multiple sensors
 //#define USE_DS18x20_LEGACY                       // Optional for more than one DS18x20 sensors with dynamic scan using library OneWire (+1k5 code)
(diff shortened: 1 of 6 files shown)


arendst:
Is there any difference in a delay of 1 or higher? I thought it was about the extra yield that would be performed within the delay routine. If so then a delay of 1 should be enough.

Did you notice any difference when using a delay of 1 or 10?

sonoff/my_user_config.h
```diff
@@ -269,6 +269,10 @@
 // -- Internal Analog input -----------------------
 #define USE_ADC_VCC                              // Display Vcc in Power status. Disable for use as Analog input on selected devices
 
+// -- BOOT Delay for drivers ----------------------
+
+#define DRIVER_BOOT_DELAY       1                // Number of milliseconds to retard driver cycles during boot-up time to reduce overall CPU load whilst Wifi is connecting
```
ascillato: What about adding this define in sonoff.h instead?

andrethomas:
It seemed that way, but let me add some debug code to code so that I can see if its actually scientifically proven insofar that wifi can connect x number of ms faster with 1 vs 10... maybe it was just my perception.

Hold back for now - let me try to measure it with micros() and see if it makes a reproducible difference that can be seen with a more scientific approach than just my perception.

sonoff/my_user_config.h
arendst: Yeah. That's better. Keep it from "normal" users.

andrethomas:
Right, I'm not sure how conclusive this is because obviously, the output timing is dependent on the driver timing as I inserted the following in support.ino within function WifiSetState()

```
if (state == 1) {
  Serial.print("wifi connect after: ");
  Serial.println(micros());
}
```

I then performed several renditions of
`restart 1`
and observed the first result returned by the Serial.println()

First a device which is known to have problems I get the following best three out of 10 results

```
0:
wifi connect after: 7221651
wifi connect after: 7221229
wifi connect after: 7221659

1:
wifi connect after: 8412190
wifi connect after: 8391596
wifi connect after: 8427615

10:
wifi connect after: 13302618
wifi connect after: 13309008
wifi connect after: 13305230
```

For this known to have issues device I had 7/10 times failing on first connect with setting on 0, failing 3/10 times with setting 1 and failing 0/10 times on setting 10 - When I say failing, I mean failing on first attempt to connect to wifi.

Then I tested the same process with a device known to not have issues connecting to wifi and got the following results:

```
0:
wifi connect after: 7220933
wifi connect after: 7221839
wifi connect after: 7221138

1:
wifi connect after: 7361852
wifi connect after: 7448294
wifi connect after: 7503575

10:
wifi connect after: 13328779
wifi connect after: 13337204
wifi connect after: 13332101
```

With this device I never had failures on first attempt to connect to wifi so it was quick to get 3 consistent results for each use case.

So adjusting it to a higher value does seem to help a device known to have wifi connection issues because going to delay(10) resulted in first time connect after each restart 1 command was issued.

The micros() reported are longer the higher the value is in both device cases but a higher value definately improves the reliability of first time connection on a device known to have issues.

I'm guessing the report of micros() is also increasing because everything is delayed by that much during startup and may be related to where I inserted the debug output - I can't be sure.

Either way, even the problematic device does eventually connect (even if it fails 2-3 times after restart 1) so I guess its questionable on whether or not we want this configurable or not.

I think there are other things at play within the SDK which we cannot directly measure using this method.

sonoff/my_user_config.h
andrethomas: For me, it does not matter where it is but it would be easier to configure it for pre-compiled binaries if it's in my_user_config.h - But let's see... may not even include it given the results below.

andrethomas: Yeah, didn't think of that... will be safer in sonoff.h ;)

andrethomas: Moved

PR #4241, opened 2018-11-01
Add mmHg conversion

eeak:


ascillato (changes requested):
Hi,

There are missing the keys for the rest of translations. It will not compile if you don't add that. Please, add them. Thanks

eeak:
Translations updated.

arendst:
Actions needed:

- Move the text "mmHg" to the correct position in the language files and define as D_UNIT_MILLIMETER_MERCURY "mmHg".
- Do not change the hPa value when read; use conversion only for presentation.
- Start using this define D_UNIT_MILLIMETER_MERCURY in the webserver GUI.

eeak:
>     * Do not change the hPa value when read; use conversion only for presentation.

In many countries, mmhg is used, and hpa is not used at all. It will be more convenient to convert when reading to get the right values ​​everywhere.

eeak:
Estonia, Russia, Ukraina, etc.... all exUSSR countries use mmHg and not use hPa.
https://weather.ee/
https://yandex.ru/pogoda/213/
https://sinoptik.ua/%D0%BF%D0%BE%D0%B3%D0%BE%D0%B4%D0%B0-%D0%BA%D0%B8%D0%B5%D0%B2

ascillato:
This case is the same as with the temperature. The SI uses C but some countries uses F.

So, I think we should add a setoption to choose between hPa (default) and mmHg

ascillato (approved):

eeak:
Can I use setoption24 ? It marked as free since v5.14.0b

PR #4242, opened 2018-11-01
Fix regression for nrg and dsp interface delays

andrethomas:
**URGENT**

After testing on POW and POW2 I find that the delay in combination with the wifi status check prevents a POW or POW2 from being set to that module type for some reason.

So reverting back to what @ascillato originally committed except for the addition of the #define for the value


andrethomas:
Thx, appreciated!

andrethomas:
Tested after merge, fresh compile - works fine. Thx again

Issue #4189, opened 2018-10-28
Hue Bridge emulation issue - Alexa say "Device is unresponsive"


wongnam:
Current TASMOTA version 6.2.1.19 has issue with "Hue Bridge" emulation - it caused Alexa say "Device is unresponsive". It also cannot control in Alexa App.

I have flash it back to version 6.2.1.18 -> it is working well.

PLEASE NOITICE: i found that TASMOTA "version 6.2.1.19" description has not been changed in code. it is still displayed "Version 6.2.1.18". 

_Also, make sure these boxes are checked [x] before submitting your issue - Thank you!_
- [x] _Searched the problem in issues and in the wiki_
- [x] _Hardware used_ :  Sonoff SV, Sonoff Touch US, Sonoff Basic
- [x] _Development/Compiler/Upload tools used_ : VSC + Platformio + espressif8266@1.5.0 (CORE-2.3.0)
- [X] _Provide the output of command_``status 0`` :
```
00:00:00 Project sonoff My PC (Topic mypc, Fallback DVES_959B44, GroupTopic sonoffs) Version 6.2.1.18-2_3_0
00:00:00 WIF: Connecting to AP1 Multimedia Wi-Fi Network in mode 11N as mypc-6980...
00:00:04 WIF: Connected
00:00:04 DNS: Initialized
00:00:04 HTP: Web server active on mypc-6980.local with IP address 192.168.12.128
00:00:04 UPP: Multicast (re)joined
00:00:04 MQT: Attempting connection...
00:00:04 MQT: Connected
00:00:04 MQT: tele/mypc/LWT = Online (retained)
00:00:04 MQT: cmnd/mypc/POWER = 
00:00:04 MQT: tele/mypc/INFO1 = {"Module":"Sonoff SV","Version":"6.2.1.18","FallbackTopic":"DVES_959B44","GroupTopic":"sonoffs"}
00:00:04 MQT: tele/mypc/INFO2 = {"WebServerMode":"Admin","Hostname":"mypc-6980","IPAddress":"192.168.12.128"}
00:00:04 MQT: tele/mypc/INFO3 = {"RestartReason":"Software/System restart"}
00:00:04 MQT: stat/mypc/RESULT = {"POWER":"OFF"}
00:00:04 MQT: stat/mypc/POWER = OFF
00:00:05 UPP: Multicast (re)joined
[...]

PR #4279, opened 2018-11-04
Correctly announce non RGBWW light

emontnemery:
Do not include level / color / color temperature if the module does not support it in Philips Hue API response.


ascillato:
Hi,

This PR brokes the Alexa support.

This has been reverted by PR #4291

@emontnemery Please, do again this PR but testing that Alexa also works. Thanks.

emontnemery:
@ascillato Unfortunately, I don't have any Alexa HW.
Do you have time to help debug why this breaks Alexa integration?

ascillato:
I don't have Alexa either.

The bug report is #4189 and it was confirmed that reverting this PR solves the issue.

PR #4296, opened 2018-11-06
tuya: fix SetOption20 behavior for Oittm/Moes

chaosmaster:
This introduces a new option SetOption22 which is related to SetOption20.
It can essentially be described as "Apply SetOption20 setting to commands from the faceplate"

See #469 for more info.


arendst:
You can't use setoption22 as it is used for mqtt_serial. What makes you think you can use this option number?

chaosmaster:
Oh is it?
The [wiki](https://github.com/arendst/Sonoff-Tasmota/wiki/Commands) states it is Reserved.
I interpreted this as "Reserved for future use".
Also the [code](https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/sonoff.ino#L736).
Makes it look like it is not used sine it says "Command Error"
Could you suggest what option could be used instead?

arendst:
Setoption54

chaosmaster:
I cant seem to find where it is used for mqtt_serial. (Besides the comment in that line).
If you could point me to it?
I'll use SetOption54 nonetheless.

arendst:
Using the right tools and a global search will reveil that it is used in the serial code of sonoff.ino as are more "reserved" options. They are reserved because you 're not allowed to change them with setoption command.

chaosmaster:
Thanks for explaining.
I must have missed it then, I tried grepping for "Settings.flag.data".
I changed it to SetOption54 now

PR #4304, opened 2018-11-07
add v6.3.0.4 config changes

curzon01:
- pack setting templates using diffs
- add more '--verbose' outputs
- fix wrong templates selection
- fix wrong filename handling
- fix --verbose filename output


arendst:
Hate to say that the settings area around drivers/sensors has been changed today. Looking at the addresses you use I think you will have to make a change there.

curzon01:
No issue - bad timing from my side, conflicts with the fixes I've found today

PR #4320, opened 2018-11-10
tuya: formatting/cleanup

chaosmaster:


sonoff/xdrv_16_tuyadimmer.ino
```diff
@@ -325,7 +326,7 @@ void TuyaSerialInput()
     else if ((tuya_cmd_status == 3) && (tuya_byte_counter == (6 + tuya_data_len)) && (tuya_cmd_checksum == serial_in_byte)){ // Compare checksum and process packet
       tuya_buffer[tuya_byte_counter++] = serial_in_byte;
 
-      snprintf_P(log_data, sizeof(log_data), PSTR("TYA: Received Packet: \""));
+      snprintf_P(log_data, sizeof(log_data), PSTR("TYA: Rcvd Packet: \""));
```
ascillato: Can you change to Rcvd to RX ? or add it to the language keys?

ascillato (changes requested):

sonoff/xdrv_16_tuyadimmer.ino
chaosmaster: you mean change "Sent" to "TX" and "Rcvd" to "RX" ?
i just changed it, so "Sent" and "Rcvd" have the same length.
Looks better in log:

00:00:00 TYA: Sent Packet: "55aa0008000007"
00:00:00 TYA: Rcvd Packet: "55aa0003000002"

ascillato: You are right that using the same length looks better in the console, but the problem I see is that rcvd is not understandable in all languages as received.

ascillato (approved):

PR #4355, opened 2018-11-15
Change C functions with () to explicitly specify (void) for no parameters

andrethomas:
Change C functions with () to explicitly specify (void) for no parameters

The Arduino compiler recipe seems to ignore () and substitute (void) anyway, but this does not seem to be the case for PIO.

This PR changes all C functions which have no parameters to explicitly define this by having void as a parameter.

### DIFF
--- a/sonoff/core_esp8266_timer.c
+++ b/sonoff/core_esp8266_timer.c
@@ -46,7 +46,7 @@ void ICACHE_RAM_ATTR timer1_isr_handler(void *para){
     }
 }
 
-void ICACHE_RAM_ATTR timer1_isr_init(){
+void ICACHE_RAM_ATTR timer1_isr_init(void){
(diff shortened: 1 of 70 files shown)


andrethomas:
Side note - These changes were done programmatically and not manually to limit the error margin - Only the following function returns were considered:

```
bool function_name() -> bool function_name(void)
boolean function_name() -> boolean function_name(void)
byte function_name() -> byte function_name(void)
char function_name() -> char function_name(void)
double function_name() -> double function_name(void)
float function_name() -> float function_name(void)
int function_name() -> int function_name(void)
int16_t function_name() -> int16_t function_name(void)
String function_name() -> String function_name(void)
uint16_t function_name() -> uint16_t function_name(void)
uint32_t function_name() -> uint32_t function_name(void)
uint8_t function_name() -> uint8_t function_name(void)
void function_name() -> void function_name(void)

```

PR #4371, opened 2018-11-17
Fix NovaSDS sensor checksum failure

curzon01:
I received a lot of SDS: Checksum failure messages. Reason was that Tasmota becomes async to received data stream and does not resync.

The changes I'd made:

1. NovaSdsSerial->flush() should be always _before_ NovaSdsSerial->write() a command  
If it is order write() / flush() and a time consuming interrupt occurs between write() and flush() you flush() results bytes from write() command.

2. checksum failure appeared when the read data stream will become async to SDS stream:  
```
09:23:30 WIF: Checking connection...
09:23:30 WIF: Connected
09:23:31 SER: Received C0 39 00 4F 00 29 3F F0
09:23:32 SER: Received C0 39 00 4F 00 29 3F F0
09:23:33 SER: Received C0 39 00 4F 00 29 3F F0
...
09:23:41 SER: Received AA C0 39 00 4F 00 29 3F
09:23:41 SDS: Checksum failure
09:23:42 SER: Received AA C0 39 00 4F 00 29 3F
09:23:42 SDS: Checksum failure
...
```
so I changed it to flush() / write() query command and get the whole stream including head/tail from NovaSDS

3. reduced query interval from 1 sec to 3 sec (`#define XSNS_20_QUERY_INTERVAL`)
[NovaSDS Control Protocol](https://cdn.sparkfun.com/assets/parts/1/2/2/7/5/Laser_Dust_Sensor_Control_Protocol_V1.3.pdf) describes at "**2) Query data command**":  
> Sensor received query data command to report a measurement data, recommended query
interval of not less than 3 seconds

### DIFF
--- a/sonoff/xsns_20_novasds.ino
+++ b/sonoff/xsns_20_novasds.ino
@@ -32,6 +32,9 @@
 #ifndef WORKING_PERIOD
 #define WORKING_PERIOD      5
 #endif
+#ifndef XSNS_20_QUERY_INTERVAL
+#define XSNS_20_QUERY_INTERVAL  3 // query every 3 seconds
+#endif
 
 TasmotaSerial *NovaSdsSerial;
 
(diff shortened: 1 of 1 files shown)


edelstahlratte:
So it seems to me that the checksum errror occured due to the high query frequency... But I never experienced this myself...

curzon01:
Don't believe the high query frequency is the issue, the data coming in on async stream state are not invalid, they are complete but misaligned
I assume serial flush() function do not remove any buffered incoming serial data, only wait for outgoing serial data to complete. I still get rare "out-of-sync" states even with the changes I'd made; so I will try to change sync manual to next command result head 0xAA (the original code before also didn't worked in that case).

Issue #4388, opened 2018-11-19
Cases in language defines not consistent (#define D_ONLINE "online"#define D_OFFLINE "offline") (de-DE)


crazyduck:
Hi, I recently changed from english release sonoff.bin to sonoff-de.bin.
Then my homeassistant was not able to receive the state. It took me a bit to find out that the defines of D_ONLINE and D_OFFLINE are not consistent between en-GB.h and de-DE.h.

After this I checked the other files and most of them use "Online" and "Offline".

To help other users, whose may follow instructions from the wiki, keep the cases consistent througout languages with same word. I followed the wiki to configure my homeassistant.

- [X] _Searched the problem in issues and in the wiki_
- [X] _Hardware used_ : Several sonoffs
- [X] _Development/Compiler/Upload tools used_ : used a file from official releases
- [X] _If a pre-compiled release or development binary was used, which one?_ : release
- [X] _You have tried latest release or development binaries?_ : latest
- [x] _Provide the output of command_``status 0`` :
```
23:32:21 MQT: stat/sonoff-wozischrank/STATUS = {"Status":{"Module":1,"FriendlyName":["sonoff-wozischrank"],"Topic":"sonoff-wozischrank","ButtonTopic":"0","Power":1,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":1}}
23:32:21 MQT: stat/sonoff-wozischrank/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"http://sonoff.maddox.co.uk/tasmota/sonoff.bin","RestartReason":"Software/System restart","Uptime":"0T06:01:15","StartupUTC":"2018-11-19T15:31:06","Sleep":0,"BootCount":35,"SaveCount":348,"SaveAddress":"F5000"}}
23:32:21 MQT: stat/sonoff-wozischrank/STATUS2 = {"StatusFWR":{"Version":"6.3.0","BuildDateTime":"2018.10.30 17:39:28","Boot":31,"Core":"2_3_0","SDK":"1.5.3(aec24ac9)"}}
[...]

PR #4392, opened 2018-11-20
Update DE lang with Online and Offline keys

ascillato:
#4388


ascillato:
The rest of the language files are ok.

PR #4393, opened 2018-11-20
Include support for Google-Kalendars

Leuselator:
It would be nice to have the possibility to switch on/off depending on Google-Kalendar as AVM does with their DECT-Powersockets... 
I use it eg. for 2 waterpumps to ensure that both are used equally - so 4h the 1st is provided with power then 4h the 2nd one and so on. This way I reduce the risk of malfunktion of a pump because of long time without movement. Could shurely be achived in other ways but google-calendar sounds easy?


ascillato2:
You are trying to merge 2 differemt branch of Tasmota.

This is not a valid PR. 

Closing.

PR #4423, opened 2018-11-24
Increase delays before web ui is refreshed

andrethomas:
The current 5-second delay before refreshing web ui after settings were changed is not always sufficient as wifi can take up to 10 seconds to re-connect and be available on some wireless networks, depending on the device's RSSI.

This change increases the delay from 5 seconds to 10 seconds which have been tested down to RSSI levels of 25 to be sufficient.

This change also increases the refresh delay when a local OTA is uploaded to 20 seconds which allows enough time for the cp: process to complete and the device to reconnected to wifi under >= 25 RSSI conditions.

### DIFF
--- a/sonoff/xdrv_01_webserver.ino
+++ b/sonoff/xdrv_01_webserver.ino
@@ -95,7 +95,12 @@ const char HTTP_SCRIPT_WIFI[] PROGMEM =
   "}";
 
 const char HTTP_SCRIPT_RELOAD[] PROGMEM =
-  "setTimeout(function(){location.href='.';},5000);"
+  "setTimeout(function(){location.href='.';},10000);"
+  "</script>";
+
+// Local OTA upgrade requires more time to complete cp: before web ui should be reloaded
+const char HTTP_SCRIPT_RELOAD_OTA[] PROGMEM =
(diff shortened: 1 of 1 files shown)


arendst:
I agree on the OTA timeout but...

I think the 5 second timeout on restart is enough considering the fact that the browser is just starting to reconnect after this 5 second and it continues to do so until a certain time that I don't know but with chrome it tries at least for 30 seconds.

Do you have different experience with other browsers that fail to keep trying at least some 30 seconds?

andrethomas:
Firefox tries to connect after the 5 second period but times out just before device obtains wifi connectivity... test scenario with RSSI of 25 makes around 9 seconds so devices close by with 6 seconds reconnect times it works fine... more remote devices it results in an unable to connect error with Firefox, but yes I we can leave it at 5 seconds since this is a very remote device scenario so it is kind of expected.

We can leave the 5-second one as is I guess if nobody else is complaining.

Will update accordingly.

Jason2866:
Tested with chrome. @arendst  you are right chrome doesnt give a error but it needs about 9 seconds
after changing something. So 10 sec. would be a good value...

arendst:
I'll make it 9 seconds then ;-)

PR #4431, opened 2018-11-24
Add main loop load average

andrethomas:
Add main loop avarage duty cycle measured against setoption36 value to telemetry data as LoadAvg

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -190,6 +190,7 @@ char mqtt_data[MESSZ];                      // MQTT publish buffer and web page 
 char log_data[LOGSZ];                       // Logging
 char web_log[WEB_LOG_SIZE] = {'\0'};        // Web log buffer
 String backlog[MAX_BACKLOG];                // Command backlog
+uint32_t loop_load_avg = 0;                 // Indicative loop load average
 
 /********************************************************************************************/
 
@@ -1575,6 +1576,8 @@ void MqttShowState(void)
   snprintf_P(mqtt_data, sizeof(mqtt_data), PSTR("%s,\"" D_JSON_VCC "\":%s"), mqtt_data, stemp1);
 #endif
 
+  snprintf_P(mqtt_data, sizeof(mqtt_data), PSTR("%s,\"LoadAvg\":%u"), mqtt_data, loop_load_avg);
+
   for (byte i = 0; i < devices_present; i++) {
     if (i == light_device -1) {
       LightState(1);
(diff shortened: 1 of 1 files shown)


synekvl:
Assume value should be in percents nevermind I get with some devices LoadAvg above 800 (as 849, 851, 829....) It is with Wemos D1, Wemos D1mini and Sonoff Touch. Another are below "100"

synekvl:
@andrethomas Thank you for your explanation. Frankly it is difficult for me to understand this figure, nevermind two devices, in my case, having this value 800+ have sensors attached - D1mini has BME280 and WeD1 has Si7021. The third above 800 - Sonoff Touch (1T) has nothing attached and has LoadAvg=820...

So they are my values and I will try to understand meaning of SetOption36 which has not been changed yet.

Cheers
Vladimir

andrethomas:
On SetOption36 of 50, if you are getting a loadavg of 820 it could be calculated that the main processing loop is taking approximately 410ms to complete which is much longer than expected - so it means there is some iteration of a driver that is consuming time in the main loop... could be indicative of a malfunction or perhaps one of the drivers in use requires optimisation. To give an indication, most sonoff basics would have a load avg of around 40 on SetOption36 = 50. So although not a direct indication that something is wrong it does seem to lend to the idea that something is not happening as fast as it would normally be expected to happen.

It is not clear from the datasheet how the ESP chip responds to thermal dynamics inside a closed housing with insufficient cooling - does it auto scale down the clock speed or not? This cannot be confirmed from the datasheet.

The general observation that we've made is that if this is way in excess of 100 you may experience secondary problems such as mqtt reconnects and/or wifi reconnects. SetOption36 therefore allows you to adjust the target loop speed depending on what the device itself is used for.

andrethomas:
Maybe this will clarify it better.

I configured a NoceMCU V2 board with http://thehackbox.org/tasmota/sonoff.bin and only configured WiFi and MQTT settings and set the device as Generic module.

SetOption36 = 50 (The default)
Before configuring I2C, I get a 6 sample load average of 38
After just selecting I2C pins in GPIO (No device connected yet), I get a 6 sample load average of 48
Connecting an LM75AD temperature sensor to the I2C pins does not change - stays at 48
Now for the interesting part - I navigate to the main web ui page where it displays the temperature reading on a 2-second interval (iirc)
For this, I now see a 6 sample load average of 78 :)

So this does compute - the more work the device is doing the higher the load average will be because it spends more time per loop cycle... because the indicated load average is still below 100 (if you want to view it as a % of time spent compared to the setoption36 value, even though this is not the intended purpose) so it means my main loop is on average still completing well under the targetted 50ms and provides a fair margin for the SDK to do its background work.

For reference, when I say 6 sample average I mean that I set the teleperiod = 10 seconds, and took the 6 values provided over that 1 minute period and averaged it out to get the above 6 sample averages.

andrethomas:
@synekvl Can you confirm that your sleep = 0 on the devices with high load averages? Having sleep > 0 will cause the main loop to be exponentially longer/slower so you will get a higher loadavg for the main loop as it is time based.

synekvl:
@andrethomas  Hi, I was quite busy today, so I could check that parameter just now. And ?? Tradaaaaaa !!! You were right as far as these three devices had the Sleep parameter set to value "20". When I set it to '0', now  D1 mini and Wemos D1 LoadAvg dropped down below '50', and Sonoff Touch T1 to even '27'...

So your idea was 100% correct. Thank you for your advices and discoveries, how it works and how the parametres are connected.

Cheers
Vladimir

andrethomas:
@synekvl You're welcome. We made some more changes so that it behaves more like a % indicator now. 

With SetOption39 set there is no need for sleep setting anymore so there's new firmware from dev binaries to which you can update to get a more % like result.

See wiki here: https://github.com/arendst/Sonoff-Tasmota/wiki/SetOption36

PR #4450, opened 2018-11-25
Support sending tele/STATUS as state/RESULT

emontnemery:
Improve Hass support by sending tele/STATUS as state/RESULT.
This make sure Hass state is synchronized with the Sonoffs after Hass restart.
See also: home-assistant/home-assistant#18703

The option will be automatically enabled if hass discovery is enabled.


emontnemery:
@hhaim Please comment if this solves your issue in home-assistant/home-assistant#18703
@arendst , @ascillato Do you think this is the right approach? The root issue is that Hass only allows one topic for state updates.

hhaim:
@emontnemery,  this will solve my issue. thanks for the quick fix

PR #4459, opened 2018-11-26
Fix math in LoadAvg

andrethomas:
Fix the math in LoadAvg calculation. To keep average we take away one Nth of loops and add one Nth of current cycle ration to the average.

Now the math makes more sense... maximum load average on setoption36 0/1 = 999 (as it should be)


andrethomas:
The numbers also make more sense now, for example, using setoption36 10 (more or less equivalent to sleep 10) it causes a loadavg of ~99, so leaving 100uS to Yield allowing a sonoff basic to just just make a connection :)

Setoption36 50 I get LoadAvg of 19, which translates to ~9.5ms duty cycle in loop as confirmed by logic analyzer.

PR #4465, opened 2018-11-27
Add support for PS-16-DZ Dimmer

chaosmaster:
Add support for yet another serial dimmer.
See #469


ascillato (approved):

arendst:
I have to re-think the use (and reservation) of serial buffers. The way this is implemented now is claiming another 256 bytes for a serial buffer while others also claimed buffer space and never will be used as they are using the same GPIO's.


I'll think of a solution. Until then I'll commit this one as it looks fine for the moment.

arendst:
Just did a commit with dynamic buffer space reservation. If possible give it a try but I suspect no problems with this solution.

chaosmaster:
Looking good on my end. Can't test on the actual device though.
It's probably also possible to reduce buffer-size, I will look into that.

PR #4481, opened 2018-11-29
Add devices Teckin SP20 US and ZooZee SA102, my first github fork and pull request...

mrMikeWatson:
Both look like BlitzWolf with relay at GPIO04 instead of GPIO15


arendst:
Aren't they both the same? Did you check they are different from already defined ones?

Every entry in this table costs flash memory and in the end will fail gui due to out of Ram.

mrMikeWatson:
Yes, the same.
So could take up one shared slot.
Name was limited to 15 char so I split into 2.
Yes, checked others, but found no energy monitor like Blitzwolf that
appeared to have relay at GPIO04.
Thanks for taking the time to look at this.
Please reject as you see fit.
Thank you for your great piece of software by the way!

Regards,
Mike

arendst:
Will use info to add Teckin US to template. Thx.

PR #4488, opened 2018-11-30
HC-SR04: switch to using NewPing libraries directly

winstona:
add NewPing-1.9.1 lib
also includes returning data on mqtt/web to 3 decimal places

### DIFF
--- a/lib/NewPing-1.9.1/README.md
+++ b/lib/NewPing-1.9.1/README.md
@@ -1,0 +1,3 @@
+# NewPing Arduino Library for Arduino
+
+## See the [NewPing Wiki](https://bitbucket.org/teckel12/arduino-new-ping/wiki/Home) for documentation
--- a/lib/NewPing-1.9.1/examples/NewPing15SensorsTimer/NewPing15SensorsTimer.pde
+++ b/lib/NewPing-1.9.1/examples/NewPing15SensorsTimer/NewPing15SensorsTimer.pde
@@ -1,0 +1,78 @@
+// ---------------------------------------------------------------------------
+// Before attempting to use this sketch, please read the "Help with 15 Sensors Example Sketch":
+// https://bitbucket.org/teckel12/arduino-new-ping/wiki/Help%20with%2015%20Sensors%20Example%20Sketch
+//
+// This example code was used to successfully communicate with 15 ultrasonic sensors. You can adjust
+// the number of sensors in your project by changing SONAR_NUM and the number of NewPing objects in the
+// "sonar" array. You also need to change the pins for each sensor for the NewPing objects. Each sensor
+// is pinged at 33ms intervals. So, one cycle of all sensors takes 495ms (33 * 15 = 495ms). The results
+// are sent to the "oneSensorCycle" function which currently just displays the distance data. Your project
+// would normally process the sensor results in this function (for example, decide if a robot needs to
+// turn and call the turn function). Keep in mind this example is event-driven. Your complete sketch needs
+// to be written so there's no "delay" commands and the loop() cycles at faster than a 33ms rate. If other
+// processes take longer than 33ms, you'll need to increase PING_INTERVAL so it doesn't get behind.
+// ---------------------------------------------------------------------------
+#include <NewPing.h>
+
+#define SONAR_NUM     15 // Number of sensors.
+#define MAX_DISTANCE 200 // Maximum distance (in cm) to ping.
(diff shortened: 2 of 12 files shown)


andrethomas:
Will need to test this to see if it yields any benefit. Previous experiments with this sensor did not consistently produce reliable/accurate values, even in 0 decimal places.

andrethomas:
@winstona I don't think the 4 hour oscillation can be mitigated in the firmware - what is most concerning about the data is the variance of up to 5CM so it essentially means the sensor could be accurate with up to 5CM deviation - This beings into question the need for fractions.

The good news is that the general use case for this sensor the 5cm deviation should not be a problem and you're definitely getting more consistent results than I previously did.

I'll run a similar test with your code over the weekend to see if I can reproduce your findings.

winstona:
@andrethomas the variance of 5cm is before these changes - it seems on average the variance is closer to about 2cm from peak to peak.

I built a 2nd one (both are using a wemos d1 mini and a 5v hc-sr04) and let it run measuring a distance within a cardboard box to avoid any other factors, here are my results so far:

screen shot 2018-12-01 at 4 38 15 pm

green - sensor_data - actual value with a teleperiod of 10s
yellow - SMA - moving average over the last 300s (5 min)
cyan - actual_distance_to_sensor - actual measured distance to the tip of the ultrasonic sensor
orange - actual_distance_to_board - actual measured distance to the side of the board the ultrasonic sensors are mounted to
blue area - 1 stddev above and below a 10 min average

Here's a closer view as well:

screen shot 2018-12-01 at 4 51 33 pm

As you can see, without the 0.5cm outliers, it looks like you can get something quite consistent, although not perfect. Yes, adding 3 decimal places doesn't guarantee that level of precision, and can be misleading, but it helps to get something closer to accurate when using a SMA or other form of smoothing.

Would be interested to see other results and how it may compare though, as I have only 2 set up and running right now.

andrethomas:
Definitely an improvement to the current codebase

image

This was done with 10 second intervals also.

Flat object placed @ 20cm and recorded hysteria of ~ 9mm.

I think its safe to say this is the best performance I've seen with a SR-04 sensor.

andrethomas (approved):

andrethomas:
@winstona Thanks for the PR and all the testing :)

winstona:
no problem @andrethomas - still working on seeing if I can reduce that 0.5 error to something smaller, since it doesn't seem to be present all the time, but haven't narrowed it down to anything yet

andrethomas:
I think its based on acoustics... possibly making a cone for the receiver will eliminate echo's off other objects but I think its one heck of an improvement on what we had and if you really want better than 5mm accuracy then you'd be better off using a time of flight sensor.

So yeah, spend time on it if you think it will do something for you but I think most users don't care too much for the 5-10mm hysteria... i mean on 20cm that's like 2 to 5 % :)

effelle:
@winstona, thanks for this PR.
@andrethomas, care to share what sensor model did you use to make your tests? 
I'm asking because I found a more recent version of the HC-SR04, called HC-SR04P, that works with 3.3V and then can solve the issue of a 5V echo. It seems to works without any kind of quartz. 

image2017-10-3 10_5_33

I've found another model too, called US-015 on AliExpress, that seems to use a quite different logic.

us015

Since I own just normal HC-SR04 I wonder if the others will works. 
I can buy one of each to test but I'll get them at least in 30-40 days.

andrethomas:
I used a HC-SR04, gave it 5V and connected the trigger and echo pins to an ESP8266 12F. I have never lost a device due to 5V on a gpio... just don't power it using 5V (The ESP I mean) - If you do it continues to work fine but loses its wifi core... so it takes some time to figure out that you can talk to the device via serial console but wifi fails because it got fried by the 5V :)

effelle:
I've always used with a level shifter, since I tested it with an ESP01. 
Anyway, I think I'll buy a couple of HC-SR04P, just to see how they works.

andrethomas:
@effelle Will be interesting to see if they perform better assuming its a newer version, thanks

effelle:
Bought on Ali. Can't tell you an ETA, but normally is between 20 and 40 days.

effelle:
@andrethomas: I got (finally!) the new model (HC-SR04P). 
Working fine but the drop @3.3V is huge, 4.2 meters on 5V and 3,1 meters @3.3V.  
Taken away this I found no issue.

PR #4490, opened 2018-11-30
Prevent sleep and SetOption36 Conflict

andrethomas:
Code change to set sleep = 0 when SetOption36 is > 0 is sent by command.

Reverse also, set SetOption36 = 0 when sleep > 0 command is sent.

Users have reported unusual results when a combination of the two is used.

### DIFF
--- a/sonoff/settings.ino
+++ b/sonoff/settings.ino
@@ -857,6 +857,7 @@ void SettingsDelta(void)
     }
     if (Settings.version < 0x0603000A) {
       Settings.param[P_LOOP_SLEEP_DELAY] = LOOP_SLEEP_DELAY;
+      Settings.sleep = 0;                   // We do not want sleep enabled when SetOption36 is active
     }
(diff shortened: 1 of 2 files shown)


andrethomas:
@arendst I'm missing something that needs to be done somehow.

SetOption36 50 works fine if sleep = 0

The idea is that if command SetOption36 50 is executed, it should set sleep to 0 but for some reason it only takes effect on reboot - until then it reports sleep like this:

21:04:23 CMD: sleep
21:04:23 MQT: stat/sonoff/RESULT = {"Sleep":"50 (0)"}

And the LoadAvg confirms that the previously set value of sleep still has an impact... this is fixed if device is restarted but can it be done without having to restart?

Thx

arendst:
Setting.sleep is the storage for sleep. There is another variable called sleep that does the actual sleep task.

If you set option 39 you will have to clear this sleep variable too.

andrethomas:
Thx!

PR #4493, opened 2018-12-01
Update xdrv_12_home_assistant.ino

mhuca:
My suggestion is to use %Profile%/%Topic%/POWER topic to publish the current status. This would give users the option to use Powerretain or HASS automations to ensure the current status on startup. Currently the %Profile%/%Topic%/RESULT is being used but no retain option is available.

While the Home Assistant automation as described in the wiki is one option... this becomes much more unmanageable when the number of devices increases. Especially when not using a single Group Topic for all devices.

### DIFF
--- a/sonoff/xdrv_12_home_assistant.ino
+++ b/sonoff/xdrv_12_home_assistant.ino
@@ -24,8 +24,8 @@
 const char HASS_DISCOVER_RELAY[] PROGMEM =
   "{\"name\":\"%s\","                              // dualr2 1
   "\"command_topic\":\"%s\","                      // cmnd/dualr2/POWER2
-  "\"state_topic\":\"%s\","                        // stat/dualr2/RESULT  (implies "\"optimistic\":\"false\",")
-  "\"value_template\":\"{{value_json.%s}}\","      // POWER2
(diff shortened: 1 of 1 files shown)


emontnemery:
This is not a good idea or necessary.
It's better to pick the power status from the RESULT JSON dictionary, then state is synced:
 - when Hass is restarted
 - when Tasmota is restarted
 - every Tele period

@mhuca if you think I misunderstood, please explain in more detail your idea

mhuca:
@emontnemery Thank you for feedback.

With the current solution... Hass restarts will cause the current status to be unknown until the Tele period expires or a Hass automation runs.. 

For environments with a large number of devices (i'm running 50+ at the moment) this becomes more challenging to manage. Having a common GroupTopic does introduce risk... (a simple mistyping of the command RESTART vs RESET  to a GroupTopic has very different results.) 

I will try shortening the telePeriod.

emontnemery:
Do you try an automation like this:
https://github.com/arendst/Sonoff-Tasmota/wiki/Home-Assistant#tip-sync-auto-discovered-devices-including-power-state

PR #4517, opened 2018-12-03
Domoticz fanspeed support

mzealey:
This patch cleans up fan support to make it easier to add new fan devices, and then introduces a fan connection to domoticz selector switch.

### DIFF
--- a/sonoff/language/bg-BG.h
+++ b/sonoff/language/bg-BG.h
@@ -387,6 +387,7 @@
   #define D_DOMOTICZ_CURRENT "Ток/PM10"
   #define D_DOMOTICZ_AIRQUALITY "Качество на въздуха"
 #define D_DOMOTICZ_UPDATE_TIMER "Период на опресняване"
+#define D_DOMOTICZ_FAN "Fan"
 
 // xdrv_09_timers.ino
 #define D_CONFIGURE_TIMER "Конфигуриране на таймер"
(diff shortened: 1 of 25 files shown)


mzealey:
(Note I am not sure how localization should work so I only included the new English string in the changeset)

ascillato:
Hi, If the key is not included in all translation files, the automatic compilation will fail, so this PR can't be merged until that is complete.

mzealey:
> Hi, If the key is not included in all translation files, the automatic compilation will fail, so this PR can't be merged until that is complete.

Updated with the define copied to all locales then.

arendst:
Thx for the idea!

As it stands it will change the settings lay-out as the unsigned long needs to be defined on a 4 byte boundary which is currently not the case.

Domoticz config on ifan02 comes with four Idx parameters; one for each relay within the the device. Idx 1 is used for the light relay; Idx 2 - 4 are the fan relays.
I'm considering the use of Domoticz Idx 2 instead of a dedicated fan memory location as that would be more appropriate and also saves a specific slot while there are plenty of not used power slots available.

I'll rewrite using your groundwork.

mzealey:
@arendst Good idea about using the idx. But why not idx 4 for the fan; in case something else gets attached in future?

A year or two back I got some ir ceiling fans with 2-speed relay and replaced the controller with an esp8266, hooked in the two pull-chains as switches and added a dht22. So I thought rather than maintaing my previous custom code, once these patches are accepted my device is about 10 lines of additional code and some rules.

arendst:
CLosing this as it has been kind of implemented ;-)

PR #4531, opened 2018-12-04
MQTT Discovery - Announce switches

emontnemery:
Announce switches through MQTT discovery.

Please also note the change to `SendKey()` to make sure suffix is controlled by `Settings.flag.device_index_enable`, and not by key or switch type.
Please let me know if it was not a bug and Switch 1 should always be suffixed, i.e. always `POWER1` instead of `POWER`.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1311,7 +1311,7 @@ boolean SendKey(byte key, byte device, byte state)
   if (Settings.flag.mqtt_enabled && MqttIsConnected() && (strlen(key_topic) != 0) && strcmp(key_topic, "0")) {
     if (!key && (device > devices_present)) device = 1;                                             // Only allow number of buttons up to number of devices
     GetTopic_P(stopic, CMND, key_topic, GetPowerDevice(scommand, device, sizeof(scommand),
-               Settings.flag.device_index_enable));                                                 // cmnd/switchtopic/POWERx
+               (key + Settings.flag.device_index_enable)));                                                 // cmnd/switchtopic/POWERx
     if (9 == state) {
       mqtt_data[0] = '\0';
     } else {
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1310,7 +1310,8 @@ boolean SendKey(byte key, byte device, byte state)
   Format(key_topic, tmp, sizeof(key_topic));
   if (Settings.flag.mqtt_enabled && MqttIsConnected() && (strlen(key_topic) != 0) && strcmp(key_topic, "0")) {
     if (!key && (device > devices_present)) device = 1;                                             // Only allow number of buttons up to number of devices
-    GetTopic_P(stopic, CMND, key_topic, GetPowerDevice(scommand, device, sizeof(scommand), key));   // cmnd/switchtopic/POWERx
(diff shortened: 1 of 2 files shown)


Jason2866:
Feature (backwards compatibility)
SetOption26  0 / off  (default) Keep using POWER without postfix for single power devices
SetOption26  1 / on  Add postfix to all POWER messages

emontnemery:
@Jason2866 Exactly, that's Settings.flag.device_index_enable, and the suspected bug is that the flag is not used in `SendKey()`

ascillato:
@emontnemery 

Thanks a lot for the work you are putting on this.

The variable `KEY `is the actual index. Please undo that modification in _sonoff.ino_. The rest of modifications are ok. Thanks

emontnemery:
> The variable KEY is the actual index. 

No, that's not correct, did you check the code?
Input parameter `key` in function `boolean SendKey(byte key, byte device, byte state)` is a bool flag indicating if the `device` is a button or a switch.
`key` is passed as input parameter `option` to function `char* GetPowerDevice(char* dest, uint8_t idx, size_t size, uint8_t option)`

Input parameter `option` is a bool flag, and when set forces postfix number for all POWER messages.
This, I suspect is a bug, because it doesn't seem to make sense to use the bool flag `key` for that.

ascillato:
You are right, I was recalling from memory. The index is the variable device, not key. Sorry. My bad.

Anyway, I still think that the modification to sonoff.ino is not necessary.

The legacy support is that when the variable `devices` is 1 (just one relay or light or led) the result stat topic will be stat/sonoff/POWER ON

The new option is that does not matter how many outputs you have, always will be POWERx

I think that the discovery should use as the user has set in setoption26

Anyway, the discovery should be resend if this setoption26 is changed (obviously deleting the previous retained messages)

emontnemery:
> The legacy support is that when the variable devices is 1 (just one relay or light or led) the result stat topic will be stat/sonoff/POWER ON
The new option is that: does not matter how many outputs you have, always will be POWERx
I think that the discovery should use as the user has set in setoption26

This is exactly the point of the modification to sonoff.ino

**Before change:**
 - setoption26 is not respected
 - button 1 is always POWER
 - switch 1 is always POWER1

**After change:**
 - button 1 is POWER or POWER1 according to setoption26
 - switch 1 is POWER or POWER1 according to setoption26

emontnemery:
Latest commit has improved logic for announcing button / switch as ON/OFF or TOGGLE.

ascillato:
Hi,

I test and your proposed fix in _sonoff.ino_ and it is not needed. In the actual code the suffix is working fine. It is user for sending POWER or POWER1 (only for 1 output case).

It has nothing to do with Switches or Buttons. I test both cases and POWER is sending respecting SETOPTION26

emontnemery:
I'll do some more testing myself, maybe it was my mistake.

Do you think they other changes look good?

emontnemery:
@arendst With your change, the logic is now:
button 1 is POWER or POWER1 in button MQTT message according to setoption26
switch 1 is always POWER1 in switch MQTT message, setoption26 is not respected

Can you please confirm this is the expected behavior?
I'll then fix the Hass discovery logic accordingly.

arendst:
Correct. Switches always had an index as controlled by the key flag.

PR #4536, opened 2018-12-05
Bitwise dev

devanl:
Add commands for bitwise operations on variables to rules:
- & operation for rule trigger
- Or<n> command against Var<n>
- And<n> command against Var<n>
- Msb<n> to get most significant bit set in Var<n>


ascillato:
Can you please post an usage example?

ascillato:
Hi, can you please post an usage example? In which cases are you using this commands?

andrethomas (commented):
I can't think of one - usually bitwise operations are done inside a sensor or driver code.

andrethomas:
I would gladly think about what you're trying to achieve on a non-working day because right now I'm a bit perplexed by the methodology.

Perhaps it will become clear to me in the not so distant future.

andrethomas:
Just attaching this from Discord Chat as to not confuse myself later :)

image

ascillato2:
As the use case explained in Tasmota Chat has another workaround and there is no use case for this new commands, it is added the label **Feature Request (hold over)** for future reference.

If there is any new idea, please ask to reopen this to review new proposals.

Thanks everyone for sharing their ideas!!!! :+1:

PR #4562, opened 2018-12-08
Enable forced local operation when button- or switchtopic is set

emontnemery:
Add new option `SetOption61` which enables forced local operation when button- or switchtopic is set

The purpose is to allow non-ambiguous switch topic for automatic while still have fast local operation.
In my case, this is needed for good user experience with Sonoff T1 switches.

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -96,6 +96,7 @@
 #define MQTT_BUTTON_RETAIN     0                 // [ButtonRetain] Button may send retain flag (0 = off, 1 = on)
 #define MQTT_POWER_RETAIN      0                 // [PowerRetain] Power status message may send retain flag (0 = off, 1 = on)
 #define MQTT_SWITCH_RETAIN     0                 // [SwitchRetain] Switch may send retain flag (0 = off, 1 = on)
+#define MQTT_BUTTON_SWITCH_FORCE_LOCAL     0     // [SetOption61] Force local operation when button/switch topic is set (0 = off, 1 = on)
 
 #define MQTT_STATUS_OFF        "OFF"             // [StateText1] Command or Status result when turned off (needs to be a string like "0" or "Off")
 #define MQTT_STATUS_ON         "ON"              // [StateText2] Command or Status result when turned on (needs to be a string like "1" or "On")
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -74,7 +74,7 @@ typedef union {                            // Restricted by MISRA-C Rule 18.4 bu
     uint32_t receive_raw : 1;              // bit 8 (v6.3.0.11)
     uint32_t hass_tele_on_power : 1;       // bit 9 (v6.3.0.13)
     uint32_t sleep_normal : 1;             // bit 10 (v6.3.0.15) - SetOption60 - Enable normal sleep instead of dynamic sleep
-    uint32_t spare11 : 1;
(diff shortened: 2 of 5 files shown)


meingraham:
Regarding SetOption61 (I don't think this is documented in the wiki yet)... The title of this post makes it obvious what the intent of this option is. What I was wondering is what is the behavior; i.e., what's the difference in "local operation" if this option is not enabled. The title implies that local operation is not affected if ButtonTopic or SwitchTopic is not set. Other than their effect on Rules (i.e., basically disabling rules) is there an explanation in the wiki of the impact on **_local operation_** of setting ButtonTopic or SwitchTopic? In reading those section, it was not obvious to me how, if so. Is there an example or use case that clarifies this? @emontnemery - perhaps an explanation as to why this option was needed for HASS auto-discovery?

There was a thread over on the Sonoff Google Groups group about some issues with timers operation because they were repeatedly attempting to reconnect to MQTT. Obviously a timer that is dependent on time of day (i.e., NTP sync) can't operate. But once time is synced, local operation of a timer ought not to be affected by a MQTT issue. Also, timers which are intervals (e.g., every n seconds) ought to never be affected by connectivity issues.

So, I mention this thread because this "local operation" topic could be broad in definition. That is why I'm trying to better understand what SetOption61 does exactly, why it should be used, and under what situations.

Regards.

Mike

emontnemery:
@meingraham:
Tasmota will always send MQTT message when a relay or light has been turned on / off.
This message is sometimes not possible to use in automations, because it's impossible to know from the MQTT message if the relay state was changed because the user pressed the button or for some other reason.
An example is a group of Sonoff T1s setup to emulate a stair switch.

- Without switchtopic set, the button or switch simply operates a relay / light, this is the default case.
- With switchtopic set, the button will send an MQTT message and NOT operate the relay
   - A home automation system, e.g. Hass, can pick up this message and it's unambiguous that the message was sent because the user pressed a button.
   - This is perfect for automations, but it may be quite slow, and as a result user experience is quite terrible for the devices with touch button (Sonoff Touch, Sonoff T1)
   - Note: In the special case of setting switchtopic to same as the relay's topic (`swithctopic 1` is a shortcut for this), it will indirectly operate the relay because the MQTT message will be sent back to the Tasmota device

With SetOption61 set, pressing a button will operate a relay AND send an MQTT message, in this way it's clear that the user pressed the button, but without any lag.

meingraham:
Thank you for the explanation! This explanation would be great to add to the wiki somewhere. Yes, I can that myself... but I'm not sure where the appropriate place to put this is. Commands? Seems a bit long for the "regular" command entry. But it would be so helpful to the understanding of the use of the option.

Recommendations?

emontnemery:
I think this should be added to the wiki page about switchtopic. That page should then also be renamed and expanded to explain both buttons and switches. I've been meaning to do that for a long time.

meingraham:
@emontnemery,
 
I may be speaking for myself, but I think I probably am a member of a larger club. The club? Those who are quite confused about what SwitchTopic and ButtonTopic are used for, how they are used, what scenarios are, how they alter the behavior of the switch, etc. If you were to expand the wiki, you'd have a lot of fans, IMO. I know that there are a few YouTube tutorials in the topic. I've watched several. I'm no less confused :( I think if I understood it better, I'd likely find that I could make use of these parameters.

I'll leave the incorporation of SetOption61 into the *Topic wiki page to you as you have the better understanding of the parameters (understatement of the day!).

Cheers!

Mike

PR #4570, opened 2018-12-09
Analog scaling for ADC analog input A0

evzone:
Hi,

The current implementation of analog input (sensor02) uses only raw values (0-1024)
This update adds the new feature of scaling the raw analog value to engineering units.

The scaling is based on the linear formula:
**analogEU = Scale * analogRaw + Offset**
where:
- analogEU: The scaled analog value in engineering units. The type is double.
- Scale: The scale factor. The type is double
- analogRaw: The analog raw value read by the ADC. The type is uint16_t
- Offset: The offset value. The type is double.

The scale and offset values are set by using the new command "sensor02".
See the comments in xsns_02_analog.ino for details about its format.

Also, the number of decimals for the scaled analog value can be set by using the "sensor02" command.

The scaling parameters are stored in flash memory and can be initialized in my_user_config.h

If this update is accepted I will create a new Wiki page on how to use the analog input with the new feature.

Thanks,
evzone


andrethomas:
I understand what you are trying to achieve but it may not be useful in real world scenario's.

To get proper use from ADC one would have to apply some form of aggregation or averaging as is usually the case with any ADC insofar that it does not consistently provide the exact value as you would theoretically or mathematically expect.

So you might say, lets add averaging...

The problem is that the strategy used to obtain useful data from an analog pin for a specific use case would change from case to case as not all strategies/filtering mechanisms would be common for all use case scenarios.

I personally do not think it is necessary to implement this option and I think it may add a layer of complexity when using the A0 value for simple rule based mechanisms.

Just my 2 cents... its obviously not my call :)

ascillato2:
@evzone 

Thanks a lot for sharing your ideas and work but It is not going to be added this feature soon as explained before. Anyway, this PR has been labeled as Request Feature (hold over) to be reviewed in the future.

If you have any comment, please feel free to find us in the Tasmota Chat.

Thanks

evzone:
This idea was driven by a thermistor IC like the MCP9700 from Microchip ([http://ww1.microchip.com/downloads/en/devicedoc/20001942g.pdf](url)

I agree that the voltage output is not accurate and it needs some kind of filtering or averaging.
I will keep this feature in a separate branch in my fork for anyone interested to use it.

Whenever you decide to add the scaling functionality just let me know and I would be glad to add some averaging too.

Thank you

PR #4598, opened 2018-12-12
Read response on WebSend instead of just fulshing it

JulienVdG:
This makes the communication between two tasmota devices reliable.

Exemple to remote control a S26 from a T1:
rule1 on Power2#State do WebSend [sonoff-2] Power %value% endon

### DIFF
--- a/sonoff/xdrv_01_webserver.ino
+++ b/sonoff/xdrv_01_webserver.ino
@@ -1981,7 +1981,16 @@ int WebSend(char *buffer)
 //AddLog(LOG_LEVEL_DEBUG);
 
         client.print(url.c_str());
-        client.flush();
+//        String rsp = "";
+        while (client.connected())
+        {
+          if (client.available()) {
+            char c = client.read();
+//            rsp += c;
+          }
+        }
+//        snprintf_P(log_data, sizeof(log_data), PSTR("DBG: Resp |%s|"), rsp.c_str());
+//        AddLog(LOG_LEVEL_DEBUG);
         client.stop();
         status = 0;                           // No error - Done
       } else {


ascillato:
Hi,

Thanks for sharing this, but this PR hungs tasmota until the reception is done. So Tasmota stops taking care of WIFI, MQTT, BUTTONS, RULES, etc.

Sorry but blocking routines can not be merged.

Instead you can use MQTT or KNX for achieving this. Thanks.

ascillato:
Your idea is nice, but as it isn't async, can not be used in Tasmota. You can wait less time, but will be waiting anyway and depending on the quality of your network and wifi connection, this time could be enough for blocking the rest of Tasks. Sorry, blocking routines can not be merged.

JulienVdG:
Thanks for your quick answer.
I understand your point. But to be fair, the client.print is probably already taking time and waiting on the network (I didn't profile it though).
I might revisit this later and make it async.

And thank you for making Tasmota !

ascillato:
You are totally right about the points of failure. You have the device itself, the router and the mqtt broker. I can not give a general statistic of reliability but speaking for myself, in all this year I have 3 or 4 times I needed to reboot the wifi router, and once some of my sonoff devices and also once my Mqtt Broker.

So, at least in my case, the wifi router was the most common to fail, and was very few times. The Mqtt broker (mosquitto on raspberry pi) is good enough (at least in my case).

You can use (as I do) mqtt + knx for some of the task that should not fail like the stair lights (2 switches that turn on/off the same light). You can send the power on command by mqtt and knx. If the reception device has for example 2 commands to power on, will power on and nothing more. It will be duplicated but will work without issues. This is not the same for toggle commands obviously

PR #4601, opened 2018-12-12
 Update fr-FR.h

localhost61:
Updated SDM220 strings and reworked to fit in the space available


localhost61:
In that specific case, Travis check is funny because it doesn't generate Sonoff-FR.bin but Sonoff.bin... that doesn't include fr-FR.h in sonoff.ino.

PR #4653, opened 2018-12-17
webserver: swap index and name in module configuration

chaosmaster:
This patch swaps index and name in module selection and gpio-configuration.

So instead of
`01 Sonoff Basic`

it is now
`Sonoff Basic (01)`

This change makes it easier to find a module/gpio because you can now simply start typing the name to jump to the correct one.
I would also suggest ordering the lists alphabetically, but I'd like to hear your thoughts on that.


arendst:
The list is currently sorted on Sonoff followed by functional identical devices so for now I would leave it that way.

PR #4660, opened 2018-12-17
platformio: fix build with core 2.4.2

chaosmaster:
`pio run` fails with the following message when building against core 2.4.2:

```
/home/chaosmaster/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld: cannot open linker script file eagle.flash.common.ld: No such file or directory
collect2: error: ld returned 1 exit status
*** [.pioenvs/sonoff/firmware.elf] Error 1
```

Removing the explicit linker-file _esp8266.flash.1m0.ld_ fixes this.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -53,7 +53,6 @@ build_flags               = ${esp82xx_defaults.build_flags}
 ; *** Esp8266 core for Arduino version 2.4.2
 platform                  = espressif8266@1.8.0
 build_flags               = ${esp82xx_defaults.build_flags}
-                            -Wl,-Tesp8266.flash.1m0.ld
                             -lstdc++ -lsupc++
 ; lwIP 1.4 (Default)
 ;                            -DPIO_FRAMEWORK_ARDUINO_LWIP_HIGHER_BANDWIDTH


arendst:
Should not be needed. You probably miss the linker file in the root. If not used the flash image won't be correct.

chaosmaster:
I have _esp8266.flash.1m0.ld_ in root and it builds fine when I build against core 2.3.0.
With core 2.4.2 it complains about not finding _eagle.flash.1m0.ld_ which doesn't exist in the repository. 
But it exists in platformio framework

> ../.platformio/packages/framework-arduinoespressif8266/tools/sdk/ld/eagle.flash.1m0.ld

arendst:
Never had this problem. I use 2.4.2 since it's release.

chaosmaster:
Strange, I wonder why it complains about _eagle.flash.1m0.ld_ when _esp8266.flash.1m0.ld_ is referenced in the command.
I am building on Linux btw.

Jason2866:
Had this problem too a time ago.... Setup of platformio is damaged

chaosmaster:
Thanks @Jason2866.
After completely reinstalling platformio using
`pip uninstall platformio && pip install platformio"
The problem still persists.
Any idea how to get it fixed?

Jason2866:
Let talk in Discord! https://discord.gg/Ks2Kzd4

PR #4667, opened 2018-12-18
Display WiFi channel on web information page

andrethomas:
Update Information web page to display channel current channel number.

image

Request eminates from https://github.com/arendst/Sonoff-Tasmota/issues/3664#issuecomment-448061327

### DIFF
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,6 +1,5 @@
 /* 6.4.0.1 20181217
  * Add support for AZ-Instrument 7798 CO2 meter/datalogger (#4672)
- * Display WiFi channel number on Web UI Information page (#4667)
  *
  * 6.4.0 20181217
  * Change GUI Configure Module by using AJAX for data fetch to cut page size (and memory use) by 40%
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,5 +1,5 @@
 /* 6.4.0.1 20181217
- * New release
+ * Display WiFi channel number on Web UI Information page (#4667)
  *
  * 6.4.0 20181217
  * Change GUI Configure Module by using AJAX for data fetch to cut page size (and memory use) by 40%
--- a/sonoff/xdrv_01_webserver.ino
+++ b/sonoff/xdrv_01_webserver.ino
@@ -1295,7 +1295,7 @@ void HandleInformation(void)
 
   func += F("}1}2&nbsp;");  // Empty line
   func += F("}1" D_AP); func += String(Settings.sta_active +1);
-    func += F(" " D_SSID " (" D_RSSI ")}2"); func += Settings.sta_ssid[Settings.sta_active]; func += F(" ("); func += WifiGetRssiAsQuality(WiFi.RSSI()); func += F("%)");
+    func += F(" " D_SSID " (" D_RSSI ")}2"); func += Settings.sta_ssid[Settings.sta_active]; func += F(" ch"); func+= WiFi.channel(); func += F(" ("); func += WifiGetRssiAsQuality(WiFi.RSSI()); func += F("%)");
   func += F("}1" D_HOSTNAME "}2"); func += my_hostname;
   if (mdns_begun) { func += F(".local"); }
   if (static_cast<uint32_t>(WiFi.localIP()) != 0) {


arendst:
Why is this needed? It's already in the JSON STATE message...

ascillato2 (approved):

andrethomas:
@Pewibe

The PR is ready but @arendst is asking why this is needed if it is already in the telemetry data.

Pewibe:
@andrethomas and others
1. I think the question "why needed" is misplaced because Tasmota is based on 3 information exchanges (MQTT, console, webserver). As long as there is no reason for a difference most information should be available on all 3 (sorry of my structured thinking).
2. WLAN data shows all access points and their channel / signal strength. The list is sorted by %. If the first list entry is the really used access point I don't know. 
3. I use TasmoAdmin (where I have requested the WiFi info, too) and a click shows the Tasmota webserver page. Because I have problems with some Sonoff modules and the usage of WiFi, every click counts (information menu 1 click, WLAN menu 3 clicks). Sometimes I have to click 5 times until a click works. 

Overall, the discussion about that change consumes more time than the change itself - right?

andrethomas:
> I think the question "why needed" is misplaced because Tasmota is based on 3 information exchanges (MQTT, console, webserver). As long as there is no reason for a difference most information should be available on all 3 (sorry of my structured thinking).

The intended purpose of the web ui is actually limited to configuration and diagnostic purposes.

> WLAN data shows all access points and their channel/signal strength. The list is sorted by %. If the first list entry is the really used access point I don't know.

If you use the correct WiFi hardware you can see which AP each device is connected to. Since I assume you do not use a hardware-based WiFi management system you can actually use the telemetry data to build a web page that will show you what you need in a more useful way than having to visit the web ui of each device and manually record this information.

> I use TasmoAdmin (where I have requested the WiFi info, too) and a click shows the Tasmota webserver page. Because I have problems with some Sonoff modules and the usage of WiFi, every click counts (information menu 1 click, WLAN menu 3 clicks). Sometimes I have to click 5 times until a click works.

Same as the previous answer.

> Overall, the discussion about that change consumes more time than the change itself - right?

It depends on your perspective (project owner, developer, or user) - yes the change is easy and quick but there are alternative ways of obtaining this information more efficiently for your specific use case. The problem with adding code is that every byte of flash and ram starts to count. Moving to 2.4.2 added about 30k of core and 2.5.2 seems to be heading in the same direction so we are in fact running out of usable flash and ram.

As for the time spent on making the PR it was my time which I give freely but in the end, it is up to the project owner as to whether or not this feature is to be added. The time it takes to respond to the points you have raised is also my time which I also give freely.

The initial request sounded nice to me which is why I did it, but in hindsight and taking your actual use case scenario into consideration I would recommend collecting telemetry data from the STATUS JSON and populating this into a web page on your HA system.

To give insight into how I handle it...

I do not do this yet as I do not have problems with my managed AP's (Managed with Capsman on Mikrotik hardware) but what I do have (and is something you should probably consider given that your request likely emanates from potential disconnects due to being connected to an incorrect or undesirable AP) is that I have a watchdog timer on node-red which alerts me when a device is no longer transmitting telemetry data (this happens so rarely that I actually induce this behaviour from time to time to ensure that the functionality is working as it should) - when telemetry data is received the watchdog timer is reset to the predefined time within which to expect the next telemetry data.

I, therefore, come to the conclusion that this PR is a nice to have but not necessary... given that it is available for you to take the line of code out if you choose to use them but in the end, it is up to @arendst whether or not this will be merged to development.

arendst:
I do not understand why you want a GUI telling you the same info as the command ``status 0`` executed from the GUI console window...

The same for command ``state`` which will tell you in the console on the GUI what channel you are on.

You'll have to understand that why someone wants this info another one wants other info. Both can be satisfied by executing the GUI console commands

andrethomas:
@Pewibe 

Firstly I take exception to the fact that you assume I do not understand the basic needs of people.

Secondly, I did not press any rewind button - The PR is here and is ready to be merged but it is entirely up to @arendst as to whether or not this is what he wants to add or not. 

This is not my project and I have no intention of running a project parallel to this one so my contribution stops at creating the PR, if merged creating the wiki information, and extend support via Issues and Discord Chat and of course engage in any changes required to previously merged code (of any origin)

Lastly, I do not yet use TasmoAdmin myself (not gotten around to using it as I have limited time) but my understanding is that the channel information would be obtained from the json via http request to the Tasmota device and not from the web ui information page which is why it is not a valid reason to have this added to the information web ui.

You can do the latter as well by using http url `http://your.device.ip.address/cm?cmnd=state`

My only part in this process was to make the small code change which was necessary to achieve what you requested but seemingly the reason behind the requirement does not have enough TAF in it :)

PR #4672, opened 2018-12-18
Add support for AZ-Instrument 7798 CO2 meter/datalogger

adebeun:
Implemented sensor interface for the AZ-Instrument 7798 CO2/temperature/humidity meter/datalogger.
See: https://www.az-instrument.com.tw/az-instrument/en/productsinfo/189.html
Tested with a Wemos D1 mini. (Can be fitted neatly inside the instrument.)

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -182,7 +182,6 @@ The following binary downloads have been compiled with ESP8266/Arduino library c
 | USE_RC_SWITCH                  | - | - | - | x | x | x |
 | USE_RF_SENSOR                  | - | - | - | - | - | x | AlectoV2 only
 | USE_DISPLAY                    | - | - | - | - | - | - |
-| USE_AZ7798                     | - | - | - | - | - | - |
 
 ## Changelog
 Version 6.4.0 20181217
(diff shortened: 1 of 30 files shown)


ascillato2 (changes requested):
Hi, thanks for sharing this. There are several points that need be addressed. This PR will not compile due problems in language files. 

Platformio.ini should not be changed. Those changes has nothing to do with your driver.
My_User_config.h should be left as default. Too much changes not related to this driver. Besides the key that enable this driver should be left commented out.
There is missing some translations and also there is a conflict in one of the files that needs to be resolved. Without all those changes, can't be merged due this can't compile.

Thanks

andrethomas2 (changes requested):
Thanks for the PR, but I agree with @ascillato2 

I'm inclined to think you should create a new PR which only has the necessary changes as opposed to trying to clean this one up.

adebeun:
I made the changes as requested, reverting my_user_config and platformio.ini to their original state, but adding the new sensor commented out. I retrieved the updated bg-BG.h and added my 2 new lines. However it is still showing as conflicted, I don't know how to fix that.

andrethomas:
The conflict needs to be resolved

image

So just click on resolve conflict and edit it accordingly.

andrethomas (changes requested):
Also, please use the original RELEASENOTES.md as this is not included in the 6.4 release and your PR is proposed for the current development codebase

andrethomas2 (changes requested):
Also, please use the original RELEASENOTES.md as this is not included in the 6.4 release and your PR is proposed for the current development codebase

andrethomas2 (approved):

ascillato (approved):

ascillato2 (approved):

andrethomas (approved):

PR #4698, opened 2018-12-20
add missing units of measurements for energy HA autodiscovery

reey:
Added a dedicated energy sensor management to report back as energy sensor and not fallback as a generic sensor in HA autodiscovery process


ascillato:
Cool. Thanks a lot.

Could you add also the counter sensor and the luminosity sensor please?

ascillato (approved):

reey:
I could do that in another pull, but what do you mean by counter sensor?

PR #4703, opened 2018-12-21
Added option for checking mqtt tls against root ca

fmeies:
Using fingerprints for tls verification can be somewhat cumbersome when using regularly changing letsencrypt certificates. This patch enables checking against a ca cert.

Signed-off-by: Frank Meies <19324766+fmeies@users.noreply.github.com>


arendst:
Will need some core version tests as core 2.3.0 does not support CA's

fmeies:
Thank you for your feedback. I suggest to have this consistency check as early as possible and abort the build in case 2.3.0 is used together with using CA's. I can add a test in my_user_config.h similar to the test for not using both USE_MQTT_TLS and USE_WEBSERVER. What do you think?

arendst:
Great improvement.

How often does this cert need to be updated? Is it a general cert? Can anyone use the same cert?

fmeies:
Yes, it's the general IdenTrust DST Root CA X3 cert which has been used to cross sign the letsencrypt intermediate certificate, see https://letsencrypt.org/certificates. It can be obtained from here : 
https://www.identrust.com/support/downloads
It can be used to validate any certificate issued by letsencrypt and is valid until 09-30-2021.

Issue #4708, opened 2018-12-22
HomeAssistant Discovery incomplete in lengthy mode


joelhaasnoot:
**Describe the bug**
_A clear and concise description of what the bug is._
When running Home Assistant auto discovery turned on (SetOption19 1) without 'abbreviations and base topic' turned on (SetOption55 0), discovery on Home Assistant works, but fails to create editable entities in Home Assistant. Looking at the 'announce' body, several keys are missing in the lengthy mode, most notably _unique_id_. Without unique_id an entity in the Home Assistant registry isn't created, and while the device appears, it's not possible to edit the entity_id in Home Assistant. 

_Also, make sure these boxes are checked [x] before submitting your issue - Thank you!_
- [x] _Searched the problem in issues and in the wiki_
- [x] _Hardware used_ : Sonoff Basic / S20
- [x] _Development/Compiler/Upload tools used_ : Various, both OTA updates and VS Code with Platform IO
- [x] _If a pre-compiled release or development binary was used, which one?_ : 
- [x] _You have tried latest release or development binaries?_ : Yes, latest version
- [x] _Provide the output of command_``status 0`` :
```
02:05:32 MQT: sonoff-5/stat/STATUS = {"Status":{"Module":1,"FriendlyName":["Sonoff"],"Topic":"sonoff-5","ButtonTopic":"0","Power":0,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":0}}
02:05:32 MQT: sonoff-5/stat/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"http://thehackbox.org/tasmota/release/sonoff.bin","RestartReason":"Software/System restart","Uptime":"0T03:44:27","StartupUTC":"2018-12-21T21:21:05","Sleep":50,"BootCount":6,"SaveCount":35,"SaveAddress":"F9000"}}
02:05:32 MQT: sonoff-5/stat/STATUS2 = {"StatusFWR":{"Version":"6.4.0(sonoff)","BuildDateTime":"2018-12-20T22:18:06","Boot":6,"Core":"2_4_2","SDK":"2.2.1(cfd48f3)"}}
[...]

PR #4710, opened 2018-12-22
Make SetOption55 enabled by default

emontnemery:
As discussed in #4708 

Edit: Closing this in favor of removing SetOption 55.


arendst:
I was just looking at the code.

How about removing all code regarding the "old" way and only provide the "short" code. This makes the code a lot smaller I guess.

Takes some more time and I'm sorry I insisted on having both until January. I didn't anticipate the easiness of HASS updates.

emontnemery:
Yes, I'm working on it now. I'll open PR in a little while.
Agree it simplifies the code and saves some program memory by removing the duplicated string constants.

PR #4711, opened 2018-12-22
Remove SetOption55 (hass_short_discovery_msg)

emontnemery:
As discussed in #4708: Remove SetOption 55 which controls use of shortened Hass MQTT discovery messages and only support the shortened version.


emontnemery:
@arendst To avoid even more questions now that behavior has changed, maybe make make an offical release (6.4.1?) including this change?

arendst:
Yes. Was thinking about it too.

Will do.

Issue #3159, opened 2018-07-08
Hue emulation Alexa works but says "this value is outside the range of the device"


yonaesp:
I'm using dual sonoff, with wemo it only detects one relay, with hue detecting both and it works perfectly when I say "Alexa, turn on button 1", normally alexa must answer "ok" but in this case he answers "this value is outside the range of the device" however, it turns on and off without problem, and when I use "Alexa, turn off the button 1" it responds correctly "ok".

My device is in Spanish since I am participating in the closed beta, it is likely that in English it will respond with a different word as indicated here: [1639](https://github.com/arendst/Sonoff-Tasmota/issues/1639)

I have also tried modifying the line indicated in this comment, [360218536](https://github.com/arendst/Sonoff-Tasmota/issues/1639#issuecomment-360218536) recompiling the firmware and installing it, but the problem persists.

Another problem that I have, from the alexa app I can turn on the relays but I can not turn them off, however using the voice command if I can do it.

this only happens to me with hue emulation.

Jason2866:
Maybe the name button 1 is for Alexa iritating. Change it to something common.
Maybe you trie light as name. In general that is not a Tasmota problem!
I use Alexa too, and it works perfect  hrre in Germany.

yonaesp:
The real name is "Arriba/Up", button 1 is only an example.

Jason2866:
"/" is not allowed just a simple usual word

yonaesp:
it's only "Arriba" ( /up is the translation that I have used here as an informative).

Jason2866:
Maybe Alexa isnt working like it should since it is in beta state in your region.
This help you not at all, but Tasmota is working correct with HueEmulation with Alexa in
countrys where is official supported

yonaesp:
Is possible, [Here](https://i.imgur.com/tPPDjD5.png) (Green = Translation to English) I attach some images of what appears in the app Alexa to see if that can indicate something.

I do not have any Skills activated, should I activate any?

It also does not allow me to edit the button settings

Jason2866:
No skill is needed to activate. Changing names edit button works here (Germany).
My setup looks like this.
![screenshot_20180708-121731](https://user-images.githubusercontent.com/24528715/42418847-66b0e1be-82a9-11e8-843e-ca52de36a2b9.png)
[...]

PR #4712, opened 2018-12-22
Remove Alexa message "this value is outside the range of the device"

gourry:
This workaround for issue #3159 is removing the message "this value is outside the range of the device" allowing me to use hue emulation with Alexa on an Amazon echo dot 3rd generation .

### DIFF
--- a/sonoff/xplg_wemohue.ino
+++ b/sonoff/xplg_wemohue.ino
@@ -568,7 +568,7 @@ void HueLightStatus1(byte device, String *response)
 {
   float hue = 0;
   float sat = 0;
-  float bri = 0;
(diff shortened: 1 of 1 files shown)


Jason2866:
This PR makes my Alexa setup fail

ciumagcatalin:
I don't see this working...This is what I see on the lights file when there is no power:
"8": {"modelid": "HASS123", "name": "Living lamp", "state": {"bri": 0, "on": false, "reachable": true}, "swversion": "123", "type": "Dimmable light", "uniqueid": "switch.sonoff_switch_01"},

Shouldn't the "bri:" be 1?

Issue #3406, opened 2018-08-04
How to configure common relay module (non-sonoff)


mattpurnell:
Make sure these boxes are checked [x] before submitting your issue - Thank you!

- [x] Searched the problem in issues (https://github.com/arendst/Sonoff-Tasmota/issues)
- [x] Searched the problem in the wiki (https://github.com/arendst/Sonoff-Tasmota/wiki/Troubleshooting)
- [x] Searched the problem in the forum (https://forum.digital-pig.com/)
- [x] Searched the problem in the discussion group (https://groups.google.com/d/forum/sonoffusers)
- [x] Hardware used : ESP01 with a relay module - with builtin voltage drop
- [x] Provide the output of command ``status 0`` :
```
STATUS 0 OUTPUT (please ignore the failure to connect to MQTT server)

17:57:42 RSL: stat/SPARE/STATUS = {"Status":{"Module":18,"FriendlyName":["SPARE","SPARE2"],"Topic":"SPARE","ButtonTopic":"0","Power":3,"PowerOnState":0,"LedState":1,"SaveData":1,"SaveState":1,"ButtonRetain":0,"PowerRetain":0}}
17:57:42 RSL: stat/SPARE/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"http://dsm.lcl:8008/api/arduino/sonoff.ino.bin","RestartReason":"Software/System restart","Uptime":"0T00:00:58","StartupUTC":"2018-08-04T05:56:44","Sleep":0,"BootCount":4,"SaveCount":17,"SaveAddress":"FB000"}}
17:57:42 RSL: stat/SPARE/STATUS2 = {"StatusFWR":{"Version":"6.1.1c","BuildDateTime":"2018-08-04T17:37:08","Boot":5,"Core":"2_4_1","SDK":"2.2.1(cfd48f3)"}}
17:57:43 RSL: stat/SPARE/STATUS3 = {"StatusLOG":{"SerialLog":2,"WebLog":2,"SysLog":3,"LogHost":"dsm.lcl","LogPort":514,"SSId":["Geogaddi_5Ghz","Geogaddi"],"TelePeriod":300,"SetOption":["00088009","55818000","00000000"]}}
[...]

Issue #1714, opened 2018-01-26
Customizing GPIOs


MacPit69:
Hello.

I flashed the tasmota firmware to an ESP8266 based relays from lctech-inc.com
It seems they used another GPIO than SONOFF to switch the Relais.
Is it possible telling the firmware the right GPIO for switching the relais?

thx
MacPit69

SodoffBaldrick:
I'm not sure if I completely understand the question, but I'll have a try.

On the web interface click Configure
Click Configure Module

Which version of firmware did you use?
If 5.11.1c or newer then under Module type select Generic
If 5.11.1b or older select Wemos D1 mini.

Click save.

Once its rebooted you'll now see a list of all the GPIOs and you can select the appropriate one and assign it to a relay.

Jason2866:
Hello,
lctech-inc.com wifi relay wotks not via gpio. It switches via commands on the serial line.
There is a pull request in which this relay is supported.

KPK75:
I did PR for 1 relay version. It won't work without changes in the code. What version do you have?

Jason2866:
Hi KPK75,
this one.
https://www.aliexpress.com/item/ESP8266-5V-WiFi-relay-module-Things-smart-home-remote-control-switch-phone-APP/32734997467.html?spm=2114.search0104.3.75.31cf7f44dPzlyy&ws_ab_test=searchweb0_0,searchweb201602_4_10320_10321_10065_10344_10068_10342_10547_10343_10322_10340_10548_10341_10193_10194_10084_10083_10618_10304_10615_10307_10302_10180_10313_10059_10184_10534_100031_10319_10103_441_10624_442_10623_10622_10186_10621_10620_10142,searchweb201603_28,ppcSwitch_5&algo_expid=cf6fc710-d9a9-4484-b117-72d0be1507a1-11&algo_pvid=cf6fc710-d9a9-4484-b117-72d0be1507a1&priceBeautifyAB=2
I unsoldered the Uart chip and connected direct one GPIO chip via the onboard Transistor to the relay.
Added some pullup resistors too. Now it works reliable.
So it is fully compatible to the Tasmota software, without any change in the software...
Just now ordered this one (i need some for a project, so better no or minor changes on the board and the have to be cheap...)
[...]

PR #1282, opened 2017-11-30
Support for LC Technology WiFi Relay

KPK75:
No buttons. ESP-01s is used to control one relay via STC15F104W PIC ( just like Sonoff Dual ).
Blue LED is connected directly to relay.
[http://www.chinalctech.com/index.php?_m=mod_product&_a=view&p_id=1204](url)

PR #4716, opened 2018-12-23
Added support for LC Tech boards

WelterRocks:
Added support for LC Technology 1, 2 and 4 channel boards, based upon ESP01 as xdrv_20. Also added RAW ESP01 for modified LC Tech boards (e.g. some clones). Added global bool suppress_serial_logging, which prevents users from setting serial log to other values than NONE and disturb serial controlled devices that way. Suppression disables serial log settings in WebUI and prevents log functions from sending over serial, if suppress_serial_logging is set to true (needed for xdrv_20 and might be useful for other serial controlled devices, too).


ascillato2:
I have deleted the `CFG_HOLDER `change you did in _my_user_config.h_

**_That change would delete all the config for anyone updating Tasmota_** (like a VIRUS !!! )

ascillato2:
Hi,

Thank you very much for sharing your ideas.

The support for LC Tech boards was previously requested in #3406 and in #1714.

Then the PR #1282 was proposed to add support for LC Tech boards, but was not merged into Tasmota (https://github.com/arendst/Sonoff-Tasmota/pull/1282#issuecomment-425886750) in order to mantain the device list in the firmware as short as possible and also due to this board can be used selecting Generic module and setting rules as explained in the wiki at https://github.com/arendst/Sonoff-Tasmota/wiki/LC-Technology-WiFi-Relay (Thanks @KPK75).

ascillato2:
@arendst @andrethomas What do you think?

arendst:
Won't merge as it changes too much core functionality just to send some serial commands to a relay board. It uses too much resources like a 32 byte entry in template for every possible relay configuration. 

There musr be an easier way to talk to these boards without soo much overhead.

andrethomas:
I agree with @arendst 

One alternative would be to use driverXX commands to control the relay states - I suggest this because technically the LC Tech is an expander in perhaps similar ways to the MCP230xx.

VerboteneZone:
@ascillato2 I understand your arguments to hold the device list as small as possible. But the primary reason, why I did it that way is, because the LC Tech boards are getting much more popular, now. I have approx ~ 60 Sonoff devices in use and bought some LC Tech boards for smaller projects, where ESP01 devices are needed. For me (and much other people - see forums e.g.) the LC Tech boards are a cheap and easily to implement alternative. Also their product portfolio grows up and is interesting for the use with tasmota (like Sonoff is now). The rule thing is primary a good idea, but if you plan to manage more than a few devices, or have to do many changes, entering rules really sucks ;), especially if you are a coding novice. This is why users want to have a native way, to use this boards by click-and-run. I think it is a better way to split the device list, in the future, in CPU boards and working (expander) boards partitions. There are thousands of use cases and currently many boards will come up with sockets for units like ESP01. Shouldn't that be up to the user to select the way, a software compiles and works? I thought therefore the switch USE_LC_TECH_RELAY exists, which can be commented out. The devices e.g. also can be moved to commented out section and made be available over IFDEF. So the users (and there are some that seem to like native LC Tech support for many reasons) can decide, what they want. The code of xdrv_20 itself isn't too much overhead, I think. I really understand the developers point of view, but there is also a user site ;). Merry X-Mas to all.

VerboteneZone:
@arendst you talked about core functionality. There is a change about suppress_serial_logging. I really like that idea, because there are other devices, controlled by serial as well. I have some use cases, where serial logging (if enabled) is preventing devices from working correctly and having some strange malfunctions. It is not really funny, if a relay starts flickering, because of some fool (like I was), turns on serial logging accidently on a serial controlled relay board. That was an expensive experience to me, therefore I implemented that variable ;)

Jason2866:
Mhh, why just not simply add rules for LC-Relais in wiki?

VerboteneZone:
@Jason2866, the idea is to prevent dangerous accidents. Pls, remember: this devices often control dangerous currents! Actually there is no protection mechanism, which stops (novice) users from using serial rules by simultaneously use serial logging, with serial controlled devices.

In case of LC Tech boards, this can (and probably will) create a situation, where a relay starts flapping. This can cause the unit to lit up or destroy other dangerous things. This situation happens to me, after accidently activated serial logging on the wrong device. 

Also, no one likes adding rules to multiple devices by hand. I am currently changing the code, so the "overhead" in sonoff_template.h can be controlled by the user itself, by #ifdeffing the affected parts and testing it against my units. I also agree arendst, but if we implement a channel selector for e.g., the overhead stuff in sonoff_template.h isnt needed anymore. If you are able to set devices_present by WebUI or user comand, you will only need one template per device and not three or more (this affects other devices, too).

ascillato:
@Jason2866 how to add this boards and the rules are already written in the wiki at https://github.com/arendst/Sonoff-Tasmota/wiki/LC-Technology-WiFi-Relay

VerboteneZone:
Commented out USE_LC_TECH_RELAY in config and added #ifdef in sonoff_template.h.

Of course, this couldn't be the final solution, but it is a way for people using these LC-Tech boards. Later, I suggest to change the template handling in general and make user changes available to devices_present, e.g.

Now, its up to you and many thanks for your suggestions and opinions.

andrethomas:
I guess we will start to see more and more non-Sonoff type devices demanding for Tasmota support.

@arendst Perhaps it could be viable to conditionally include or exclude GENERIC support by splitting these kinds of functionality off into a sonoff-generic.bin perhaps... dunno, just asking.

So #define BE_GENERIC or USE_GENERIC would result in all the normal Sonoff device support to be removed and only drivers etc for generic use to be listed for GPIO pins.

Just throwing a stone into the bush... sometimes something comes out... other times nothing :)

arendst:
Any ifdef in the template module list will make current user devices using different module ids then they used to do. So this needs a major overhaul which I have on my plate since a while but I still haven't found a useable solution.

I hate ifdefs as they mean different binaries which is against the one binary for all policy I once envisioned (but as we already have six different binaries I guess I have to let go that policy too).

Anyway. This won't be implemented yet as it breaks backward compatibility and it only provides good functionality for LC Tech users.

VerboteneZone:
@arendst, agree partially. I hate ifdefs too and the template stuff needs an overhauling, but I dont see the negative effects for the other devices (and users).

Anyway, at least it is your decision and I respect it.

For users which need the LC-Tech implementations: I keep that path up a while in my fork https://github.com/verbotenezone/Sonoff-Tasmota, till we find a better (template and overhead) solution. 

But @arendst I highly recommend to think about the suppress_serial_logging solution. I know you dont mention this as a problem, but a short circuit and a smoking device will change anyones mind.

Thanks and merry X-Mas, again.

wolfgangr:
Hope it is OK to continue here with testing of  @VerboteneZone 's fork.
can't create an issue there.
I merged it with curren development.
First impression was great, but relais did not switch (4CH device).
Changing to Generic device and sending serialsend5 commands DO work - without any resetting of the  relais controller.
LogLevel4 reveals  a trailing 0x0a as compared to the sendstrings.
`Jan 12 00:37:35 sonoff-7353 ESP-LCT: Sent new serial state to relay 0 (state: 0): 0xa0, 0x01, 0x00, 0xa1, 0x0a`
May this be the issue? 
Or is serialsend5 adding <CR> as well, anyway?
Tried to remove the trailing 0x0a, but did not help.

wolfgangr:
continue testing....
Trying it with unchanged fork from @VerboteneZone . same picture
removing the trailing `0x0a`from send buffer - same picture

However - adding additional 0x0a to serialsend does NOT hurt
``serialsend5' -m '0xa00400a40a0a'
``
So the trailing `0x0a` are not the culprit, obviously.

Tried to debug the serial output - no success yet:
Conntecting the naked ESP-01S to the serial adaptor, I can see patterns of scrap on the arduino serial moitor both with `serialsend` and with `xdrv_20_lc_tech_relay.ino` , however, they look different.
At lest I can say, that SOMETHING is sent.

Could not find yet a way to precisely debug the hex stream, at best with timing information. 
Can't switch arduino serial monitor to "hex" mode. 
PlatformIO appears to call minicom, which crashes. 
`cat /dev/ttyUSB0` finishes immediately.
Still playing with unbuffered serial reaing with perl - no success yet.
can't find 'hex serial monitor' or sth like that on google.

stty settings?
This is what I get after I ran arduino serial console on it:
```
...$ stty -F /dev/ttyUSB0 
speed 115200 baud; line = 0;
min = 0; time = 0;
-brkint -icrnl -imaxbel
-opost -onlcr
-isig -icanon -iexten -echo -echoe -echok -echoctl -echoke
```

Any Idea?

One thing that comes to mind is a timing issue.  
I remeber stumbling across this in one of the other issue threads around here, but can't find it any more. Since `xdrv_20_lc_tech_relay.ino` always updates all four relays, not just the one switched, commands may come to fast.
Delays are forbidden in a preemptive pseudo RTOS multitasking environment - I understand.

What about keeping relay state in a buffer and updating only one channel every 25 or 100 ms in the respective loops?

VerboteneZone:
Hi Wolfgang, I enabled the issue tracker and thanks for responses.

I did some tests, with the boards in my stock. You are partially right, I can acknowledge, that all 4 channels are switched on one of my boards. 


My theory is, that I got a four channel board, which is not really LC-Tech. They seem to be equal at frist glance, but on my testing device, there are some extra diodes and resistors. So it is possible, I bought a clone, which doesnt behave like the genuine ones.


Pls use issue tracker at verbotenezone, so its a better way to find a solution here.

wolfgangr:
OK, continued the issue there: https://github.com/VerboteneZone/Sonoff-Tasmota/issues/4
Did some rewrite of the code and got it working on the end https://github.com/wolfgangr/Sonoff-Tasmota/blob/LCT_noblock/sonoff/xdrv_20_lc_tech_relay.ino

heikoh81:
> 
> 
> @Jason2866 how to add this boards and the rules are already written in the wiki at https://github.com/arendst/Sonoff-Tasmota/wiki/LC-Technology-WiFi-Relay

Tried the wiki rule for 4ch Tasmota.
Does not work for relay 1.

Also, examples for Relays 2-4 are missing.
I'm not an experienced programmer, so I would need a step-by-step guide or a precompiled .bin to flash to my ESP01S.

andreas-bulling:
> @arendst, agree partially. I hate ifdefs too and the template stuff needs an overhauling, but I dont see the negative effects for the other devices (and users).
> 
> Anyway, at least it is your decision and I respect it.
> 
> For users which need the LC-Tech implementations: I keep that path up a while in my fork https://github.com/verbotenezone/Sonoff-Tasmota, till we find a better (template and overhead) solution.
> 
> But @arendst I highly recommend to think about the suppress_serial_logging solution. I know you dont mention this as a problem, but a short circuit and a smoking device will change anyones mind.
> 
> Thanks and merry X-Mas, again.

I am one of the new users of Tasmoto and I have one of these LC modules I believe (ordered from Aliexpress). I've read through diverse issues/wiki pages but it remains unclear to me what the recommended way of getting them to work is. The rules approach in the wiki?

heikoh81:
I got LCTech 4ch relay boarding working by switching to tasmota-fork called WESPOTA.
You can select 1ch, 2ch, 4ch LCTech from the dropdown list and it works like charm.

However, there are no releases yet on github, so I had to download the nightly from
https://wespota.verbotene.zone/nightly/

meingraham:
https://github.com/VerboteneZone/WESPOTA

AntonioCorrenti:
So guys, basically the solution to this issue is still to use Wespota by @VerboteneZone, right?

andrethomas:
For now yes...

Perhaps now that we have user defined templates as an option one of the dev's will add this if they have one of these boards... unfortunately I do not.

PR #4738, opened 2018-12-26
Add INPUT_PULLUP option for single DS18B20

phreakmonkey:
This feature allows users to connect a single DS18B20 directly to a GPIO pin without an external pull-up resistor.  (The internal pullups on the ESP8266 are strong enough for a single directly connected DS18B20.)

Reference:  https://wp.josh.com/2014/06/23/no-external-pull-up-needed-for-ds18b20-temp-sensor/


lalo-uy:
the reference is for a 5V arduino, doing the math for 3.2V  it leaves 100p for wire => ~2m for the sensor.
Nice

lalo-uy:
The reference is for a 5V AVR arduino . Esp is on 3.2V and 100k worst case pullup.
I do the math and there is room for a DS18B20 and 100pF of wire ~2m.
Nice.

phreakmonkey:
Yes, it's a tighter margin than in the referenced article, and I've only tested it with single "directly connected" DS18B20.  That's why I only put the change in the single DS18B20 code and not the broader multi-DS18x20 modules (where it's more likely to be problematic if the pull-up is too weak).  I have it running this way on three different boards though, works great.  ;)

PR #4739, opened 2018-12-26
Temperature offset directive for BME680

phreakmonkey:
The BME680 is known to return a temperature reading that is 3 - 4 C too high due to it's internal heater for gas measurement.  This directive allows compensation for that issue inside Tasmota instead of dealing with it after the fact.

I'm running with this mod on three different units (two indoors, one outdoor) and have found -4.0 C. to be fairly consistently accurate across all three.

Context: https://github.com/pimoroni/bme680-python/issues/11

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -293,6 +293,7 @@
   #define USE_HTU                                // Enable HTU21/SI7013/SI7020/SI7021 sensor (I2C address 0x40) (+1k5 code)
   #define USE_BMP                                // Enable BMP085/BMP180/BMP280/BME280 sensors (I2C addresses 0x76 and 0x77) (+4k4 code)
 //    #define USE_BME680                           // Enable support for BME680 sensor using Bosch BME680 library (+4k code)
+      #define BME680_OFFSET	-4.0                 // Temperature offset for BME680 (-4 seems to be average)
   #define USE_BH1750                             // Enable BH1750 sensor (I2C address 0x23 or 0x5C) (+0k5 code)
 //  #define USE_VEML6070                           // Enable VEML6070 sensor (I2C addresses 0x38 and 0x39) (+1k5 code)
     #define USE_VEML6070_RSET    270000          // VEML6070, Rset in Ohm used on PCB board, default 270K = 270000ohm, range for this sensor: 220K ... 1Meg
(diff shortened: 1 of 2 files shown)


andrethomas:
Would this be a solid 4 deg C in all scenarios or would it depend on the actual ambient temperature (i.e. transposed as a % of the temperature measurement to get a calibrated value)

Jason2866:
Do your PCBs with the BME680 does have a Vreg on it?
If yes that is your issue not the BME680!
The Vreg does heat the PCB. So the board design is to blame....
A example of bad design:
image

phreakmonkey:
> Do your PCBs with the BME680 does have a Vreg on it?
> If yes that is your issue not the BME680!
> The Vreg does heat the PCB. So the board design is to blame....
> A example of bad design:
> image

Entirely possibly.  I was just going on the consensus I saw online, which matched my experience.  So, my description of why I need the offset may be wrong, but that doesn't obviate the need.

ascillato2:
As a workaround for adjusting sensors you can use rules as explained in the following example of the wiki: https://github.com/arendst/Sonoff-Tasmota/wiki/Rules#8b-adjust-the-value-of-a-sensor-and-send-it-by-mqtt

ascillato:
Hi,

All this sensors has that offset?

I think that should be better that the user measure it and adapts the measurement to his/her own case.
I don't think that a general -4 degrees for all batches is correct.

There is the example posted before to adapt any sensor to an user-defined "calibration"

ascillato2:
Hi @phreakmonkey 

Thanks a lot for sharing your ideas. But this PR is not going to be merged. Sorry.

From the datasheet, the BME680 has an internal compensation:

image
_(thanks to @andrethomas for finding the table)_

It is reported that there is a problem of self-heating from the internal voltage regulator, making the device to need, at least, a linear correction that should be experimentally adjusted for each case, not being a scalar offset value for the exact correction. We don't know if all batches of these sensors has that issue and in if the correction would be the same for all.

Also there is already an example using rules to make a user-defined "calibration".

Thanks again for sharing your ideas.

phreakmonkey:
Ah makes sense. I didn't know about the rule capability to modify sensor
output, thanks!

Cheers!

K.C.

Issue #4727, opened 2018-12-24
Moving average for Power Measurement


rasti70:
**Have you look for this feature in other issues and in the wiki?**
=> YES

**Is your feature request related to a problem? Please describe.**

=> I have a few power measurement devices (Blitzwolf/Sonoff/Gosund) and noticed that some of them show really big fluctuations, e.g. I measured a 15W bulb and get measurements values of 13-19 Watts. This happens in particular with Gosund 111 / Blitzwolf SHP6 . On the contrary, Sonoff Pow, Gosund1, Gosund SHP2 seem to be pretty stable (fluctuations of +/- 1 Watt only)

**Describe the solution you'd like**
_A clear and concise description of what you want to happen._

=> I would like to have a moving average filter. The current sample rate or update rate of the values shown in the webinterface seems to be around 1-2 Hz. I would suggest a moving average filter which calculates the average of e.g. 10 measurements before displaying or publishing it. Obviously the refresh rate of the data shown in the webinterface will be lower, but getting an update e.g. every 10 or 20 seconds would be acceptable for me.  This filter could be adjustable through the console, e.g. "movingaverage 0" or "movingaverage 50" .  Maybe an IIR filter also will do the job.

**Describe alternatives you've considered**

=> I tried to implement a moving average filter in home automation software FHEM.
This works, however I had to set the teleperiod to a very low value (10s) which imposes 
a lot of traffic on my mqtt/fhem server 

**Additional context**

=> I posted this on fhem forum, too : 
https://forum.fhem.de/index.php/topic,90220.msg875650.html#msg875650
In the subsequent posts and elsewhere on the same forum I found at least 3 users more which had the same problem I had.

**(Please, remember to close the issue when the problem has been addressed)**

ascillato:
Hi,

Your idea of doing some type of data processing for the power measurement is nice, but there are some points to be taken into account:

If the 15W bulb you are measuring is a cheap one, it will have a lot of noise that it is being injected in your mains AC. Also if you have other devices of that type, you will be having much more noise. There are some of these cheap ones that have a noise switching power supply and I saw others that have a capacitive power supply that produces a huge amount of noise that it is just an additive noise, not a random noise.

So, the moving average filter won't work in an additive noise environment. If the noise is just white noise type or pink noise type, a moving average filter will do the job. In an environment where you have an asymmetrical noise, a moving average filter will give you all the time a higher value than the real one.

[...]

PR #4741, opened 2018-12-27
Add SensorFilter Library

andrethomas:
Add SensorFilter library for Median and Average use.

A use case for Median is demonstrated by the change made in xsns_26_lm75ad.ino

**Usage for SensorFilter_Median**
``` 
 SensorFilter_Median MyMedian(x,y);  // where x = number of elements e.g. 9 and y the index which will be used e.g. 5
 newvalue = MyMedian.NewMedian(z);   // where z is the new value to be added to the sorted array to determine the median - new median is returned by the function
```
**Usage for SensorFilter_Average**
```
 SensorFilter_Average myAverage(x);  // where x = number of elements over which average will be calculated
 newvalue = myAverage.NewAverage(y); // where y is the new number to be added to the array on which average will be based - new average is returned by the function
```


rasti70:
Thank you

Can you please show an example how to implement this routine for power measurement ?
I assume it must go somewhere in sonoff\xdrv_03_energy.ino but I have no clue how.

andrethomas:
I'm guessing **_if_** Theo implements it will most likely be there, or perhaps device specific (e.g. xnrg_01_hlw8012) but I honestly have no clue... you'd have to wait for him to decide if, where and how he wants it :)

arendst:
Pls hold your horses...

the way the HLW8012 driver works is that if enough power is used, like say >10W, it takes ten probes and provides the average to the user by simply adding all measurements and divide by ten. This way no additional RAM is used to store intermediate results. Works great on all but SHP6.

For the SHP6 this works great for Voltage; It's very stable. For both power and current it does not provide stable results.

Power is measured constantly from a dedicated power pin which provides a PWM pulse where it's pulse width is a measure of power. For some reason this single pin provides unstable data on the SHP6.

Current is measured alternated from another single pin with voltage. The code switches between current and voltage measurement every 10 probes taking into account the time it takes the chip to provide "stable" info. This works fine for all but SHP6. Apparently the SHP6 is not stable once switches to current measurement. It needs further analysis of samples when it can provide stable Current measurements.

Considering the tiny SHP housing and the fact that all is connected to AC this is a nice task for rubberman.

So to go forward 
- the SHP6 current measurement will need further analysis before deciding the signal is stable. This is a better approach then starting to uses average or median right now.
- the SHP6 power measurement is a different ballgame and having some kind of probe scanner could make life easier.

andrethomas:
Does rubberman still have his rubber underpants? :)

Wish I had an SHP to debug it from a hardware perspective.

andrethomas:
Mother of cow... I only now see what the thing looks like.

image

arendst:
Right! If we can get it stable it's a great small device.

Let's continue here #4727

PR #4749, opened 2018-12-27
Add variable %timestamp% to be used in rules

ascillato:
This PR adds a new variable `%timestamp%` to be used in a rule to allow the user to include the Time Stamp as Tasmota has for Status, Sensors, etc, for example:

Command:

`publish stat/topic/sensor {"Time":"%timestamp%","mysensor":"%var1%"}`

Output:

`{"Time":"2018-12-27T12:52:57","mysensor":"1"}`

(https://github.com/arendst/Sonoff-Tasmota/issues/4734)

### DIFF
--- a/sonoff/xdrv_10_rules.ino
+++ b/sonoff/xdrv_10_rules.ino
@@ -167,6 +167,10 @@ bool RulesRuleMatch(byte rule_set, String &event, String &rule)
     if (rule_param.startsWith(stemp)) {
       rule_param = String(GetMinutesUptime());
     }
+    snprintf_P(stemp, sizeof(stemp), PSTR("%%TIMESTAMP%%"));
+    if (rule_param.startsWith(stemp)) {
+      rule_param = GetDateAndTime(DT_LOCAL).c_str();
+    }
 #if defined(USE_TIMERS) && defined(USE_SUNRISE)
     snprintf_P(stemp, sizeof(stemp), PSTR("%%SUNRISE%%"));
     if (rule_param.startsWith(stemp)) {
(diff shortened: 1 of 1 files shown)


andrethomas2 (approved):

scotty264:
Hello:
Does the addition of the %timestamp% variable only apply to the SONOFF version? Is it not available in the default build? 

I'm running "13.4.0(tasmota)" and issuing the command "publish stat/DIS/sensor {"Time":"%timestamp%"}" from the TASMOTA console on my device results in the message "18:05:38.997 MQT: stat/DIS/sensor = {"Time":"123mestamp"}" being published.

In my application, I am running a drip-irrigation system running on a timer with a rule that triggers POWER1 relay for 90 seconds then turns it off using a RULESTIMER. The only information published is "{"POWER":"ON (or) OFF"}" with no time/date stamp. For testing purposes, I wish to capture the time/date that POWER1 is turned on and when it is turned off. Is there a means to do this? 

Thanks.

Peace and blessings,
JQ

sfromis:
Rule variables only works from rules, not when issuing commands by other means. As the thread title says "to be used in rules".

It should be simple for your client to "know" what time it is when receiving a message via MQTT. Still, your rules could do a publish when acting.

scotty264:
Hi:
Thanks for responding. I was not comprehending that the %timestamp% would only work in rules. I used the above example in a rule and it works properly.

I also figured out that I can apply PulseTime to the relay when triggered by a timer. Using a rule, I can also use the Time command to publish the date/time when the relay turns on then when it turns off, which suits my purposes.

Thanks again.

Peace and blessings,
JQ

PR #4764, opened 2018-12-30
MAX31855 thermocouple support

ipm79:
Support for the MAX31855 thermocouple chip. Currently only (embedded) SW SPI is supported (I will add HW SPI support later on).
This is a quite simple implementation that currently only supports one MAX31855 chip. I will add support for multiple chips as soon as I get mine delivered. Support for MAX6675 is also planned (unfortunately I don't have one but they are already on the way from China).
The driver has no dependency on external libraries.


andrethomas:
@arendst Which standard are we following for additional SPI drivers - We keeping to the already defined D_SENSOR_SSPI_MISO, D_SENSOR_SSPI_MOSI, D_SENSOR_SSPI_SCLK, D_SENSOR_SSPI_CS, D_SENSOR_SSPI_DC or are we going to add new pin dropdowns for each sensor/driver we add?

I like the idea of being able to select pins for a specific device because it makes false detection of a different device impossible - how does this weigh up against ram considerations?

The major challenge with SPI devices is that they generally require their own CS/SS pin so the current model of having a fixed set of pin defines for any particular device will make it really hard to know exactly which device we are actually receiving responses from.

Your thoughts on this please?

Thanks

andrethomas:
PS - I think using the the pre-defined SPI/SSPI MOSI, MISO, CLK would be fine but we need to find a way to handle the CS/SS and DC (Data Control - Read or Write) pins to be definable as a bare minimum.

arendst:
Me too (;-)) have been thinking about the soft SPI interface and what to keep common and what not.

For now I'll merge this and keep an eye on future soft SPI implementations.

AlbertoLopSie:
Hi!

Looking to interface a MAX31865 (RTD to Digital Converter, PT100/PT1000) to a Sonoff/Tasmota.

I haven't looked to the datasheets yet... Do you think it could be compatible with 31855 support code?

Thanks!!

meingraham:
Yes, please refer to the datasheets. If you have further questions, please address this to the [Tasmota Support Discord Chat](https://discord.gg/Ks2Kzd4) first to discuss and test. The chat is a better and more dynamic channel for helping you. Github issues are best used for Tasmota **_software feature requests and bug reporting_**. Troubleshooting and setup assistance is more effective using an interactive forum.

Please check the [Contributing Guideline and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md) and the [Support Guide](https://github.com/arendst/Sonoff-Tasmota/blob/development/SUPPORT.md).

Thanks.

### Support Information
See [Wiki](https://github.com/arendst/Sonoff-Tasmota/wiki) for more information.
See [Chat](https://discord.gg/Ks2Kzd4) for more user experience.
See [Community](https://groups.google.com/d/forum/sonoffusers) for forum.
See [Code of Conduct](https://github.com/arendst/Sonoff-Tasmota/blob/development/CODE_OF_CONDUCT.md)

AlbertoLopSie:
Thanks meingraham!  And I'm sorry if I missplaced my question. At first sight I supossed a single digit in the chip could mean an easy (or none) change. After reading both datasheets I realize it is non trivial. MAX31855 and 865 are two very different animals.

Support for the MAX31865 could be surely interpreted as a feature request, but perhaps on its's own thread.

I'll check on the support chat anyway.

Thanks!

une18:
Hi! What is the current status? MAX31865 is in the supported component list.
Tasmota.bin (7.1.2.4) does not contain max31865 sensor, should I try other bin file instead?

andrethomas:
It is supported but not included by default in the release binaries. You will need to enable its compiler directive as currently commented out in `my_user_config.h` or add it to `user_config_override.h` and then compile your own binary from source.

une18:
Took some time for me to test it. Found the right components in tasmota-sensors.bin. The only problem is, the sensor is not showing right temperature. My current room temperature is 22 degrees celsius, but it gives me 8 degrees celsius. Is there a way to calibrate it, like HX711 can be calibrated?

12:49:44 MQT: tele/31855/SENSOR = {"Time":"2020-03-10T12:49:44","MAX31855":{"ProbeTemperature":8.0,"ReferenceTemperature":96.0,"Error":0},"TempUnit":"C"}

Can I change the reference temperature somehow, or....?

andrethomas:
From the datasheet:

image

Are you sure that the chip type is correctly matched with the type of thermocouple you are using?

Other than that the only other thing I can think of is to do post-processing on your automation software side to adjust for accuracy.

aik:
Trying MAX6675 with Sonoff TH, the readings are ~+22C higher than expected. Is there a way to adjust it somehow, or a custom build is the answer?

PR #4819, opened 2019-01-04
New Triggers for Rules: VARx#STATE and MEMx#STATE

ascillato:
Added new triggers for rules: `VARx#STATE` and `MEMx#STATE`

These new triggers allows to execute commands when **VARs** or **MEMs** change their values.

This is useful for automations where the parameter sent to Tasmota by MQTT is a _SetPoint_ for example or if after the usage of ADD, SUB, MULT or SCALE, it is required to execute commands.

Before these triggers, the approach was using an `EVENT` to store a value but some components in home automation softwares can't use `cmnd/sonofftopic/EVENT SETVAR=25`, where `cmnd/sonofftopic/EVENT` is the mqtt message and `SETVAR=25` is the payload.

With this new triggers, it will be as simple as `cmnd/sonofftopic/VAR 25` so any automation software can use it.

_(Using `cmnd/sonofftopic/EVENT SETVAR=` as mqtt message and `25` as payload is not understood by Tasmota)_

Usage in Tasmota:

```
rule1 1
rule1 on VAR1#STATE do [commands] endon
```

or 

```
rule1 1
rule1 on VAR1#STATE>27 do [commands] endon
```

### DIFF
--- a/sonoff/xdrv_10_rules.ino
+++ b/sonoff/xdrv_10_rules.ino
@@ -95,6 +95,14 @@ uint8_t rules_teleperiod = 0;
 
 char event_data[100];
 char vars[MAX_RULE_VARS][33] = { 0 };
+#if (MAX_RULE_VARS>8)
+#error MAX_RULE_VARS is bigger than 8
+#endif
+#if (MAX_RULE_MEMS>8)
+#error MAX_RULE_MEMS is bigger than 8
+#endif
+uint8_t vars_event = 0;
+uint8_t mems_event = 0;
 
 /*******************************************************************************************/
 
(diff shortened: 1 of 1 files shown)


benzino77:
Adrian is it enough to change `xdrv_10_rules.ino` the way below to have 16 vars/mems and still got the possibility to use proposed triggers? It compiles well and new triggers works as expected but maybe there is something else?
```CPP
diff --git a/sonoff/xdrv_10_rules.ino b/sonoff/xdrv_10_rules.ino
index df2db974..b6ca637a 100644
--- a/sonoff/xdrv_10_rules.ino
+++ b/sonoff/xdrv_10_rules.ino
@@ -95,14 +95,14 @@ uint8_t rules_teleperiod = 0;
 
 char event_data[100];
 char vars[MAX_RULE_VARS][33] = { 0 };
-#if (MAX_RULE_VARS>8)
-#error MAX_RULE_VARS is bigger than 8
+#if (MAX_RULE_VARS>16)
+#error MAX_RULE_VARS is bigger than 16
 #endif
-#if (MAX_RULE_MEMS>8)
-#error MAX_RULE_MEMS is bigger than 8
+#if (MAX_RULE_MEMS>16)
+#error MAX_RULE_MEMS is bigger than 16
 #endif
-uint8_t vars_event = 0;
-uint8_t mems_event = 0;
+uint16_t vars_event = 0;
+uint16_t mems_event = 0;
```

If this is a proper solution, is there a chance to change your limits to allow 16 vars/mem in Tasmota (I realize that this consumes 4 instead of 2 bytes of memory)?

PR #4845, opened 2019-01-06
fix empty prefix issue

engrbm87:
When switchtopic is set to a different topic the prefix will be empty so its length will be zero. In the Shorten function the strncmp will give a zero result because the prefixlen is zero and will add "\~" even though there is no prefix.
Adding prefixlen != 0 in the if condition fixes the issue. Also skip adding the prefix "\~" on line 292 by checking if the prefix length is greater than zero.


emontnemery:
This is fixed in home-assistant/home-assistant#19501, which is included in Hass 0.85 so this workaround can be removed some time from now.

PR #4868, opened 2019-01-09
Added N0DY_RELAY module

n0dyjeff:
Create new module to support "N0DY Wifi Relay" device, with two relays.

Device can be found at:
https://www.n0dy.com/product/web-controlled-dual-relay/
https://www.amazon.com/dp/B072MKV8ZM

Hardware is open-source on github: n0dyjeff/WiFiRelay

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -1705,9 +1705,16 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
      GPIO_USER,        // GPIO03 Serial TXD or Optional sensor on J2 TXD (if not using serial io)
      GPIO_REL2_INV,    // GPIO04 Relay 2 (active low)
      GPIO_REL1_INV,    // GPIO05 Relay 1 (active low)
-     0, 0, 0, 0, 0, 0, // Flash connection
-     0, 0, 0, 0, 0,    // Unused GPIO 12-16
-     0                 // ADC0 Analog input
+                       // GPIO06 (SD_CLK   Flash)
+                       // GPIO07 (SD_DATA0 Flash QIO/DIO/DOUT)
+                       // GPIO08 (SD_DATA1 Flash QIO/DIO/DOUT)
+     0,                // GPIO09 (SD_DATA2 Flash QIO or ESP8285)
+     0,                // GPIO10 (SD_DATA3 Flash QIO or ESP8285)
+                       // GPIO11 (SD_CMD   Flash)
+     0,                // GPIO12
+     0,                // GPIO13 Button
+     0,                // GPIO14 IR Transmitter
+     0, 0, 0
   }
 };
 
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -275,6 +275,7 @@ enum SupportedModules {
   MANZOKU_EU_4,
   OBI2,
   YTF_IR_BRIDGE,
+  N0DY_RELAY,
   MAXMODULE };
 
 /********************************************************************************************/
@@ -538,6 +539,7 @@ const uint8_t kModuleNiceList[MAXMODULE] PROGMEM = {
   WION,
   SHELLY1,
   SHELLY2,
+  N0DY_RELAY,
   BLITZWOLF_BWSHP,    // Socket Relay Devices with Energy Monitoring
   TECKIN,
   TECKIN_US,
@@ -1693,6 +1695,19 @@ const mytmplt kModules[MAXMODULE] PROGMEM = {
      GPIO_KEY1,        // GPIO13 Button
      GPIO_IRSEND,      // GPIO14 IR Transmitter
      0, 0, 0
+  },  
+  { "N0DY Relay",      // N0DY Wifi Dual Relay (ESP-07)
+                       // https://www.n0dy.com/product/web-controlled-dual-relay/
+                       // https://www.amazon.com/dp/B072MKV8ZM
+     GPIO_KEY1,        // GPIO00 PROG Button
+     GPIO_USER,        // GPIO01 Serial RXD or Optional sensor on J2 RXD (if not using serial io)
+     0,                // GPIO02 
+     GPIO_USER,        // GPIO03 Serial TXD or Optional sensor on J2 TXD (if not using serial io)
+     GPIO_REL2_INV,    // GPIO04 Relay 2 (active low)
+     GPIO_REL1_INV,    // GPIO05 Relay 1 (active low)
+     0, 0, 0, 0, 0, 0, // Flash connection
+     0, 0, 0, 0, 0,    // Unused GPIO 12-16
+     0                 // ADC0 Analog input
   }
 };
 


arendst:
Just wondering how much percentage is donated to Tasmota.

ascillato:
Hi, this can be configured using generic module. No need to add another entry to the already super long list.

@n0dyjeff

Please add this configuration in the wiki under other devices. Thanks

andrethomas:
That thing looks dangerous though... maybe a photo of the bottom of the pc board will settle me a little but the top of the board scares the sht out of me :)

n0dyjeff:
@andrethomas 

I'm curious as to the detail of your concerns. High voltages (anything greater than 20V) are well separated from other traces and the traces themselves are of minimal length. Unlike a sonoff (which passes 120V thru to the outputs, in this design the relay outputs are isolated switch contacts only. 120V only goes about an inch to the power converter module, where it's dropped down to 5V to energize the relays.

Can you please share more details about what your concerns are? I've attached jpgs of the front and back side layouts.

Thanks!
Jeff

front
back

n0dyjeff:
@arendst 
Hi Theo,

Much to your credit, the Paypal links are not prominent on the README. In fact, so subtle that I didn't notice them until I read your comment.

The N0DY Wifi Relay is pretty much a hobby project. I've sold a total of 11 of them in the last 18 months. Haven't come close to recouping my costs for the initial purchase of parts for 50 units. Nevertheless, I acknowledge the benefit I receive from your work and have just donated about 10% of sales (USD $20) via the Paypal link.

Regards,
Jeff

andrethomas:
@n0dyjeff 

Constructive feedback... I'll be soft becaue I don't have a physical device to criticize :)

1. Good to see you're at least not feeding the outputs from the input side from an engineering perspective but it does present the need for extra wiring to feed the relays on the other side of the board by wire I would guess - The first part is what I was most concerned about but at least that is not happening - you do, although, have the risk of a user inadvertently flying an AC cable right over the PCB which may cause more interference than what it would have been if just routed and air gapped correctly but I understand that you want to leave the option open for user to switch what ever source load they choose so good job on that one... **phew**

2. The copper traces on the relay switch side of are not going to survive a 10A load if my guesstimate by their width is accurate... I don't think they will survive 10A at 12V without inadvertent heating or instantanious vaporisation, but since I don't have a physical device to see for myself it would be hard to say what the actual threshold would be.

3. The proximity of the flyback diode (D1) and Q1 (switching transistor) to the ESP8266's antenna is not ideal. One would have to bench it but I'm pretty sure there will be some interference on the wifi side of things that would limit the range of the ESP8266 device significantly, or at best cause dropped wifi frames when switching occurs.

4. You run the 5V supply feed to the LD1117-3.3 directly under the AC to DC module (which is switch mode) but given the thickness, length and direction of the track you probably need a little more decoupling than just the C1 you have in place there. Keep in mind that the ESP8286 can be very unpredictable when it comes to power demand as it spikes up and down when using wifi.

5. The external antenna is conditional insofar that the on board chip antenna needs to be disconnected by removing the surface mount decoupling capacitor next to the uFL connector - unless of course you want to use both antenna's at the same time which has its own unique "features" that comes with it... not sure if it would be considered useful by a user who cannot remove a surface mount capacitor.

6. Good job on open sourcing the hardware though - its just unfortunate that some important input could not be added during the initial design process - I could use this to switch things like routers running on 12V 2.5A but I most certainly will not use it to switch a 10A load.

I think other than the above its just concerning that the photos of functionality provided on the two weblinks you gave is showing Sonoff-Tasmota firmware being used but I do not see any clear indication of  this and the "Sonoff-Tasmota x.x.x.x by Theo Arends" has been removed/omitted from all the images so it kind of brings into question the ethics behind using open source software for projects with potential commercial gain - again not directed at you personally... just highlighting the fact that it does not go down with open source developers to not be credited for their work.

I understand your statement of not having sold a lot of them but I think if you want to use the firmware then it should be done with due reference and credit to whichever open source firmware you use on your product.

*These are of course my own views and may not be representative of the collective view of Sonoff-Tasmota's owner and other community contributors be it through donations, coding, providing support here on Github or through the Discord Chat system, and lets not forget all the marketing being done by members of the public through blogs, facebook posts and of course a very long list of Youtube video's*

n0dyjeff:
@andrethomas 
Hi Andre,

Thanks for the constructive comments - and for being gentle :)

In response to some of your comments:
_2. The copper traces on the relay switch side of are not going to survive a 10A load..._
Ugh. You're correct. The traces are good for only 2.5A at best. If I do a V1.2 layout, that will get fixed. In the meantime, I'll update the documentation / product description to accurately reflect a 2.5A capacity.

_3. The proximity of the flyback diode (D1) and Q1 (switching transistor) to the ESP8266's antenna is not ideal._
Another good point that would have been easy to address during layout. I'm thinking it's not a huge issue since switching would be relatively rare (at least relative to wifi packets). But an oversight and good catch. Thanks.

_4.  you probably need a little more decoupling than just the C1 you have in place there_
I was thinking that the 100 uF C2 on the output side of the LM-1117 would cope with the ESP's spiky power demand, and that the 10 uF C1 on the input was sufficient. No?

_5. The external antenna is conditional insofar that the on board chip antenna needs to be disconnected by removing the surface mount decoupling capacitor next to the uFL connector - unless of course you want to use both antenna's at the same time._
Yes, that was an intentional design decision to allow use of either. I had originally designed with an ESP-07S, but didn't want to force the user to come up with an external antenna if they didn't want it. The diversity antennas haven't shown themselves to be an issue in my testing.

_Sonoff-Tasmota firmware being used but I do not see any clear indication of this and the "Sonoff-Tasmota x.x.x.x by Theo Arends" has been removed/omitted from all the images_
You might not have noticed it, but also on github under "n0dyjeff" is WifiRelay-Tasmota, which gives full credit. You're correct in noting that the WiFiRelay repository doesn't mention it. I guess my thought was to have one repository for HW and another for SW. 

As far as removing the copyright, I don't think I did that. Those images are at least 18 month old, if not older. I don't think there was the same copyright notice that there is today. If there had been, knowing me, I'm too lazy to remove it anyway. At the very least, there was no intent to hide credit due to the originator. Now that I've finally gotten around to merging the latest code, it's time to update the product images as well. When I do, you can be assured that the copyright statements will be there.

Thanks again for your thoughtful remarks and insights.

Jeff

andrethomas:
@n0dyjeff 

> I was thinking that the 100 uF C2 on the output side of the LM-1117 would cope with the ESP's spiky power demand, and that the 10 uF C1 on the input was sufficient. No?

You need the ESP8266 to function yes, but in order to achieve that you must also have the LM1117 function as close to perfect as possible.

Implementing 5V to 3.3V regulation has two objectives in your case.

1) Receive a stable 5V supply rail which is free from fluctuations and free from noise. The 10uF is good to mitigate low-frequency ripple current but I would decouple it with a 100nF ceramic capacitor also just to make sure that any high-frequency noise is contained. In addition to this, I would use wider tracks to deliver the up to 250mA peaks to the LM1117 since it is a linear regulator and any noise introduced would effectively carry through to the output stage because the ground pin is used to reference comparison between input and output voltage.

2) Provide a stable 3.3V supply to the ESP8266 - The 100uF is more than sufficient to handle the impulsive current demands from the ESP8266 but it is important to remember that the LM1117 uses the voltage it reads on the output pin against the ground pin to correctly bias the internals in an attempt to maintain a stable 3.3V regulated output voltage. Again I would add a 100nF close to the LM1117 and another 100nF as close as possible to where VCC is provided to the ESP8266 - And of course the same would apply for that section of the power supply - it needs to be decent sized traces to be less susceptible to voltage drops caused by the irregular current demand during wifi usage.

You will see a lot of circuits put 0.1uF capacitors close to power pins of IC's - This is to decouple high frequency transcients from the component itself. Most people think its just useless components and on breadboard they get away with it because all the tracks are parallel to each other and a breadboard does have capacitive properties of its own.

Many people learn this when building a prototype on breadboard it works and when they make their custom PCB weird things happen.

If you do a new board I would also recommend rotating the ESP-07 90 degrees anti-clockwise so that the antenna is close to the end of the PC board which will improve wifi reliability by bringing the antenna to the edge of the board taking it as far away from interference as possible - AND this way the small capacitor next to the uFL connector can quite easily be removed by a newbie with a large Walmart soldering iron.

> Yes, that was an intentional design decision to allow use of either. I had originally designed with an ESP-07S, but didn't want to force the user to come up with an external antenna if they didn't want it. The diversity antennas haven't shown themselves to be an issue in my testing.

If you look at the signal strenth reported by your router and that which is reported by Tasmota and compare with and without the capacitor you will see that you gain a fair amount of dBm by implementing it correctly.

Anyway, we're way off topic here so I will leave you with that but I'm on Discord Chat from time to time if you want more information.

PR #4884, opened 2019-01-10
Fix GPIO16 Pull-down

netpok:
Currently the pull-up settings are not applied on GPIO16, this pull request fixes this behavior and allows the internal pull-down to be disabled.

### DIFF
--- a/sonoff/support_button.ino
+++ b/sonoff/support_button.ino
@@ -54,7 +54,7 @@ void ButtonInit(void)
   for (byte i = 0; i < MAX_KEYS; i++) {
     if (pin[GPIO_KEY1 +i] < 99) {
       buttons_found++;
-      pinMode(pin[GPIO_KEY1 +i], (16 == pin[GPIO_KEY1 +i]) ? INPUT_PULLDOWN_16 : bitRead(key_no_pullup, i) ? INPUT : INPUT_PULLUP);
+      pinMode(pin[GPIO_KEY1 +i], bitRead(key_no_pullup, i) ? INPUT : ((16 == pin[GPIO_KEY1 +i]) ? INPUT_PULLDOWN_16 : INPUT_PULLUP));
     }
   }
 }
--- a/sonoff/support_switch.ino
+++ b/sonoff/support_switch.ino
@@ -121,7 +121,7 @@ void SwitchInit(void)
     lastwallswitch[i] = 1;  // Init global to virtual switch state;
     if (pin[GPIO_SWT1 +i] < 99) {
       switches_found++;
-      pinMode(pin[GPIO_SWT1 +i], (16 == pin[GPIO_SWT1 +i]) ? INPUT_PULLDOWN_16 : bitRead(switch_no_pullup, i) ? INPUT : INPUT_PULLUP);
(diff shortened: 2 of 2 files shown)


andrethomas:
Just a side note - This pin has a special function which is why it has internal pull-down... it is used to wake up from deep sleep by the internal timer interrupt pulling the pin high so that the ESP8266 resets and starts running.

Yes, I know this is not used in Tasmota but the challenge is that in a lot of cases (ESP8266-12E, for example) the GPIO16 pin is actually hard-wired to the RESET pin.

Caution is therefore advised when using this as an input pin because it can have adverse implications on some ESP8266 based modules.

If used for a push button that pulls GPIO16 down on a module that has GPIO16 hard-wired to the RESET pin this would cause a hard reset on the ESP8266 when the button is released.

andrethomas:
Try this from the command line console

`GPIO16 1`

It will cause an immediate reset with the restart reason being "Software/System restart" if connected to the reset pin (as seems to be the case on the ESP-07 also which I just tested it with)

netpok:
`GPIO16 1` will cause software reset, because it sets the function of gpio 16 and then restart the device to reload the config, executing `GPIO14 1` will also cause a system restart.

I made this pull request because I have a Digoo SP202 smart socket, which uses GPIO16 as a button via an external 10k resistor, leaving the internal pull-down active causes ghost switching, because the pull-down and the pull-up resistor forms a voltage divider with a ~2.5V output.

Jason2866:
It is a very bad idea to change it in general for all devices. Many issues will occur.

netpok:
It's not changing it in general.

I think this was a bug:
Old behavior:

- Normal gpio configured as `KEY1`: Pull-up enabled
- Normal gpio configured as `KEY1_NP`: Pull-up disabled
- gpio16 configured as `KEY1`: Pull-down enabled
- gpio16 configured as `KEY1_NP`: Pull-down **enabled**

New, fixed behavior:

- Normal gpio configured as `KEY1`: Pull-up enabled
- Normal gpio configured as `KEY1_NP`: Pull-up disabled
- gpio16 configured as `KEY1`: Pull-down enabled
- gpio16 configured as `KEY1_NP`: Pull-down **disabled**

As `_NP` stands for no pull, I think this is the correct way to handle buttons, also in the template there are no switch or button defined on gpio16 with no pull, so it should not change anything for currently supported devices.

andrethomas:
From the datasheet

image

GPIO pin will be triggered as HIGH with a voltage of > 0.75v - 0.80v

You would need a voltage lower than 0.75 volts to cause ghost switching.

netpok:
Maybe I measured it wrong, but disabling the internal pull down fixed the problem. Also based on my previous comment I think this is a general bug, since `_NP` version off switches and buttons are used only if someone wants to disable the internal pull resistors.

andrethomas:
@netpok I guess if it solves it the measurement was maybe on the wrong end of the 3.3V - i.e. 3.3V - the 2.5 you measured = 1.3V - maybe it dips below that at some point due to instabilities inside the Digoo SP202 to match a low condition on the GPIO pin. 

So all the information is here now... now it's up to @arendst whether he wants to merge or not.

If we start getting users with random boot loops we'll know why :)

andrethomas2 (approved):
Approving based on the code being fine - Cannot guarantee any implications but we can observe the current development codebase binary usage and see if other users report unpredictable behaviour based on this PR.

PR #4887, opened 2019-01-11
Add support for Mi Desk Lamp

dh-harald:
Hi,

This is my initial commit for Mi Desk Lamp... It's WIP, no color temperature switch yet...
I'd like to add it for review, because I'm pretty sure, that you have coding standards, recommendations, etc, that I need to modify...

Status:
- [x] Add Mi Desk Lamp Module
- [x] Add support for rotary switch
- [x] Control brightness with rotary switch
- [x] Control color temperature with button press & rotary switch

Any reviews or suggestions are welcome

### DIFF
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,8 +1,4 @@
-/* 6.4.1.9 20190111
- * Add support for rotary switch
- * Add support for Mi Desk Lamp
- *
- * 6.4.1.8 20190107
+/* 6.4.1.8 20190107
  * Change sonoff_template.h layout regarding optional module flags like ADC0
  * Add command SetOption62 1 to force no Button/Switch pullup on dedicated modules. Currently only supported on Shelly2 (#4841)
  * Fix Display exception 28 when JSON value is NULL received
@@ -15,7 +11,9 @@
  * Add support for Luminea ZX2820 Smart Socket with Energy monitoring (#4921)
  * Add define MDNS_ENABLE to control initial mDNS state (#4923)
  * Add split interlock part 1 (#4910)
- *
+ * Add support for rotary switch
+ * Add support for Mi Desk Lamp
+ * 
  * 6.4.1.7 20190106
  * Fix HLW8012, HJL01 and BL0937 based energy sensors low Power (below 10W) measurement regression from 6.4.1.6
  * Add Power status functionality to LED2 when configured leaving LED1 for Link status indication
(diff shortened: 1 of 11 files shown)


andrethomas2 (changes requested):
The `uint16_t rotary_debounce;` could be `uint8_t rotary_debounce;` as it is unlikely that you would need a debounce of more than 255ms on a rotary encoder.

Also, @arendst will need to decide if the rotary support should be in a separate driver file or not so that it may be conditionally included or excluded for wider use cases where users just want to use a rotary encoder with any other project.

Jason2866:
Would be cool to see a separate general rotary driver for Tasmota

andrethomas:
Is there any particular reason why "Control color temperature with button press & rotary switch" is not included in your initial PR ?

dh-harald:
@andrethomas I've used `uint16_t` because the other debounce variables (button_debounce, pulse_counter_debounce, switch_debounce) are the same type.

dh-harald:
@andrethomas Because I'm not a developer, I'd like to start the review process as soon as possible to taylor any coding style problems... 
Abouth the "Control color temperature with button press & rotary switch" feature, I'm thinking about, how can I handle that... I need to figure out, how can I enable a "multipress" thing with the rotary switch, then hopefully, it won't turn off the LEDs, when the color temperature setting has been finished.

Jason2866:
Have you looked here?
https://github.com/xoseperez/espurna/blob/dev/code/espurna/encoder.ino

arendst:
In the future we might have to move the device specific code elsewhere but for now it can stay where it is.

To make things simpler I would like to get rid of the special rotary debounce code and use the key debounce instead. I do not expect deviations between the two.

Have a separate handler for the rotary switch is also a good idea for now as it allows changes to be kept to this file only.

For double press you might find some hints in the button.ino file.

arendst (changes requested):
See my last comment

dh-harald:
@arendst Did you think about something like my last commit?

arendst:
Yes.

andrethomas2 (approved):
Happy for the moment, so just need to see how @arendst wants to proceed.

I think the initial state of the rotary can also be pushed by home automation software, but may not be necessary.

dh-harald:
I've added the color temperature change feature.
We should consider the following things:
- KEY1 is hardcoded now to switch between color temp and brightness (maybe we could use a configuration)
- Rotary button is working only with Mi Desk Lamp (if some other projects are using rotary switch, this could be changed)

PR #4907, opened 2019-01-12
Add support for MAX44009 ambient light sensor

llagendijk:
Here is a driver for the MAX44009 ambient light sensor. This is a very low power device, that outputs the light level as a float. In order to support that I had to add a new HTTP_SNS_ILLUMINANCE_S as I needed to output a float (is this the right way?).
I have used xsns_91 for now pending allocation of an own driver number.
The size of the driver is 5k7. mainly caused by the calculations on line 57 in the MAX44009 driver. Is there a way to do float calculations that do not use so much code space?

Thanks for considering to include this in tasmota!
BR, Louis


andrethomas (commented):
You can re-use HTTP_SNS_ILLUMINANCE - you do not need to define a new HTTP_SNS_ILLUMINANCE_S

andrethomas2 (changes requested):
In addition to not defining HTTP_SNS_ILLUMINANCE_S and re-using the existing HTTP_SNS_ILLUMINANCE please look at the unique values listed for default after reset/power-on and make your detection routine a little more robust using this to validate that you're actually finding a MAX44009 as opposed to just finding something on that particular I2C address.

image

andrethomas2:
In addition, I don't think missed readings would apply because although the chip is configured in continuous measurement mode it will only provide whatever it has in its register on the 1-second interval you're reading it at... so maybe skim off a few bytes by using it without that - unless there's some technical or fundamental reason why you would want that logged output (which I can't imagine, but then I do not have the sensor so I am only reliant on what I see in the datasheet)

llagendijk:
> @llagendijk
> 
> Hi,
> 
> Any news on this?

Yeah, I am working on it: changes prove to be not so simple as I thought and I am a bit short on time. Please give me a few more days

llagendijk:
I made the requested changes:
- Added checks for improved MAX44009 detection
- removed HTTP_SNS_ILLUMINANCE_S (show only integer in web-interface)
- removed missed readings and repeated detection
Please comment if this solves your concerns

llagendijk:
Changes implemented. In the detect function I had to split the 16 bit read into 2 8 bit reads as the MAX44009 supports 16bit reading only for luminance.
Left a bit of if() else if for the precision determination as I find that so much easier to read.

andrethomas2 (approved):

PR #4923, opened 2019-01-14
Add a define for controlling the default MDNS state.

hackbar:
This only affects the first boot when flash is empty or invalid. After
flash been written once, MDNS is controlled bia SetOption55.


arendst:
Define needs to be in my_user_config.h

I'll update and release soon.

arendst:
Will release today.

PR #4957, opened 2019-01-16
PN532: Prevent non AlphaNumberic on DATA

andrethomas:
Block 1 (Used for text DATA) on a blank or newly formatted card may contain character values other than alpha-numeric which is the only characters we really want to use so I added a small loop to just ignore those cases so that we don't get garbage on the DATA JSON when such a card/tag is scanned.

Issue https://github.com/arendst/Sonoff-Tasmota/issues/4941 bears reference.


andrethomas2:
Hang on, need to fix an attempted write also.

andrethomas:
I suppose you could make a PR to add specific characters to be allowed if you are that way inclined.

andrethomas:
https://github.com/arendst/Sonoff-Tasmota/pull/5034 should do it.

qingz2004:
Great!
Thanks!

PR #4989, opened 2019-01-20
Update README.md

rogervdf:
text clarity


arendst:
As stated it tells the user to select the correct Arduino IDE Flash Size option being **1M (no SPIFFS)**

Your change removes the Arduino IDE special option and makes the text more general. Nice for non Arduino users.

I prefer to keep it this way.

Issue #4486, opened 2018-11-29
CalcRes "Command":"Unknown"


calcipher49:
I wish to use command CalcRes as described in Wiki Commands list:
Show / Set number of decimals used in commands ADD, SUB, MULT and SCALE

I use Tasmota last stable version **6.3.0**

But when I run command: **CalcRes** with or without parameter
Result is always: _{"Command":"Unknown"}_

So seems there was an ommission or mistake in Commands list.

Thanls if you can help.

Issue #2293, opened 2018-03-29
New Feature: Support for RGB WIFI LED (TYWE3S) Issue. Bulb only show white.


helmar74:
Make sure these boxes are checked before submitting your issue - Thank you!

- [ WIFI LIGHT BULB AILIGHT TYWE3S] What hardware you are using
- [X ] Provide the output of command ``status 0``
22:36:47 CMD: status 0
22:36:47 MQT: stat/sonoffB1TYWE3S/STATUS = {"Status":{"Module":27,"FriendlyName":"Sonoff","Topic":"sonoffB1TYWE3S","ButtonTopic":"0","Power":1,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"ButtonRetain":0,"PowerRetain":0}}
22:36:47 MQT: stat/sonoffB1TYWE3S/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffB1TYWE3SStandardGruppe","OtaUrl":"http://192.168.1.50:8080/static/firmware/sonoffB1TYWE3S.ino.bin","Uptime":"0 00:00:53","Sleep":0,"BootCount":15,"SaveCount":77,"SaveAddress":"F7000"}}
22:36:47 MQT: stat/sonoffB1TYWE3S/STATUS2 = {"StatusFWR":{"Version":"5.12.0","BuildDateTime":"2018.03.29 21:41:02","Boot":4,"Core":"2_4_0","SDK":"2.1.0(deb1901)"}}
22:36:47 MQT: stat/sonoffB1TYWE3S/STATUS3 = {"StatusLOG":{"SerialLog":2,"WebLog":2,"SysLog":0,"LogHost":"192.168.1.254","LogPort":514,"SSId1":"XXXXX","SSId2":"sonoffap","TelePeriod":300,"SetOption":"00008009"}}
22:36:47 MQT: stat/sonoffB1TYWE3S/STATUS4 = {"StatusMEM":{"ProgramSize":490,"Free":512,"Heap":19,"ProgramFlashSize":1024,"FlashSize":1024,"FlashMode":3}}
[...]

Issue #1822, opened 2018-02-07
Lyasi/Meamor bulbs


Jolanrensen:
So I bought a Lyasi bulb on Amazon since it also had an ESP8266 chip to try if I could get Tasmota to run on it. However when opening it I was greeted by the lack of Rx/Tx etc contact points.
[Photo album](https://photos.app.goo.gl/AA7OtqQ10U5CEupP2)
As you can see only the (gp)io0, (gp)io2 and RET (whatever that might be, Rest? Rx?) are available. Others might be as well where the green board is soldered to the blue one, but this is unlabled so I don't know. Is it possible to flash Tasmota using only those ports? And if so, how? Or does anyone know the layout of the board?
The bulbs are btw based on the Tuya protocol out of the box.

reloxx13:
google esp8266 pins
solder cables to the vcc,rx,rx,gnd pins directly on the chip
flash

Jolanrensen:
I would do that, however as you can see in the pictures the Rx/Tx chips on the top right of the chip are behind that cylinder thing. I can't properly see it, let alone solder a cable to it...
I did however get a response from Lyasi themselves and they probably will send me the board schematics (after the Chinese new year)

samuscherer:
Hey, did you receive the schematics? If so, do you think it's possible to upload them so we all can have a look at how the bulb is build? Thanks!

helmar74:
Depending on your Picture, I think I have the same bulb:
[https://github.com/arendst/Sonoff-Tasmota/issues/2293]

It's running Tasmota and I controll white LED, but not the RGB.
Good informations can be found:
[https://community.home-assistant.io/t/cheap-uk-wifi-bulbs-with-tasmota-teardown-help-tywe3s/40508/39]

Jolanrensen:
I think the AI light is slightly different as it does have labeled pins on the top, right? Mine almost has no labeled pins, so I asked Lyasi and got a specsheet from them. I'll try to flash tasmota on it in a week or so. Any idea how to backup the current software btw?

Jolanrensen:
Using esptool and the latest sonoff.bin I managed to get the server running! However, the lamp won't stop blinking, which makes sense as it's an unknown device

samuscherer:
Sounds like you fried your chip 😆 Do you think you'll get a new one or just buy a non-lyasi bulb?

helmar74:
Maybe this helps:
[...]

Issue #1146, opened 2017-11-07
Command "Dimmer 100" breaks H801 device


dimitrystd:
*Maybe it's related to #1133*
Spent a couple days to understand why H801 hangs from time to time.

* Use `v5.9.0a`
* Added my custom configuration [user_config_override.h](https://github.com/dimitrystd/Sonoff-Tasmota/blob/h801_bedroom/sonoff/user_config_override.h) to `/sonoff` folder
* Compiled and flashed my H801
* Run `SetOption15 1` in console and waited for restart
* Send via mqtt `cmnd/h801_bedroom/Dimmer 5`. Everything looks good
* Send again 100 via mqtt `cmnd/h801_bedroom/Dimmer 100`

Device starts to restart in infinite loop. It doesn't connect to wifi so i canno open console. In serial i see only
```

 ets Jan  8 2013,rst cause:4, boot mode:(3,7)

wdt reset
load 0x4010f000, len 1384, room 16
tail 8
chksum 0x2d
csum 0x2d
v09826c6d
~ld

 ets Jan  8 2013,rst cause:4, boot mode:(3,7)

wdt reset
load 0x4010f000, len 1384, room 16
tail 8
chksum 0x2d
csum 0x2d
v09826c6d
~ld
....
```
Turn off then on doesn't help. I have to flash device with blank image and then flash with firmware.

dimitrystd:
If run command in console
```
10:12:49 CMD: Dimmer 100
10:12:49 MQT: stat/h801_bedroom/RESULT = {"POWER":"ON", "Dimmer":100, "Color":"FFFFFF"}
```
Then it just restart once
```
10:13:06 MQT: tele/h801_bedroom/INFO3 = {"RestartReason":"Hardware Watchdog"}
```
So you should use mqtt for reproducing.

dimitrystd:
Hmm, repeated test with console, namely
```
speed 3
fade on
dimmer 100
```
and now it's died again.

arendst:
What is your PWMfrequency and PWMRange?

dimitrystd:
```
10:37:02 CMD: PWMfrequency
10:37:02 MQT: stat/h801_bedroom/RESULT = {"PWMFrequency":880}
10:37:10 CMD: PWMRange
10:37:10 MQT: stat/h801_bedroom/RESULT = {"PWMRange":1023}
```
Just tested on `5.9.0b` and bug is still here
```
[...]

PR #4991, opened 2019-01-20
Add support for the SM16716 RGB LED controller

gsimon75:
It is used in the Sunyesmart SYF05/07 (aka. Fcmila) smart bulbs, it controls the RGB LEDs via a synchronous serial protocol.

The code works for me in its current form, but as I'm new to the Tasmota codebase, most probably there is a lot to fix about it :), so I'd appreciate any constructive feedback.

[The bulb](https://www.flipkart.com/fc-mila-bxav-xs-ad-smart-bulb/p/itmf85zgs45fzr7n)

[SM16716 datasheet](http://www.datasheet-pdf.com/PDF/SM16716-Datasheet-Sunmoon-932771)

[A forum discussion about it](https://community.home-assistant.io/t/cheap-uk-wifi-bulbs-with-tasmota-teardown-help-tywe3s/40508/37)

[My notes](https://github.com/gsimon75/Tasmota-SYF05-Notes)


Jason2866:
Where is or are this devices available?  
I didnt found sources to get it easily or for a decent price

kueblc:
#2293 #1822 https://github.com/xoseperez/espurna/issues/568 http://fastled.io/docs/3.1/class_s_m16716_controller.html

This would be great to have for a number of bulbs and chipsets. Would love to have an option to set color clock and data pins from generic module, and also to be able to use PWM to control the white LEDs, controlled separately from the RGB in my bulb. I've been trying to work it out but I'm too new to the Tasmota codebase to know what I'm doing yet. Let me know how I can help.

kueblc:
> Where is or are this devices available?
> I didnt found sources to get it easily or for a decent price

In the US, Walmart sells a cheap (~$13) Tuya based RGBW bulb branded under Geeni / Merkury Innovations

gsimon75:
> Where is or are this devices available?
> I didnt found sources to get it easily or for a decent price
For example from AliExpress, just search for "Fcmila"

[Fcmila Smart LED Bulb WiFi](https://www.aliexpress.com/item/Fcmila-Smart-LED-Bulb-WiFi-Lights-Multicolored-E27-E26-E14-B22-LED-Lamps-Dimmable-Daylight-Home/32954055216.html)

gsimon75:
> ...Would love to have an option to set color clock and data pins from generic module
That's exactly what I did :).
Added 'SM16716_CLK' and '..._DAT' to the GPIO Pin config choices, so we can select them for any GPIO pin.  

>... and also to be able to use PWM to control the white LEDs, controlled separately from the RGB in my bulb.
That seems to work in my local build. On MQTT 'cmnd/sonoff/dimmer NN' controls the white PWM, while 'cmnd/sonoff/color RRGGBB' controls the RGB leds.

What I failed to achieve is exactly the opposite: To have one unified 'RGBW' color setting, from which the 'RGB' controls the colors and 'W' controls the PWM of the white leds.
I think I saw something similar in 'sonoff/xdrv_04_light.ino': In addition to PWM it can call through to the WS2812 module (search for 'Ws2812' in that file, with this letter cases), which is another color management driver.
Though I'm somewhat confused, because it is implemented by adding a new LightType, but it's strongly related to how the color components are parsed and converted, etc., so it looks very hard to extend to SM16716 with being backwards compatible.

In fact, this is why I kept it as separate controls, I'm looking for some advice from someone more experienced with this codebase :).

"I've been trying to work it out but I'm too new to the Tasmota codebase to know what I'm doing yet."
That makes us two for this category :D.

"Let me know how I can help."
To avoid darting off in a wrong direction I'd wait a few days for the opinions of some veterans, but if noone throws a veto, I think we might start experimenting a bit with my branch.

If you have devices with this controller, it would definitely help a lot to test the code with them, but for that I've got to locate why did the CI build fail and fix it first.

Oh, and my second TODO is to figure out how to **re**-assemble it in a safe and tidy way, like I've got to find some thermal adhesive to glue back the led panel to the heatsink :).

ascillato2 (changes requested):
Code don't compile. Check travis log please. Thanks

andrethomas2 (changes requested):
Please add void parameter to functions which do not have any parameters, e.g. void `SM16716_Update()` should be `void SM16716_Update(void)`

Confirm the logic of

```
else if (SM16716_Parse_RRGGBB(data, data_len)) { // RRGGBB
  }
  else { // rrr,g,bb
```

Conditional checks such as these

```
    if (tok)
      return false; // junk at the end
```

Should be clearly coded in {} so the following would be preferable

```
    if (tok) {
      return false; // junk at the end
    }
```

kueblc:
@gsimon75 Thanks for your work and your detailed write up, your section on compiling Tasmota without an IDE was especially helpful. I think it would be a great addition to the wiki.

Zebble:
Nice work!  Looking forward to trying this.  The Feit WiFi bulbs from Costco (Canada) have the SM16726 that is protocol-compatible.  Might snag some Merkury bulbs while im at it...

Jason2866:
Nice tutorial how to build Tasmota without a local IDE. Could you add it to wiki?

gsimon75:
> Nice tutorial how to build Tasmota without a local IDE. Could you add it to wiki?

[Done](https://github.com/arendst/Sonoff-Tasmota/wiki/Create-your-own-Firmware-Build-without-IDE), also added a link in the [Tutorials](https://github.com/arendst/Sonoff-Tasmota/wiki/Tutorials) list.

Zebble:
@gsimon75 In case you haven't seen the thread talking about the tool from @SynAckFin that enables flashing Tuya devices with Tasmota over the air, here it is:

https://github.com/codetheweb/tuyapi/issues/49

Github for the tool:

https://github.com/SynAckFin/TuyOTA

@kueblc also has a tool to speed up the process through MQTT calls to force an upgrade

https://github.com/kueblc/mocktuyacloud

Between these tools and your support for SM16716, the Feit bulbs should work nicely with Tasmota without ever having to crack open the bulb!

damondins:
Is there a link to the current bin for this? I have some merkury lights i'd like to test on

Zebble:
Looks like the Feit bulbs work in white mode (warm/cool adjustments also work) but color doesn't.  I suspect the SM16726 isn't no the same pins so I'll need to play around.  I don't see the CLK and DAT options in the GPIO pulldowns when I select the Generic module.  Any chance of adding this?  Then I can more easily experiment and add another template definition.

gsimon75:
> @gsimon75 In case you haven't seen the thread talking about the tool from @SynAckFin that enables flashing Tuya devices with Tasmota over the air, here it is:

Nice :) ! I've added a link for his description to the [Tutorials](https://github.com/arendst/Sonoff-Tasmota/wiki/Tutorials) list.

> @kueblc also has a tool to speed up the process through MQTT calls to force an upgrade

@kueblc : Could you pls add a link to the wiki? It would be very helpful for those who want to upgrade an old Tasmota version to a new one. (The method by SynAckFin works for the original firmware, if I get it right.)

gsimon75:
> Only seems to play with the white/warm white values and output of the color command gives something that only seems to change 2 values:
> 00:07:45 RSL: stat/sonoff/RESULT = {"POWER":"ON","Dimmer":19,"Color":"3018","HSBColor":"0,0,19","Channel":[18,9],"CT":326}

That status message comes from [xdrv_04_light.ino:590](https://github.com/arendst/Sonoff-Tasmota/blob/dcabb9c6dc1668209a6a770446253b79efbe1cc5/sonoff/xdrv_04_light.ino#L590), `void LightState(uint8_t append)`.

I'm new to this codebase, but it seems that some sort of color processing is done in xdrv_04_light.ino, in case when the R-G-B components are controlled by separate PWMs or by a WS2812 controller.

I don't know your 'Feit' bulb, but if it has an SM167x6, then xdrv_04_light.ino has nothing to do with the colors, but as the cold- and warm-white uses PWM1 and 2, those still belong to it.

Now I see two main reasons:
- Somehow the color command is intercepted and it never reaches xdrv_20_sm16716.ino
- It does, but there's still some bug in xdrv_20_sm16716.ino

I'd like to ask you to do some debugging, if you don't mind :).

1. First please try to configure your bulb with the 'Generic' module, and assign only the cold-white (PWM1) pin and the two of the SM16716, and give it a try with this setup.
My bulb here has only cold-white, so this is the setup I've tested it with. The presence of the warm-white shouldn't matter, but "shouldn't" and "doesn't" may be different :).

If that doesn't work, then

2. Please enable the logging by un-commenting the definition of [D_LOG_SM16716](https://github.com/gsimon75/Sonoff-Tasmota/blob/f833dd07ad02b2fd2c39d37cb95b864f0337c672/sonoff/xdrv_20_sm16717.ino#L32)

With that we'll see what exactly the driver is sending to the chip (`"Update; pwr=%02x, rgb=%02x%02x%02x"`), on what pins (`"ModuleSelected; clk_pin=%d, dat_pin=%d)"`), what command it got (`"Command; topic='%s', data_len=%d, data='%s', code=%d"`), what argument it tries to parse (`"Parse_Color; data='%s', data_len=%d"`), etc.

I hope this will narrow down the problem, so we can catch it in a few iterations :).

Jason2866:
@gsimon75  OTA upgrading is in Tasmota a included feature...
As far as i understand is the method from @kueblc another possibilty to replace the original firmware with Tasmota

kueblc:
> @kueblc : Could you pls add a link to the wiki? It would be very helpful for those who want to upgrade an old Tasmota version to a new one. (The method by SynAckFin works for the original firmware, if I get it right.)

> @gsimon75 OTA upgrading is in Tasmota a included feature...
> As far as i understand is the method from @kueblc another possibilty to replace the original firmware with Tasmota

@Jason2866 is correct, the MockTuyaCloud framework includes an example script that flashes any firmware over the stock firmware on Tuya devices using the vendor OTA upgrade system.

gsimon75:
> Only seems to play with the white/warm white values and output of the color command gives something that only seems to change 2 values

This is going to take some time to solve in a nice way.

Digging into xdrv_04_light.ino (especially at LightCommand()) now I see why it doesn't work with your Feit device and why it worked with my Fcmila.

If the bulb is capable of more than a simple white (that is, light_subtype > LST_SINGLE), then xdrv_04_light.ino takes on the command, handles it the best way it can, and returns true, signalling that the request has been serviced and no need to call further modules. That's why the SM16716 module hasn't even received the color commands on your Feit bulb. If you define only the PWM1 pin and the two SM16716 pins, then light_subtype will be LST_SINGLE (incorrectly, btw, but now it's good for us), and it shall work. But only by coincidence, because it's still a bug!

On my Fcmila exactly this was the case (one white only, plus the SM16716), that's why it worked.

And that's why I had to parse the color command myself in the SM16716 module, now I see.

So first of all I have to extend the light_type setting algorithm in sonoff.ino, so if there are SM16716 pins defined, then add the RGB capabilities, so light_type will be LST_RGBW or LST_RGBWC.

Then I have to add some code to xdrv_04_light.ino so that when it tries to set some RGB value and the SM16716 is present, then call the SM16716-specific functions to set the color.

The function 'LightCommand()' is 315 lines long, with quite a complex business logic in it, and I need to find the right place to add that rgb-setting call. Now I'm at LightAnimate(), which is not just for animating the colors, but for simply setting them as well (light_update=1). We'll see :)...

Jason2866:
Another OTA way to flash Tuya devices with Tasmota
https://github.com/ct-Open-Source/tuya-convert

gsimon75:
I'm sort of confused, so I'd welcome any advice. Sorry, this will be somewhat long.

Here is what I found, pls correct me if I'm wrong at something:

- To keep all additional features (schemes, color animation, etc.), this specific RGB controller driver must be integrated into xdrv_04_light.ino, and not must not parse and implement the commands by itself.

- The key to the logic of xdrv_04_light.ino is 'light_type' and 'light_subtype', they control what path the execution takes for any command.

- 'LightTypes' is somewhat of mixed purpose: it contains the *capabilities* of the device (Cold- and Warm-White, RGB), but also the *hw implementation* (like RGB is implemented via PWMs, or via My92x1, or via WS2812, or now via SM16716). In certain cases 'LightTypes' (more or less) means just the number of PWMs. The enum numbers are also kind of sensitive, as their lower 3 bits have special meaning (LightSubtypes).

- PWMs are just numbered (PWM1..5), but their roles are situation-dependent, like PWM1 may be WarmWhite, but in another situation it may be Red, it depends on light_type.

This Fcmila bulb I have here has only ColdWhite (but that via PWM1), and RGB via the SM16716 controller. The Feit bulb of @Zebble seems to have Cold- and WarmWhite (via PWM1 and 2), and RGB via this SM chip.

If we don't want RGB support, the light_type is LT_PWM1 and LT_PWM2, and respectively light_subtype is LT_SINGLE and LT_COLDWARM, and it all works well via xdrv_04_light.ino, but there is no color support.

To employ all the features of xdrv_04_light.ino, I tried to change LT_PWM1 to LT_RGBW and LT_PWM2 to LT_RGBWC if the SM-specific pins are defined, and then begun the trouble.

Now that I said there is color support, xdrv_04_light.ino caught CMND_COLOR (very correctly), so it didn't reach xdrv_20_sm16716.ino, so my original approach stopped working.

Then I tried (similarly to the my92x1 support) add an sm-specific call to `LightAnimate()`, because that is the point where the colors are actually updated.

And the RGB support started working. But the PWM support stopped :( !

It seems hardcoded in the logic of xdrv_04_light.ino that *if there is a non-trivial RGB support* (eg. light_type = LT_RGBW), *then there is no PWM controlling*.

It seems to be the consequence of that LightTypes mixes *capabilities* (like 'have warmwhite', 'have coldwhite' and 'have rgb') with *implementations* of them (like 'rgb is controlled by 3 PWMs' or 'rgb is controlled by WS2812').

As there is no (or at least I didn't recognise any) code path that contains both PWM control and RGB control, I think I'm out of luck with this case.

Certainly I don't want to turn everything inside out after spending only a few days with the code :D, but it seems that my problem would need some separations in xdrv_04_light.ino:

- separating capabilities (can_rgb) from implementations (rgb_via_pwms, rgb_via_etc.)
- separating high-level control (schemes, animation, hsb-to-rgb conversion, wakeup) from low-level control ('set these intensities to the leds, the way your hardware requires')
- separating pwm functions from numbers, so have PWM_WW, PWM_CW, PWM_R, _G, _B instead of _0.._5
- separating color components (r, g, b) from other controllable values (cw, ww) in variables like 'light_color[]'

Well, seeing so many things to change is like thinking that everyone drives the wrong way on the highway: it's more probable that I'm the one going in the wrong direction :D...

So I'd be thankful if someone could give me some advice (or link some example) how, at what point could I integrate a call to my 'sm16716_setrgb(uint8_t, uint8_t, uint8_t)' into xdrv_04_light.ino, with keeping the white controls at PWM1 and PWM2.

Zebble:
@gsimon75:  Thanks for digging into the details on this.  I briefly went through some of the code as it pertained to RGB and found the same things you did.  Sounds like @Jason2866 and @andrethomas did as well.  It does appear the RGB code will need to be refactored...

I'll help where I can, just let me know.  Happy to donate some bulbs if that will help as well.

kueblc:
Also got stuck around the same place as @gsimon75 and @Zebble 

Also open to assist with refactoring with project owner's direction

andrethomas:
> Sounds like @Jason2866 and @andrethomas did as well. It does appear the RGB code will need to be refactored

Only had a very brief discussion with @mike2nl on Discord on this so I'm not totally sure if it is in line with the project's mandate but I can comment on it I guess...

I think a rework will need careful consideration with the main objective being to segregate the code into separate callbacks.

Possibly something in the order of the main driver maintaining a set of global variables/parameters which are derived from settings with a secondary driver to maintain the light modes whether its just normal dimming or schemes and then individual drivers for each device type I guess.

I say careful consideration because although they could be segregated into separate files/drivers these will need to share some common goals which mean the use of global variables so that will have to be viewed as a potential tradeoff of flash size and ram utilisation against usability and future scalability.

Use case scenarios are pushing us to face implementing the code in ways which goes beyond just providing an alternative for Sonoff devices which is over-extending its original mandate in many ways but I do not see this as a bad thing, but the question is whether or not it is out of context given this projects main objective.

It's entirely up to @arendst and whichever direction he goes it will get my support for sure.

arendst:
The current light code base was initiated with as much functionality using as less resources as possible. This lead to the construction of either RGB or PWM support which is fine until manufacturers started to produce bulbs with both RGB and PWM. 

So just don't buy this hardware or rewrite the light driver.

Making the requested changes like both RGB AND PWM and more blinky blinky will extend the code beyond it's available resources. It needs to become either a dedicated light package or the current multipurpose package.

I think one needs to consider either support RGB AND PWM to be used as simple light control as is the current implementation (with some twinklings like fade and color cycle) and still keep the multipurpose package. This would mean a reasonable rewrite of the light driver.

The other features are unexhaustive as everyone wants the Mexican fire light functionality in another way resulting in a major code bloat making it just another package to control fancy lights.

For Tasmota I prefer the first option as it could be made using minor changes. As @andrethomas also noted the current implementation heavily rests on re-use of resources which is also making it impossible to use both RGB and PWM at the same time. Any deviation from this will make the code a lot larger. 

Doing it like the display driver using function calls makes it even worse as there is much code duplication and some displays just use a lot of RAM resulting in enabling/disabling compile time defines.

So for implementing both PWM and RGB keep in mind to use less resources as possible while making it extendable for future hardware while keeping legacy support. It would be great to just remove the current light driver with the new, extended one.

I won't be starting a rewrite but if some one does try to remember the above sentence.

andrethomas:
I suppose one could add to the current driver to add the ability to run schemes which are not PROGMEM'd into the firmware but uploaded to the SPI flash using a Scheme Manager (Software on Desktop/Notebook).

Jason2866:
> I suppose one could add to the current driver to add the ability to run schemes which are not PROGMEM'd into the firmware but uploaded to the SPI flash using a Scheme Manager (Software on Desktop/Notebook).

Or could be done in Tasmota too. Scheme upload via upgrade function (like the sonoff bridge firmware)

andrethomas:
> Or could be done in Tasmota too. Scheme upload via upgrade function (like the EFM bridge firmware)

Yeah but we'd still need an interface of sort for a user to create his/her custom binary blob... just a thought.

gsimon75:
> This lead to the construction of either RGB or PWM support which is fine until manufacturers started to produce bulbs with both RGB and PWM.

I see. It certainly would broaden the scope of the project, and if it's not (yet) a goal, then it indeed isn't worth the instability of a concept-level overhaul.

Considering this, I think I'll postpone this issue _until_ Itead also introduces such a mixed-control device, and these changes get more priority.

> So just don't buy this hardware or rewrite the light driver.

The problem with not buying such a hardware is that the vendors rarely ship a schematic along with the bulb, so one can tell whether it's supported or not only after buying and inspecting it.
Maybe it would be a good idea to keep a list of unsupported devices (along with the supported ones) to spare the struggle for the others.

Meanwhile I made an attempt to do some cleanup in xdrv_04_light.ino (at least on a sketch-level) **and** maintain full compatibility, but as I expected, I couldn't make the concept flexible enough and keep it unchanged enough at the same time. Maybe someone else with a broader view on the codebase will achieve it.

So I think I close this issue now; even if we couldn't solve it, at least we found the obstacle.

Thanks to everyone for the efforts and the feedback, and let's have better luck with the next hardware :) !

ascillato2:
Added labels enhancement and feature hold over for future reference.

kueblc:
I did give it a try with two, one, and none at all PWM pins. Still doesn't seem to be recognized. I'll try to dig a little more today, just wanted to see if I had missed something, like making sure the format of the command is correct.

gsimon75:
Got it, 'mea culpa' !
I misplaced a '#endif' in xdrv_20_sm16716.ino, so `SM16716_Command()` wasn't called at all unless the debug logging was enabled.
I pushed the fix to github:
```
commit a770f54e9022a45b6a3b538f375ed4a23b293550
Author: Gabor Simon <gabor.simon75@gmail.com>
Date:   Tue Jan 29 06:58:52 2019 +0000

    misplaced #endif fixed
```

damondins:
I'm getting the following trying to compile

sonoff.ino: In function 'void GpioInit()':

sonoff:2285:25: error: 'GPIO_KEY1_INV' was not declared in this scope

       else if ((mpin >= GPIO_KEY1_INV) && (mpin < (GPIO_KEY1_INV + MAX_KEYS))) {

                         ^

sonoff:2289:25: error: 'GPIO_KEY1_INV_NP' was not declared in this scope

       else if ((mpin >= GPIO_KEY1_INV_NP) && (mpin < (GPIO_KEY1_INV_NP + MAX_KEYS))) {

                         ^

E:\_Home Automation\tasmota dev bulb\Sonoff-Tasmota-development\sonoff\support.ino: In function 'bool GetUsedInModule(uint8_t, uint8_t*)':

support:716:15: error: 'GPIO_KEY1_INV' was not declared in this scope

   if ((val >= GPIO_KEY1_INV) && (val < GPIO_KEY1_INV + MAX_KEYS)) {

               ^

support:719:15: error: 'GPIO_KEY1_INV_NP' was not declared in this scope

   if ((val >= GPIO_KEY1_INV_NP) && (val < GPIO_KEY1_INV_NP + MAX_KEYS)) {

               ^

exit status 1
'GPIO_KEY1_INV' was not declared in this scope

gsimon75:
> I couldn't get the latest commit to compile

There was a cleanup of types in the upstream branch (boolean -> bool, among others), that affected the interface for the 'xdrv_*' plugins:

> commit 304ac6fe44c9ba9aca4478036555e9eaa442dc9b
> Author: Theo Arends
> Date:   Mon Jan 28 14:08:33 2019 +0100
> 
>     6.4.1.12 code changes (byte/boolean)
>     
>     6.4.1.12 20190128
>      * Change code use of boolean to bool and byte to uint8_t
>      * Change code uint8_t flags to bool flags

I've applied the changes on my branch as well, so please update it and I hope it'll compile now.

meyer-solutions:
Thanks for you work and this great enhancement. I'm using Tasmota for several outlets and now also for my lamp which has no name or number, it is just a low budget tuya wifi lamp. Works great, but only as Generic. I will post my setting later...

kueblc:
> > I couldn't get the latest commit to compile
> 
> There was a cleanup of types in the upstream branch (boolean -> bool, among others), that affected the interface for the 'xdrv_*' plugins:
> 
> > commit [304ac6f](https://github.com/arendst/Sonoff-Tasmota/commit/304ac6fe44c9ba9aca4478036555e9eaa442dc9b)
> > Author: Theo Arends
> > Date:   Mon Jan 28 14:08:33 2019 +0100
> > ```
> > 6.4.1.12 code changes (byte/boolean)
> > 
> > 6.4.1.12 20190128
> >  * Change code use of boolean to bool and byte to uint8_t
> >  * Change code uint8_t flags to bool flags
> > ```
> 
> I've applied the changes on my branch as well, so please update it and I hope it'll compile now.

Perfect, compiles fine now. Thank you again Gabor!

seidler2547:
Works for me, also Fcmila branded cheap bulb. Pins are
GPIO4 = SM16716 DAT
GPIO5 = PWM1
GPIO14 = SM16716 CLK

I can use "Dimmer" commands for the white LEDs and "Color" commands for the RGB. There are some issues, for one the labels aren't right, somehow I ended up with one label named "Button4inSM16716 CL" and number 131 (the real CLK) has no label at all. But it still works :-) Also, I think when I have the white LEDs above 60%, the lamp will reboot after a while. Maybe a design fault or a bad lamp, but I wanted to mention it.

andrethomas:
> Also, I think when I have the white LEDs above 60%, the lamp will reboot after a while

Are there any hints in the logs after booting as to the reason for the last restart such as this under status 1

`"RestartReason":"Software/System restart"`

kueblc:
> There are some issues, for one the labels aren't right, somehow I ended up with one label named "Button4inSM16716 CL" and number 131 (the real CLK) has no label at all.

There is a missing `"|"` in `sonoff_template.h`. Change line 224 from `D_SENSOR_BUTTON "1in|" D_SENSOR_BUTTON "2in|" D_SENSOR_BUTTON "3in|" D_SENSOR_BUTTON "4in"` to `D_SENSOR_BUTTON "1in|" D_SENSOR_BUTTON "2in|" D_SENSOR_BUTTON "3in|" D_SENSOR_BUTTON "4in|"`

> Also, I think when I have the white LEDs above 60%, the lamp will reboot after a while. Maybe a design fault or a bad lamp, but I wanted to mention it.

I noticed a similar issue on my bulb when color and white LEDs are active at the same time. The stock firmware only allows you to have either color or white on at a given time.

kueblc:
@gsimon75 I've integrate your code into `xdrv_04_light.ino` to bring it in line with our discussion here, and seems to work great. There is also a lot less duplicated code. Command processing is handled already, so you can do things like set the color and white levels simultaneously, fade between colors, and use the built in schemes. I also added a pin for SM16716 select, as is the case in my bulb. I'll post what I have shortly.

damondins:
I uploaded the firmware to a Walmart Merkury bulb and can get the RGB colors working but nothing with white is working. I don't see anywhere to set the pwm1 so I assume its pre-programmed. I'm guessing the pre-programmed PWM1 and PWM2 needs to be adjusted for this bulb. Let me know if there is anything I can do to test that.

gsimon75:
> I uploaded the firmware to a Walmart Merkury bulb and can get the RGB colors working but nothing with white is working.

If you used the 'SYF05' template, then for the time of experimenting please reconfigure it to 'Generic', it allows a lot more flexibility.

RGB works, that means that the SM16716 pins are correct.

The TYWE3S module has the following pins, with the assignment we have found:

GPIO16: Not Connected
GPIO14: SM16716_DAT (correct for your bulb too)
GPIO12: WarmWhite PWM
GPIO13: SM16716_SEL for @kueblc, Not Connected for me
GPIO5: ColdWhite PWM
GPIO4: SM16716_CLK (correct for your bulb too)
GPIO0: Button (always: the controller boots to firmware upgrade mode if pressed when powered on)
GPIO2: Not Connected
GPIO15: **hardwired to GND**

So if your bulb doesn't have a PWM on GPIO5 or GPIO12, then
- maybe GPIO16, GPIO2 or GPIO13 (unlikely) is worth a try
- I **wouldn't recommend GPIO15**, because if yours are also wired to GND, then it'll be a **short circuit**

If neither of those lights up the whites, then it's even possible that your bulb is of a different build and uses a non-PWM method to control them.

A photo of the controller board and maybe some wire tracking could tell a lot more, but I assume you haven't physically disassembled the bulb yet.

By the way, to test individual channels, go to the 'Console' menu and
1., set the dimmer to 100% via "cmnd/sonoff/dimmer 100" (the lights shall go to 100%)
2., control the channels directly via "cmnd/sonoff/color RRGGBBCCWW"

RR, GG, BB are the color components in hex, CC is the cold- and WW is the warm-white component

This way you can send the exact test vector you'd like to :).

seidler2547:
Thanks. Some testing showed that my lamp will cut out at around 55°C on the outer shell, which is probably a lot more inside. I never took my lamp apart (thanks to you guys for doing it to yours) because I was afraid I would damage the (already bad) thermal bond.

The lamp also reaches this temperature when I do not have the RGB leds on, only the white. So actually, I think I'll limit the maximum PWM1 value to around 60 in Home Assistant and be content with that. Since the is highly dependent on the type and quality of the bulb, would it be enough to add a warning somewhere to limit the maximum brightness manually?

damondins:
@gsimon75 Ok so I was able to get it to work on generic finally
Here are my pinouts for the Murkery Bulb

GPIO14: SM16716_DAT
GPIO13: SM16716_SEL
GPIO5: PWM1 Warm White ??
GPIO4: SM16716_CLK

The only issue that I have now is when Dimming the White is Flashes brighter after it changes and then is the correct dimmed setting on all settings except 0 and 100. Not sure what is causing that but probably not good.

gsimon75:
> Some testing showed that my lamp will cut out at around 55°C on the outer shell, which is probably a lot more inside.

It is. I experienced exactly the same (cut-out at at 54.8°C, after 40 min uptime), and the temperatures that time on the inner parts:
- white led controller 76
- rgb led controller 73
- tywe3s module 76
- transformer 85.5
- something behind/beneath it, probably the voltage stabiliser: 91.3

That indeed qualifies for 'a lot more' :)...
(I'm doing a counter-check with a Sonoff B1 right now, will plot a graph in around an hour.)

gsimon75:
>  when Dimming the White is Flashes brighter after it changes 

I'm not sure if it's relevant, but I found this in the code:
```
xdrv_04_light.ino:955:              cur_col[i] = 0xFC;   // Fix unwanted blinking and PWM watchdog errors for values close to pwm_range (H801, Arilux and BN-SZ01)
```
It's an old commit, it forces an upper limit of 0xFC to the PWM values to prevent some sort of 'unwanted blinking', its comment is "Fix PWM watchdog timeout if Dimmer is set to 100 or Color set to 0xFF (#1146)"

Does this flashing occur also when you are changing the dimmer value only between 'safe' values?
Do you see any correlation between the intensity/length of the blinking and the value or the change of the dimmer?

I'm not sure what I can do about this, but the more informations we have, the better.

gsimon75:
Here is the [temperature graph](https://raw.githubusercontent.com/gsimon75/Tasmota-SYF05-Notes/master/bulb_temps.png)
(The time on the X axis is in minutes, I've fixed the plot, it just needs time to propagate.)

Both bulbs grew gradually hotter, the increase is slowing, so there may (or may not) be an equilibrium temperature, but the SYF05 (Fcmila, red) couldn't reach it as its thermal protection shut it down at a chassis temperature of 54.8 Celsius.

The reason is not its worse thermal coupling (it's disassembled), because the hottest part was the power supply unit, and within that the ?stabiliser? being over 90 Celsius, and that area doesn't have a connection to the heatsink anyway.

The reference bulb, an unmodified Sonoff B1 produced a similar curve, though less steep. Even after an hour it was 'only' 50 Celsius, approx. 10 degrees cooler than the SYF05 would've been that time.
Maybe within another hour or two it would've overheated as well, but I didn't wait for that.

As this bulb wasn't disassembled, I have no information about its inner details. Its current consumption is around 42mA, that is 13.5% more than SYF05 and its heatsink area is smaller, and it still builds up the heat slower, so it seems more efficient in design.

Conclusion:
It seems to me that neither bulbs are designed for sustained 100% intensity, so our firmware should provide some way to throttle it to a safe value.
But that safe value is not a function of the LED controllers, but more of the model, so it either should be a template-bound parameter, or a configurable setting.

Both these are conceptional issues, so (as usual :)) I'd welcome suggestions and insights from the more experienced Tasmota developers.

digiblur:
Watching the work done on this one.....great job so far!  Might have to go pick me up a Merkury bulb to play with as well.  I've got a couple other oddballs that have the MY9291/MY9231 LED drivers that don't fit the profile of the Ailight/Sonoff B1 here and have been looking through code thinking of ways to easily template them as they channels can vary from model to model.

kueblc:
> Conclusion:
> It seems to me that neither bulbs are designed for sustained 100% intensity, so our firmware should provide some way to throttle it to a safe value.
> But that safe value is not a function of the LED controllers, but more of the model, so it either should be a template-bound parameter, or a configurable setting.
> 
> Both these are conceptional issues, so (as usual :)) I'd welcome suggestions and insights from the more experienced Tasmota developers.

We seem to have come to an understanding that the limits are highly dependent on the specific model. My bulb seems to have no problem running white at full intensity, or even color and white at moderate intensity. I only encounter an issue when both are high intensity.

That said, I'm more in favor of the approach suggested by @seidler2547, enforcing limits in your home automation setup rather than hard coded in firmware.

> The lamp also reaches this temperature when I do not have the RGB leds on, only the white. So actually, I think I'll limit the maximum PWM1 value to around 60 in Home Assistant and be content with that. Since the is highly dependent on the type and quality of the bulb, would it be enough to add a warning somewhere to limit the maximum brightness manually?

And yes, probably make a note to warn others they may have to find and set their own safe limits.

If restrictions like this are something we end up wanting to enforce in firmware, I'd say it's outside of the scope of this PR and should be developed as a separate feature. For now, I'd just like to get support for this chip family stable so we can push it upstream :+1:

damondins:
@gsimon75 
> > when Dimming the White is Flashes brighter after it changes
> 
> xdrv_04_light.ino:955:              cur_col[i] = 0xFC;   // Fix unwanted blinking and PWM watchdog errors for values close to pwm_range (H801, Arilux and BN-SZ01)

It is definitely unwanted blinking :) 

> Does this flashing occur also when you are changing the dimmer value only between 'safe' values?

It blinks only on white 000000ff It blinks no matter where on the slider or on the dimmer command. 
Ex. Dimmer 10 will blink and then immediately after dimmer 20 will blink. So yes even in the "safe" areas.

> Do you see any correlation between the intensity/length of the blinking and the value or the change of the dimmer?

No
 
> I'm not sure what I can do about this, but the more informations we have, the better.

More than happy to give you any info I can get from it. I have 2 others of the same bulbs that I can try as well.

maricnet:
Is there a link to the current bin, I wold like to test couple bulbs? Currently I’m on 6.4.1 basic and have PWM1 on GPI05 working fine.

kueblc:
@damondins Unfortunately I was unable to reproduce your issue. I welcome any outside input on this.

@maricnet Prebuilt binaries are not available for this yet, we're indev. You can build it yourself using any of the very helpful guides on the wiki to do so, or be patient and hopefully it will be available in the next release :smile:

@andrethomas2 @ascillato2 I pushed some final tweaks over to @gsimon75, and I'd like to request a review at your convenience to finish up this PR.

maricnet:
@kueblc Thanks 😉 Is this one of the helpful guides? "Create your own Firmware Build without IDE"

kueblc:
> @kueblc Thanks wink Is this one of the helpful guides? "Create your own Firmware Build without IDE"

Yes, @gsimon75 wrote that guide and I highly recommend it

damondins:
@kueblc Second bulb is flashed and colors work but white is not. Merkury bulb and works on GPIO5 set as PWM1 by itself but when using the RGB sel clk and dat is doesnt work. Very strange...

Trying to load the 3rd one now

damondins:
@kueblc Since we both have Merkury bulbs I set the pin settings that you used but no white. 
So the question is what commands are you using to switch between white and RGB?
Any test sequences  I can run to test. Right now I'm using Color ff00 / Color 00ff / Color 0000ff / Color 000000ff To view all the colors. The Last one just turns off the light.

kueblc:
> @kueblc Second bulb is flashed and colors work but white is not. Merkury bulb and works on GPIO5 set as PWM1 by itself but when using the RGB sel clk and dat is doesnt work. Very strange...
> 
> Trying to load the 3rd one now

Very strange indeed. So to confirm,
- When using generic, if GPIO5 is set to PWM1, white works, color does not
- When using generic, if SM16716 pins are set, color works, white does not

Do I have that right?

Have you tried the latest commit with the SYF05 module, setting GPIO13 to SM16716 SEL?

> @kueblc Since we both have Merkury bulbs I set the pin settings that you used but no white.
> So the question is what commands are you using to switch between white and RGB?
> Any test sequences I can run to test. Right now I'm using Color ff00 / Color 00ff / Color 0000ff / Color 000000ff To view all the colors. The Last one just turns off the light.

Yes looks like you're using the correct commands, `Color <rrggbbww>`.

kueblc:
> > That said, I'm more in favor of the approach suggested by @seidler2547, enforcing limits in your home automation setup rather than hard coded in firmware.
> > And yes, probably make a note to warn others they may have to find and set their own safe limits.
> 
> I agree, so I've added a note for it to the [First Steps](https://github.com/arendst/Sonoff-Tasmota/wiki/Initial-Configuration) list.

Excellent, thanks @gsimon75 looks great! With that, just awaiting feedback from maintainers.

damondins:
@kueblc 

Ok, so this gets stranger... 

Flashed the bulb to minimal and white works fine no flashing when changing dimming. Can turn on and off the bulb.

Flashed the new version of the dev firmware and set to just GPIO5 -> PWM1 and generic(18)
The bulb turns on and off fine. flashes after every change on the dimmer

Changed type to SYF05(68) set GPIO12 -> SM16716 SEL
tried dimmer and toggle and buld is off will not turn on
issued command color ff00 and I get a dim cold white
issued command color ff00 again and I get no change
issued command color 00ff and get green
issued command color ff00 and get red
issued command color 0000ff and get blue
issued command color 000000ff and light is off but power is at 100 and dim is 100

here is the log
```
20:49:58 RSL: STATE = {"Time":"2019-02-02T20:49:58","Uptime":"0T00:10:18","Vcc":1.124,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":19,"POWER":"ON","Dimmer":100,"Color":"FF000000","HSBColor":"0,100,100","Channel":[100,0,0,0],"Scheme":0,"Fade":"OFF","Speed":1,"LedTable":"OFF","Wifi":{"AP":1,"SSId":"WLAN-Devices","BSSId":"34:8F:27:55:F7:48","Channel":11,"RSSI":80}}
20:50:09 CMD: color ff00
20:50:09 RSL: RESULT = {"POWER":"ON","Dimmer":100,"Color":"FF000000","HSBColor":"0,100,100","Channel":[100,0,0,0]}
20:50:13 CMD: color ff00
20:50:13 RSL: RESULT = {"POWER":"ON","Dimmer":100,"Color":"FF000000","HSBColor":"0,100,100","Channel":[100,0,0,0]}
20:50:18 CMD: color 00ff
20:50:18 RSL: RESULT = {"POWER":"ON","Dimmer":100,"Color":"00FF0000","HSBColor":"120,100,100","Channel":[0,100,0,0]}
20:50:21 CMD: color ff00
20:50:21 RSL: RESULT = {"POWER":"ON","Dimmer":100,"Color":"FF000000","HSBColor":"0,100,100","Channel":[100,0,0,0]}
20:50:24 CMD: color 0000ff
20:50:24 RSL: RESULT = {"POWER":"ON","Dimmer":100,"Color":"0000FF00","HSBColor":"240,100,100","Channel":[0,0,100,0]}
20:50:32 CMD: color 000000ff
20:50:32 RSL: RESULT = {"POWER":"ON","Dimmer":100,"Color":"000000FF","HSBColor":"0,0,0","Channel":[0,0,0,100]}
```

andrethomas2 (approved):
Approved in principal, will request approval from @arendst - Not sure if @ascillato has any further inputs.

kueblc:
> ps... Travis is still not happy with you for the changes on sonoff_template.h - likely because other smart bulbs have been merged.

I'm sure it'll be a quick fix, I'll take a look this evening. Maybe we'll wait a day or so to see if we can figure out what's causing issues for @damondins

andrethomas2:
> Maybe we'll wait a day or so to see if we can figure out what's causing issues for @damondins 

Good idea...

gsimon75:
> Maybe we'll wait a day or so to see if we can figure out what's causing issues for @damondins

I have no specific idea, but that SM16716_SEL signal is ... weird. On my bulb it is not connected,  I cannot do tests with it, so all I have are speculations.

According to what google translate produced from the datasheets, the 726 has two data protocols, and the SEL signal chooses between them.

SEL=1 is the 716-compatible mode:
- Preamble: >50 zero bits
- For each chained chip: 1 start bit (1) + 24 bit rgb data
- Trailing: as many additional 0-s, as the number of chips (???) (NOTE: this wasn't enough in my case: as many *pixels* were needed)

SEL=0 is the new mode:
- Preamble: >52 zero bits
- For each chip: 1 start bit (1) + 1 D/#S bit (see below) + 24 bit data
- Trailing: as many additional 0-s, as the number of chips (???)

That additional D/#S bit chooses what to write: 1=RGB Data, 0="Status register data" (controls output gain).

The datasheets:
- [SM16716 datasheet](https://datasheetspdf.com/pdf-file/932771/Sunmoon/SM16716/1)
- [SM16726 datasheet](https://cdn.datasheetspdf.com/pdf-down/S/M/1/SM16726-Sunmoon.pdf)

So the two modes are **incompatible**, because in the SEL=0 mode an additional zero bit is required both in the sync preamble, and also before the rgb data.

The current code sends the rgb data in the old protocol (1 start + 24 rgb), so I see two problems here:
- if we send 1+24 bits of data when SEL=0, that is, at 'SM16716_Update(0, 0, 0)', it is misinterpreted, unterminated and puts the state machine out of sync
- if it's out of sync, then the subsequent writes are also misinterpreted

(Maybe this is the reason why that 'SM16716_Init()' is needed after setting SEL to HIGH?)

If this SEL is not hardwired to HIGH on your devices, then just not configuring SM16716_SEL may not be enough: perhaps it should be explicitely set to HIGH in '_Init()'?

@kueblc: Could you please do a test with a local build that sets SEL to HIGH at the beginning and leaves it that way?

kueblc:
@gsimon75 My naming of that pin is the source of confusion here, it is not the same as the SM16726 SEL described. Perhaps a better name would be SM ENABLE?

I came up with that not from a datasheet, but based on my experience with the stock firmware and by mapping the traces on my bulb with a SM726EB.

The power to the SM726EB and color LEDs is gated by pin 13 and in the stock firmware, power is cut while in white mode. I suspected this was to conserve power while not using color, does that make sense?

So while it might work to leave that pin HIGH all the time, I'm guessing it would draw more power, generate more heat, and ultimately cause more unexpected shutdowns.

gsimon75:
>  it is not the same as the SM16726 SEL described

Then it indeed makes no sense to keep it always high. (It was just a wild guess :))

kueblc:
[SM726EB datasheet](http://chinaasic.com/product.jsp#item=series#style=21#id=104). GPIO13 gates pin 5, VDD, as well as the color LEDs themselves. This helps explain the occasional brief flash that occurs when switching to color mode. The color LEDs start getting power before the SM726EB is fully initialized so it displays a "cool white" for a few microseconds before the driver kicks in.

elzershark:
"Channel":[0,0,0,100]}
if a 100 = white... if a 0 = color
the first 3 values ​​are ok
100 = channel4 (or channel3? i down't now)
the 100 = level.color.temperature
from Alexa (Amazon Echo) is the number 250 for White (level.color.temperature)
when the number = 250  then the number must be changed to 100
 (not 0,0,0,250.. change to 100 (0,0,0,100)
if the number is not 100 then do nothing
then can make in tasmota Emulation Hue Bridge
and then everything works
sorry, i'm german

kueblc:
Sorry @elzershark I'm not understanding. If you'd like to post your question or comment in German I can use a translator.

kueblc:
I noticed yesterday that schemes were no longer working well on my bulb and I've tracked it down to 737cd7a2d013929b30a706f02d70d1ff40e068dd, merging from development. It seems to be constantly dropping off MQTT, and even the animation seemed janky.

Can anyone confirm that the development build misbehaves with schemes on other types of bulbs? Is upstream aware of this, or have a workaround?

kueblc:
@damondins could you try 267896914ca4f458d297811086a65739d01a45a0 and see if your issues persist?

damondins:
I will try in about an hour

gsimon75:
> I noticed yesterday that schemes were no longer working well on my bulb and I've tracked it down to 737cd7a, merging from development.

With a fresh build the schemes seem to work on my bulb ('cmnd/sonoff/scheme 2'), and I haven't experienced troubles with MQTT either.
The upstream change merged in https://github.com/arendst/Sonoff-Tasmota/commit/737cd7a2d013929b30a706f02d70d1ff40e068dd seems to do something with the config location within the flash and it also added a new item to Settings.

Did you update the firmware via OTA, or just rewrote the whole flash via esptool (like me)?

gsimon75:
@elzershark:
I'm having some trouble to interpret what you wrote, so as @kueblc said, feel free to write it in German if you're more comfortable with it. Sag's auf Deutsch, wir werden versuchen es zu verstehen :D.

Meanwhile:

> "Channel":[0,0,0,100]}
> if a 100 = white... if a 0 = switch to color, other = dimming

Yes, Channel4 is the (cold) white channel, and its range is [0..100], so 100=full white, and 1..100 is dimming.
However, color and white LEDs can work simultaneously, so you don't need to switch the white leds off to control the color ones.

For example, if you go to the 'Console' menu of the web interface:
```
cmnd/sonoff/power off
# now every led shall be switched off
cmnd/sonoff/channel1 50
# now the red leds shall be at 50%
cmnd/sonoff/channel4 10
# now the white leds shall be at 10%
cmnd/sonoff/channel1 0
# now the red leds shall be off, only the whites remain on
cmnd/sonoff/channel4 0
# everything shall be off again
```

(That 'sonoff' is the default topic name of the bulb, you can change it in Configuration/MQTT/Topic.)

> the 100 = level.color.temperature

No, it's just the intensity % of the white LEDs.

Color temperature is a higher-level control, so if you specify a temperature value ('cmnd/sonoff/ct', and its argument is [153..500]), then from that, and from the dimmer value ('cmnd/sonoff/dimmer') are calculated the intensity values for cold white (channel4) and warm white (channel5).

But that works only if your bulb has **both** cold- and warm-white LEDs!

> if the number is not 250 then do nothing.

Yes, Channel values must be in the [0..100] range, other values are invalid, they are ignored, and only the current setting is reported back.

If you want to use 0..255 intensity **values** (and not percentages), than maybe the 'color' command is what you're looking for:
```
cmnd/sonoff/color 16,32,64,8,12
cmnd/sonoff/color 102040080c
cmnd/sonoff/color #102040080c
```
These all set the red to 0x10=16, green to 0x20=32, blue to 0x40=64, cold-white to 0x08=8, warm-white to 0x0c=12, but please note that these values are still affected by the 'dimmer' value, so if 'dimmer=50', then the LEDs will get half the intensity that you configure here. (If you don't need some of the white values, you may specify only 3 or 4 bytes.)

If you'd like to experiment with the commands, I suggest the Console menu on the web interface (or an MQTT client), and here you can find the description of the [MQTT commands](https://github.com/arendst/Sonoff-Tasmota/wiki/Commands).

elzershark:
Super schwer zu erklären. Emulation Hue Bridge in Tasmota. das funktioniert nicht. Weil level.color.temperature = 250 ist. Wenn man auf weiß stellt... Diese 250 werden in Channel":[x,x,x,100]} geschrieben. Channel":[x,x,x,250]}. das funktioniert nicht. das ist falsch. Sobald level.color.temperature = 250 ist, schreibe es um in 100 und erst dann schreibe in Channel":[x,x,x,100]} die 100 rein. Nur bei 250.  bei keiner anderen Zahl. Dann funktioniert die Lampe so wie es sein soll. Alle Farbe, Weiß, Dimmen u.s.w.

elzershark:
https://www.youtube.com/watch?v=3IG4KjpqAnU

seidler2547:
Ok, I finally managed to check out the latest code. Works for me. The only issue I have is that the ESP restarts from time to time. The log says:

    "RestartReason":"Fatal exception:3 flag:2 (EXCEPTION) epc1:0x4010011d epc2:0x00000000 epc3:0x40000f68 excvaddr:0x40035718 depc:0x00000000"

Is that anything to do with your code?

elzershark:
weiß = 250 = rgbw
warm weiß = 455 = rgbww
weiches weiß = 370 = rgbww
tageslicht = 182 = rgbww
kühles weiß = 143 = rgbww

gsimon75:
> weiß = 250 = rgbw + rgbww

Ah, I see. So the range of the color temperature is OK, that's good news.

The bad news is that color temperature handling is available only for RGBWC bulbs, here is the [code](https://github.com/arendst/Sonoff-Tasmota/blob/16ec25bdff43df24f1972f000d5d468db2960cbf/sonoff/xdrv_04_light.ino#L980)

Just for testing, I have an idea:

In the Module Configuration try to assign PWM2 to a GPIO port that is not connected to anything else,
with this we lie a bit and say that there is a warm-white Channel5 as well.
Even if there is actually no such LED in your bulb, Tasmota will gladly control that unconnected pin as if it were.

This way your bulb will count as RGBWC, so if your home assistant sends color temperature, it will be used for controlling the (existing) cold-white and the (fictional) warm-white channels. Sure, you'll only see the cold component from the 'cold+white' mix, so the colder the color, the more useful it will be, but at least we'll see if the concept works or not.

As of which GPIO pin to use for this: On some other SM16716-based bulbs it **seems** that GPIO12 is reserved for warm-white, but I can't say for sure that is safe for your bulb as well, it's your risk.

kueblc:
> > I noticed yesterday that schemes were no longer working well on my bulb and I've tracked it down to [737cd7a](https://github.com/arendst/Sonoff-Tasmota/commit/737cd7a2d013929b30a706f02d70d1ff40e068dd), merging from development.
> 
> With a fresh build the schemes seem to work on my bulb ('cmnd/sonoff/scheme 2'), and I haven't experienced troubles with MQTT either.
> The upstream change merged in [737cd7a](https://github.com/arendst/Sonoff-Tasmota/commit/737cd7a2d013929b30a706f02d70d1ff40e068dd) seems to do something with the config location within the flash and it also added a new item to Settings.
> 
> Did you update the firmware via OTA, or just rewrote the whole flash via esptool (like me)?

I uploaded OTA, that is the only option for me right now. I haven't been able to get one of these bulbs fully opened, only the plastic dome was easy to remove without total destruction. It was originally flashed from the stock firmware OTA via my MockTuyaCloud framework. This also means I'm very careful not to brick the thing as there is no easy recovery.

If you think it might be something residual in flash, perhaps a call to `Reset 2` would help?

This recent instability is proving difficult to track down. Still haven't a clue what's causing issues for @damondins on the white channel.

kueblc:
> Ok, I finally managed to check out the latest code. Works for me. The only issue I have is that the ESP restarts from time to time. The log says:
> 
> ```
> "RestartReason":"Fatal exception:3 flag:2 (EXCEPTION) epc1:0x4010011d epc2:0x00000000 epc3:0x40000f68 excvaddr:0x40035718 depc:0x00000000"
> ```
> 
> Is that anything to do with your code?

> You'd have to check the .map file which was generated for a specific compile to see if the origin of epc1:0x4010011d can be traced to a specific function or functional area.

Please let us know what you find @seidler2547 :+1:

Jason2866:
Alexa HueSimulation in Tasmota does not support color.
It supports on / off and dimm.

elzershark:
yes. GPIO0 = pwm 2. Then work Emulation Hue Bridge to 80%. Color dimmen and white is ok. but is first color and then white, is the white 60%. and this can not change to 100...

with my attitude. So with the level.color.temperature everything works as it should be

elzershark:
> Alexa HueSimulation in Tasmota does not support color.
> It supports on / off and dimm.

I think, this is not right. is support color. my lamp make color, white, dimm. but is color, and i say white, then is the color white. but 60%. 100% is impossible. i down't now wy.

gsimon75:
> yes. GPIO0 = pwm 2.

Good choice for a bulb! It's the pin for the external button if there is any, but on a bulb it's indeed not used.

> is the white 60%. and this can not change to 100...

Yes, because now your home assistant is setting the color temperature to 250.
The exact value should be 72%, because the range is 153..500, so 250 is at (250-153) / (500-153) = 28% to the (non-existing) warm white and (500-250)/(500-153) = 72% to the cold-white.

This experimental 'hack' proved that the initial problem indeed occured because on RGBW bulbs we don't handle the color temperature.

But that is a generic question that affects every RGBW bulb, not only those with this SM16716 controller, so I don't think we can solve it in this controller-specific topic.

As of the problem in general, well, your home assistant system still tries to switch to white light by setting the color temperature, which, at least for now, is just as impossible and invalid as setting colors on a plain white-only bulb: such a command cannot be treated as 'completed successfully', so Tasmota is right in refusing it.

It seems that others found this issue too: [Issue #4486](https://github.com/arendst/Sonoff-Tasmota/issues/4886)

What @gitolicious wrote implies that Home Assistant first detects if a bulb is capable of color temperature handling by asking its status first, and then it uses color temp **only** if it found it in the status, otherwise (correctly) it uses Channel4 for white.

Now if I ask the status of my bulb (RGBW, no fake warm-white) in the Console of its web interface:
```
18:47:57 CMD: cmnd/sonoff/status 0
18:47:57 MQT: stat/sonoff/STATUS = {"Status":{"Module":18,"FriendlyName":["Sonoff"],"Topic":"sonoff","ButtonTopic":"0","Power":1,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":0}}
18:47:57 MQT: stat/sonoff/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"http://thehackbox.org/tasmota/release/sonoff.bin","RestartReason":"Power on","Uptime":"0T00:02:59","StartupUTC":"2019-02-04T17:44:58","Sleep":50,"BootCount":27,"SaveCount":195,"SaveAddress":"F9000"}}
18:47:57 MQT: stat/sonoff/STATUS2 = {"StatusFWR":{"Version":"6.4.1.12(sonoff)","BuildDateTime":"2019-02-01T03:37:19","Boot":31,"Core":"2_4_2","SDK":"2.2.1(cfd48f3)"}}
18:47:57 MQT: stat/sonoff/STATUS3 = {"StatusLOG":{"SerialLog":3,"WebLog":2,"SysLog":0,"LogHost":"","LogPort":514,"SSId":["agifules2",""],"TelePeriod":300,"SetOption":["00008009","558180C0","00000000"]}}
18:47:57 MQT: stat/sonoff/STATUS4 = {"StatusMEM":{"ProgramSize":522,"Free":480,"Heap":24,"ProgramFlashSize":1024,"FlashSize":1024,"FlashChipId":"1440C8","FlashMode":3,"Features":["00000809","0FDAE794","240383A0","23B617CE","00003BC0"]}}
18:47:57 MQT: stat/sonoff/STATUS5 = {"StatusNET":{"Hostname":"sonoff-2802","IPAddress":"*","Gateway":"*","Subnetmask":"*","DNSServer":"*","Mac":"*","Webserver":2,"WifiConfig":4}}
18:47:57 MQT: stat/sonoff/STATUS6 = {"StatusMQT":{"MqttHost":"*","MqttPort":1883,"MqttClientMask":"DVES_%06X","MqttClient":"DVES_E02AF2","MqttUser":"*","MqttType":1,"MAX_PACKET_SIZE":1000,"KEEPALIVE":15}}
18:47:57 MQT: stat/sonoff/STATUS7 = {"StatusTIM":{"UTC":"Mon Feb 04 17:47:57 2019","Local":"Mon Feb 04 18:47:57 2019","StartDST":"Sun Mar 31 02:00:00 2019","EndDST":"Sun Oct 27 03:00:00 2019","Timezone":"+01:00","Sunrise":"08:16","Sunset":"17:51"}}
18:47:57 MQT: stat/sonoff/STATUS10 = {"StatusSNS":{"Time":"2019-02-04T18:47:57"}}
18:47:57 MQT: stat/sonoff/STATUS11 = {"StatusSTS":{"Time":"2019-02-04T18:47:57","Uptime":"0T00:02:59","Vcc":3.096,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":19,"POWER":"ON","Dimmer":1,"Color":"02020201","HSBColor":"0,0,1","Channel":[0,0,0,0],"Scheme":0,"Fade":"OFF","Speed":1,"LedTable":"OFF","Wifi":{"AP":1,"SSId":"*","BSSId":"*","Channel":4,"RSSI":92}}}
18:50:12 MQT: tele/sonoff/STATE = {"Time":"2019-02-04T18:50:12","Uptime":"0T00:05:14","Vcc":3.366,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":19,"POWER":"ON","Dimmer":1,"Color":"02020201","HSBColor":"0,0,1","Channel":[0,0,0,0],"Scheme":0,"Fade":"OFF","Speed":1,"LedTable":"OFF","Wifi":{"AP":1,"SSId":"*","BSSId":"*","Channel":4,"RSSI":92}}
```

There is no color temp data (like '..."CT":499...' in @gitolicious' issue) neither in the STATUS11 line, nor in the STATE line, but the  "Color" values contain 4 bytes of RGBW.

Could you please:
1. Disable that fake warm-white pin assignment
2. Try this `cmnd/wled/status 0` command in the Tasmota web ui Console
3. Check for '"CT":..." in the output STATUS11 line?

If it isn't present, and the '"Color"' values are 8 digit long, then maybe it would be worth deleting and re-adding this bulb to your Home Assistant config (as it happened with @gitolicious) ?

elzershark:
`19:49:51 CMD: cmnd/wled/status 0
19:49:51 MQT: stat/wled/STATUS = {"Status":{"Module":69,"FriendlyName":["Wohnlicht"],"Topic":"wled","ButtonTopic":"0","Power":1,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":0}}
19:49:51 MQT: stat/wled/STATUS1 = {"StatusPRM":{"Baudrate":115200,"GroupTopic":"sonoffs","OtaUrl":"http://thehackbox.org/tasmota/release/sonoff.bin","RestartReason":"Software/System restart","Uptime":"0T00:28:57","StartupUTC":"2019-02-04T18:20:54","Sleep":50,"BootCount":21,"SaveCount":557,"SaveAddress":"F7000"}}
19:49:51 MQT: stat/wled/STATUS2 = {"StatusFWR":{"Version":"6.4.1.13(sonoff)","BuildDateTime":"2019.02.03 13:27:07","Boot":4,"Core":"2_4_2","SDK":"2.2.1(cfd48f3)"}}
19:49:51 MQT: stat/wled/STATUS3 = {"StatusLOG":{"SerialLog":2,"WebLog":2,"SysLog":0,"LogHost":"","LogPort":514,"SSId":["elzershark",""],"TelePeriod":300,"SetOption":["00008009","558180C0","00000000"]}}
19:49:51 MQT: stat/wled/STATUS4 = {"StatusMEM":{"ProgramSize":526,"Free":476,"Heap":21,"ProgramFlashSize":1024,"FlashSize":1024,"FlashChipId":"1440C8","FlashMode":3,"Features":["00000407","0FDAE794","240383A0","23B617CE","00003BC0"]}}
19:49:51 MQT: stat/wled/STATUS5 = {"StatusNET":{"Hostname":"wled","IPAddress":"192.168.178.26","Gateway":"192.168.178.1","Subnetmask":"255.255.255.0","DNSServer":"192.168.178.1","Mac":"60:01:94:48:D9:FE","Webserver":2,"WifiConfig":4}}
19:49:51 MQT: stat/wled/STATUS6 = {"StatusMQT":{"MqttHost":"192.168.178.28","MqttPort":1884,"MqttClientMask":"wled","MqttClient":"wled","MqttUser":"Sonoff","MqttType":1,"MAX_PACKET_SIZE":1000,"KEEPALIVE":15}}
19:49:51 MQT: stat/wled/STATUS7 = {"StatusTIM":{"UTC":"Mon Feb 04 18:49:51 2019","Local":"Mon Feb 04 19:49:51 2019","StartDST":"Sun Mar 31 02:00:00 2019","EndDST":"Sun Oct 27 03:00:00 2019","Timezone":"+01:00","Sunrise":"08:16","Sunset":"17:51"}}
19:49:51 MQT: stat/wled/STATUS10 = {"StatusSNS":{"Time":"2019-02-04T19:49:51"}}
19:49:51 MQT: stat/wled/STATUS11 = {"StatusSTS":{"Time":"2019-02-04T19:49:51","Uptime":"0T00:28:57","Vcc":3.444,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":19,"POWER":"ON","Dimmer":100,"Color":"FF0000FF","HSBColor":"0,100,100","Channel":[100,0,0,100],"Scheme":0,"Fade":"OFF","Speed":1,"LedTable":"OFF","Wifi":{"AP":1,"SSId":"elzershark","BSSId":"9C:C7:A6:8E:11:3E","Channel":11,"RSSI":80}}}`

kueblc:
> the problem is easy... when is color.. then is all ok.. is white.. then must a 100 in channel4.
> if no 100 is written, it will not work on any lamp. no matter what

This works as intended. Channel 4 is white. What did you expect it to do?

elzershark:
this tasmota script write not 100 in the channel4. this is only manuel

kueblc:
> this tasmota script write not 100 in the channel4. this is only manuel

Sounds like indeed this is not relevant to this pull request. We're working on adding support for a particular family of LED drivers. If you can manually set channel 4 to 100, and your white LEDs turn on, then it works as intended. If you have an issue with your home automation configuration, this is not the place to ask.

elzershark:
my english is so bad.. 
original tasmota script = esp-01 + led strip ws2812b   Emulation Hue Bridge.  color, white, dimm works fine.
this script with a luminea bulb + Emulation Hue Bridge has not the level.color.temperature. it does not work either
it is not in a home automation

gsimon75:
> It seems to be constantly dropping off MQTT

@kueblc: May it be related to [this](https://github.com/arendst/Sonoff-Tasmota/issues/5109), more precisely to [this note](https://github.com/arendst/Sonoff-Tasmota/wiki/Troubleshooting#wifi-issues-arduino-core-versions-and-expressif-sdk) provided by @Jason2866 ?

My installation procedure now is to erase flash and write the image via esptool, just as mentioned in the note, and I experience no MQTT problems. Maybe that `reset 2` you mentioned, or that `sleep 0` from the note would be worth a try?

@seidler2547 :
 
> "RestartReason":"Fatal exception:3 flag:2 (EXCEPTION) epc1:0x4010011d epc2:0x00000000 epc3:0x40000f68 excvaddr:0x40035718 depc:0x00000000"

- Exception 3 is "LoadStoreErrorCause" according to the [Xtensa reference](https://0x04.net/~mwk/doc/xtensa.pdf#G9.1921073)
- epc1 is the address of the faulty instruction, it's in the framework code function `umm_assimilate_up`, so by itself it's not Tasmota-build-specific, but the call stack most probably is
- excvaddr is the address of the failed load/store operation, according to the [memory map](https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map) it is indeed not valid

Btw, others have run into [this exact exception](https://github.com/esp8266/Arduino/issues/2957), so I think it's from the [Arduino framework libs](https://github.com/esp8266/Arduino/blob/09826c6d87e54c15422da90f060377c9ef188947/libraries/ESP8266WebServer/src/Parsing.cpp#L54).

Their code uses Arduino 2.3.0 libs, ours (by default) 2.4.2 but that `readBytesWithTimeout` implementation is still the same, so most probably you've caught the same problem.

It has been completely overwritten in [2.5.0-beta1](https://github.com/esp8266/Arduino/blob/9c1e03a1f99d3d6caafc534438734999597c4442/libraries/ESP8266WebServer/src/Parsing.cpp#L52), so if the problem persist, it may be worth trying to rebuild the firmware with 2.5.0 by enabling it in [platformio.ini](https://github.com/arendst/Sonoff-Tasmota/blob/b839d7e80e962adcc21ac0103b58024e80f17030/platformio.ini#L121)

It seems that the size of the network packet is related to the issue, so if you can dump the traffic between the bulb and whatever is communicating with it, maybe you could check if there are any packets larger than usual when the crashes occur.

kueblc:
@gsimon75 you may be right, looks like I can't reliably reproduce the issue. I'll circle back if I gain any more insight into the issue.

@ascillato2 @arendst Merge/review status? Would like to close out this PR if the additions are acceptable, moving tangential issues to their own issues/PRs

arendst (changes requested):
All looks fine except:

- The language files are out of sync with current state as new translations have been added.
- Change D_SENSOR_SM16716_POWER text to "SM16716 PWR" as many translations are way too long.

When those have been fixed it's ready to merge.

Once done I will have a look at the used define in light.ino.

gsimon75:
>     * The language files are out of sync with current state as new translations have been added.
Resolved, tests are running.

>     * Change D_SENSOR_SM16716_POWER text to "SM16716 PWR" as many translations are way too long.
I'm on it, need a few hours.

kueblc:
> * Change D_SENSOR_SM16716_POWER text to "SM16716 PWR" as many translations are way too long.

@gsimon75 Don't waste time doing this manually, I can make that change via script

gsimon75:
Thanks everyone for all the help and contributions :) !

kueblc:
Thrilled to help make this happen :)

elzershark:
SYF05 Modul

if rgbww = #ffffff...
how can I achieve that in channel 4 a 100 is written. in the the script. automatically..
if rgbww = #ffffff then... (color x,x,x,100)

gsimon75:
> how can I achieve that in channel 4 a 100 is written. in the script. automatically..
> if rgbww = #ffffff then... (color x,x,x,100)

The code of Tasmota only executes commands that are sent to it.
So if it receives an MQTT command like `cmnd/wled/color ffffff`, then it will set its channels 0..2 all to 100.
To set Channel 4, either that channel must set explicitely by the command `cmnd/wled/channel4 100`,
or such a color command must be sent that contains information for channel 4, like `cmnd/wled/color 000000ff`.

Such commands are sent by your home automation assistant software, that is (according to the video), by [ioBroker](https://github.com/ioBroker).

Maybe if you set that `wled.Color` to `#000000ff` (8 digits, not just 6), perhaps it is sent to Tasmota in this form, or you may try to set `wled.WW` to `100`, but that is also just a guess according to the [ioBroker-sonoff](https://github.com/ioBroker/ioBroker.sonoff/blob/4792135442cbd6eb3ab623b9670fe8b6169a80bb/lib/server.js#L547) module sources.

I'd like to emphasize that this is already deep in the ioBroker part of your setup, and I'm not familiar with it in any way.

But all this is still just curing the symptoms, because the real cause of your problem is that *why* is your setup trying to set the lights to white by setting  `250` to that object `test1` (that your're showing at 1:48 in your video)? It should just set `100` to `wled.Channel4`, or `000000ff` to `wled.Color`, because that is what you need.

I've tried to find this value '250' in the ioBroker-sonoff and ioBroker-iot (the Alexa-interface of ioBroker), but found nothing, so it must be somewhere in your configuration.

All I could suggest is to take a look at 'Skripte/Alexa2 Adapter', and try to find the point where it assigns that '250' to that 'test1' object, and instead of that, set `wled.Channel4`, just as you did in the 'Dummy2' script.

That's how far I could trace this in ioBroker, so I'm afraid I can't do anything more in this case. Perhaps you should ask the ioBroker developers, they are more familiar with that part, so maybe they have some more helpful ideas.

elzershark:
hi, but it has nothing to do with iobroker. If the tasmota device is placed on a simulated hue lamp, that will not be set for rgbw or rgbww bulbs. Because only the command #ffffff is sent instead of #ffffff and channal 4. if you could get that channel4, then the simulated hue would work. I know how the lamp works with iobroker and alexa.

elzershark:
mqtt not on. but hue
http://89.163.196.39/sonoff/Clipboard1.jpg
this is the Modul
http://89.163.196.39/sonoff/Clipboard2.jpg
this send tasmota when i say "make warm white" she send FFFFFF00..
this is for rgb. nor for rgbw ore rgbww
can one write that in the tasmota script not so, as soon as FFFFFF00 comes, which stands also in the channel4 100?

gsimon75:
> Because only the command #ffffff is sent instead of #ffffff and channal 4.

Exactly, that is the problem. Only '#ffffff' (3 'ff'-s) is sent, and not '#ffffffff' (4 'ff'-s).

But '#ffffff' is not the same as '#000000ff', nor as '#ffffffff', because 3 'ff'-s means that the rgb leds must be at 100% but the white ones must be off, the '000000ff' means that the rgb leds must be off and the white leds fully on, and 4 'ff'-s mean that every led must be fully on.

These are different situations, so on the Tasmota side we cannot just substitute 'ffffff' with 'ffffffff'.
(Think about the use case when someone is just gradually moving from yellow 'ffff00' towards rgb-white 'ffffff': until 'fffffe' just the rgb leds are used, everything is nice and smooth, then it reaches the rgb-white 'ffffff'. If there were such a substitution that you suggest, then at this point suddenly the white leds would turn on to 100%, turning the smooth transition into a harsh full-intensity floodlight.)

So the best would be to solve it on the sender side: if 4 'ff'-s are needed, then 4 'ff'-s should be sent.

> If the tasmota device is placed on a simulated hue lamp

I don't know about this 'simulated' lamp concept, but as far as I know, there is no such thing on the Tasmota side, neither is any scripting. This code runs on the bulb itself, so it tries to be minimalistic, it only receives the commands and sets the led intensities.

Any scripting (like that example 'Dummy2' that you showed) is done on the ioBroker side.

> *hue* lamp

Now, this reminds me of a limitation, but that is the limitation of the HSB color space concept, so it's unlikely that we can do anything about it.

If you set the color by its hue+saturation+brightness components (`cmnd/wled/hsb hue, sat, bri`), then it controls only the rgb leds, and not the white ones. [source code](https://github.com/arendst/Sonoff-Tasmota/blob/1bd3bcc9be383d1646b6bc70671d8d66ee05eabd/sonoff/xdrv_04_light.ino#L1112)

Theoretically it would be possible to use the white leds, but then we needed to know how much brighter is the 'cold-white' than the 'rgb-white', which is different from one model to another, and we don't know it.
(And when the 'warm-white' enters the picture with the rgbwc bulbs, the equations start getting really ugly.)

So, if this 'simulated hue lamp' of yours is actually sending HSB colors (and not RGBW ones), then unfortunately the answer is no: At this point there is no way to control the white leds via HSB colors, and a lot of concept-level thinking is needed even to considering how it could be done.

elzershark:
ah, ok.. :(

qingz2004:
How to set up the Walmart Merkury Innovations bulb? It uses TYLC2 and SM726EB.

kueblc:
Set module to SYF05 and then set GPIO13 to SM16716 PWR

qingz2004:
@kueblc Works great! Thanks!

jtroberts1:
I've noticed that the pre-compiled versions haven't been updated since December.  So I assume to use the enhancements discussed here I need to compile in PlatformIO, is that correct? I just compiled a new bin file based on the development branch.  Hope I'm doing this right, I've just never used the DEV branch before.  I should be able to test tomorrow.  Thanks again.

andrethomas:
@jtroberts1 pre-compiled development binaries are updated all the time to http://thehackbox.org/tasmota/ so if its something you need from development that was not included in the last release (https://github.com/arendst/Sonoff-Tasmota/releases/tag/v6.4.1) you can download or OTA from the first mentioned link.

Jason2866:
@gsimon75  Can you test PR #5702 please? Thx!

@elzershark  You can test PR #5702  Color control is implemented via HueEmulation (Alexa)
AND driver is refactored (solving rounding issues)

PR #4998, opened 2019-01-21
extra char size in mqtt_host

fzoccara:
In case of nead of some extra char size in mqtt_host configuration. :)

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -199,7 +199,7 @@ struct SYSCFG {
 
   byte          free_1D5[20];              // 1D5  Free since 5.12.0e
 
-  char          mqtt_host[33];             // 1E9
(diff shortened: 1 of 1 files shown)


ascillato:
Hi, this change is backwards incompatible. Sorry. If someone updates using this, their config will be scrambled.

ascillato2:
Sorry, it can't be merged.

Anyway, thanks for sharing your ideas.

fzoccara:
You are 100% right! Better to generate a shorter url for mqtt broker instead!
Sorry for not having thinking about this incompatibility before.. :)
Best.
Francesco.

andrethomas:
Another thing to consider is just using the ip address - Depending on a DNS name adds running the risk of not resolving due to DNS servers being down whereas this could have been avoided by just using the ip address... just a thought.

PR #5022, opened 2019-01-25
Make provision to store experimental drivers

andrethomas:
Create a new folder named experimental so that we can place experimental drivers used for testing or other use cases which are not included in the development codebase but may be useful to some users who like to self compile and tinker with drivers that were created for debugging purposes but will never be merged into the compiled codebase releases.

### DIFF
--- a/experimental/xdrv_91_WiFi_setOutputPower.ino
+++ b/experimental/xdrv_91_WiFi_setOutputPower.ino
@@ -1,0 +1,70 @@
+/*
+
+  EXPERIMENTAL DRIVER - PROVIDED FOR TINKERERS BUT NO SUPPORT PROVIDED
+
+  How to use: Simply copy file into sonoff folder and compile - The driver will then be included automatically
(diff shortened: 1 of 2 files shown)


arendst:
That's why I introduced the sonoff-tasmota-playground repository.

Once in sync with this repository it could be used for your drivers.

Now if we only could get that repository in sync with this one...

andrethomas:
Hmm, good idea. Let me think about how to do that - do I have write access to it with my other userid?

andrethomas:
It would overwrite any changes made to existing files though...

arendst:
I did sent an inviite some minutes ago...

arendst:
The current state is as of october so no problem updating it now once.

andrethomas:
Updated

andrethomas:
Now need to find a way to automate it... closing this one for now

andrethomas:
Think I am able to sync it now - just did and it worked.

PR #5030, opened 2019-01-26
added Board definition for the RGBWW controller

pljakobs:
Controler development is here: https://github.com/pljakobs/esp_rgbww_controller/tree/v2.3
This should add the correct definitions for this controller.


arendst:
I'll add them to the optionals section as I don't expect a large user base.

pljakobs:
probably not compared to the industrially manufactured products, no.

thanks!

pj

PR #5064, opened 2019-01-29
Setting cleanup -- remove deprecated eeprom API

hhaim:
Remove unused EEPROM layer from Setting code 


arendst:
You couldn't know this feature was added a few weeks ago. It needed further attention but while working on it it became a lot like the already available core SPIFFS version.

For now it stays.

PR #5088, opened 2019-02-01
MHZ19: Allow enable and disable of ABC

andrethomas:
Add commands to Sensor15 (MHZ19(B)) to allow setting of automatic baseline correction (ABC) during runtime.

It is enabled by default and may be changed with Sensor15 command as follows:

```
Sensor15 0 // Disable ABC
Sensor15 1 // Enable ABC
```

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -163,6 +163,20 @@ typedef union {
   };
 } Mcp230xxCfg;
 
+typedef union {
+  uint8_t data;
+  struct {
+    uint8_t spare0 : 1;
+    uint8_t spare1 : 1;
+    uint8_t spare2 : 1;
+    uint8_t spare3 : 1;
+    uint8_t spare4 : 1;
+    uint8_t spare5 : 1;
+    uint8_t spare6 : 1;
+    uint8_t mhz19b_abc_disable : 1;        // Disable ABC (Automatic Baseline Correction for MHZ19(B) (0 = Enabled (default), 1 = Disabled with Sensor15 command)
+  };
+} SensorCfg1;
+    
 /*
 struct SYSCFG {
   unsigned long cfg_holder;                // 000 Pre v6 header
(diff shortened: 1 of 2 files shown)


Mich-b:
@andrethomas thanks for your work. It seems from the code the default behaviour is that ABC is disabled instead of enabled, right? The setting in settings.h is:

`uint8_t mhz19b_abc_disable : 1;  `

And this is used at line 239 of the ino file. For me it’s fine to have it disabled, I just wanted to confirm since the docs say ABC is enabled by default.

Mich-b:
@andrethomas the docs and his pull request say the default behaviour is that ABC is enabled. However looking at line 240 of https://github.com/arendst/Tasmota/blob/development/tasmota/xsns_15_mhz19.ino and 218 of https://github.com/arendst/Tasmota/blob/development/tasmota/settings.h it seems as if ABC is _disabled_ by default.

For me that’s fine, just wanted to confirm that my interpretation is correct.

Mich-b:
@andrethomas 

Thanks! I see, that makes sense indeed. I did not realize `mhz19b_abc_disable` would always be 0 by default on boot. 

That leaves me wondering, I could issue the command `Sensor15 0` to disable ABC, but I guess I would have to automate this to be sure ABC would not accidentally be enabled again in case the device would restart, right? 

To avoid this, could I do the following to have a configuration variable in user_config_override.h which I can use to disable ABC by default?

**settings.ino:**
`Settings.SensorBits1.mhz19b_abc_disable= ABC_DISABLED;`

**user_config_override.h**
```
#undef  ABC_DISABLED
#define ABC_DISABLED          1
```

Guess that should work? Since Tasmota is new for me and since this pull request indicates you have thought this over, I wanted to check if there are better ways to achieve this.

andrethomas:
It will be 0 on first boot on a new flash when all settings or the entire flash has been erased. Once configured to either 0 or 1 this will be stored with all the other settings in flash so it only needs to be set once (which ever option you choose) and it will retain that setting unless you change it again with the `Sensor 15` command.

So it's a persistent setting like all the other settings in Tasmota.

Mich-b:
That clarifies things, thanks!

PR #5093, opened 2019-02-02
Add support to EFUN, a Tuya-based smart plug.

t0ny-peng:
EFUN only uses three pins:

GPIO0 => GPIO_LED2_INV  
GPIO13 => GPIO_KEY1  
GPIO14 => GPIO_REL1

This smart plug is well sealed and doesn't have any other available pin exposed to the user.

### DIFF
--- a/sonoff/sonoff_template.h
+++ b/sonoff/sonoff_template.h
@@ -281,6 +281,7 @@ enum SupportedModules {
   APLIC_WDP303075,
   TUYA_DIMMER,
   GOSUND,
+  EFUN,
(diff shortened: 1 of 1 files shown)


ascillato:
Hi,

As this plug can be configured using generic and that the device list is too long, could you please add this configuration to the wiki for other users? 

Thanks.

arendst:
This PR will invalidate current module numbering as the new module is not inserted at the end of the list.

As the GPIO's can easily be added using module Generic I agree with @ascillato to add this device to the wiki with the correct GPIO information.

PR #5105, opened 2019-02-04
decode-status.py update to 6.4.1.14

jziolkowski:
added descriptions for known options from range 32..49
rewritten the function parsing the setoptions registers
did some cleanup and made the code bit more pythonic


jziolkowski:
Great, couldn't find the descriptions myself.

PR #5122, opened 2019-02-05
Enhencement: Add new compare operators ("==", "!=" ,">=" and "<=") for rule

laurentdong:
Introduce new compare operators for rules and did some optimization as well.
While still keep the original "=" operator, the new introduced "==" operator act as a real number comparison instead of the original "=" operator which is doing string comparison that is result in FALSE for "1 + 1 = 2". For example:
rule1 on event#test do backlog var1 1;add1 1; event CompareWith2=2 endon on event#CompareWith2=%var1% do ledpower on endon
ledpower off
rule1 on
event test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ build
 .vscode/.browse.c_cpp.db*
 .vscode/c_cpp_properties.json
 .vscode/launch.json
+*.bak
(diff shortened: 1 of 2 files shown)


ascillato:
Hi, sorry, duplicated issue. This was already requested and dismissed. Sorry.

PR #5133, opened 2019-02-06
Make free pins available for user on WAGA life module

netpok:
This commit changes the accessible free pins to GPIO_USER to be used with sensors or external switches.


Jason2866:
Very bad idea. It is dangerous.
All powering metering devices have similar circuits....
So same thing as for POW
https://github.com/arendst/Sonoff-Tasmota/wiki/Sonoff-Pow-and-Pow-R2

netpok:
As someone who "works" with ac for quite some time I don't really think that it's a very bad idea, of course you don't want to connect a metal shielded MHZ19B but a simple IR transmitter or receiver should not be a problem. Using them as external switches should be even less problem, the same principle is applied in Shelly2. But if this keeps the fires and electrocutions in bay then I have no problem with your standpoint.

ascillato:
We would be irresponsable if we let that pins available for people who don't know in deep how to use mains AC. Don't take it personal. If you know, great! But most people don't. So, that is why this is explained in the wiki with a WARNING message.

PR #5134, opened 2019-02-06
Make free pins available for user on DIGOO SP-202 module

netpok:
This commit changes the accessible free pins to GPIO_USER to be used with sensors or external switches.


Jason2866:
Very bad idea. It is dangerous.
All powering metering devices have similar circuits....
So same thing as for POW
https://github.com/arendst/Sonoff-Tasmota/wiki/Sonoff-Pow-and-Pow-R2

PR #5162, opened 2019-02-07
PN532: Change from I2C to HSU (Serial)

andrethomas:
The current implementation of I2C is inefficient and slow in comparison to using HSU (Serial) so this PR removes I2C support in favour of using the device in HSU (High-Speed UART / Serial) mode.

This is after some testing was done and I2C was found to be somewhat unpredictable at times, possibly because if the slowness of the I2C support provided by the PN532 or perhaps related to the I2C implementation in the ESP8266 Arduino core - Either way, the Serial interface appears stable and runs at a lower load average than the initial I2C implementation.


arendst:
This is indeed a lot more stable. I lost the I2C comms after some hours. This one still works.

andrethomas:
Yup, same experience... and it does not need hardware serial since there is only data transmission when the driver actually transacts with the PN532.

Jason2866:
It does not work for me!
Your fork does

andrethomas:
@Jason2866 Will chat on discord as I've removed the original branch which you were testing on so that I can get a copy of the driver file you are using to compare what I changed but its working fine for me in its current form from the development branch.

Jason2866:
The funny change of RX and TX did it :-)

andrethomas:
Yeah @arendst was cleaning up my messy code :)

qingz2004:
It does not work for me.
It used to work with I2C mode with old firmware.
Changed #define USE_PN532_I2C to #define USE_PN532_HSU in my user_config_override.h file and compiled the new code.
Changed the dip switches to "00". Flashed new firmware. Changed I2C SDA to PN532 TX and change I2C SCL to PN532 Rx in configure menu.

After reboot, it does not detect any of my NFC tags.

Jason2866:
Swap tx and rx in config

qingz2004:
@Jason2866 That works. Thanks!
It's a bug or intended?

Jason2866:
Theo explicit changed it, so as intended

PR #5175, opened 2019-02-09
Added smart plug with power metering and rgb night light

blakadder:
Don't know if its worth to include this in the device list but it is a specific product that has power monitoring and an RGB light.


arendst:
Updating the wiki with this information would be enough.

blakadder:
Soooooon...

PR #5183, opened 2019-02-10
[Please ignore this, I have new pull request] Introduce Expression in Rules

laurentdong:
The expression can be used in 3 commands for now:
VAR<x>, MEM<x> and RULETIMER<x>
The format is same with before but instead of only allow constant parameter, we are now supporting using an expression as parameter. For example:
VAR3 3.14 * (MEM1 * (10 + VAR2 ^2) - 100) % 10 + TIME / (2 + MEM2)
The expression can include constants (integer or float number), variables (var, mem, time, uptime, sunrise, sunset), operators and round brackets.
Here is a list of 6 math operators we supported for now, by the order of priority from Higher to Lower:
"^"  power
"%"  modulo
"*" and "/"
"+" and "-"
The space between different element is not mandatory and variables are case insensitive.  You are allowed to have expression like: Uptime+120-Mem3/2
Please do NOT use % to bracket the variable name, like %Var2% because it will be replaced with its content before the command running. So the previous result will not affect the followed command in backlog.


arendst:
Just wondering, what is the impaact on code and memory size?

localhost61:
@laurentdong 
So you drop %timestamp% ? Because I was interested to publish the duration between 2 timestamps (From Wifi#Connected to Wifi#Disconnected), but I think it'll be more efficient to hack the code than run a rule for that.

laurentdong:
@arendst Added 3392 Bytes total.

laurentdong:
@localhost61 No. You still can use %timestamp% like before. Just the expression can not support timestamp variable because it is a string. I would like to add some new variables (constant) like utctime and localtime.

Jason2866:
Just my 2 cents...  
3K code "just" for expression in rules (rules function needs now about 4k5)?
Imho this stuff should be done in a Home Automation system.
Maybe adding it optional like sunrise calculation (#define USE_SUNRISE) and default is off.

meingraham:
@Jason2866,

USE_RULES?

Rules are critical. And I know that you're not proposing taking this capability completely away. I just want to make the case for how valuable they are. Although I use a home automation hub, there are some scenarios where I depend on Rules to make an implementation more robust/foolproof. Some of this is that the device knows exactly what state it's in and thus it's more certain as to what the logic must perform. Some of it that the communication lag makes the hub less than ideal for some controls. There are many boot-up scenarios where Rules are critical. I also use device level logic to provide backup to functionality if there are issues with my hub (or, more specifically, the connection to Google Assistant) where the logic on the device allows me to provide an RF remote and still retain the automation. This is more of an issue for the rest of my family and for guests.

Some of my Rules are complex and difficult to maintain or remember how they work when I need to touch them after many months. A good portion of this is because they are coded to have to handle GE/LE/NE with lots of "extra" Rule segments.

So, I'm in favor of adding GE/LE/NE. It will simplify many Rules and make Rules more user friendly. Yes, I know the old saying, "If it's hard to write, it ought to be hard to use!". Of course that's a bit of programmers' humor. What we all know is that we write software for use by others to help them do something they can't do themselves and we try make it so that it is as easy/simple as possible, weighing the cost/benefits, of course.

What we need to consider is the benefit of this new feature vs. the program memory cost (P.S., perhaps there are efficiencies to be gained by detailed peer review of the code?). We know the cost. It is the benefits, like those to users that I mentioned. Making it easier to understand and use is a great benefit. But there are additional benefits to making Rules easier. I'm referring the the numerous support requests specifically around Rules. No, this won't eliminate all the questions. And yes, there are always users who will not RTFM. But I have seen lots of Rules related queries. If simplifying expressions reduces some of that, then that's a great benefit.

My $0.01 (my opinions are not worth full value) :stuck_out_tongue_winking_eye: 

Respectfully,

Mie

laurentdong:
@Jason2866 Actually my implement of expression is pretty small and clear. The reason that used 3K code space is because the new introduced library - LinkedList which is implemented pretty small as well but anyway, it is a new data type. I think we might can optimize some other codes with this new introduced type - LinkedList? Even the expression can be used by some other places other than rules too?
I knew while some people are likely to implement major logic on local and some might want to put them on server. So if the code space is a concert, I think we'd better have a new switch like #define USE_EPXRESSION to control it? @arendst

Jason2866:
@meingraham @laurentdong  I use rules a lot. Same reason than yours probably.
> What we need to consider is the benefit of this new feature vs. the program memory cost
That is the question!
I would like (as i said in my comment) a switch to choose 
> #define USE_EPXRESSION
Thats it. No different view at all :-)

meingraham:
@Jason2866 

My concern is that `USE_EXPRESSION` to only affect GE/LE/NE expressions yet allow GT/LT/EQ would be error prone. And then the question that a noob might ponder "is my trigger statement an expression?" or any of a countless number of "what is an expression?" questions.

Mike

laurentdong:
@meingraham The GE/LE/NE I implemented recently does not require the expression and actually does not support expression for now. 
If you disable expression by #undef USE_EXPRESSION, you will have to use a constant (Of course, still can use %<Variable>% which is not affected by expression as well) as the parameter for VAR, MEM and Ruletimer command.
BTW, I'm planing to add the expression into EVENT command but I do not like to support expression in event trigger like (ON EVENT#TEST=?) because it will be called very frequently and slow down the whole system.

meingraham:
@laurentdong,

> because it will be called very frequently and slow down the whole system.

The overhead of using these comparisons will cause slower code execution?

I thought the `Event` **command** only **_assigned_** values? Isn't it the `Event#` **trigger** that uses **_comparisons_**? I could easily be wrong. Then, if I'm not mistaken, expressions are on the trigger side of Rules, i.e., `ON trigger DO command ENDON`. If so, and a compile time #DEFINE is the direction this is going, can these new comparisons be called something other than expressions? EVENT_COMPARISONS?

So, to clarify these new comparisons would **_not_** be available for these triggers:
- [SensorName]#[ValueName]
- Tele-[SensorName]#[ValueName]
- Time#Initialized
- Time#Minute
- MEMx#State
- VARx#State

If that's the case, the cost of adding these comparisons for "limited" use as you are proposing seems to have limited benefits. And then there's the added confusion of that they'll try to be used for all triggers and the support queries are likely to escalate.

Maybe, if it's a compile decision, make these new comparisons be available to **_all_** triggers... and a user can enable or disable for all trigger expressions?

I would concur that the addition of the LinkedList library be assessed for use by other parts of the code as you proposed earlier. Perhaps there is an opportunity for a large portion of the ~3K added by this comparison feature to be offset by code size reductions/efficiencies in other modules.

Mike

andrethomas:
This is an interesting addition but I have reservations as others above have expressed with regard to program flash and ram utilisation.

I also think one needs to decide where the boundary line is between stand-alone use (within rules etc on the device itself to operate autonomously) and those functions which should rightfully be handled by a home automation solution.

I believe a good question to answer would be if this necessitates implementation for autonomous use, which if yes, then it could be in contrast with what the firmware tries to achieve measured against how easily operations such as these may be implemented on the home automation side.

Is there a generic use case scenario (or preferably a number of them) which would be convincing enough to trade off program and ram on the device vs. implementing on a home automation platform?

@arendst is continuously mitigating resource limitations and the last rework on the web ui to do just that should not be traded off against features that are in contrast the project's objective... my opinion of course as @arendst may view it differently.

That being said if the enhancement is only included conditionally by a compiler directive (#define in my_user_config.h) then it immediately sidelines what I believe to be the majority of users who just download released or pre-compiled binaries for a really simple reason - self-compiling is too daunting or complex or overkill for what they need to do... would having this addition make things any easier for such users? I doubt it.

laurentdong:
@meingraham All the trigger are been checked every 50ms, so evaluate the expression during the comparison might affect the performance a lot unless the expression is very simple. But supporting expression in EVENT command should not bring too much affection. So this is my next plan.

meingraham:
@laurentdong,

Can you give an example ofan Event Command using these comparison directives?

laurentdong:
@meingraham I'm planing to make the EVENT command can accept expression as parameter too. For example, you can use:
EVENT TEST=VAR1+(MEM1 * 10)
to trigger an event TEST. The expression will be evaluated when execute the EVENT command. Then you can have a trigger to do the comparison like:
ON EVENT#TEST>=100 DO ....
Like what I said before, I don't want to let the event checker do expression evaluation. So this will NOT work:
ON EVENT#TEST>=VAR2 + (MEM1 * 2)

andrethomas:
And what would you put in VAR1 and MEM1 ?

ascillato:
@laurentdong

Hi,

Please, clean this PR. Now it has changes not related to what you want to add. In the actual form can not be merged. It has now 65 files with changes.

meingraham:
So, the new _**comparisons**_ ARE on the `Trigger`, not on the `Command`. The _**expression**_ is on the `Command`. Thus, is this enhancement to allow expressions such as `VAR1+(MEM1 * 10)` on the command? I don't have any Rules with this kind of calculation so I don't know if this is the enhancement you are adding. Or if this kind of expressions capability is already implemented? If your PR is indeed adding the capability to use expressions with the `Event` `Command`, then I would propose separating the `Command` enhancement from the `Trigger` enhancement. Is that possible? That is, `USE_EVENT_EXPRESSIONS` (note the distinction to include "**EVENT**" so it is clearer what the compile parameter does) will add capability to the `Event` `Command`. However, the ability to use ">="/"<="/"!=" would **_NOT_** be dependent on `USE_EVENT_EXPRESSIONS` - these comparisons would be available "always". Possible?

Thanks.

Mike

andrethomas2 (changes requested):
Needs cleanup to allow testing against a branch which is up to date with development.

laurentdong:
@arendst @ascillato Sorry. I'm pretty new on github and still learning how to play with it. For this new feature, I created the pull request then try to merge code from base and finally messed up everything :(
 OK. I'm going to create a new fork and create a new pull request to include everything.

laurentdong:
@meingraham No. I did not implement expression support for both EVENT and TRIGGER yet. I said this in on my plan.
The enhancement for comparison (GE, LE, NE, EE) has no dependency with expression and yes, it is always on if you enabled rules.

meingraham:
So, to confirm...

PR 5122 is a standalone enhancement which adds GE/LE/NE/EE. This has already been merged and is available in 6.4.1.14.

Your new PR 5210 (which replaces this PR - 5183) is the enhancement to add expressions to the Event Command... and is not related to nor dependent upon the trigger comparisons your introduced in 5122.

Did I finally get this straight in my mind?

Thanks.

Mike

andrethomas2:
To be continued at https://github.com/arendst/Sonoff-Tasmota/pull/5210

andrethomas2:
Just a reminder for those who made comments on this PR to follow the new PR as you will not be automatically notified unless you comment or subscribe to https://github.com/arendst/Sonoff-Tasmota/pull/5210

Issue #5151, opened 2019-02-07
Color/PWM inconsistency issue collection


armageddon421:
**Describe the bug**

There is no way to accurately set RGB/RGBW values, nor is there a way to set multiple PWM values at once to achieve a desired result. I've collected some unresolved but closed issues describing the various problems.

- [x] _Searched the problem in issues and in the wiki_
- [x] _Hardware used_ : irrelevant, all hardware is affected
- [ ] _Development/Compiler/Upload tools used_ :
- [x] _If a pre-compiled release or development binary was used, which one?_ : latest sonoff_DE from releases
- [x] _You have tried latest release or development binaries?_ : yes
- [ ] _Provide the output of command_``status 0`` : please see below issues for details

**To Reproduce**
repeatedly set channel 2, other channels decrease each time
issue: #4558 
```
'stat.dhaled3.color':'F8007D'
'cmnd.dhaled3.channel2':'98'
'stat.dhaled3.color':'F6F97C'
'cmnd.dhaled3.channel2':'98'
'stat.dhaled3.color':'F5F97B'
...
```

issue: #4901 
Delay between backlog commands of 50ms is definitely visible on an RGB or RGBW stripe, so there's no way to set multiple PWM channels at the same time in setOption15 0 mode.
Also, using that mode would completely lose any ability to dim or fade, so it's not really an option.

issue: #4797 
describes the accuracy error due to "rounding from 255 to 100 to 255", to reproduce

> transmit via MQTT the color (201,204,0) the console shows
> `{"POWER":"ON","Dimmer":80,"Color":"200,204,0","HSBColor":"61,100,80","Channel":[78,80,0]}`

duplicate issue: #4803 
> Behaviour seen on H801.
> Start with all channels 0.
> Set 1st channel by "channel1 30" -> response includes "Channel":[28,0,0,0,0]
> Set 2nd channel by "channel2 30" -> response includes "Channel":[27,28,0,0,0]
Here you can see both problems. The value of 30 is not accuarately set, and channel1 is influenced by setting channel2

**Expected behavior**
There should be a consistent way to reliably and accurately set the RGB(W) values with full resolution of 0 to 255, without influencing other channels in case only one channel was changed.

**Additional context**
- One use case is where I want to use the separate channels for separate white lights, so only one might be changed at a time, while still using fading (hence not using legacy PWM mode). The channels should be independent and not be influenced by the other channels, especially not continually decreasing.
[...]

PR #5207, opened 2019-02-12
fixed color/dimmer/channel inconsistencies

armageddon421:
PR related to #5151 

I think I was able to fix everything.
Steps taken:
- Not touching user-supplied color value, storing in Settings as is
- calculate new dimmer value from highest color channel
- dimmer is now 255 resolution internally instead of a percentage (commands still work the same)
- when the dimmer value changes, "normalized" to 255 color is calculated and dimmer applied to that.
- replaced some floats by uint16_t to save memory and performance while offering precise and predictable calculations
- made sure forward/backward mapping of 100 and 255 range for the user interface is consistent (related to the result always being one less issue)

The only remaining inconsistency is that if you set a color and at a later point set the dimmer value to the same dimmer value that was calculated from that color, it can still be a little bit off. That's just inherently caused by the percentage resolution of the dimmer.

I tested it with an RGBW stripe and couldn't find any problems using the commands color, dimmer, channelX and the fading settings. Shouldn't have affected any of the animations, since the way light_current_color works hasn't changed at all. I tried setting up a dummy device with only 2 channels, simulating Warm/Cold white devices, but I can't really verify it everything's working properly with that, so maybe someone should check.

From a performance and memory standpoint, I think, if anything, I probably improved the situation by getting rid of some floats. The only increase is that when changing the dimmer, the highest channel value has to be determined, which shouldn't be much of an impact.

### DIFF
--- a/sonoff/xdrv_04_light.ino
+++ b/sonoff/xdrv_04_light.ino
@@ -627,7 +627,7 @@ void LightSetDimmer(uint8_t myDimmer)
   }
 
   //determine highest color to premultiply dimmer with
-  uint8_t highest = 0;
+  uint8_t highest = 1;
   for (uint8_t i = 0; i < light_subtype; i++) {
     if (highest < Settings.light_color[i]) {
       highest = Settings.light_color[i];
--- a/sonoff/xdrv_04_light.ino
+++ b/sonoff/xdrv_04_light.ino
@@ -612,26 +612,34 @@ uint16_t LightGetColorTemp(void)
 
 void LightSetDimmer(uint8_t myDimmer)
 {
-  float temp;
+  uint16_t temp;
 
   if (PHILIPS == my_module_type) {
     // Xiaomi Philips bulbs use two PWM channels with a different scheme:
-    float dimmer = 100 / (float)myDimmer;
-    temp = (float)Settings.light_color[0] / dimmer; // channel 1 is intensity
+    temp = (uint16_t)Settings.light_color[0] * myDimmer / 255; // channel 1 is intensity
     light_current_color[0] = (uint8_t)temp;
-    temp = (float)Settings.light_color[1];          // channel 2 is temperature
+    temp = (uint16_t)Settings.light_color[1];          // channel 2 is temperature
     light_current_color[1] = (uint8_t)temp;
     return;
   }
   if (LT_PWM1 == light_type) {
     Settings.light_color[0] = 255;    // One PWM channel only supports Dimmer but needs max color
   }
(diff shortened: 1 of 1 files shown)


armageddon421:
found a divisoin by zero error, fix is easy, chance line 630 to default to 1 instead of 0.

arendst:
Does this still work with Hue emulation enabled?

armageddon421:
> Does this still work with Hue emulation enabled?

I'm not sure exactly what you mean by hue emulation. If you mean emulating the philips hue bridge, I don't know why it wouldn't. Color values were 0 to 255 before, and the if (PHILIPS == my_module_type) sections seem to only access that directly. The only occasion where the dimmer is used is properly updated to use the new 0 to 255 scale, for example line 619. I didn't find any other special treatment.

If you mean the HSB implementation, it should also be updated in the relevant spot, which is only lines 1134, 1142 and 1157, where the dimmer value is updated according to the HSB brightness.

arendst:
At least LightSetHsb and LightGetHsb are bing used by xplg_wemohue.ino so it's functionality should still be guaranteed by this update.

armageddon421:
I only quickly checked if the hsbcolor command works and it seemed plausible. If you want, I can do some more tests later. Maybe someone else should also try to play around with the commands and look for any problems, just in case I forgot to try something out. I also only tried with 3 and 4 channel devices, as I have no warm/cold white lights here nor any 5 channel setup. Could just add/remove PWM channels in the generic device config, but I wouldn't be able to see the actual resulting light properly.
I deployed to two devices already and had no issues so far.

arendst:
Needs more testing before implementing

armageddon421:
Please let me know if and how I can help.

blakadder:
Do you need testing on a RGBCCT strip or do you need to test the hue compatibility? i can provide RGBCCT testing

sillyfrog:
I have been using this branch intensively with 4 channels and it's been working great. I would love to see this merged in also.

s-hadinger:
Complete refactor of lights management in PR #5702, this should supersede this PR.

New version proposes two interfaces: Dimmer (0..100) or Brightness (0..255). Use whatever you need.

Jason2866:
Please close this PR since it is outdated and obsolet. Thx for your work and ideas which where used in
https://github.com/arendst/Sonoff-Tasmota/commit/c0a4e08f4c184c195515e84d4c261628b1835d00

ascillato2:
Closing this PR as the feature has been added.

Thanks a lot for all your contributions. Very appreciated :+1:

PR #5210, opened 2019-02-13
Introduce Expression in Rules - Updated

laurentdong:
Support use an expression as paramter in some rule commands, include Var<x>, Mem<x> and Ruletimer<x>.
The format is same with before but instead of only allow constant parameter, you are now allowed to use an expression as parameter. For example:
VAR3 3.14 * (MEM1 * (10 + VAR2 ^2) - 100) % 10 + TIME / (2 + MEM2)
The expression can include constants (integer or float number), variables (var, mem, time, uptime, sunrise, sunset), operators and round brackets.
Here is a list of 6 math operators we supported for now, by the order of priority from Higher to Lower:
"^" power
"%" modulo
"*" and "/"
"+" and "-"
The space between different element is not mandatory and variables are case insensitive. You are allowed to have expression like: Uptime+120-Mem3/2
You are still able to use variable placeholder, like %var1%, %sunrise% in rules. But please remember those placeholder will be replaced with its content before the command running. So the previous result will not affect the followed command which is different than expression.

Update:
You can comment "#define USE_EXPRESSION" in my_user_config.h to disable this feature and save 3k code space.

### DIFF
--- a/lib/LinkedList-1.2.3/LICENSE.txt
+++ b/lib/LinkedList-1.2.3/LICENSE.txt
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Ivan Seidel
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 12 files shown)


laurentdong:
> 
> 
> So, to confirm...
> 
> PR 5122 is a standalone enhancement which adds GE/LE/NE/EE. This has already been merged and is available in 6.4.1.14.
> 
> Your new PR 5210 (which replaces this PR - 5183) is the enhancement to add expressions to the Event Command... and is not related to nor dependent upon the trigger comparisons your introduced in 5122.
> 
> Did I finally get this straight in my mind?
> 
> Thanks.
> 
> Mike

@meingraham Almost correct! My new enhancement of expression do not support Event command for now, which only support 3 commands: VAR, MEM and Ruletimer

andrethomas2 (approved):
Code looks fine - up to @arendst if he wants to enable it by default or not.

PR #5227, opened 2019-02-14
New on this

Danifont:
HI !
I'm new on this ... I'm just configuring, and using sonoff dual and TH modules with tasmota, and the world is great and all is fun.
Recently I bought a sonoff T1 UK 3 buttons and begin the nightmare.

I configure it and test an all seems to be good. I use your pictures to  locate the correct wires.

But now, when I configure the appropriate  module type, it restart and after a few moment restart with the configuration cleared to Sonoff basic module.

Any idea? I see it's the same circuit board than you show in the pictures. 

Any new version on tasmota that can I use for this ?

I use the last tasmota  6.4.1(sonoff)-2_4_2
Where do you find the version  6.4.1.5
It works with this board ?

Thanks and regards !


ascillato2:
Invalid pull request. Closing.

For support, please use the Tasmota Support Chat. Thanks

PR #5242, opened 2019-02-16
Remove wifi password restriction

spock64:
(Edit to remove refs to android!)

This PR removes the restriction where an asterisk can't appear in a wifi password.

The changes provide a prompt of four asterisks ("****") in the web form.
If the user changes this value, the value is stored. If the user does not change this value, nothing new is stored.

The limitation imposed by this change is that a wifi password consisting of 4 asterisks will be ignored. There is no impact of this restriction as WPA2 passwords must be of 8 octets or more.

I have tested this using Arduino core 2.5, and have separately verified that there is no issue with asterisks in passwords using android core 2.3.

Note: In i18n.h, I have left the definition of D_ASTERIX as is as it used in the mqtt module. Instead I have defined D_ASTERISK_PWD to be the constant used.

Side note: ASTERIX is a French comic book character. ASTERISK is the name of a character in ISO 646 and many other character sets !

(This is the first time I have made a public PR - please let me have any feedback and I can resubmit if needed)

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -419,6 +419,7 @@
 /********************************************************************************************/
 
 #define D_ASTERIX "********"
+#define D_ASTERISK_PWD "****"
 
 #ifndef MY_LANGUAGE
   #include "language/en-GB.h"
--- a/sonoff/xdrv_01_webserver.ino
+++ b/sonoff/xdrv_01_webserver.ino
@@ -280,9 +280,9 @@ const char HTTP_FORM_WIFI[] PROGMEM =
   "<fieldset><legend><b>&nbsp;" D_WIFI_PARAMETERS "&nbsp;</b></legend>"
   "<form method='get' action='wi'>"
   "<p><b>" D_AP1_SSID "</b> (" STA_SSID1 ")<br/><input id='s1' name='s1' placeholder='" STA_SSID1 "' value='{s1'></p>"
-  "<p><b>" D_AP1_PASSWORD "</b><br/><input id='p1' name='p1' type='password' placeholder='" D_AP1_PASSWORD "' value='" D_ASTERIX_PWD "'></p>"
+  "<p><b>" D_AP1_PASSWORD "</b><br/><input id='p1' name='p1' type='password' placeholder='" D_AP1_PASSWORD "' value='" D_ASTERISK_PWD "'></p>"
   "<p><b>" D_AP2_SSID "</b> (" STA_SSID2 ")<br/><input id='s2' name='s2' placeholder='" STA_SSID2 "' value='{s2'></p>"
-  "<p><b>" D_AP2_PASSWORD "</b><br/><input id='p2' name='p2' type='password' placeholder='" D_AP2_PASSWORD "' value='" D_ASTERIX_PWD "'></p>"
+  "<p><b>" D_AP2_PASSWORD "</b><br/><input id='p2' name='p2' type='password' placeholder='" D_AP2_PASSWORD "' value='" D_ASTERISK_PWD "'></p>"
   "<p><b>" D_HOSTNAME "</b> (" WIFI_HOSTNAME ")<br/><input id='h' name='h' placeholder='" WIFI_HOSTNAME" ' value='{h1'></p>";
 
 const char HTTP_FORM_LOG1[] PROGMEM =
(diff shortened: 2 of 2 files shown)


andrethomas:
Thx

> I have tested this using android core 2.5, and have separately verified that there is no issue with asterisks in passwords using android core 2.3.

From the download statistics we know that most users currently use 2.4.2 - Did you test on 2.4.2?

> ASTERIX is a French comic book character

:)

andrethomas:
I can confirm that this works on 2.3.0, 2.4.2 and 2.5.0

andrethomas2 (approved):

meingraham:
@andrethomas,

> From the download statistics we know that most users currently use 2.4.2

I think you can only say that the current **_downloads_** are mostly 2.4.2 Core - not that most users are currently using that Core. A lot of these downloads are probably due to the fact that the default download links (both release and development) point to the 2.4.2 Core binaries thus it's the defacto version that most users are grabbing. With all of the Wi-Fi connectivity issues with 2.4.x, one of the primary resolutions is to use the 2.3.0 Core. I think a major portion of users are probably on that core both from an installed base and due to the instability issues associated with 2.4.x. Hopefully 2.5.0 will alleviate some of the fragmentation and slowly bring users onto this new Core.

But your point is taken, the PR needs to be tested against that version of the Core since it is being used.

Mike

arendst:
Reviewed and Merged!

I love these tiny PR's providing so much joy!!

andrethomas:
> But your point is taken, the PR needs to be tested against that version of the Core since it is being used.

Someone should probably test all the other special characters so we can have a definitive list of which ones are not supported - The information that some special characters are not supported is based on an observation of what some users experienced in the past.

I agree that 2.4.2 is the current defacto download pointer but it may remain like that for some time.

I attempted to upgrade my 66+ devices with 2.5.0 core code and three of my units are giving a strange result insofar that they connect perfectly to wifi and work as expected but they refuse to connect to the mqtt server (as if the host cannot be reached) which is configured as a static ip address on my network so its not DNS related but something else going on there... the same binary works fine on same models of device elsewhere so its not device specific either.

spock64:
> Someone should probably test all the other special characters so we can have a definitive list of which ones are not supported - The information that some special characters are not supported is based on an observation of what some users experienced in the past.

@andrethomas I searched the issues in the Arduino core repo and could only find a mention of underscore (_) being an issue at one point. The main challenge is that any octet is a valid one for a Wi-Fi password - limitations get imposed as a consequence of the UIs provided by the Wi-Fi router vendors. (For example the following hex string - in binary - would be a valid password ... 0011223344AABBCCDDEEFF - and that’s not easy to enter in an html page!). Perhaps we should simply warn users in the documentation that some characters can be a problem rather than trying to fix it in the GUI ? If there were a definitive published list of issues from the Arduino Core guys, that would be different I guess?

joba-1:
my 2 cents on this:

There are several domains involved, you cannot control all of them.
1) It is possible that a user has no control over the wifi password: let
him enter what he has to use.
2) It is possible that wifi routers choke on some characters: ignore this,
you cannot save the world from stupid.
3) It is possible that the esp core cannot handle some characters.
  a) we know for sure this wont work: reject the password, telling why.
  b) it might be (there were some reports) it wont work: warn about this,
but in the end accept it (e.g because the user might have fixed it in his
core).
4) Not a html expert, but aren't there encodings for html forms so more or
less any input can be safely handled (I think base64, uuencode, ...)?

regarding 1) in conjunction with any 32 octets are allowed by the standard:
How about a hex-checkbox. If checked then interpret the password as up to
32 pairs of hex nibbles (0-9, a-f).



Am So., 17. Feb. 2019 um 11:27 Uhr schrieb spock64 <notifications@github.com

andrethomas:
The intention is not for the firmware to prevent special characters from being used... just that if we find them that they can be known to us on a list somewhere in the Wiki.

Adding exception handling to prevent or warn a user when a possibly unacceptable password is entered is not within the scope of this project.

I know we came across some on Discord also but it may just have been typo's from a user when performing the initial config causing failed connections which were then resolved by changing the password on both ends.

ascillato:
The Tasmota code for mqtt password had the exact same checking code than wifi password. That one should be addressed too. It is not good to have one set of rules for wifi password and different restrictions to * for mqtt.

Can you address that also? Thanks

spock64:
Sure - will get to it overnight

andrethomas:
Awesome, thanks.

PR #5250, opened 2019-02-17
Add WiFi down monitoring

localhost61:
This PR offers a way to publish MQTT messages gathering information on the period of time spent offline. The message is sent at first when MQTT connects and is refreshed every teleperiod.
`01:59:21 MQT: stat/[topic]/INFO = {"Offline":"0T00:37:08","Online":"2019-02-17T02:59:20"}`

It uses 2 new Time values (i.e. 8 bytes) that appear in the Information webpage too:
- `Offline` is a time duration that gather all offline periods since startup, it is reset when NTP first sync.
- `Online` is the timestamp of last connection. In fact the same timer holds WiFi Up/Down events but Down timestamps are not published (because MQTT is off too).

A command is provided (in the console or by URL) to reset both values to begin a fresh monitoring session or to retrieve current values without waiting for the teleperiod occurence.

Command          | Payload      | Description                                                               
-----------------|--------------|--------------------------------------------------------------------------------
OfflineStat       |              | Show current time values that describe periods spent offline
OfflineStat       | 0            | Reset both Wifi offline time values

This feature costs (448 bytes of code, 8 bytes of mem) but it can be opted out by commenting out 
`#define USE_OFFLINESTAT` in `my_user_config.h` but for Travis purpose it's let defined.


ascillato2:
Hi,

A PR without all the definition keys for all languages can't be merged due it will break the automatic compilation for development branch binaries.

So, this PR in the current form can't be merged. 

Anyway, before working further on this, please first wait for Theo's answer.

localhost61:
Hi Adrian,
Tell me what words I have to add in language files. I think it's not necessary because I intentionally reused the words `Online` and `Offline` for the web page and for MQTT it has to stay in English.
For me the job is done and deployed at home on all my 22 devices.

andrethomas:
`delay(my_activity /2);` is actually 1/3 delay.

If my_activity = 100 (for example) then the delay would be 100/2 = 50 towards a total of 150 which is 1/3. So the comment should actually read

`If wifi down and my_activity > sleep then force loop delay to 1/3 of main loop target delay`

localhost61:
@andrethomas 
Corrected. thanks !

arendst:
Why not add the information to status 5?

This alleviates the need (and code space) for yet another status command.

arendst:
BTW as Uptime is already is already available why not only log wifi downtime?

In that case add it to Status 11

localhost61:
`Online` is a timestamp, it tells you since when you got the WiFi connection, a value that can be different than LWT timestamp in case of MQTT connections issue .
`Uptime` is the duration since Startup (or Restart), it's not related to WiFi status.
`Offline` is a duration too, it collects all the WiFi offline durations since startup (or `Offline 0` command).

I agree that those data could fit in the `Status 11` message. `Status 5` seems to me more devoted to WiFi setup information, isn't it? My first thought was to put them in the TELE message but it can already be overcrowded (I already tested a device with 8 DS18B20 ;-) )
My attempt was to have a mean to monitor WiFi connection efficiency because I have some devices that have a very good RSSI >95% but frequently drop the line and others with often less than 50% than stay online. This is a recurrent issue I guess.

So what do you think if:
- I add the data in `Status 11` and drop the STAT/../INFO message
- I keep them in the /in webpage
- I let to the domotic server the job to issue the `Status 11` command to all devices to retrieve the data. and remove the self-publishing at MQTT.connected and Teleperiod.   
  
What about the `#define USE_OFFLINESTAT` in `my_user_config.h` I commit commented out or not?

arendst:
Is the timestamp when your last wifi re-connect occurred really important? Isn't the online duration not enough? Using the uptime and the online duration you'll get a good indication how bad/good the wifi connection is I think.

I just want to keep it small. Adding Online time to GUI is fine. Adding last wifi re-connect to GUI is overdone.

So I suggest
- Add ``,"Online":"0T00:11:41"`` to Status 11 just after uptime
- Add it to GUI just below uptime
- No need for define as it is always available.

andrethomas:
@arendst Can we get a down times counter since last reboot in the STATE message also? ;)

```
"Wifi":{"AP":1,"SSId":"Wireless","BSSId":"DE:AD:00:00:BE:EF","Channel":3,"RSSI":88,"LinkDown":0}
```

Or maybe as a continuous counter like the boot counter - might be useful for finding devices that are unstable.

andrethomas:
Or just on the information page

image

localhost61:
I wonder if it can happen to drop the WiFi connection between two MQTT messages without MQTT to be aware of the signal loss. Can you tell me? In that case LWT timestamp recorded by the domotic server could be enough for me and I could do without publishing the 'Online` Timestamp, but I still need the value to process the `Offline`` duration.
I agree that the LinkDown counter would be a interesting indication too.

arendst:
I'll add the linkdown after restart counter.

localhost61:
Implemented with commit [Add property Downtime](https://github.com/arendst/Sonoff-Tasmota/commit/3c3b5bb8ca1eaaa1684b6c042326a8fa2fdd9eb0)

PR #5258, opened 2019-02-17
counter minor issue

hhaim:
1. Counter is uint32_t should not be printed as double. 
2. Better to use uint32_t over long which is platform specific 32/64 bit. 
3. Minor cleanup in the vicinity

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,7 +11,7 @@
 src_dir = sonoff
 
 ; *** Uncomment one of the lines below to build/upload only one environment
-;env_default = sonoff
+env_default = sonoff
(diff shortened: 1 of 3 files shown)


andrethomas:
unsigned long is used because that is how it comes from the core - all time-related variables use unsigned long from the SDK and core afaik including millis() and micros()

I can't think of any logical reason why it needs to be changed... not to mention be type cast back and to unsigned long in the future... at least this way we know it is always an unsigned long.

hhaim:
Not sure how sdk timers related to my diff. Have a second look.

PR #5265, opened 2019-02-18
 Counter is uint32_t should not be printed as double

hhaim:
Fix the old PR 
1. Counter is uint32_t should not be printed as double.
2. Better to use uint32_t over long which is platform specific 32/64 bit.
3. Minor cleanup in the vicinity

### DIFF
--- a/sonoff/settings.h
+++ b/sonoff/settings.h
@@ -309,7 +309,7 @@ struct SYSCFG {
   unsigned long energy_kWhtotal;           // 554
   char          mqtt_fulltopic[100];       // 558
   SysBitfield2  flag2;                     // 5BC
-  unsigned long pulse_counter[MAX_COUNTERS];  // 5C0
+  uint32_t      pulse_counter[MAX_COUNTERS];  // 5C0
   uint16_t      pulse_counter_type;        // 5D0
   uint16_t      pulse_counter_debounce;    // 5D2
   uint8_t       rf_code[17][9];            // 5D4
@@ -364,7 +364,7 @@ struct RTCMEM {
   uint8_t       ota_loader;                // 293
   unsigned long energy_kWhtoday;              // 294
   unsigned long energy_kWhtotal;              // 298
-  unsigned long pulse_counter[MAX_COUNTERS];  // 29C
+  uint32_t      pulse_counter[MAX_COUNTERS];  // 29C
   power_t       power;                     // 2AC
   uint8_t       free_020[60];              // 2B0
                                            // 2EC - 2FF free locations
--- a/sonoff/xsns_01_counter.ino
+++ b/sonoff/xsns_01_counter.ino
@@ -22,6 +22,11 @@
 \*********************************************************************************************/
 
 #define XSNS_01             1
+#define GPIO_LEVEL (99)
+
+static inline bool is_gpio_cnt_low(int index){
+   return ((pin[GPIO_CNTR1 + index] < GPIO_LEVEL)?true:false);
+}
 
 unsigned long last_counter_timer[MAX_COUNTERS]; // Last counter time in micro seconds
 
@@ -66,7 +71,7 @@ void CounterUpdate4(void)
 void CounterSaveState(void)
 {
   for (uint8_t i = 0; i < MAX_COUNTERS; i++) {
-    if (pin[GPIO_CNTR1 +i] < 99) {
+    if (is_gpio_cnt_low(i)) {
       Settings.pulse_counter[i] = RtcSettings.pulse_counter[i];
     }
   }
@@ -78,7 +83,7 @@ void CounterInit(void)
   function counter_callbacks[] = { CounterUpdate1, CounterUpdate2, CounterUpdate3, CounterUpdate4 };
 
   for (uint8_t i = 0; i < MAX_COUNTERS; i++) {
-    if (pin[GPIO_CNTR1 +i] < 99) {
+    if (is_gpio_cnt_low(i)) {
       pinMode(pin[GPIO_CNTR1 +i], bitRead(counter_no_pullup, i) ? INPUT : INPUT_PULLUP);
       attachInterrupt(pin[GPIO_CNTR1 +i], counter_callbacks[i], FALLING);
     }
@@ -97,13 +102,14 @@ void CounterShow(bool json)
   uint8_t dsxflg = 0;
   uint8_t header = 0;
   for (uint8_t i = 0; i < MAX_COUNTERS; i++) {
-    if (pin[GPIO_CNTR1 +i] < 99) {
+    uint32_t cnt = RtcSettings.pulse_counter[i];
+    if (is_gpio_cnt_low(i)) {
       char counter[33];
       if (bitRead(Settings.pulse_counter_type, i)) {
-        dtostrfd((double)RtcSettings.pulse_counter[i] / 1000000, 6, counter);
+        dtostrfd((double)cnt / 1000000, 6, counter);
       } else {
         dsxflg++;
-        dtostrfd(RtcSettings.pulse_counter[i], 0, counter);
+        snprintf_P(counter, sizeof(counter), PSTR("%lu"), (unsigned long)cnt);
       }
 
       if (json) {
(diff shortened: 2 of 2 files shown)


andrethomas:
If its not supposed to be printed as a double then why do:

`dtostrfd((double)cnt / 1000000, 6, counter);`

and if its supposed to be a uint32_t instead of an unsigned long why do you do:

`snprintf_P(counter, sizeof(counter), PSTR("%lu"), (unsigned long)cnt);`

Also note that on the xtensa architecture there is no bit difference between uint32_t and unsigned long so if you're going to typecast it anyway then I don't understand what this PR does.

hhaim:
1. uint32_t vs unsigned long 

you are right, uint32_t and unsinged long on ESP82xx (32bit) is actually the same (32bit). However, it is better use uint32_t when you want to keep the variable as 32bit in all platform types (e.g.  future 64bit Tensilica or new SonOff that will use future ARM) 

2. The reason I've changed the Setting. pulse_counter to uint32_t is because this is an API (MQTT/HTTP) and we should keep it 32bit even in future 64bit platform. The SDK time infra could use native long (64bit or 32bit) but not an external API (MQTT/HTTP)

3. Looking into Tensilica asm output (objdump -D), I see emulation for double (ESP82xx should support single floating point hardware block) so converting uint32_t to double cost many cycles and I'm not sure what is the accuracy and implementation of the emulation layer. so better to print nativly as 32bit

4. regarding this
``snprintf_P(counter, sizeof(counter), PSTR("%lu"), (unsigned long)cnt);``

The reason I'm casting to unsigned long again (from uint32_t) is because printf functions require unsigned long (64bit on 64bit platform) and this line will emit an error/warning in those  platform (not on ESP82xx).  By converting always to unsigned long, there is no issue. 

BTW the alternative is to use something like this %"PRIu32" but I don't like it.

see this link https://stackoverflow.com/questions/3168275/printf-format-specifiers-for-uint32-t-and-size-t

 5. 

dtostrfd((double)cnt / 1000000, 6, counter);

This line print freq in double and it is OK to use dtostrfd (expensive) however, there is no need to do that with native 32bit 


**so to sum**:

1. Upper layer software with interaction to API should use uintxx_t and not long/short that is platform dependent 

2. To print uintxx_t this you should use either this ugly %"PRIu32" or just to cast 

3. When you need to print uint32_t you should not convert it to double -- it is expensive and there is a need to look if we are loosing precision with current Tensilica gcc implementation.

arendst:
Just quickly made some of these changes to the code hoping it would make the code smaller but alas, it doesn't. As performance is no issue here (it's called only at show time and trigger time) I see no reason to implement these changes other than the replacement of the dtostrfd staying at non float.

SO all I implement is this line:
```
        snprintf_P(counter, sizeof(counter), PSTR("%lu"), RtcSettings.pulse_counter[i]);
```
Thx for your input.

hhaim:
@arendst 
Would you able to elaborate why in this instance (pulse_counter), unsigned long is used, and in another instances uint32_t is used? Is it a mistake or is it on purpose? 
It is very clear why the size is bigger the reason is this line ```snprintf_P(counter, sizeof(counter), PSTR("%lu"), RtcSettings.pulse_counter[i]);``` which require more code to convert the args and const string. into the calling function (bigger than converting to double) and not the other cleanup I did (which I can guess reduce the code size).

arendst:
unsigned long vs uint32_t
This project has been on the road a long time and many features are added by others than me so for obvious reasons you'll see sometinms unsigned long and other times uint32_t. It is also depended on loaned code people add to the project. So it's not on purpose, it's just a matter of directing.

I didn't say the code was bigger, I said I hoped the code size would be smaller. It's the same size (on 16 byte granularity). Changing pulse_cointer[i] to cnt didn't make a difference either.

andrethomas:
> which I can guess reduce the code size

We do actually test code proposed for PR's to see if they yield any benefit before commenting.

The latter in itself is especially disappointing if it gets you excited that you might shave off a few bytes of flash or ram only to find out after spending 15 minutes recreating the codebase inclusive of the proposed PR and another 15-30 minutes testing whether or not it yields any practical / tangible benefit.

To extend on @arendst comments about unsigned long being inherited from other code - This is a fact insofar that writing libraries which need to work across multiple platforms one does tend to use something which would be applicable on all platforms. Not all platforms have uint32 and uint64 defined but most (if not all) definitely have the good old unsigned long so most users and amateur programmers out there learn these type definitions sometimes by purpose and sometimes by accident, but in this particular case it actually makes no difference imho.

I'm not too concerned about whether we're using uint32_t or unsigned long - just that the benefit of one over the other is outweighed by the amount of time spent to change it because it basically has no impact on the current codebase.

If a 64bit Tensilica was ever released I do not see any problem in keeping it as an unsigned long because all that will happen is that the theoretical maximum number you can store in it will increase. A 64bit micro controller is not going to process a uint32_t any faster than a uint64_t so the reason for the change is purely academic and has little practical value.

The answers on PR's is usually not this detailed because we simply do not have the time to explain in fine detail why a specific option is taken over another.

I love practical and well staged PR's though - what I don't like very much is when the time it takes to test and explain why its not viable takes longer than all the clock cycles the PR was supposed to save over the lifetime of a ESP82xx...

hhaim:
Understood. I had in mind that there is a good reason for using that. I come from a place that the code is compiled to many platforms and there is a good chance that this project will be ported to more one arch/64bit. My two cent is **not** to use long/short, only if there is a good reason to (in 99% of the time it is wrong to use it) -- this would be a good start in this direction.  
Many duplications of ```RtcSettings. pulse_counter [I] ``` on the same short function does not look good too, but maybe I'm wrong..

hhaim:
> If a 64bit Tensilica was ever released I do not see any problem in keeping it as an unsigned long because all that will happen is that the theoretical maximum number you can store in it will increase. A 64bit micro controller is not going to process a uint32_t any faster than a uint64_t so the reason for the change is purely academic and has little practical value.


This is 100% true . The counter using MQTT is an API. The client of this API (e.g. HASS) should know if the wrap happens in 32bit or 64bit.  The client code would be **different** in case it is wrapped in 64bit.  But as long you are confined this project to 32bit/ ESP8266 there is no good reason to change it other than many inconsistencies in the code even in the same struct 
The fact that many writing that does help 


```
struct SYSCFG {
..
  unsigned long save_flag;                 // 004
  unsigned long version;                   // 008

...

  uint32_t      ip_address[4];             // 544

  uint32_t      monitors;                  // 7A0

..

  unsigned long pulse_counter[MAX_COUNTERS];  // 5C0

```

But again we don't need to agree

hhaim:
> The latter in itself is especially disappointing if it gets you excited that you might shave off a few bytes of flash or ram only to find out after spending 15 minutes recreating the codebase inclusive of the proposed PR and another 15-30 minutes testing whether or not it yields any practical / tangible benefit.
> 

I think you missed the point of the PR. It wasn't to reduce the ram/flash size. It wasn't either performance improvement. It was correctness issue.
I've tried to fix some other things in the same vicinity, but it seems you were distracted by that. 
Won't do that again :-)

andrethomas:
> The counter using MQTT is an API. The client of this API (e.g. HASS) should know if the wrap happens in 32bit or 64bit.

Again, incorrect assumption - data is transmitted as text over MQTT.

I agree it would be an issue if the data was sent in binary format, but it isn't.

hhaim:
This is not true. Let me give you an example to better explain it

Let's say the counter has a value of 0xFFFF,FFFF and it was sent in MQTT string as 4294967295.

What would be the next value when there is another **1** pulse?

Now there are two options: 

1. 64bit option (long on 64bit): 4294967296
2. 32bit option (long on 32bit): 0

If the client knows that the counter is 32bit it can do 

(32bit)new-(32bit)old and it get diff==1 (OK)

But if the client will assume the counter is 64bit and actually it will come as 32bit there would be a problem 
beacuse (64bit)new-(64bit)old would be a huge diff of 2^32

The API is important, we can't change it  because the platform was changed 

this is the reason I suggest to use uint32_t which is 32bit on all platform and it is more readable 
and consistence

andrethomas:
You're forgetting that we are using a 32bit platform... so again yes I get your point but its purely academic at this point and has no impact or value for the project in its current form.

In either event, in your proposed change to uint32_t, the next pulse would be 0 whether its unsigned long or uint32_t on the current platform so you're kinda detracting from your original idea which is somewhat confusing but I'll accept that as your prerogative :)

PR #5266, opened 2019-02-18
Update Bulgarian translations

znanev:


arendst:
I think your translation of D_SENSOR_USER will make the GUI unusable. 

Did you try and noticed the consequence of this long name.

Pls reconsider.

znanev:
To be honest I didn't pay attention to this particular translation. I blindly translated it, because I "missed" this key the previous time.

The translation is abbreviated now and should fit the GUI.

PR #5272, opened 2019-02-18
Introduce Reset 6

andrethomas:
Requested from Discord Chat support

Introduce reset 6 command which is a variation of reset 5 but also retains mqtt host, port, username, password and topic.

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1309,7 +1309,7 @@ void MqttDataHandler(char* topic, uint8_t* data, unsigned int data_len)
       case 2:
       case 3:
       case 4:
-      case 5:
+      case 5 ... 6:
         restart_flag = 210 + payload;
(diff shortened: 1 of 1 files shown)


andrethomas:
Hold on, need to add mqtt_fulltopic also...

andrethomas:
Nevermind, they decided against retaining fulltopic :)

majherek:
What about mqtt fingerprints?

andrethomas:
@majherek Thought about that but if mqtt fingerprints were configured in my_user_config(override).h it would be restored to that anyway... but this feature is mostly aimed at for users who do not have anything like that configured and are using the standard release / development binaries but are experiencing a bad data flash scenario so the support guys on Discord usually try `reset 5` or now `reset 6` as a first method which does seem to solve the problem in a number of cases thereby avoiding the need to pull a device from its physical location to do a serial flash.

Being able to carry the mqtt details over as per this PR also allows users to continue sending commands over mqtt after a reset has been done. We've seen cases where the web ui seemingly stops working but the device is still reachable via mqtt allowing for an upgrade to minimal and then back to a previous version or core if necessary.

So yes, you can't get all the possible scenario's but at least having some options in contrast to a physical recovery is better than having nothing at all.

PR #5273, opened 2019-02-18
Support HttpGet command

laurentdong:
HttpGet command send a HTTP Get request to specified URL and return the response from website or error message if failed.
Note: This command support URL encoding, so you do not have to do encode by yourself. For example replacing all " " with %20 is no necessary.
Format:
"httpget \<url\>"
With HttpGet command you can do a lot of things.
For example:
- Retrieve your physical location:
httpget http://ipinfo.io/geo
Result is a JSON object
{
  "ip": "8.8.8.8",
  "city": "Mountain View",
  "region": "California",
  "country": "US",
  "loc": "37.3860,-122.0840",
  "postal": "94035",
  "phone": "650"
}
- Control another Sonoff switch directly:
httpget http://192.168.1.130/cm?cmnd=power off

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -272,6 +272,7 @@
   #define D_JSON_GPIO "GPIO"
   #define D_JSON_FLAG "FLAG"
   #define D_JSON_BASE "BASE"
+#define D_CMND_HTTP_GET "HttpGet"
 
 // Commands xdrv_01_mqtt.ino
 #define D_CMND_MQTTHOST "MqttHost"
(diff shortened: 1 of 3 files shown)


arendst:
Have a look at websend...

laurentdong:
@arendst Sorry, I'm newbie here. Did not realize that we already have it! :)

localhost61:
I wanted to test the `websend` command with curent version 6.4.1.17
and got the expected result (that is toggle the relay on the target device) with:
`websend [192.168.61.130] /cm?cmnd=power1 on`

but with
`websend [ipinfo.io] /geo`
I didn't receive the expected Json, only  the `WebSend : "Done"` acknowledge.
Same result from the URL:
`192.168.61.120/cm?cmnd=websend [ipinfo.io] /geo'

arendst:
Websend was never intended to receive data. The goal was to initiate commands on another tasmotized device without the need for MQTT.

That worked fine but I noticed today that it doesn't work fine anymore. It worked very intermittent.

Today I rewrote the websend code using part of @laurentdong which btw is an update of the core ESP8266HTTPClient example. This lead to a stable command again and also reduced code size. my major task in life.

Returning received data to the user is still not implemented as I see no use for it and it adds code.

PR #5274, opened 2019-02-18
double socket

detafun:
Hello, 

I have a double socket. The two relays are switched via D2 (2) and D6 (1). A blue LED over D1, a green LED over D6. There is still a red LED, this should be switched on and off via the button.

Problem is, if I use the relays and LEDs, the relays switch correctly, but the corresponding LED does not. The only thing that lights up better is the blue LED (D1) for both switches in the web interface, not just the blue (D1) LED for relay D6, but also for the green LED D0.
How can I manage to switch the blue LED (D1) only with relay (D6) and switch the green LED (D0) to relay (D2)?

cu Deta


ascillato2:
Sorry, invalid pull request. You are trying to merge 2 different branches of Tasmota

Closing

Issue #5286, opened 2019-02-20
tasmota reset without button


elzershark:
tasmota reset without button.
a led Bulb (Tuya) have not a Button.
can the software reset without button? by bulbs reset is 5x 
without electricity by tuya.
can you install that?

by wrong password in wifi, and the lamp can no longer be used. ore wrong ssid. ore wrong password in web interface

wrong password wifi.. = in router delete the wlan password. bulb is ok
wrong ssid = not to use lamp
wrong web password interface = not to use lamp

netpok:
I don't really understand your message sorry, but:
wrong wifi ssid and/or password: if you know the ssid or password you entered just set that temporarily in router
wrong admin password: if you have mqtt configured you can reset it via that

Also since tasmota uses GET request for forms you should check your history, maybe it's there look for entries with name of `Save Configuration`.

for example the following link
`http://sp10/co?t1=%7B%22NAME%22%3A%22Generic%22%2C%22GPIO%22%3A%5B23%2C22%2C24%2C17%2C134%2C132%2C0%2C0%2C131%2C52%2C21%2C0%2C0%5D%2C%22FLAG%22%3A0%2C%22BASE%22%3A67%7D&p1=SecretPassword&b1=on&a1=Sonoff&a2=Sonoff2&a3=Sonoff3&a4=Sonoff4&b2=0&save=`
contains the password (`SecretPassword` in this case) for web interface in the `p1` parameter.

Any other cases you are out of luck and you probably have to open the case (or find a bug in the code, but then please report it).

elzershark:
no, you can not send a mqtt command, when you not have the password, ssid, wlan ore what ever 😊
For lamps, there should be a possibility that reset over "de-energized". for example Switch 5x (de-energized)

ascillato:
Hi

At this moment, Tasmota has a bootloop protection routine that if Tasmota starts and resets (or loose power) in less than 2 seconds, and this happens several times in a row, Tasmota will start changing settings to default, one by one. I think that 10 cycles in a row put mostly settings to default. But wifi and mqtt settings will remain.

May be, can be added an extra configuration key at compiling time to make Tasmota full reset if it is performed 11 cycles of less than 2 seconds of power ON for each cycle.

What do you think?

netpok:
[...]

PR #5296, opened 2019-02-21
Wifi password validation

netpok:
In light of issue #5286, I noticed that currently an invalid wifi password can be saved, rendering the device inaccessible.

This pull request adds validation to the password so it has to be at least 5 character long which is the minimal length of a WEP password (in my opinion nobody should use WEP, but the current firmware supports it) so a mistyped password can be restored without opening the device.

In case of an invalid password the device will keep the current settings, as it does with the `****` default password used in the web interface.


andrethomas:
@netpok Can we retain support for * (D_ASTERISK_PWD) as part of a password - This was added with PR https://github.com/arendst/Sonoff-Tasmota/pull/5242 as we confirmed that * may be used on all three cores we currently support.

andrethomas:
Ah, wait it will be supported anyway.

andrethomas:
Confirmed, a new password will update if blank or at least 5 characters in length.

andrethomas2 (approved):

PR #5297, opened 2019-02-21
WifiConfig 7, a reset only WifiManager for smart bulbs

netpok:
This is an implamentation of `WifiConfig 7`, which starts the WifiManager but the only available option is to reset the device. This was created on the base of #5286, and maybe it should be sat as default mode, as it does not expose the configuration of the device.

In this mode after the device enters to WifiManager mode, the user can reset the device without the need of the web admin password.

Finished and waiting for testing.


elzershark:
I do not understand the function. How can I reset a device with it? how do you do that?

netpok:
If the device can not connect to the access point it will start up in AP mode, just as the first time you configured it, but now only a reset button is available, to securely erase the all configurations without exposing the credentials.

Here is a screenshot:
image

andrethomas2 (approved):
Thanks @netpok 

Works as expected and uses 336 bytes of flash.

Up to @arendst for further comment.

elzershark:
Ich werd es testen ob man resetten kann, auch wenn man "Daten" vergessen hat.

elzershark:
Perfect. works wonderfully

Issue #5310, opened 2019-02-23
Need option to change RGB color to BGR (SM16716)


dgranz:
The color bytes are in BGR order on the Feit Electric RGB bulbs (which use the SM16716)
Please add an option to select a BGR ordering.

Jason2866:
@dgranz  SM16716 implementation is made from @gsimon75 
Maybe he can add this as a option....

dgranz:
Around line 435 of xdrv_04_light.ino
changed to:

  // send 24-bit rgb data
  SM16716_SendByte(duty_b);
  SM16716_SendByte(duty_g);
  SM16716_SendByte(duty_r);

makes the Feit Electric bulb colors work.

Unfortunately I don't have the time to figure out how to make it a configurable option.

daninfuchs:
I suspect you could use Generic or the new Template feature (which I've just learned of) to swap PWM pin order to self correct this - I've been doing similarly (Generic, that is) for a set of unsupported WoMo Smart LED controllers for a while now.

dgranz:
> I suspect you could use Generic or the new Template feature (which I've just learned of) to swap PWM pin order to self correct this - I've been doing similarly (Generic, that is) for a set of unsupported WoMo Smart LED controllers for a while now.

Unfortunately those don't work as the color order is hardwired within the SM16716 code.

gsimon75:
Right, sm16716 doesn't use pwms, it's a completely different controller.
I've added a config option (like `#define USE_SM16716_RGB_ORDER     SM16716_RGB` in `my_user_config.h`), but I'd like to test it before submitting the pull request :).

DavinKD:
MY9231 has the same issue. It would be good if these were handled as well. I have some Lohas bulbs that have two like the AILights, but in a different order for color channels.

arendst:
Discussed with @gsimon75 (#5322) who will provide a SetOption version.

gsimon75-temp:
There is a problem with the color mapping code recently merged: when a domoticz server contacts the bulb, for some reason the `color_remap[]` array gets overwritten.

Not a solution, but a workaround is to move its declaration from below kFixedColdWarm to below light_signal_color:

```
diff --git a/sonoff/xdrv_04_light.ino b/sonoff/xdrv_04_light.ino
index c606425..d765a9f 100644
--- a/sonoff/xdrv_04_light.ino
+++ b/sonoff/xdrv_04_light.ino
[...]

PR #5322, opened 2019-02-24
sm16716: added config option for custom rgb order

gsimon75:
It seems that there is a need to make the RGB order configurable ( [Issue #5310](https://github.com/arendst/Sonoff-Tasmota/issues/5310) ), I propose a `USE_SM16716_RGB_ORDER` compile-time option and `SM16716_{RGB,RBG,...,BGR}` constants for this purpose.

The data protocol of SM16716 uses strictly RGB, but if a manufacturer wires the red leds to the 'blue' output of the controller (and vice versa), then all we can do in the firmware is to send the data in BGR order.

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -421,6 +421,7 @@
 //  #define USE_ALECTO_V2                          // Add support for decoding Alecto V2 sensors like ACH2010, WS3000 and DKW2012 weather stations using 868MHz RF sensor receiver (+1k7 code)
 
 #define USE_SM16716                              // Add support for SM16716 RGB LED controller (+0k7 code)
+  #define USE_SM16716_RGB_ORDER     SM16716_RGB  // SM16716 RGB order (SM16716_RGB, SM16716_RBG, SM16716_GRB, SM16716_GBR, SM16716_BRG, SM16716_BGR) 
 
 /*********************************************************************************************\
  * Debug features are only supported in development branch
--- a/sonoff/sonoff.h
+++ b/sonoff/sonoff.h
@@ -171,6 +171,13 @@ typedef unsigned long power_t;              // Power (Relay) type
 
 #define LT_SM16716             16           // Lights that use SM16716 will have this bit set in light_type
 
+#define SM16716_RGB            0
+#define SM16716_RBG            1
+#define SM16716_GRB            2
+#define SM16716_GBR            3
+#define SM16716_BRG            4
+#define SM16716_BGR            5
+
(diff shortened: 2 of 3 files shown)


blakadder:
Will there be a SetOption for this?

gsimon75:
> Will there be a SetOption for this?

Options, especially param options, are like global resources, and there are pretty few of them still unassigned (options 37..49), so they are for global settings, and not for such controller-specific ones.

On the other hand, such rgb re-mapping might make sense in a global scope as well, and it wouldn't be that difficult to add.

It's definitely worth a try, thanks for the idea :) !

arendst:
@gsimon75 keep it this way for now.

The SetOption implementation is a good one but needs to be implemented with minimal footprint for the My92x1 too. Preferably using an index in a RGB array. In any case no 6 calls to SM16716_SendRGB.

Also one SetOption for both MY and SM chips as they will never be available on the same device (for now.)

gsimon75:
> @gsimon75 keep it this way for now.
> The SetOption implementation is a good one but needs to be implemented with minimal footprint for the My92x1 too. Preferably using an index in a RGB array. In any case no 6 calls to SM16716_SendRGB.

Definitely not :).

I started adding a single-byte param, that is interpreted as a 3-digit ternary number, each digit telling where the corresponding rgb component is mapped *from* (digit 0 being for red, 1 for green, 2 for blue, to conform the channel ordering), so '210' (=21) would mean the identity, '012' (=5) would mean BGR, '221' (=25) would mean BBG, etc.

I thought to add it to xdrv_04_light.ino, after the cur_col[...] gets assigned, somewhere [here](https://github.com/arendst/Sonoff-Tasmota/blob/ff68b201484b7c27f9d1ad4f0b5f82f6fa82553e/sonoff/xdrv_04_light.ino#L982). The loop has to be split in two, one for setting cur_col[] and one for using it, and the remapping could be done in between.

But sure, I haven't thought it over, how it would affect everything else (schemes, etc.), so it indeed may cause a lot more trouble than it seems.

Just for fun, I'll give it a try with a local build, but even if it seems working, it'll be another pull request...

arendst:
The 3-digit ternary number is great.
The location probably too. This might also solve the WS2812 colors.

Pls test and continue as you seem fit.

digiblur:
Nice! Was just talking about how this would be great for the My9231/9291 as well. Some bulbs have multiple chips then some have odd orders along with say two channels of WW and two of CW.

DavinKD:
I was just going to mention CW and WW. Hopefully 5 is the extent of it.

Issue #4987, opened 2019-01-20
Obi Socket 2 green wifi led does not work


devZer0:
i'm using latest tasmota devel on obi socket 2.

apparently, i can enable both led's (red for relay status and green for wifi status), but i can't configure tasmota to have the green led reflect the wifi state or the mqtt state - as it seems that this is tied to the red relay status led. 

it's very suboptimal, that relay status/red-led has a double function whereas green led remains unused. this way it simply isn't complete.

how do i "route" the wifi and mqtt state to the green led?

i would have re-opened https://github.com/arendst/Sonoff-Tasmota/issues/4567 as it seems to be the same issue

devZer0:
yes i have tried it and i would not have opened an issue if it did work correctly

i looked at the code in "sonoff" file  - i don't understand the meaning of "led_pin" ? what sense does it make to be able to map gpio to led-functions/numbers when digitalWrite() containing offset-information being added to the adress?

void SetLedPower(uint8_t state)
{
  if (state) state = 1;

  uint8_t led_pin = 0;
  if (pin[GPIO_LED2] < 99) { led_pin = 1; }
  digitalWrite(pin[GPIO_LED1 + led_pin], (bitRead(led_inverted, led_pin)) ? !state : state);
}

void SetLedLink(uint8_t state)
{
  if (state) state = 1;
  digitalWrite(pin[GPIO_LED1], (bitRead(led_inverted, 0)) ? !state : state);
}

Issue #4567, opened 2018-12-08
OBI socket v2 parameter assignment


gitterman:
I frequently see those screen shots of module parameter assignments like here: https://github.com/arendst/Sonoff-Tasmota/wiki/OBI-Socket-2
Is there a list which describes the meaning of the selectable parameter values?

I have flashed the latest 6.3.0 with https://github.com/arendst/Sonoff-Tasmota/releases/download/v6.3.0/sonoff-DE.bin and selected the generic device.
The OBI v2 has the relay on GPIO4 and the button S1 on GPIO5.
There is a green LED driven by GPIO12 which is labelled "Link" and a red LED driven by GPIO13 which is labelled "Power" and which shows the relay status.
I tried various assignments for GPIO12 and GPIO13 but none worked as expected.
With the setup from the 1st link the green LED never lights and the red LED is always blinking no matter what the relay status is.
I ended up assigning "Relay 1" to the red LED on GPIO13 to make it go along with the relay.
Alas for the wifi link I cannot find a suitable setting.

BTW: is there an US-english binary which corresponds to sonoff-DE.bin?

ascillato:
> Is there a list which describes the meaning of the selectable parameter values?

Yes, in the wiki

> is there an US-english binary which corresponds to sonoff-DE.bin?

Yes, **sonoff.bin** (The difference between files are explained in the release notes)

If you need further support, please find us in the Tasmota Support Chat at https://discord.gg/Ks2Kzd4 Thanks

andrethomas2:
Closing this issue as it has been answered.

### Support Information
See [Wiki](https://github.com/arendst/Sonoff-Tasmota/wiki) for more information.
See [Chat](https://discord.gg/Ks2Kzd4) for more user experience.

chaosmaster:
@gitterman I suggest you assign GPIO12 to LED1. That will be used to indicate connection state.
See https://github.com/arendst/Sonoff-Tasmota/wiki/Commands#main (LedState) to see when it will blink.
For the power LED (GPIO13) it seems you have already found a viable solution.
If that configuration works for you as expected, you may want to update the information in the [wiki](https://github.com/arendst/Sonoff-Tasmota/wiki/OBI-Socket-2)

devZer0:
@gitterman, did that work for you? can you independently use power and wifi led?

gitterman:
@devZer0, no it did not work. GPIO13 needs to be connected to LED1 to show the correct power state.

[...]

PR #5327, opened 2019-02-24
OBI Socket 2: invert LED1

alarmschaben:
I noticed the Link LED being inverted on my OBI v2 socket.


Jason2866:
Fixes #4987

arendst:
How about #4567

alarmschaben:
Indeed. I wasn't aware that there are conflicting feelings about which way around the LED should work. Plus, in my initial research, I didn't find those two issues,

Nevermind, then. Just thought it was helpful. Thanks anyway for the great software! :-)

PR #5328, opened 2019-02-24
Augmented clock so that for even width (e.g. 2) more precise display …

jrw77:
Augmented clock so that for even width (e.g. 2) more precise display is possible with e.g. 30 or 6 LEDs. For width 2, alternately 1 and 2 LEDs will light.
With 30 LEDs, can indicate 60 positions, with 6 LEDs, with this change you can indicate 12 positions.
With odd width (1, 3, etc), the behaviour is exactly as before. Before, width 2 acted just like width 1. Now it acts like width 1.5, alternating between 1 LED and 2 LEDs.


jrw77:
Note that commit 25a2f392b1018ffcc226c38886e82d03e6c82168 allows width 0 which disables the corresponding hand without resorting to color 000000. I use this to switch off the "seconds" hand.  For small pixel numbers, setting width2 0 enables special code that (if pixels <=24) adds extra pixels in the color3 ("seconds") after the minute hand to indicate the time more precisely. 
Commit 7bfdcca091e322c3cdbf83cac2275133797f0ecb fixed the marker code to small pixel numbers to indicate only 4 positions instead of 12.

arendst:
Adding the settings print out makes this PR a no go.

The code impact is huge while the same information can already be returned by the tools decode-config.py and decode-status.py from the tools folder.

jrw77:
Sorry, I made a mistake in not creating a new branch. Commit 59707c2f14a3c0494a71237b7f7b042e915e4b74 should not have been included in this pull request. How do I undo this? Nevertheless I am glad to know about the tools, I will try them.

jrw77:
Here are some images that show what a 12 pixel clock shows with this code at different times.
12Pixel13_20
12Pixel13_23
12Pixel13_05
12Pixel13_57

PR #5334, opened 2019-02-24
Add Support for MFRC-522

AFUman:
Is it already in the planning to support the MFRC-552 RFID reader?


localhost61:
@AFUman 
On `github`, a `Pull Request` is a way to submit new code to the project owner.
I guess this is not you case, you just want to ask a question, then `Issue` would be more appropriated (even if it's not really an issue).

FabianHM:
I would also be very happy if Tasmota supported the MFRC-522. If anyone in the community could support this, I would be happy.

Thanks, Fabian

andrethomas2:
This PR is invalid as it proposes a merge between two branches.

Please do not open PR requests between branches of the same project.

PR #5361, opened 2019-02-27
ADD ADC configuration in WebUI

Easyrunner:
Hello
Is it possible to add an configuration of the ADC in the WebUi like selecting a GPIO port.
And a scaling factor would be nice.

Like EasyESP it has.

Thank you!


andrethomas:
This PR is invalid as it proposes merges between branches which are not in sync with each other. For a feature request please open a new issue.

Thanks

PR #5417, opened 2019-03-08
Fix bad snprintf in format()

emontnemery:
## Description:
Fix illegal calls to snprintf_P where output string is same as format string in format() which causes ChipID to be incorrectly formatted with core 2.5.0.

Core 2.5.0:
image

Core 2.3.0 / 2.4.2:
image

I tried to find this pattern elsewhere in the code with this regex, but it seems to be the only case:
https://regex101.com/r/lZa25j/1

Background:
snprintf_P has previously copied the format string to a temporary buffer:
``` C++
int vsnprintf_P(char* str, size_t strSize, PGM_P formatP, va_list ap) {
    int ret;

    size_t fmtLen = strlen_P(formatP);
    char* format = new char[fmtLen + 1];
    strcpy_P(format, formatP);

    ret = vsnprintf(str, strSize, format, ap);

    delete[] format;

    return ret;
}
```

This is no longer the case in core 2.5.0:
https://github.com/esp8266/Arduino/commit/6280e98b0360f85fdac2b8f10707fffb4f6e6e31
https://github.com/igrr/newlib-xtensa/commit/61f08d03e875a64629deaaa0b975e179bcfa4ea0#diff-41d58351e2a3ed465ccfa03961a37263

## Checklist:
  - [x] The pull request is done against the dev branch
  - [x] Only relevant files were touched (Also beware if your editor has auto-formatting feature enabled)

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -200,12 +200,13 @@ char* Format(char* output, const char* input, int size)
     if (token != NULL) {
       digits = atoi(token);
       if (digits) {
+        char tmp[size];
         if (strchr(token, 'd')) {
-          snprintf_P(output, size, PSTR("%s%c0%dd"), output, '%', digits);
-          snprintf_P(output, size, output, ESP.getChipId() & 0x1fff);       // %04d - short chip ID in dec, like in hostname
+          snprintf_P(tmp, size, PSTR("%s%c0%dd"), output, '%', digits);
+          snprintf(output, size, tmp, ESP.getChipId() & 0x1fff);            // %04d - short chip ID in dec, like in hostname
         } else {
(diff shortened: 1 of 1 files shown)


ascillato2:
Nice bug catch!

This fix is backwards compatible with core 2.3.0 and 2.4.2? Or we need to add a precompiler define key?

emontnemery:
Backwards compatible.

ascillato (approved):

arendst:
Any idea why line 205 changed from snprintf_P to snprintf and line 209 still keeps snprintf_P?

I know the difference but I like 205 to stay at snprintf_P as it makes the code smaller...

emontnemery:
@arendst just copy/paste mistake, sorry

Issue #5369, opened 2019-02-28
USE_EXPRESSION breaks string variable support


netpok:
**Describe the bug**
_A clear and concise description of what the bug is._
Firmware compiled with USE_EXPRESSION can't store string variables as it's defaults to 0.000.

_Also, make sure these boxes are checked [x] before submitting your issue - Thank you!_
- [x] _Searched the problem in issues and in the wiki_
- [x] _Hardware used_ : All affected
- [x] _Development/Compiler/Upload tools used_ : not relevant
- [x] _If a pre-compiled release or development binary was used, which one?_ : not relevant
- [x] _You have tried latest release or development binaries?_ : yes
- [x] _Provide the output of command_``status 0`` : not relevant

**To Reproduce**
Executing `var1 TEST` will result in value `0.000`

**Expected behavior**
var1 value: `TEST`

**Additional context**
Currently can not think to a right solution so I'm leaving this here maybe someone has an idea.

[Cause of problem is:](https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/xdrv_10_rules.ino#L1172)
```
#ifdef USE_EXPRESSION
      dtostrfd(evaluateExpression(XdrvMailbox.data, XdrvMailbox.data_len), Settings.flag2.calc_resolution, vars[index -1]);
#else
      strlcpy(vars[index -1], ('"' == XdrvMailbox.data[0]) ? "" : XdrvMailbox.data, sizeof(vars[index -1]));
#endif      //USE_EXPRESSION
```

localhost61:
Yes, the feature has its limits, it can operate on timestamps neither, it's one of the reason I couldn't use it to monitor Wifi down periods and made a PR for that.

netpok:
My current ideas to solve this is:

 1. to introduce 2 new commands:
`VarE` variable with expression evaluation and
`MemE` memory with expression evaluation
 2. Introduce a character for the same:
 for example values starting with `\` is handled as expressions, but then it needs it's own escape sequence

Note: for backward compatibility with release versions the expression ones should get the change.

I'm waiting for feedback and better ideas.

laurentdong:
Actually I knew that the implementation of expression will block string assignment to variable but I'm really new on Tasmota ( and any home automation projects ) so I didn't realize some people really need the string support.
[...]

PR #5421, opened 2019-03-08
Only do eveluate for expression while Var/Mem command follow a '='

laurentdong:
## Description:
Related issue: [issue #5369] USE_EXPRESSION breaks string variable support 

A "=" is required to do expression evaluate in command Var/Mem:
**Var1 = (Mem1 - 2) / 100**

Command:
**Var1 Mem1**
will do the exactly same thing like previous version - assign string "Mem1" to Var1

## Checklist:
  - [X ] The pull request is done against the dev branch
  - [X ] Only relevant files were touched (Also beware if your editor has auto-formatting feature enabled)
  - [X ] The code change is tested and works.
  - [X ] The code change pass travis tests. **Your PR cannot be merged unless tests pass**


netpok:
Maybe it would be better to introduce a new command for this like `VarE` (var expression) or something in code size it wouldn't really matter but it would allow the use of full character range in the value.

laurentdong:
@netpok 
Require a "=" actually is basically same idea like "VarE" but looks much better and easy to understand, isn't it? For example
**Var1=Mem1 * 2**
vs
**VarE1 Mem1 * 2**

If we really want to introduce new command, I would prefer to have a "Let" command. For example:
**Let Var1 = Mem1 * Var2 + 1**

netpok:
The difference is that in case of var1 =something, the = sign is part of the argument so you can't start the string with = sign (probably would't matter for anyone), on the other hand in case of the E is part of the command and leaves the argument constraint free.

Also I have a new idea: Currently only doubles are returned from expressions, so maybe we should make a new command for it and store them as doubles this would allow the usage of more variables. Like: Var1...5 and DVar1...40 would take up the same memory space allowing 8 times of variables and 2 times of memory spaces. The only constraint is that VAR1 must not be used with DVAR1...8 as they are pointing the same memory space. I will try this out (probably I won't have time till the weekend) and report back.

ascillato (approved):

windischb:
Is there a chance to merge this in near future?

arendst:
I'll check it now...

arendst:
Merged just now.

PR #5433, opened 2019-03-10
Command should be able to end with other characters

laurentdong:
## Description:
New implementation of ExecuteCommand()
With previous version, all the commands have to end with a space (' '). So
**Var1=Mem1+1**
will become an invalid command.
With this new implement, a command can be split up by not only spaces but also symbols like '=', '(', ',', etc. Actually commands can only use letters, digits and underscores, so any other characters can end the command. ("/" might be included in commands as a start of command)

There are some other improvement:
- Leading spaces are allowed.
- Performance should be a little bit better
- Less memory usage

## Checklist:
  - [X ] The pull request is done against the dev branch
  - [X ] Only relevant files were touched (Also beware if your editor has auto-formatting feature enabled)
  - [X ] The code change is tested and works.
  - [X ] The code change pass travis tests. **Your PR cannot be merged unless tests pass**


arendst:
I'll need to dive into this as currently it doesn't fix mqtt provided commands.

laurentdong:
> I'll need to dive into this as currently it doesn't fix mqtt provided commands.
Why? It should be able to process commands from MQTT and HTTP as well.

ascillato:
more time

ascillato2:
More time

PR #5467, opened 2019-03-14
Update he-HE.h

rt400:
Add more translate to language file


ascillato:
@rt400 

Hi Yuval, 

I have done PR #5471 to solve a compiling issue. In line 65 of this translation file, there was an extra " that prevents the compilation. When I make the edition I could not preserve all hebrew letters on that line. My editor didn't allow it. So, please, re check if that line needs a fix. Thanks.

PR #5584, opened 2019-04-05
MiDesk Lamp not default activated  

Jason2866:
## Description:

Since Mi Desk Lamp isnt used by most users make it off by default

## Checklist:
  - [ x] Only relevant files were touched (Also remember to update _changelog.ino_ file)
  - [ x] The code change is tested and works.
  - [x ] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).


arendst:
Will look in to it as there are some more hooks to rotary which can also be disabled when some code change is done.

Thx for the idea. I will follow up.

markus1540:
Any reason for this change other than just disabling it out of "unpopularity"?
Just tried out the latest binary from http://thehackbox.org/tasmota/ and discovered the lack of support for the Mi desk lamp in these builds.

arendst:
@markus1540 every piece of code uses part of precious flash/ram.

Re-enable it and re-compile yourself should solve your issue.

PR #5627, opened 2019-04-12
RF transmitter emulation for DC1602 blind controllers and similar blinds and shutters etc.

nosaj66au:
Hi Guys have never contributed code to tasmota before so I'm hoping this meets spec.
The hardware requirement is a single RF433 tx module requiring 1 output pin. 
I will enable a blind or shutter using a DC1602 or similar RF remote of which there are several to control up to 15 blinds or shutters with a single command or individually. I have added this to DRV91 but this will likely need to be moved I suspect. The command is given is JSON and there is examples in the code.
Jason

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -69,7 +69,6 @@
 #define D_JSON_FLASHCHIPID "FlashChipId"
 #define D_JSON_FLASHMODE "FlashMode"
 #define D_JSON_FLASHSIZE "FlashSize"
-#define D_JSON_FLOWRATE "FlowRate"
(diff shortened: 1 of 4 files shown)


nosaj66au:
OK I think I can see the process works after reviewing the build report. I make some changes and get it sorted.

Jason2866:
Isnt it possible to control the DC1602 via the supported Sonoff Bridge?

nosaj66au:
That is what I would assumed that as well. After 3 weeks of trying flashing and sniffing the protocol. I hooked up my scope to a receiver and bit banged it. Working coded in 8 hours. Happy days.

PR #5648, opened 2019-04-15
Dooya RF Protocol for Blind and Shutters

nosaj66au:
Hi Guys have never contributed code to Tasmota before but this is my second attempt so I'm hoping this meets spec.
The protocol is for Dooya devices which is used on just about all blinds and shutter from China.
The hardware requirement is a single RF433 tx module requiring 1 output pin.
I will enable a blind or shutter using a DC1602 or similar RF remote of which there are several to control up to 15 blinds or shutters with a single command or individually. I have added this to DRV91 but this will likely need to be moved I suspect. The command is given is JSON and there is examples in the Driver code .

Jason

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -424,6 +424,14 @@
 #define D_CMND_LATITUDE "Latitude"
 #define D_CMND_LONGITUDE "Longitude"
 
+// Commands xdrv_91_timers.ino
+#define D_JSON_RF433_GROUP "Group"
+#define D_JSON_RF433_BLIND "Blind"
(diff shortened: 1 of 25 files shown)


ascillato:
Hi,

Sorry, 

Your changes can't be followed due to the way you uploaded the files were first deleting the files you want to modify and then it were uploaded, so, there is not a control change. Se other Pull Requests for examples. 

Also, you need to complete the checklist for PRs in order to be merged.

nosaj66au:
That's ok I'm still learning the process. I'll take a look later today.

Issue #5741, opened 2019-05-04
Combining RGB with W1 or W2 is not possible any more


mtmayr:
### BUG DESCRIPTION

Combining RGB with W1 W2 is not possible any more using the 5 channel H801 (20) hardware. You can see below, that RGB switches from a value to 0,0,0 when a value for W1 is added. Also some wrong values for W1 and W2 get stored: W1=ee, W2=dd got set, and the resulting color is W1=84 W2=7B. The actual color shown seems to suit the values in RESULT; I tried more modules to verify this.

This bug is probably related to issue  #5702 or  #5704, but I am not sure. Below is the same commands on February version, which is working fine. I changed the version from back then, because of permanent reboots.

### REQUESTED INFORMATION
- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Sonoff-Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in issues (https://github.com/arendst/Sonoff-Tasmota/issues)
- [x] Searched the problem in the wiki (https://github.com/arendst/Sonoff-Tasmota/wiki/Troubleshooting)
- [x] Searched the problem in the forum (https://groups.google.com/d/forum/sonoffusers)
- [x] Searched the problem in the chat (https://discord.gg/Ks2Kzd4)
- [x] Device used (i.e. Sonoff Basic) : H801
- [x] Tasmota binary firmware version number used : 6.5.0.9(sonoff) / self-compiled
- [x] Development IDE - Compiler / Upload tools used : Arduino 1.8.9 IDE
- [x] Provide the output of command ``status 0`` :
```
02:30:19 MQT: stat/sonoff_6F95E6/STATUS = {"Status":{"Module":20,"FriendlyName":["Sonoff"],"Topic":"sonoff_6F95E6","ButtonTopic":"0","Power":1,"PowerOnState":3,"LedState":0,"LedMask":"FFFF","SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":0}}
[...]

PR #5702, opened 2019-04-25
Refactored management of lights, using classes and integers instead of floats.

s-hadinger:
## Description:

This is a refactoring of the way Lights are managed.
 - all manipulation are encapsulated in classes.
 - all values are stored and computed as integers, avoiding expensive float calculations and avoiding rounding errors. Only exception: RGBtoXY uses floats
 - Brightness is now managed at full range 0..255, Dimmer is kept in sync with range 0..100
 - Hue/Sat and RGB are always kept in full sync. RGB is stored with full brightness (i.e. at least one of the components is 255), channels are then mixed with Brightness to get actual PWM values
 - Except if Option37 is set, light is either in RGB/Hue-Sat mode or in CT (white with colortone) mode. If CT is zero, then RGB mode, else CT mode.

Please use the following singletons:
 - light_state: to query light state (do not use setters unless you know exactly what you're doing)
 - light_controller: to change the values of the light. It will automatically recompute all channels, and store settings. Note: it does not send any MQTT message.

Final good news, code is 1.1KB smaller.

**Related issue (if applicable): 

## Checklist:
  - [x] The pull request is done against the latest dev branch
  - [x] Only relevant files were touched (Also remember to update _changelog.ino_ file)
  - [x] Only one feature/fix was added per PR.
  - [x] The code change is tested and works.
  - [x] The code change pass travis tests. **Your PR cannot be merged unless tests pass**
  - [x] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

PR #5667, opened 2019-04-17
Support for color and colortone in Philips Hue emulation via Alexa

s-hadinger:
## Description:

This provides full support for colors and white color-tone to the Philips Hue emulation via Alexa devices or Alexa smartphone app.

Specifically the Alexa app is very picky about receiving the exact same value that it sent initially. For ex, if it sends Brightness of '43', sometimes due to rounding errors it would send back '42'. The new version sends exactly the same value to the app.

 - to avoid receiving XY colors, don't send any XY colors in HUE lights status
 - adding rounding when converting from float to uint8 to minimise rounding errors when converting from RGB to HSB and back
 - memorize value sent by the app and send back the exact same value if the computed value is within 1%
 - store colors with full brightness in Settings.light_color, and use the Dimmer to adjust brightness. This avoids big conversion errors when converting from RGB to Hue/Sat
 - size code increase is less than 800 bytes

**Related issue (if applicable):** fixes #5600 #4809

## Checklist:
  - [x] The pull request is done against the latest dev branch
  - [x] Only relevant files were touched (Also remember to update _changelog.ino_ file)
  - [x] Only one feature/fix was added per PR.
  - [x] The code change is tested and works.
  - [x] The code change pass travis tests. **Your PR cannot be merged unless tests pass**
  - [x] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

PR #5748, opened 2019-05-05
Allow all 5 PWM channels individually adressable with LEDs.

s-hadinger:
## Description:

This is the last stage of lights managemenent refactoring of #5667, #5702, #5742. This change allows for individual setting of each of the 5 PMW channels, whatever color or CT mode. The behavior corrects a regression from #5702.

Exemple:

You can now issue the command (with 5 PMW devices)

`Color #1020304050`

Result:
```
17:25:01 CMD: Color #1020304050
17:25:01 RSL: stat/sonoff/RESULT = {"POWER":"ON","Dimmer":31,"Color":"1020304050","HSBColor":"210,67,19","Channel":[6,13,19,25,31],"CT":307}
```

**Related issue (if applicable): #5741

Implementation:

Internally, the Light state moved from implicit color-mode (based on CT, brightness and Option37) to a simpler and more convenient explicit color-mode through flags.

## Checklist:

### DIFF
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -4,6 +4,8 @@
  * Fix Shelly 2.5 overtemp
  * Set gamma correction as default behavior, ie "Ledtable 1"
  * Refactored management of lights, using classes and integers instead of floats.
+ * Extend PWM resolution from 8 to 10 bits for low brightness lights
+ * Allow all 5 PWM channels individually adressable with LEDs. (#5741)
  *
  * 6.5.0.8 20190413
  * Add Tuya Dimmer 10 second heartbeat serial packet required by some Tuya dimmer secondary MCUs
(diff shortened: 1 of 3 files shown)


s-hadinger:
@blakadder I'm not sure we need to update the wiki, as we are simply back to previous behavior.

The PR #5702 prevented from chosing levels of PMW4 and PMW5 separately, because they were always linked by CT (white color temp). With this PR, you can now specifiy each PMW value separately.

Normal operation is the following:
 - `HSBColor1 xx` to `HSBColor3 xx` will automatically reset channels 4 and 5 - we switch to RGB mode.
 - `CT xx` will automatically reset channels 1 to 3 - we switch to White mode.
 - `Color #112233` will automatically reset channels 4 and 5 - they are implicitly set to zero like if you had typed `Color #1122330000`

But the following commands will allow to have both RGB and White at the same time:
 - `Color #1122334455` to will set all five LEDS at once
 - `Channel1` to `Channel5` will also set the corresponding value for one LED without changing the others.

Now if you SetOption 37 to something greater than 128, then `HSBColor` will keep channels 4-5 unchanged, and `CT` will not change channels 1-3.

blakadder:
Ah great! Thanks a lot! Great work on the color management

PR #5808, opened 2019-05-17
MQTT latency improvement

sislakd:
## Description:

The current implementation of MQTT PubSubClient process only one incoming MQTT message per one loop() call (e.g. it could be just MQTT ping response). In combination with default Dynamic 50 ms sleep of the main Tasmota loop this increase latency of execution of incoming MQTT messages especially if there are several cumming within 50ms interval. The latency is dependent on the number of messages.

This PR proposes simple change removing this negative effect by small change in PubSubClient implmentation.

## Checklist:

### DIFF
--- a/lib/PubSubClient-EspEasy-2.6.09/src/PubSubClient.cpp
+++ b/lib/PubSubClient-EspEasy-2.6.09/src/PubSubClient.cpp
@@ -303,7 +303,7 @@ boolean PubSubClient::loop() {
                 pingOutstanding = true;
             }
         }
-        if (_client->available()) {
+        while (_client->available()) {
             uint8_t llen;
             uint16_t len = readPacket(&llen);
             uint16_t msgId = 0;
--- a/lib/PubSubClient-EspEasy-2.6.09/src/PubSubClient.cpp
+++ b/lib/PubSubClient-EspEasy-2.6.09/src/PubSubClient.cpp
@@ -346,6 +346,7 @@ boolean PubSubClient::loop() {
                 // readPacket has closed the connection
                 return false;
             }
+            optimistic_yield(1000);
         }
         return true;
     }
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,4 +1,7 @@
-/* 6.5.0.10 20190513
+/* 6.5.0.11 20190517
+ * Reduced latency of Tasmota for incominig MQTT commands
+ *
+ * 6.5.0.10 20190513
  * Enable ADC0 by default in my_user_config.h (#5671)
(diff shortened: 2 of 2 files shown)


joba-1:
I understand 50ms/mqtt message might sometimes be too slow (not in my usecases though). But processing only one message per loop iteration is a deliberate choice to guarantee multitasking (wifi connection, button press latency, ...) and prevent reset by mqtt flooding. I think at least there should be a time limit for the message loop.

sislakd:
I've added optimistic_yield to avoid issues related to wifi connection, wdt and other background tasks.

arendst:
From tests I conducted not calling loop too often improves MQTT connection. Servicing incoming MQTT messages need more time between loop checks.

This has been implemented in xdrv_02_mqtt.ino at lines 830.

Since this change (only every 50 mSec instead of every loop) MQTT reconnects have almost disappeared in my environment where they were abundant (have to say I never use sleep).

sislakd:
MQTT reconnects disappear if you use latest Arduino with fixed implementation of ClientContext where there were deadlocks resulting in MQTT disconnections. On other hand, if you have Tasmota devices not handling rare incoming messages but frequent ones (e.g. sensor values send among various devices), increased latency of command execution from user (e.g. from OpenHab) decreases user experience with the system.

Jason2866:
@sislakd  the core with zero reconnects is core 2.3.0. All other cores have now and than mqtt reconnects
This is the experience of many users which had mqtt or wifi reconnects. For nearly all users the same Tasmota version with core 2.3.0 solved this issue.

meingraham:
"If you find the reason why core 2.4.2 and 2.5.x behaves not as good than core 2.3.0 it would be awesome!" :+1:  :+1:  :+1:  :+1:  :+1:  :+1:  :+1: 

Because of newer core features etc. Like having to use newer core for software serial, for example (but suffer link reconnects a lot, I mean a lot!). Plus all the other things the new core brings... like any newer software version.

sislakd:
@Jason2866  I know that default is Dynamic sleep with 50 ms. That is the reason why this PR was proposed. With more intensive MQTT communication (e.g. sensor data sent from one ESP to another) I've noticed that sometime it takes more than 1 second to execute user command from OpenHab sent via MQTT. The reason is that Tasmota handles only 1 MQTT incoming message every 50ms. That includes also internal message like Ping response which is sent every 10 seconds. Thus user experience was unacceptable while reacting so slowly to user input. 

With the latest stage core where there I've submitted critical fixes around WiFiClient and WiFiClientSecure I don't have any stability problems. The only issues there are around wifi_sleep_mode which doesn't work well in all NONOS SDKs.

Jason2866:
Why does core 2.3.0 have no reconnect? Same Tasmota code.

PR #5809, opened 2019-05-17
Add missing delay in main loop

sislakd:
## Description:

If Dynamic sleep is configured to time shorter than is needed to handle drivers and other tasks in the main loop, there is no delay called at all. This cause various issues related to not handled core tasks. 

This PR, proposes change in the loop which for such situation call delay(0) to yield esp context to run its job.

## Checklist:

### DIFF
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,4 +1,7 @@
-/* 6.5.0.10 20190513
+/* 6.5.0.10 20190517
+ * Add missing delay in main loop when configured Dynamic sleep is shorter than other tasks executed during the loop
+ *
+ * 6.5.0.10 20190513
  * Enable ADC0 by default in my_user_config.h (#5671)
  * Add user configurable ADC0 to Module and Template configuration compatible with current FLAG options (#5671)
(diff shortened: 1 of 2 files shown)


joba-1:
delay(0) at the end of the main arduino loop is redundant

sislakd:
Hmm, than I don't understand why it remove unresponsiveness of Tasmota when it was not there. NP will keep this change in my fork only.

sislakd:
Closing PR as there is no interest for this change.

joba-1:
Intentions don't travel well over the internet, so just to make sure I dont discourage you in general:

1) PRs are always welcome (there is a need or they wouldn't exist)
2) I am not tasmota - me commenting the change is redundant does not mean there is no interest ( rather on the contrary: Even I wouldnt comment if I'm not interested :) ).
3) I am not always right ;)

But since docs https://arduino-esp8266.readthedocs.io/en/latest/reference.html#timing-and-delays say
  "...handle any pending events each time the loop() function completes, OR when delay is called"
a test case that prooves the need would be good (my opinion).

Jason2866:
Mhh @joba-1  documentation says not needed but not everything in the world works as described 🤔
@dsislak  How can it be checked that the change you suggest is a good choice and documentation is wrong?

PR #5810, opened 2019-05-17
Save RAM - move ledTable from RAM to FLASH

sislakd:
## Description:

Offloading big _ledTable in xdrv_04_light.ino from RAM to FLASH.

## Checklist:

### DIFF
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,4 +1,7 @@
-/* 6.5.0.10 20190513
+/* 6.5.0.11 20190517
+ * Offload big constant ledTable from RAM to FLASH
+ *
+ * 6.5.0.10 20190513
  * Enable ADC0 by default in my_user_config.h (#5671)
  * Add user configurable ADC0 to Module and Template configuration compatible with current FLAG options (#5671)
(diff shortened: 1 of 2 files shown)


arendst:
Thx.

The table is in RAM as it is used frequently (for every byte) and some hardware just cannot cope with many flash accesses due to bad power design.

So will stay in RAM for now.

sislakd:
I don't see there so high frequency of reads from this table. The whole Tasmota program is read from flash as well and frequency of reads is incomparable. I haven't noticed any issues while moved to FLASH even with enabled light correction.

arendst:
-) Code is read from flash during execution; There is just not enough RAM to hold all Tasmota code.
-) Using WS2812 every pixel value will be read from this table. Flash reads always use four bytes at least so there is overhead too.

sislakd:
That was the point, if read of code from FLASH doesn't make issues on a device, very small increase of reads due to light correction is not causing issues. On other hand, issues related to lack of ram for runtime as it is consumed by big constant table makes more troubles for all devices even they are not using this driver.

s-hadinger:
I believe the best option would be to use #ifdef to move to Progmem only when S2812 is disabled.

arendst:
@s-hadinger won't work as non WS2812 devices also use gamma correction.

s-hadinger:
Indeed. But my point is that non-WS2812 only recalculate gamma at most 4 times only when you change colors. So it should be rare enough not to stress the power.

Issue #5579, opened 2019-04-03
ADD:Humidity compensation for SGP30


TONGXD:
What should I do?

The SGP30 features an on-chip humidity compensation for the air quality signals (CO2eq and TVOC) and sensor raw signals (H2-signal and Ethanol_signal). To use the on-chip humidity compensation an absolute humidity value from an external humidity sensor like the SHTxx is required.

ascillato2:
Hi, 

According to the [SGP30 datasheet] (https://www.mouser.com/pdfdocs/Sensirion_Gas_Sensors_SGP30_Datasheet_EN-1148053.pdf) that compensation is possible. 

Tasmota is using the adafruit library 1.0.2 that does not have the commands for applying that compensation, but it is supported on the new adafruit library 1.0.3 from https://github.com/adafruit/Adafruit_SGP30

So, it is needed to update the library in Tasmota codebase and modify the SGP driver to allow using humidity as input. I think a new sensor command is needed to choose to apply or not this compensation.

Adding the label enhancement to this issue. 

Thanks for sharing your ideas :+1:

arendst:
Tasmota has a hidden feature which stores temp and hum of a connected device in a global struct. The info in this struct can be used by any sensor needing this information for whatever reason.

So all that is needed is adding some code to the SGP30 sensor driver to get this global data. Sounds simple ;-)

gemu2015:
unfortunately it is not that easy! the SGP30 needs ABSOLUTE humidity with is very complicated to calculate and needs relative humidity, temperature and air pressure.

gemu2015:
as it turns out that absolute humidity (as in contrast to relative humidity) is the factor that determines if viruses can survive in an environment (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2893471/) it would be very interesting to have it. since the bme280 provides all the necessary  input, absolute humidity should be implemented there. it then can also be used for compensation of the sgp30.

the formula however  is a monster and uses some flash and time consuming functions =>

[...]

PR #5817, opened 2019-05-18
Sgp30 compensation

gemu2015:
## Description:

ADD:Humidity compensation for SGP30 #5579

and additionally generates absolute humidity as an extra output of the sgp30

the issue with disturbing clock stretched devices be resetting the clock stretch back to 230 could be resolved by removing wire.begin from the adafruit driver. however some other drivers call this also.

my suggestion would be to set the clock stretch limit for the ccs811 and other devices that need longer stretching upon every call to these devices. this call simply sets the clock stretch variable and  should have no impact on performance.

## Checklist:
  - [ ] Only relevant files were touched (Also remember to update _changelog.ino_ file)


s-hadinger:
Actually just realized there is a more compact FastPrecisePow()

gusman80:
Hi, great enhancement! Can't absolute humidity be added to all sensors providing all the necessary inputs like BME280 as well?

PR #5898, opened 2019-05-31
Save some bytes at web interface

tammo0:
## Description:
By removing the "name"-tag of input, select, textarea and button elements and add them via javascript we save some bytes. The name-tag will be set if the element has an id-attribute and no name tag already.
Also `<br/>` can be `<br>`

## Checklist:
  - [?] The code change pass travis tests. **Your PR cannot be merged unless tests pass**


arendst:
Thx.

PR #5923, opened 2019-06-05
Added support for AWS IoT

s-hadinger:
## Description:

Thanks to the use of BearSSL and aggressive optimization, it is now possible to add TLS and AWS IoT support with only 6.0k of memory footprint during normal operations, and 6.8k memory footpring during connection (TLS handshake).

This opens a wide area of uses, like automation of your home without requiring a local gateway, or adding Alexa Skills to control all sorts of device beyond Philips Hue / Wemo emulation.

Full documentation here: https://github.com/arendst/Sonoff-Tasmota/wiki/AWS-IoT

This is still experimental, and needs long run stability testing. You will need to custom compile your firmware with your own Private Keys and certificates. There is no impact to your Tasmota code until you enable `#define USE_MQTT_AWS_IOT`

## Checklist:


EYScudeler:
That is an AWESOME work man! Congratz... Waiting to see this on the next release...

arendst:
This only works with what core version(s)?

arendst:
Just read the wiki. 

I suggest to make a seperate platformio.ini entry for this environment as it relies on many changes and (probably) a single supported core version. Better yet, it allows for your control over just one core version which would benefit form stability too.

arendst:
Why is the stripped down BearSSL enabled in core 2.5.2 but not in stage?
Why is the pubsubclient keepalive in stage but not in 2.5.2?

s-hadinger:
Good suggestions. I forgot to change stage - good point.

It works only with **core_2_5_2**. I agree that making a separate platform.ini entry makes sense.

To my understanding, changing changing `MQTT_KEEPALIVE` in `sonoff_post.h` has no effect on PubSubClient, unless you force the define with a gcc -D option. This is why I put it as `-DMQTT_KEEPALIVE=30`. Happy to know if there's a cleaner way to do it.

Actually I had the same issue with `WEB_LOG_SIZE` which was defined in `sonoff.h` and relying on #defines in `my_user_config.h` - but had no effect work because `sonoff.h` did not include `my_user_config.h`. So I changed it to a #define instead of a const and overloaded the #define in `my_user_config.h`

arendst:
Thx for the clarifications. I noticed the overloading too. You're right that it probably never worked as I intended...

I'll investigate the ramafication of changing KeepAlive from 10 to 30 overall. I don't think it's a big problem. The way it is done now means it won't work with Arduino IDE compiler.

arendst:
And pls find ways to keep it working with a released core version as core stage sometimes results on bricked devices only to be revived by serial firmware uploading.

So I suggest for now make a [core_2_5_2_aws] entry.

arendst:
Just released changes:
- Keepalive is set to 30 in pubsubclient
- Check for weblog size is now in sonoff_post.h allowing you to simply add the AWS define

s-hadinger:
@arendst Sorry I just realized there is still a conflict in `my_user_config.h`.

You need to delete the last lines:
```
#if defined(USE_MQTT_TLS) || defined(USE_MQTT_AWS_IOT)
  #undef WEB_LOG_SIZE
  #define WEB_LOG_SIZE (2000)   // reduce log buffer size when using TLS
#endif
```

It's already merged. Can you do the change, or do I submit another PR?

arendst:
Yes, I'm working on it.

Already changed locally.

Now fixing core 2.4.2 and 2.3.0 compiles.

Will update soon.

arendst:
Done

EYScudeler:
Question: why naming it AWS_IOT? I believe it will work on any MQTT with TLS enabled, right? Can't it just replace the original USE_MQTT_TLS?

s-hadinger:
Indeed but with strong limitations.

My first goal was to connect to AWS IoT which requires Client certificate on top of regular TLS. TO trim down the code size and memory requirements, I limited to TLS 1.2 and to a single cipher.

This may work with other TLS servers as long as they support **TLS 1.2**, **MFLN** and **TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256**

Btw I think we need to add an option to bypass client certificate if TLS is only what you want.

Also keep in mind that AWS IoT does not support MQTT Retained message, so they are marked as non-retained. If you send a Retained MQTT message to AWS IoT, it will drop the MQTT connection.

s-hadinger:
Actually if you are using letsencrypt, certificates are RSA 2048 + SHA 256. So as long as your server supports ECDHE (very common today) and TLS 1.2 (also very common) it should work.

It's more uncertain with regards to MFLN. If the server does not support it, it will still work as long as no message sent by the server is greater than 1024 bytes.

s-hadinger:
Actually I just tested mosquitto with letsencrypt certificate and it works like a charm.

It does make sense to replace the old TLS code relying on deprecated AxTLS with the new one, considering the limitations above.

@arendst Are you ok to dump the old TLS code and use this new one? Limitations seems worth with the low impact on memory (6k + 7k).

I do have an issue: AWS IoT domain names exceed 32 chars, and I know we cannot expand this field without breaking changes. So I concatenated MqttUser and MqttHost. But if we use the same code for TLS, we need back MqttUser and MqttPassword.

@arendst any idea where to store a 63 chars domain name to replace MqttHost?

arendst:
Considering the large amount of users stuck on core 2.3.0 or 2.4.2 (I am as the code size is just below OTA limit making updates easier and I haven't found any advantage in core 2.5.2 for me yet).

- TLS on core 2.3.0 never worked great.
- Does AxTLS work fine with core 2.4.2? 
- I suppose the new TLS code only works from core 2.5.2 right?

I think not many users are using TLS due to lack of memory. If it is possible to make a stable TLS environment from core 2.5.2 I think it would be a good decision to abandon AxTLS and use the new one.

Regarding the 63 chars domain I suggest to use the area at the end of rule3. So rule3 length will decrease by 64 bytes. If you need more use the same procedure. I do not want to start using the last 512 bytes of the flash page yet as it will increase RAM use too (active Settings area is in RAM).

s-hadinger:
@arendst I have good news, with minor changes, BearSSL TLS works fine with core2_4_2. PR will be available tomorrow.

It may be possible to retrofit to core 2_3_0, but the WifiClient diverged too much. I will try to evaluate the code impact.

For 63 char domain, I'm reluctant to reduce Rule3, I'll try to package better the use of MqttHost+MqttUser to make it transparent from Web or Command use.

s-hadinger:
Even better news, it does work on core_2_3_0 and changes are reasonable. That's 12k less code compared to 2.4.2.

My main concern is that I need to add the following line (only in 2.3.0): 

```
  .irom0.text : ALIGN(4)
  {
[...]
    *libbearssl.a:(.literal .text .literal.* .text.*)
[...]
  } >irom0_0_seg :irom0_0_phdr
```
 in the linker scripts that lies in platform.io. I'll try to find a way to store the linker for 2.3.0 into the Tasmota project. @arendst Maybe you know a clean way to do it?

arendst:
Alas, I don't know an easy way to make that change but it will certainly make Arduino IDE compilations impossible.

So for now I would refrain from using core 2.3.0 with bearssl.

s-hadinger:
Good advice. I did more testing with 2.3.0 and TLS behavior is erratic and randmon. Let's drop it for TLS.

So I'll stick to 2.4.2 and 2.5.2, it works fine on both.

PR #5971, opened 2019-06-21
Changed from double to float in Sunrise and sunset, -3.4k

s-hadinger:
## Description:

Converted all `double` to `float` which are more than enough for this calculation. Since `float` takes only 1 register, it saves a lot of housekeeping code. Also switched to `sinf()` and other `float` equivalents of trigonometric functions.

My German is too rusty to translate code in English :)

Saves 3.4k of flash.

## Checklist:
  - [ ] Only relevant files were touched (Also remember to update _changelog.ino_ file)

### DIFF
--- a/sonoff/xdrv_09_timers.ino
+++ b/sonoff/xdrv_09_timers.ino
@@ -60,11 +60,11 @@ int8_t timer_window[MAX_TIMERS] = { 0 };
  *         Rewrite for Arduino by 'jurs' for German Arduino forum
 \*********************************************************************************************/
 
-const double pi2 = TWO_PI;
-const double pi = PI;
-const double RAD = DEG_TO_RAD;
+const float pi2 = TWO_PI;
+const float pi = PI;
+const float RAD = DEG_TO_RAD;
 
-double JulianischesDatum(void)
+float JulianischesDatum(void)
 {
   // Gregorianischer Kalender
   int Gregor;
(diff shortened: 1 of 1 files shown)


arendst:
I love this!!!

s-hadinger:
I realize we don't have a wiki page with guidance about how to keep code small. I didn't find any on the web for ESP8266. I'm happy to start one when I find time.

Things like:

- don't use `double` use `float`
- don't concatenate `char[]`, use `String`
- sometimes using uint32_t is better than uint8_t because register are 32 bits anyways
...

meingraham:
@s-hadinger perhaps an expansion to code development in general and then [this article](https://github.com/arendst/Sonoff-Tasmota/wiki/Sensor-API) as a reference to Tasmota drivers in general?

arendst:
I prefer char over string as it doesn't use heap which in turn increases memory fragmentation. Code is also smaller as (esp8266) string is a wrapper over char anyway

s-hadinger:
Sorry, I was unclear. When adding to a `String` object, I first did the following:

```
      light_status += "\"xy\":[" + String(x, 5) + "," + String(y, 5) + "],";
```

then realized it's better to do so:

```
      light_status += "\"xy\":[";
      light_status += String(x, 5);
      light_status += ",";
      light_status += String(y, 5);
      light_status += "],";
```

Btw I cannot use sprintf because converting float to string is disables to save code space.

s-hadinger:
I posted the tips for compact code at the end of the page: https://github.com/arendst/Sonoff-Tasmota/wiki/Sensor-API#keeping-esp8266-code-compact

Also added a table of content as the page is getting very long.

arendst:
Thx Stephan, just came back from a short holiday and read your comments about code optimization.

I guess I have to change my use of uint8_t ...

BTW for float in printf I supply dtostrfd

s-hadinger:
@arendst I'm still chasing some other double to be converted to float, and plan to change the stdlib to a more compact version of sin/cos/tan... It will save another 3k or so.

I also found an issue in the way `String` are used especially in the Hue emulation code. Whenever there is:

```
s->replace("{re", "some string");
```

Two `String` objects are silently created from the string literals, and destroyed just after. This adds up to 600 bytes of completely useless code. I tried overloading String and adding additional methods to accept `(const char *)` parameters, but I'm not super happy with the result. I'm considering re-writing parts of the Hue code with more compact code.

PR #6005, opened 2019-07-01
Converted double to float in rules, and replaced trigonometric functions from stdlib with smaller versions.

s-hadinger:
## Description:

**Code reduction: 7.6k**

Continuing to chase down `double` in favor of `float`

- Changed `CharToDouble` to `CharToFloat`
- Changed `double` to `float` in Rules
- Replaced all trigonometric functions (sin, cos, tan, asin, acos, atan) with compact code designed for embedded systems

Note: the result is slightly less precise but gives at leas 5-6 digits of precision, which is more than enough in most use cases - especially calculating the sunrise/sunset.

## Checklist:


arendst:
Whoow!!

Just compiled my stripped down version and it even saved 7.5k code and 1k RAM. 

This is really great.

meingraham:
Stephan - you are definitely on Theo's Christmas card list ;-)

arendst:
But considering #6012 we do have to find a workaround for the unavailable fmodf function in core 2.3.0

PR #6019, opened 2019-07-04
Add custom build environment

claybar:
## Description:

Currently to build a custom build the platformio.ini file needs to be edited.  This method has a few downsides,
- The edited file needs to often be stashed when doing a git pull
- The final build is still labeled according to the build env (sonoff, display etc).  Makes it difficult to establish the build is actually custom
- Custom builds are difficult to perform on an automated system because of the editing required

Some of these issues could be solved by having a copy of the repo, but then needs to be rebased regularly.

This PR attempts to solve this by adding a new env to the build `sonoff-custom`, automatically setting USE_CONFIG_OVERRIDE.

There is one current issue, the build will fail on this new environment because the file `user_config_override.h` does not exist by default.  I haven't got a solution to this as yet, but will work on this is if the general idea of this PR is acceptable.  I welcome any input on this.

## Checklist:
  - [x] Only relevant files were touched (Also remember to add to _changelog.ino_ file without updating version)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -18,6 +18,7 @@ src_dir = sonoff
 ;env_default = sonoff-knx
 ;env_default = sonoff-sensors
 ;env_default = sonoff-display
+;env_default = sonoff-custom
 ;env_default = sonoff-BG
 ;env_default = sonoff-BR
 ;env_default = sonoff-CN
--- a/platformio.ini
+++ b/platformio.ini
@@ -276,6 +276,20 @@ upload_resetmethod = ${common.upload_resetmethod}
 upload_speed = ${common.upload_speed}
 extra_scripts = ${common.extra_scripts}
 
+[env:sonoff-custom]
+platform = ${common.platform}
+framework = ${common.framework}
+board = ${common.board}
+board_build.flash_mode = ${common.board_build.flash_mode}
+board_build.f_cpu = ${common.board_build.f_cpu}
+build_unflags = ${common.build_unflags}
+build_flags = ${common.build_flags} -DFIRMWARE_CUSTOM -DUSE_CONFIG_OVERRIDE
+monitor_speed = ${common.monitor_speed}
+upload_port = ${common.upload_port}
+upload_resetmethod = ${common.upload_resetmethod}
+upload_speed = ${common.upload_speed}
+extra_scripts = ${common.extra_scripts}
+
 [env:sonoff-BG]
 platform = ${common.platform}
 framework = ${common.framework}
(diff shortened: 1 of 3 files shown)


claybar:
As expected this PR did not build successfully on Travis because of the missing file issue of the new `sonoff-custom` env.

It also failed because of `sonoff-sensors`, but this is because of an issue in the upstream `development` branch.  Once this is sorted I will merge it into this PR.

meingraham:
@claybar 

There are occasions when one wants to build sonoff, sonoff-sensors, sonoff-minimal, etc., while also using user_config_override.h. Thus "custom" is not the only time the compiler directive is used.

The only line in platform.ini that needs to be modified is to uncomment -DUSE_CONFIG_OVERRIDE in the [common] section. This then gives you the option of compiled any of the firmware variants while also using one's own custom settings.

I would vote down adding this section as "custom" can have differing meanings for users and may only add confusion. When one retrieves a fresh copy of the code, all you need to do is edit the one line and now you can compile any "custom" variant.

Of course, you are always free to maintain your own fork which gives you the option of doing it the way you prefer.

Regards.

Mike

arendst:
I think Mike stated it well.

It's almost the same procedure I use (I just keep my own copy of platformio.ini and merge changes when the come along).

ascillato2:
Closing this PR as it is not going to be merged.

Thanks for sharing your ideas.

claybar:
While frustrating, I can see some logic behind this.  Although I agree changing one line is all it takes (and this still remains in this PR), this is still not ideal for automated systems.  You also end up with a binary without a name indicating it is not a 'standard' build.

I will move forward with a custom fork I think.  And hope there isn't a CODE_IMAGE 7 added to the main  repo.

meingraham:
@claybar 

You may want to get in touch with @andrethomas. He developed and maintains the scripting that builds all of the Tasmota firmware variants in the hackbox repository. I believe he uses PlatformIO and generates all of the binaries for each of three different cores with all of the associated variants (sonoff, sonoff-classic, sonoff-sensors, sonoff-minimal, etc.). He may be able to provide some guidance on the automation of binary compilation while providing your desired customized settings as well as specific binary naming options.

Cheers!

Mike

claybar:
Thanks, I will keep that in mind..

I wasn't referring to the filename of the binary, that is easily changed with a script.  I am more meaning the string contained within the binary, the one that is displayed on the 'information' page of the webserver.  This will display something like `(sonoff)` or '(display)` even when the config has been changed.  This string comes from `i18n.h` and is selected by the value of CODE_IMAGE, hence my comment above about using a new value of 7.  Even if I do this in a custom fork it may get clobbered one day if the main repo ever uses 7 for something else.

meingraham:
@claybar 

re: filename & image name - Andre's build script incorporates both.

Issue #5704, opened 2019-04-25
Change calculation of white on RGBW LED strips


stefanbode:
in xdv_04_light.ino i changed following:
```
  light_current_color[0] = (uint8_t)(r * 255.0f);
  light_current_color[1] = (uint8_t)(g * 255.0f);
  light_current_color[2] = (uint8_t)(b * 255.0f);
  if(light_ct_rgb_linked){
    //stb mod
    light_current_color[3] = (uint8_t)( ( (r < g && r < b) ? r : (g < b) ? g : b )    * 255.0f);
    //end
    light_current_color[4] = 0;
  }
}
```
If using RGBW LED stripes with hue or any other app only RGB is transmitted and you do not use the benefits of the white light. The change set the white LED to the minimum value of all colors. This increases light emission and still allow to use pure red/green....

s-hadinger:
Very interesting. Calculation of channels changed in PR #5702. I could easily retrofit this one.

Open question, should we add a new Option for this behavior or take it as the standard behavior?

stefanbode:
The fix is a one-line fix and does not really take into consideration the intensity and color of the white light compared to the RGB. The change is only 100% correct if the emitted color of the RGB is the same as the white one, which is for sure not true. 

Therefore I assume it is not a solution for everyone. A SetOption should make it valuable for everyone. I can confirm, that for indirect lightning with the iOS App Huemote the visual effect is great and feel natural.

s-hadinger:
Afterthoughts I can see 3 possible behaviors:

1. (current implementation) RGB and W are independent. Until you trigger the CT mode (ex: "ct 500" and use dimmer), the White channel is not used. Pros: white in CT mode is pure. Cons: RGB-mode white is not as pure as white leds.

2. (your implementation) in RGB mode, set W channel to the min of RGB - ie. (255-sat). Pros: adds some pure white to the RGB-white, Cons: white is still not as pure as white leds, and you might double the brightness. It's also probable the you will desaturate non-fully saturate colors.

3. (alternative) in RGB mode, set W channel to the unsaturated value (ie min of RGB, or 255-sat), then substract this value to RGB channels. As a side effet, RGB is never used for white and one component is always 0. This would gradually move from RGB to W. Pros: you keep the full gamut of colors and white is pure. Cons: it depends on the compared brightness of RGB and W leds.

[...]

PR #6032, opened 2019-07-07
Add blend RGB leds with White leds for better whites #5895 #5704

### LINKED ISSUE #5895: Enhancement of the color command (RGBW)
**Have you look for this feature in other issues and in the wiki?**
yes

The real white of a SL6812 RGBW LED can not be controlled with the color 12 (rgb white) command

An enhancement of the color command:
color 13 (using the real white: 000000FF)
color 14 (using the rgb & real white: FFFFFFFF)
If possible, control via alexa to the new colors

**Additional context**

s-hadinger:
## Description:

When setting last argument of `rgbwwtable`to zero, like this `rgbwwtable 255,255,255,255,0`, color will be automatically blended between RGB and White. Any White component will be taken off RGB and used on White leds.

Changing the 4th argument can be used to calibrate White vs RGB: often white leds are more bright than RGB. For example, Sonoff B1 would require `rgbwwtable 255,255,255,35,0`.

Works only with lights with 4 (RGBW) or 5 (RGBCW) channels.

@arendst I bumped to version to 6.6.0.1.

**Related issue (if applicable):** fixes #5895 #5704

## Checklist:
  - [x] Only relevant files were touched (Also remember to add to _changelog.ino_ file without updating version)

### DIFF
--- a/sonoff/_changelog.ino
+++ b/sonoff/_changelog.ino
@@ -1,4 +1,7 @@
 /*********************************************************************************************\
+ * 6.6.0.1 20190707
+ * Add blend RGB leds with White leds for better whites #5895 #5704
+ *
  * 6.6.0 20190707
  * Remove support of TLS on core 2.3.0 and extent support on core 2.4.2 and up
  * Remove MQTT uptime message every hour
--- a/sonoff/xdrv_04_light.ino
+++ b/sonoff/xdrv_04_light.ino
@@ -101,23 +101,27 @@
  *
  * 3.  In LightAnimate(), final PWM values are computed at next tick.
  *  .a If color did not change since last tick - ignore.
- *  .b Apply color balance correction from rgbwwTable[]
- *  .c Extend resolution from 8 bits to 10 bits, which makes a significant
+ *  .b Extend resolution from 8 bits to 10 bits, which makes a significant
  *     difference when applying gamma correction at low brightness.
- *  .d Apply Gamma Correction if LedTable==1 (by default).
+ *  .c Apply Gamma Correction if LedTable==1 (by default).
  *     Gamma Correction uses an adaptative resolution table from 11 to 8 bits.
- *  .e For Warm/Cold-white channels, Gamma correction is calculated in combined mode.
+ *  .d For Warm/Cold-white channels, Gamma correction is calculated in combined mode.
  *     Ie. total white brightness (C+W) is used for Gamma correction and gives
  *     the overall light power required. Then this light power is split among
  *     Wamr/Cold channels.
- *  .f Gamma correction is still applied to 8 bits channels for compatibility
+ *  .e Gamma correction is still applied to 8 bits channels for compatibility
  *     with other non-PMW modules.
- *  .g Avoid PMW values between 1008 and 1022, issue #1146
- *  .h Scale ranges from 10 bits to 0..PWMRange (by default 1023) so no change
+ *  .f Apply color balance correction from rgbwwTable[].
+ *     Note: correction is done after Gamma correction, it is meant
+ *     to adjust leds with different power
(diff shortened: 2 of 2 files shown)


meingraham:
@s-hadinger @emontnemery 

Would it be possible to have a write-up of this command so that I can add it to the wiki?

I can start with [this](https://github.com/arendst/Sonoff-Tasmota/pull/3933#issue-219152426) but it may not be comprehensive enough. Plus, the latest modifications likely need special explanation as well.

Thanks.

Mike

s-hadinger:
Yes of course. Will do this week. :)

meingraham:
@s-hadinger any progress on a description for the `rgbwwtable` command for inclusion in the wiki?

s-hadinger:
Here is the blurb. I would appreciate you to add it to the wiki, I edited twice the Tasmota Commands page but I my Github account got suspended twice - maybe the page is too long.

-------

RGBWWTable

-------

Adjusts the range of each PWM channel: 255 is full range, any lower number will reduce the range of the channel. It can be used to adjust the power of each led in case some leds are brighter than others. You normally don't need to use this.

Default: `RGBWWTable 255,255,255,255,255` all channels use full power.
Ex: `RGBWWTable 51,51,51,51,51` limit max power to 25% for all channels.

_Note: range adjustment is computed **after** Gamma correction._

**White blend mode** _for 4 and 5 channels (RGBW and RGBWC)_

In this mode, any white component will be removed from RGB leds and sent to White leds, whites look much better. For ex: if you set `color 30508000`, it will be internally converted to `color 0020503000` (0x30 is substracted from R,G,B channels and added to White channel)

You enable White blend mode by setting the last channel to zero: `RGBWWTable 255,255,255,255,0`

Generally, white leds are brighter than RGB, if you want to keep the same brightness, you need to calibrate the white level. This is done with `RGBWWTable 255,255,255,<n>,0` where <n> lowers the white channel.

To calibrate:

1. `Color FFFFFF00`
2. `RGBWWTable 255,255,255,255,255` - reset to RGB mode
3. `RGBWWTable 255,255,255,<n>,0` - start with n=150
4. If too bright lower <n>, if too dark increase <n>
5. go back to 2. and iterate

Example:

- Sonoff B1: `RGBWWTable 255,255,255,35,0`
- Teckin SB50: `RGBWWTable 255,255,255,80,0`

meingraham:
Will take care of adding this.

meingraham:
@s-hadinger it says 51 would be a 25% setting. Did you mean 20%?

s-hadinger:
Yes you're right. Let's put 64 instead avec 51.

meingraham:
Broke it up and moved some portions around...
https://github.com/arendst/Sonoff-Tasmota/wiki/Commands#rgbwwtable
https://github.com/arendst/Sonoff-Tasmota/wiki/Commands#color
https://github.com/arendst/Sonoff-Tasmota/wiki/white-blend-mode

meingraham:
@s-hadinger 

Is there any interaction between rgbwwtable and [pwmrange](commands#pwmrange)? That command allows the "range" to be extended. I want to avoid any confusion with the use of the term "range" in rgbwwtable and either "link" the topics so a user considers both or explicitly disassociate the documentation of the two settings.

s-hadinger:
Good question. Actually no, it's unrelated. PWM is done in software on ESP8266n, so there is a loop at `PMWFrequency` that sets the Output to 1 for `PWM<x>` cycles out of `PWMRange`cycles.

PMWRange only defines the length of the internal loop. Actually it would be better named **PWMResolution**.

This is how you can lower from 10 bits PWM to 8 bits PWM - although there is little reason you want this.

digiblur:
@s-hadinger would it be possible to still allow the full white channel brightness when pure white is called? This new feature works great for colors except for losing some white brightness of a light.

PR #6063, opened 2019-07-10
Use latest SDK22y for core_stage

Jason2866:
Switch default FW to "2.2.2-dev(38a443e)" (menu:2.2.1+100) 

* enable by default latest 2.2.x firmware, including fixed espnow
* LittleFS: avoid crash when FS size is 0
* flash size defaults: 1M for generic board, not empty FS for all

## Description:

## Checklist:

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -99,8 +99,10 @@ build_flags               = ${esp82xx_defaults.build_flags}
 ; Code optimization see https://github.com/esp8266/Arduino/issues/5790#issuecomment-475672473
                             -O2
                             -DBEARSSL_SSL_BASIC
+; nonos-sdk 22y
+                            -DPIO_FRAMEWORK_ARDUINO_ESPRESSIF_SDK22y
 ; nonos-sdk 22x
-                            -DPIO_FRAMEWORK_ARDUINO_ESPRESSIF_SDK22x
+;                            -DPIO_FRAMEWORK_ARDUINO_ESPRESSIF_SDK22x
 ; nonos-sdk-pre-v3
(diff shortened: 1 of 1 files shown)


ascillato:
I'm testing this new version of the SDK on several devices at once and there was zero mqtt reconnects.

Can anyone also confirm this?

In a TP-LINK wifi router I have zero mqtt disconnections using core 2.3.0 but with 2.4.2, 2.5.0, 2.5.1 and 2.5.2 I have several mqtt disconnections per hour. Now with SDK2.2y I don't have issues.

Jason2866:
It performs well for me too. With core 2.5.2 (22x) i had some reconnects.
On a test device since yesterday with stage core (22y) i had none (AP LANCOM LN830acn)

arendst:
A pity 22y is using another 328 bytes of iram.

In the end we have a great core environment without any features ;-)

ascillato:
I continue testing this new core. It is way more stable than 2.5.2.

I found that if RSSI <38 it has wifi/mqtt reconnects, but with a RSSI >38 it works fantastic.

So, for future troubleshooting, if a device has RSSI >20 and <38 and can't be improved, using core 2.3.0 is a workaround for that. It is very slow in that conditions but it don't have mqtt disconnections.

araczkowski:
hi, I'm testing this too
6.6.0.4(sonoff)
Core 2_5_2
SDK 2.2.2-dev(c0eb301)
my own build from yesterday
It looks really promising, no mqtt disconnects so far, on more than 20 devices (different types).

**hope this really will be  the great core environment** :)
thanks and regards!

image

image

PR #6179, opened 2019-08-04
Change Store AWS IoT Private Key and Certificate in SPI Flash avoiding device-specific compilations

### LINKED ISSUE #5860: Best place to store AWS IoT client private_key+certificate?
Apologies if it’s the wrong place to discuss a development question.

I’m reviving an old topic: **connecting Tasmota to AWS IoT for remote MQTT, and still using Web/Hue Emulation**.

I've been working on it for the last weeks and I have a working prototype using as low **6.3KB of heap memory** in nominal state (+ 1KB of stack), and an additional **transient 8.0KB** during TLS handshake (connect or reconnect). TLS handshake takes 1.3s at 80MHz or 0.7s at 160MHz.

### Here is my question:
Each device needs to have a distinct TLS Private Key and Certificate. They can be in string (PEM) format but would ideally be stored as binary format to spare some decoding code. Size is ~1.5KB. What would be the best option to store them?

1. Custom compile each firmware with key burnt in it?
2. Would it make sense to burn a 4KB PROGMEM zone, and use spi_flash_write to update this Flash block? Did somebody already do this?
3. SPIFFS? But I know we don't have SPIFFS on Tasmota

### For the curious:
For the curious, here are some implementation details about memory optimization.

It is now possible to have both Web/Hue and TLS because of the following (see #2031)
1. Web interface now uses chunks (thanks @arendst), hence a reduced need for heap memory.
2. Arduino moving from AXTLS to BearSSL, it allows to dramatically reduce RAM requirements. 

Even if BearSSL use little memory, the stock WifiSecureClient takes at least 18KB of heap.
I was able to aggressively reduce the memory requirements:
-	**MFLN support**. Normal SSL/TLS requires 16KB buffers – which is huge for an ESP8266. Fortunately AWS IoT supports TLS1.2 and MFLN (Max Fragments Length). MFLN allows to reduce send/receive buffers to 2x1024 bytes. Note: using 512 bytes buffers causes packet drops and very slow TLS handshake.
-	**Elliptic Curve (EC) Private Key**. Default Private Key in AWS IoT uses RSA 2048 bits keys. Switching to EC 256 bits key saves a lot of space, for increased security and slightly faster handshake.
-	**ECDHE_RSA_WITH_AES_128_GCM_SHA256 cipher**. To save code and RAM, it only supports the previous algorithm. ECDHE with EC key, RSA certificate on server side, AES 128 which is 40% faster than AES256, GCM instead of CBC.
-	**Fingerprint validation**. The server certificate is not validated against the CA. Instead there are two options: either ignore server certificate (not super secure), or learn fingerprint of first connection. The fingerprint is calculate on the server’s public key instead of certificate, this is more robust in case of a certificate renewal of the server with the same public key.
-	**Custom Thunk Stack**. Arduino uses a secondary stack of 5.6KB mainly used during handshake. Instead of constantly using this memory, it is only allocated during handshake. For normal TLS communication after handshake, AES GCM 128 does not need more than 1KB, so the normal ESP8266 stack is enough.
-	**P256 curve**. To save code size , it only supports the P256 elliptic curve.

s-hadinger:
## Description:

Following discussion on #5860, the AWS IoT credentials (~800 bytes) are now moved from source code to SPI Flash in unused location in bank FF.

This means that you will not need to compile device-specific firmwares. The same AWS IoT enabled firmware can be used on multiple devices, and the credentials (Private Key and Certificate) can be loaded with the new `TLSKey` command.

Added base64-1.1.1 Arduino library, used only when AWS IoT is enabled.

Code impact: +500 bytes when AWS IoT is enabled. No impact on standard firmware.

**Related issue (if applicable):** fixes #5860

## Checklist:
only 2.4.2 and 2.5.2


s-hadinger:
Note: I will change the wiki page after the PR is merged.

arendst:
Thx for the implementation.

As it stands it will only work with 1M devices; the location has to move with other flash sizes.

I will refactor with correct offset based on used linker script (as I do with OTA too) and move the code also to settings.ino as there the flash routines are currently stored.

s-hadinger:
Thanks Theo. I didn’t think of 4M devices. Appreciated.

PR #6186, opened 2019-08-06
support for the CHIRP soil moisture sensor

Staars:
This is a non-blocking driver for for the "CHIRP soil moisture sensor" -https://wemakethings.net/chirp/. 

No external libs are needed.

It should support multiple sensors, but I only have one for development, so no real-world-test for this feature yet.

I encountered recurrent problems with the (then frozen with error:4) I2C-bus, especially regarding sleep/wake. That's why the auto-sleep-wake feature is currently deactivated.

The driver will scan the whole I2C-bus for sensors, because the address can be changed and 0x20 is only the default address. So we only configure I2C-pins and the rest shall work automatically.

Sensors will be enumerated starting with 0. 
The active sensor can be setected with CHIRPSELECT 0 .... x. Then this sensor can receive further commands like CHIRPRESET, CHIRPSLEEP, CHIRPWAKE and CHIRPSET. The last commands needs an I2C-address (in decimal) as an argument to change the sensor address (which will survive a device reset). CHIRPSET 31 will change address to 0x1f.

Tele-message-example:
RSL: tele/sonoff/SENSOR = {"Time":"2019-08-06T21:12:30","Epoch":1565122350,"CHIRP0":{"Moisture":252,"Temperature":26.8,"Illuminance":"5356},"TempUnit":"C"}

The sensor-read-cycle will be synced with the tele period.

I know that some things are left to do, before a merge is possible, but I think it is easier to discuss things with the code on the table.

Naming conventions? 
Shall we name it MOISTURE, HUMIDITY or CAPACITANCE?
The light data is dimensionless and not even close to LUX. A higher value means less light, so the use of ILLUMINANCE with lx (at least in the WEB GUI) is misleading.

At the moment there are lots of debug messages left in the driver, which shall be commented out in the final version.

The values for Wire.setclockstretchlimit and Wire.setclock are subject to change. Maybe it will be necessary to use special I2C-functions instead of the used generic TASMOTA-versions (I2CWrite and so on), where a delay(X) must be inserted here and there.

### DIFF
--- a/sonoff/xsns_48_chirp.ino
+++ b/sonoff/xsns_48_chirp.ino
@@ -124,7 +124,7 @@ void ChirpReset(uint8_t addr) {
 /********************************************************************************************/
 
 void ChirpResetAll(void) {
-    for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+    for (uint32_t i = 0; i < chirp_found_sensors; i++) {
       if (chirp_sensor[i].version) { 
         ChirpReset(chirp_sensor[i].address);
         }
@@ -146,7 +146,7 @@ void ChirpSleep(uint8_t addr) {
 /********************************************************************************************/
 
 // void ChirpSleepAll(void) {
-//     for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+//     for (uint32_t i = 0; i < chirp_found_sensors; i++) {
 //       if (chirp_sensor[i].version) { 
 //         ChirpSleep(chirp_sensor[i].address);
 //         }
@@ -156,7 +156,7 @@ void ChirpSleep(uint8_t addr) {
 // /********************************************************************************************/
 
 // void ChirpAutoWakeAll(void) {
-//     for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+//     for (uint32_t i = 0; i < chirp_found_sensors; i++) {
 //       if (chirp_sensor[i].version && !chirp_sensor[i].explicitSleep) { 
 //         ChirpReadVersion(chirp_sensor[i].address);
 //         }
@@ -178,7 +178,7 @@ void ChirpSelect(uint8_t sensor) {
 /********************************************************************************************/
 
 bool ChirpMeasureLight(void) {
-  for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+  for (uint32_t i = 0; i < chirp_found_sensors; i++) {
     if (chirp_sensor[i].version && !chirp_sensor[i].explicitSleep) { 
       uint8_t lightReady = I2cRead8(chirp_sensor[i].address, CHIRP_GET_BUSY);
       AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: busy status for light for sensor %u", lightReady);
@@ -195,7 +195,7 @@ bool ChirpMeasureLight(void) {
 /********************************************************************************************/
 
 void ChirpReadCapTemp() { // no timeout needed for both measurements, so we do it at once
-    for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+    for (uint32_t i = 0; i < chirp_found_sensors; i++) {
       if (chirp_sensor[i].version && !chirp_sensor[i].explicitSleep) { 
         AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: now really read CapTemp for sensor at address 0x%x", chirp_sensor[i].address);
         chirp_sensor[i].moisture = I2cRead16(chirp_sensor[i].address, CHIRP_GET_CAPACITANCE);
@@ -208,7 +208,7 @@ void ChirpReadCapTemp() { // no timeout needed for both measurements, so we do i
 
 bool ChirpReadLight() {   // sophisticated calculations could be done here
   bool success = false;
-  for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+  for (uint32_t i = 0; i < chirp_found_sensors; i++) {
     AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: will read light for sensor %u", i);
     if (chirp_sensor[i].version) {
       if (I2cValidRead16(&chirp_sensor[i].light, chirp_sensor[i].address, CHIRP_GET_LIGHT)){
@@ -352,7 +352,7 @@ void ChirpEverySecond(void)
 
 void ChirpShow(bool json)
 {
-  for (uint8_t i = 0; i < chirp_found_sensors; i++) {
+  for (uint32_t i = 0; i < chirp_found_sensors; i++) {
     if (chirp_sensor[i].version) {
       // convert double values to string
       char str_moisture[33];      
--- a/sonoff/xsns_48_chirp.ino
+++ b/sonoff/xsns_48_chirp.ino
@@ -168,10 +168,10 @@ void ChirpSleep(uint8_t addr) {
 void ChirpSelect(uint8_t sensor) {
   if(sensor < chirp_found_sensors) { //TODO: show some infos
     chirp_current = sensor;
-    AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: Sensor %u now active.", chirp_current);
+    AddLog_P2(LOG_LEVEL_DEBUG, PSTR("CHIRP: Sensor %u now active."), chirp_current);
   }
   if (sensor == 255) {
-    AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: Sensor %u active at address 0x%x.", chirp_current, chirp_sensor[chirp_current].address);
+    AddLog_P2(LOG_LEVEL_DEBUG, PSTR("CHIRP: Sensor %u active at address 0x%x."), chirp_current, chirp_sensor[chirp_current].address);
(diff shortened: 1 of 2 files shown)


s-hadinger:
Thanks for the PR.

Checks failed here:

```
/home/travis/build/arendst/Sonoff-Tasmota/sonoff/xsns_48_chirp.ino: In function 'void ChirpClockSet()':
/home/travis/build/arendst/Sonoff-Tasmota/sonoff/xsns_48_chirp.ino:138:5: error: 'Wire' was not declared in this scope
     Wire.setClockStretchLimit(4000);
     ^
*** [.pioenvs/sonoff-minimal/src/sonoff.ino.cpp.o] Error 1
```

sonoff/xsns_48_chirp.ino
```diff
@@ -0,0 +1,484 @@
...
+
+
+  ---
+  1.0.0.0 20190608  started   - further development by Christian Baars  - https://github.com/Staars/Sonoff-Tasmota
+                    forked    - from arendst/tasmota                    - https://github.com/arendst/Sonoff-Tasmota
+                    base      - code base from arendst and              - https://github.com/Miceuz/i2c-moisture-sensor
+
+*/
+#define USE_I2C
+#define USE_CHIRP
```
s-hadinger: I believe you forgot to remove

```
#define USE_I2C
#define USE_CHIRP
```

sonoff/xsns_48_chirp.ino
```diff
@@ -0,0 +1,484 @@
...
+
+/********************************************************************************************/
+
+void ChirpSelect(uint8_t sensor) {
+  if(sensor < chirp_found_sensors) { //TODO: show some infos
+    chirp_current = sensor;
+    AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: Sensor %u now active.", chirp_current);
+  }
+  if (sensor == 255) {
+    AddLog_P2(LOG_LEVEL_DEBUG, "CHIRP: Sensor %u active at address 0x%x.", chirp_current, chirp_sensor[chirp_current].address);
```
s-hadinger: You can use PSTR() to force text in PROGMEM, like:

`AddLog_P2(LOG_LEVEL_DEBUG, PSTR("CHIRP: Sensor %u now active."), chirp_current);`

sonoff/xsns_48_chirp.ino
```diff
@@ -0,0 +1,484 @@
...
+        chirp_sensor[i].temperature = I2cRead16(chirp_sensor[i].address, CHIRP_GET_TEMPERATURE); 
+        }
+    }
+}
+
+/********************************************************************************************/
+
+bool ChirpReadLight() {   // sophisticated calculations could be done here
+  bool success = false;
+  for (uint8_t i = 0; i < chirp_found_sensors; i++) {
```
s-hadinger: You can use `uint32_t` in loops, in sometimes make the code smaller, see https://github.com/arendst/Sonoff-Tasmota/wiki/Sensor-API#user-content-keeping-esp8266-code-compact

s-hadinger (commented):
Great PR. I just put some remarks and suggestions.

Staars:
Thank's a lot for responding. 
The API-documentation in the WIKI has improved a lot, I will take a read!!

Slightly off-topic: 
Do we have a common mechanism to build a DEBUG-driver? I tend to use the LOG a lot while developing and most of these messages are not useful in the "final product", so in the end I uncomment them. This is no big deal, but would it make sense to you to introduce something like a conditional compiling ADDLOG_DEBUG() and a new build target like sonoff-debug (or sonoff-verbose), which an end user will only flash, if he runs into trouble and wants to communicate things to the developer?

s-hadinger:
Thanks for the changes.

I will let Theo decide whether `#define USE_CHIRP` should be enabled by default, or custom compiled.

I agree about debugging, I ended up putting `#define DEBUG_xxx` at the head of the file, but I didn't find a consistent way to do it.

arendst:
I'll have to find a way to easily add debugging enable/disabling at compile time. I see problems when this is enabled globally as there will be too many debugging messages popping up leading to stack/heap overflows.

I guess it must be sensor/driver specific.

It's on my to do list...

Staars:
I hope it is okay to ask a tiny question here.

To achieve the debugging functionality at compile time I tried out a macro, which seems to work (to some extent).

It goes like:

#define DEBUG_DRIVER
#ifdef DEBUG_DRIVER
  #define AddLog_Debug(X,Y,...)   AddLog_P2(X,Y,##__VA_ARGS__)
#else
  #define AddLog_Debug(X,Y,...)   
#endif

I get a warning message:
warning: __VA_ARGS__ can only appear in the expansion of a C99 variadic macro [enabled by default]
   #define AddLog_Debug(X,Y,...)   AddLog_P2(X,Y,##__VA_ARGS__)

But it seems to work. 
My rough idea is, that we have a new build target for (highly verbosive) debugging drivers, where rarely needed/debug log messages are compiled and the "normal" build target gets only the vital log messages.
This should not break anything old and must actively added to older code, just by search and replace AddLog_P2 to AddLog_Debug, where it makes sense.

arendst:
Thx.

I've been looking at the Arduino core code to see how they did it in ESP8266HTTPClient.h:

```
#ifdef DEBUG_ESP_HTTP_CLIENT
#ifdef DEBUG_ESP_PORT
#define DEBUG_HTTPCLIENT(...) DEBUG_ESP_PORT.printf( __VA_ARGS__ )
#endif
#endif

#ifndef DEBUG_HTTPCLIENT
#define DEBUG_HTTPCLIENT(...)
#endif
```
Looks the same as yours and doesn't give the C99 warning probably due to ``__VA_ARGS__``. When changed for Tasmota it may look like this:
```
#define DEBUG_DRIVER
#ifdef DEBUG_DRIVER
#define ADDLOG_DEBUG(...) AddLog_Debug(__VA_ARGS__)
#else
#define ADDLOG_DEBUG(...)
#endif
```
with an additional handler like:
```
void AddLog_Debug(PGM_P formatP, ...)
{
  va_list arg;
  va_start(arg, formatP);
  vsnprintf_P(log_data, sizeof(log_data), formatP, arg);
  va_end(arg);

  AddLog(LOG_LEVEL_DEBUG);
}
```
which keeps parameters at a minimum.

The only problem with this is the warnings I get of the redefine due to Arduino INO processing:
```
C:/Users/Theo/Google Drive/Work/VisualStudio/Tasmota/dev/gd-dev-v6.6.0/sonoff/xdrv_09_timers.ino:54:0: warning: "ADDLOG_DEBUG" redefined [enabled by default]
 #define ADDLOG_DEBUG(...)
 ^
C:/Users/Theo/Google Drive/Work/VisualStudio/Tasmota/dev/gd-dev-v6.6.0/sonoff/xdrv_09_timers.ino:52:0: note: this is the location of the previous definition
 #define ADDLOG_DEBUG(...) AddLog_Debug(__VA_ARGS__)
 ^
```

I have to find out why these warnings do not appear with the Arduino code but it's probably due to the .cpp file instead of .ino.

Investigating...

Staars:
Good luck!
(ATM I have no better idea ... but will try to find a solution)

Some redefinition warnings may be related to not doing a 'make clean' before.

arendst:
Got it!

Moved the defines to sonoff_post.h and the enable define to either my_user_config.h or user_config_override.h

That way it allows control over it while not showing warnings.

Now I have to decide of making several types of these defines. I think for now I just leave it at two: DEBUG_SENSOR and DEBUG_DRIVER.

Staars:
Cool.
In the end it is a design decision and a bit bit of personal preference.

I am not sure if I understand the this handler correctly 
`void AddLog_Debug(PGM_P formatP, ...)
{
  va_list arg;
  va_start(arg, formatP);
  vsnprintf_P(log_data, sizeof(log_data), formatP, arg);
  va_end(arg);

  AddLog(LOG_LEVEL_DEBUG);
}`

How do we avoid, that this compiles?

"My" macro will let the debug-log-code line simply disappear, if DEBUG (or something) is not defined.

arendst:
The same happens with my solution too as long as you do not use the AddLog_Debug function directly; if no define is made the function disappears too as it is not linked in by the compiler.

So soluton for now will be:

I'll add this to sonoff_post.h
```
#ifdef DEBUG_TASMOTA_DRIVER
#define DEBUG_DRIVER_LOG(...) AddLog_Debug(__VA_ARGS__)
#else
#define DEBUG_DRIVER_LOG(...)
#endif

#ifdef DEBUG_TASMOTA_SENSOR
#define DEBUG_SENSOR_LOG(...) AddLog_Debug(__VA_ARGS__)
#else
#define DEBUG_SENSOR_LOG(...)
#endif
```
I'll add this to my_user_config.h
```
#define DEBUG_TASMOTA_DRIVER                     // Enable driver debug messages
//#define DEBUG_TASMOTA_SENSOR                     // Enable sensor debug messages

```
Then as an example one could add the below line to xdrv_09_timers.ino
```
DEBUG_DRIVER_LOG(PSTR("TIM: Timer 1 XTimerTime %d, Window %d, SetTime %d"), xtimer.time, timer_window[i], set_time);

```
When define DEBUG_TASMOTA_DRIVER is enabled the log message will be compiled. If not no message will be shown and no code will be compiled.

PR #6234, opened 2019-08-15
add support for PAJ7620 gesture sensor

Staars:
## Description:

Add support for (another) gesture sensor: PAJ7620 (or PAJ7620U2)

## Checklist:
  - [x ] The code change is tested and works on core 2.3.0, 2.4.2 and 2.5.2
  - [x ] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

### DIFF
--- a/sonoff/sonoff_post.h
+++ b/sonoff/sonoff_post.h
@@ -134,7 +134,7 @@ char* ToHex_P(const unsigned char * in, size_t insz, char * out, size_t outsz, c
 //#define USE_VL53L0X                           // Enable VL53L0x time of flight sensor (I2C address 0x29) (+4k code)
 //#define USE_MLX90614                          // Enable MLX90614 ir temp sensor (I2C address 0x5a) (+0.6k code)
 //#define USE_CHIRP                             // Enable CHIRP soil moisture sensor (variable I2C address, default 0x20)
-#define USE_PAJ7620                           // Enable PAJ7620 gesture sensor (I2C address 0x73) (+2.5k code)
+//#define USE_PAJ7620                           // Enable PAJ7620 gesture sensor (I2C address 0x73) (+2.5k code)
 
 #define USE_MHZ19                             // Add support for MH-Z19 CO2 sensor (+2k code)
 #define USE_SENSEAIR                          // Add support for SenseAir K30, K70 and S8 CO2 sensor (+2k3 code)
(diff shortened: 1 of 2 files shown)


Staars:
PAJ7620U2 is an integrated gesture recognition sensor from PixArt-Imaging Inc. based on infrared. It also has a built-in proximity detection and can sense various properties like position (x,y,z) and speed.
Compared to the APDS-9960 this sensor is roughly 3-4 times the price, but delivers its data without the need for expensive computations on the ESP8266. Gestures are read from one single I2C-register und only light post-processing is applied.
This drivers tries to expose as much of the capabilities of the hardware as possible and will introduce some new concepts.
No external libraries are used. Delay() was avoided as much as possible.

### Usage:
Simple I2C-connection with the usual configuration in Tasmota.

### Commands:
Sensor50 0…5:

#### 0 - no readings / muted sensor

####1 - Gesture Mode: 
gesture names: Up, Down, Left, Right, Near, Far, CW (clockwise rotation), CCW (counter-clockwise rotation)
TELE-message: …{Up:1}
			…{Left:3} -> 3 times „Left“ in a row, without any other gesture in between

#### 2 - Proximity Mode:
Arbitrary value between 0 (far away) and 255 (very near) are given. Exit from the sensor field will always give at least one „zero message“. TELE is only triggered, when the value has changed
TELE-message: …{Proximity:255}  - virtually touching the sensor in close proximity
			…{Proximity:0}	 - object has left the sensing volume

#### 3 - Corner Mode:
Quarters of the sensing volume(area) are organised like this:
1|2
—-
3|4

An object in one of the corners will trigger the corresponding number, a bit like a pie menu in a GUI.
TELE-message: …{Corner:2} - upper right corner

#### 4- PIN mode:
Based on the corner mode a fluent movement of an object through a given sequence of corners (e.g. 1,2,3,4 -> begin upper left, move right, move diagonally down left, move right) will trigger a valid „PIN“
TELE-message: …{PIN:1} - valid PIN

#### 5- cursor mode:
Shows x- and y-coordinates, mainly intended for debugging and „seeing“ the sensing area. This reads only the upper 5-bit-values, which automatically removes much of the jitter, giving values between 0 and 15.
TELE-message: …{x:1, y:15} - upper left corner



My impressions of this sensor are very positive. Gesture recognition seems to be more stable than on the APDS-9960, which on the other hand is a lot cheaper.
As expected NEAR and FAR gestures are tricky and you have to train your movements to catch them. Sometimes the sensor reports NEAR and FAR at once (which will be discarded). I tried to add some postprocessing to allow the object (hand or finger) to move into the sensing area and delay the initial direction report (up,down,left,right) to give the chance to trigger (the intended) NEAR or FAR movement. 
As always it is a very subjective decision, what to report in the TELE-message and - even more - in what combination and how often.
I hope my experimental introduction of the corner mode and PIN-mode makes sense and maybe this is extendable in the future.
The sensor provides some more goodies, like velocity of an object, so if someone has a fancy use case for this, please let me know.
Of course it would be possible to mix the modes, but this can produce TELE-messages like crazy. I would suggest to add this later upon user request (based on real world use cases).

s-hadinger:
Nice. This can make an interesting gesture remote controller for other devices.

Staars:
Yes, I hope so.
The hard part is always to create the whole UI in a logical way, which means to give a quick response to the user for every event. Less is often more, but most of the time it is nice to play with.
I see the current state as a baseline and will wait for further user suggestions.

BTW, I will add a wiki entry.

PR #6236, opened 2019-08-15
Add MAC address to Home Assistant Discovery Message

### LINKED ISSUE #6226: Add MAC address to Home Assistant Discovery Message
Yes, nothing found.

Home Assistant is using identifiers like the MAC address or serial numbers to match up devices which are reported by multiple integrations.
I have a device tracker integration set up (Unifi), which creates Home Assistant devices for each device found on the network, identified by their MAC address. Because the MAC address is not provided in the MQTT discovery message by Tasmota, these Home Assistant devices cannot be matched with the devices created by the MQTT integration.
As a result, I now have two entries in my Home Assistant device_registry for every Sonoff on my network.

The MAC address of Tasmota devices should be included in the payload of the Home Assistant MQTT discovery message as an additional attribute `cns` (see https://www.home-assistant.io/docs/mqtt/discovery/ section _Supported abbreviations for device registry configuration_).

JSON: `"cns": [ ["mac", "$DEVICE_MAC_ADDRESS"] ]`
where $DEVICE_MAC_ADDRESS is the Sonoff's MAC address in FF:FF:FF:FF:FF:FF form (lowercase is also possible).
It needs to be a nested list, because the attribute is itself a listing of identifiable connections and each connection is given as a list with its type as first and value as second element.

Leave it like it is and live with multiple device_registry entries for the same device.

caphm:
## Description:

**Related issue (if applicable):** fixes #6226

## Checklist:

### DIFF
--- a/sonoff/xdrv_12_home_assistant.ino
+++ b/sonoff/xdrv_12_home_assistant.ino
@@ -129,6 +129,7 @@ const char HASS_DISCOVER_SENSOR_HASS_STATUS[] PROGMEM =
 const char HASS_DISCOVER_DEVICE_INFO[] PROGMEM =
   ",\"uniq_id\":\"%s\","
   "\"device\":{\"identifiers\":[\"%06X\"],"
+  "\"connections\":[[\"mac\",\"%s\"]],"
   "\"name\":\"%s\","
   "\"model\":\"%s\","
   "\"sw_version\":\"%s%s\","
@@ -136,7 +137,8 @@ const char HASS_DISCOVER_DEVICE_INFO[] PROGMEM =
 
 const char HASS_DISCOVER_DEVICE_INFO_SHORT[] PROGMEM =
   ",\"uniq_id\":\"%s\","
-  "\"device\":{\"identifiers\":[\"%06X\"]}";
+  "\"device\":{\"identifiers\":[\"%06X\"],"
+  "\"connections\":[[\"mac\",\"%s\"]]}";
 
 const char HASS_DISCOVER_TOPIC_PREFIX[] PROGMEM =
   ",\"~\":\"%s\"";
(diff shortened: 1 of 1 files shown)


meingraham:
This is required because HA recently introduced a device registry? Just pinpointing the reason why this is now a requirement for proper operation of the HA auto-discovery feature already in Tasmota.

If a user is using an older version of HA, will having the additional MAC attribute in the payload be ignored? In other words, does this change retain backward compatibility?

How much additional program memory does this change require?

P.S. Are we at the point where Tasmota HA integration needs to be off by default and HA users need to compile their own binaries in order to add HA support? With Gitpod, this is not a huge burden.

caphm:
> This is required because HA recently introduced a device registry? Just pinpointing the reason why this is now a requirement for proper operation of the HA auto-discovery feature already in Tasmota.

HA auto-discovery will still work without this, but there will be mutliple devices created in HA because it is missing the MAC as an identifier to merge potential duplicaes in the device registry. See https://developers.home-assistant.io/docs/en/device_registry_index.html under _Defining devices_.

> If a user is using an older version of HA, will having the additional MAC attribute in the payload be ignored? In other words, does this change retain backward compatibility?

Yes. The additional attribute is part of the `device` object within the payload, which is ignored entirely if HA version doesn't support it (no device registry present).

> How much additional program memory does this change require?

I don't know how to determine that. Do you mean how much larger the binary will become or the memory usage during operation?
I can't compare the binary sizes right now, but it should be minimal. There's 58 characters being added to the `const char[]` in total (29 each), plus an additional call to `WiFi.macAddress().c_str()`. It should only be a couple of bytes in total.

> P.S. Are we at the point where Tasmota HA integration needs to be off by default and HA users need to compile their own binaries in order to add HA support? With Gitpod, this is not a huge burden.

That's not for me to provide an opinion on.

meingraham:
Thanks for the information.

re: my P.S. - Yup, meant for Theo... since he'll evaluate this PR an your notes as well.

ascillato:
> P.S. Are we at the point where Tasmota HA integration needs to be off by default and HA users need to compile their own binaries in order to add HA support? With Gitpod, this is not a huge burden.

You are right, but I think that the amount of Tasmota users that also use Home Assistant is enough so as to mantain it enabled by default.

emontnemery:
This change is not needed because serial number is included as identifier in the device info already:
https://github.com/arendst/Tasmota/blob/956f56e0e17f89c79455d1ebcd4c45f477799501/tasmota/xdrv_12_home_assistant.ino#L107

Reverted as part of #7901.

PR #6256, opened 2019-08-18
Container to build all or custom builds with one run.

eddyhub:
## Description:
Container to build all or custom builds with one run. No IDE, no python or platformio needed. Install docker, build the container on your machine or use the one already published on docker hub and start building customized firmware. See README.md for more information.

I just needed a simple solution to build some binaries without setting up python or platformio on my build machine. Maybe you find this addition useful. It is also possible to add esptool to flash the generated firmware to the device, but I would add this feature later.

## Checklist:

### DIFF
--- a/build-container/Dockerfile
+++ b/build-container/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:2
+
+LABEL author="Eduard Angold"
+
+# Install platformio. To be able to build tasmota <=v6.6.0 (and later)
+# we have to use version 3.6.7 of platformio.
+RUN pip install --upgrade pip &&\ 
+    pip install -U platformio==3.6.7
+
+# Init project
+COPY init_pio_tasmota /init_pio_tasmota
+
+# Install project dependencies using a init project.
+RUN cd /init_pio_tasmota &&\ 
+    pio run &&\
+    cd ../ &&\ 
+    rm -fr init_pio_tasmota &&\ 
+    cp -r /root/.platformio / &&\ 
+    chmod -R 777 /.platformio
+
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/bin/bash", "/entrypoint.sh"]
+
--- a/build-container/README.md
+++ b/build-container/README.md
@@ -1,0 +1,26 @@
+# Docker container for tasmota builds
+This Container will setup a proper build environment for [Sonoff-Tasmota](https://github.com/arendst/Sonoff-Tasmota)
+
+## Create container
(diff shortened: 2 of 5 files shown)


meingraham:
@eddyhub 

You can build any Tasmota variant you wish, with any available Core with any custom settings you desire without installing any software by using [Gitpod](../wiki/Compiling-Tasmota-on-Gitpod)

eddyhub:
@meingraham 
But than I have to pass my credentials for wifi (that would be ok) and credentials to my MQTT broker, which is online and available.

meingraham:
Your choice... you can set up everything except credentials in the `user_config_override.h` and compile. Then just set up Wi-Fi using Wi-Fi Manager after flashing.

I have a config that has all of my settings except Wi-Fi, MQTT, and Lat/Lon. I compile, flash, and then connect it to Wi-Fi using Wi-Fi Manager followed by a `Backlog` pasted in to the Console for MQTT & geolocation.

digiblur:
Interesting..will have to check this out. Gitpod is great but I myself also prefer stuff to be local when I can.

meingraham:
I wonder how @andrethomas has his hackbox server set up. Could be some collaboration opportunities or some opportunity for @eddyhub to incorporate some of the vast build scripting Andre has done.

arendst:
Both @andrethomas and me use Arduino CLI on linux with local scripts to compile and release (and in the case of Andre) upload the binaries to the website.

Using Arduino CLI is just a legacy solution solving code flash memory alignment errors in the past.

I locally use platformio for development and compile.

andrethomas:
The development binaries are the result of a fairly uninteresting build method.

The script always starts with un unmodified version of `my_user_config.h`

It then makes changes to the `my_user_config.h` to include or exclude compiler directives using sed from command line, something like:

`sed '/FIRMWARE_CLASSIC/ s?^//??' my_user_config.h`

This is done for each of the target binaries from the development branch and binaries for all core variants are compiled in parallel at the same time... if there were no new merges the compile is not performed though.

Compiling is done using portable arduino installs (one for each core) and is executed by the compile script from command line without using the GUI (in fact the container does not have a GUI)

For development I use platformio as @arendst does.

ascillato2 (approved):

Jason2866:
I think this line should be changed https://github.com/arendst/Sonoff-Tasmota/pull/6256/files#diff-fb8e6a4c933df3485b0daa0166252b30R8
to use latest platformio. Since we found the reason why platformio v.4.x failed with compiling and
we changed platformio.ini to the new needs of v.4.x

PR #6267, opened 2019-08-20
Initial DALI bus support

agners:
Initial support for DALI (Digital Addressable Lighting Interface) bus
support. Currently only two commands are implemented which send out
commands using bus broadcasts. This is sufficient to control a single
DALI capable device.

The DALI library has been borrowed from the Simple DALI Controller
project:
https://create.arduino.cc/projecthub/NabiyevTR/simple-dali-controller-506e44

## Checklist:

### DIFF
--- a/lib/DALI/Dali.cpp
+++ b/lib/DALI/Dali.cpp
@@ -1,0 +1,517 @@
+
+#include "Dali.h"
+
+
+
+Dali::Dali() //constructor
+{
+  applyWorkAround1Mhz = 0;
+}
+
+
+void Dali::setTxPin(uint8_t pin)
+{
+  TxPin = pin; // user sets the digital pin as output
+  pinMode(TxPin, OUTPUT); 
+  digitalWrite(TxPin, HIGH);
+}
+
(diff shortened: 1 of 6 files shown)


ascillato:
Hi,

These routines uses delay function that can interfere with other Tasmota routines. 

Have you experienced any problem with wifi or mqtt stability?

ascillato2:
Travis tests fail due to missing key in all language files.

arendst:
In addition adding a GPIO in the middle of the GPIO list will break backwards compatibility too

agners:
> Hi,
> 
> These routines uses delay function that can interfere with other Tasmota routines. Some delays are greater than half second.
> 
> Have you experienced any problem with wifi or mqtt stability?

So far I have not, but also only used it for 2 hours or so.

> 
> I think the library should be rewritted to a non-blocking routine.

Since the protocol is using a fixed data rate (1200 bit/s) and we use bit banging, timing is somewhat crucial... Maybe we could buffer and bitbang using a timer or similar?

agners:
Will address the other comments sometime next week.

PR #6273, opened 2019-08-21
Fix: Tuya Switches are detected as dimmers.

shantur:
Tuya switches are detected as dimmers even after setting SetOption65 to 0.
Currently SetOption65 just hides the dimmer from Web UI for Tuya switches with SetOption65 to 0 but they are advertised as dimmer to HASS.
With this change we mark the switch to LT_BASIC instead of dimmable when option 65 is set. Makes sure its just an On-off switch.

## Description:

## Checklist:

### DIFF
--- a/sonoff/support_command.ino
+++ b/sonoff/support_command.ino
@@ -621,6 +621,9 @@ void CmndSetoption(void)
           if (18 == pindex) { // SetOption68 for multi-channel PWM, requires a reboot
             restart_flag = 2;
           }
+          if (15 == pindex) { // SetOption65 for tuya_show_dimmer requires a reboot
+            restart_flag = 2;
+          }
         }
       }
       else {                   // SetOption32 .. 49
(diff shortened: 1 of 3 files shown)


shantur:
@arendst can you please merge it in, I have more fixes coming in based on top of this change.

arendst:
I would love to but I cannot see if it works for other people.
- I do not own any tuya device like this
- Others have made the change to make the dimmer optional by SetOption.
- These devices have always used the serial type as what I imagine this also directs the code to Tuya driver
- I expect your changes will break previous defined devices.

I may be wrong of course but I do not want to merge a PR related to tuya every day

arendst:
And there we go again....

shantur:
Ok, In that case, I will make a PR with all the Tuya serial related fixes. But still, can you please confirm, I what I am thinking is right. 
The current implementation of `SetOption65` and wiki seems to be opposite.

https://github.com/arendst/Sonoff-Tasmota/wiki/Commands#setoption65

https://github.com/arendst/Sonoff-Tasmota/blob/development/sonoff/xdrv_16_tuyadimmer.ino#L152

arendst:
I cannot confirm as I didn't implement SetOption65.

But what I think that should happen is as all Setoptions are default 0 then with 0 it should show the dimmer and as 1 it should not show the dimmer. That way backward compatibility was garanteed.

PR #6288, opened 2019-08-24
ibeacon added

gemu2015:
## Description:

this driver scans for ibeacon devices with ble module hm17
it simply reports the RSSI when a device was found and reports the RSSI  
 
it scans every IB_UPDATE_TIME_INTERVAL seconds 
if a device is no longer present (IB_TIMEOUT_INTERVAL in seconds) the device is discarded and an RSSI of zero reported

these 2 intervals should be moved to settings => (already prepared)

IB_UPDATE_TIME
IB_TIMEOUT_TIME

a debug mode may be enabled
#define IBEACON_DEBUG
which allows for inspecting all cmds to and reactions from the module

2 kinds of beacon devices are supported
1. real fully ibeacon compatible devices (battery lasts about a year) completely passive, no beeper  
2. and key fobs with a beeper (battery lasts about a month only) 

type 2 may be made beeping by sending an RSSI of 99 to the driver

## Checklist:

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -420,6 +420,7 @@
 //  #define USE_PN532_DATA_FUNCTION                // Add sensor40 command support for erase, setting data block content (+1k7 code, 388 bytes mem)
 //  #define USE_PN532_DATA_RAW                     // Allow DATA block to be used by non-alpha-numberic data (+ 80 bytes code, 48 bytes ram)
 //#define USE_RDM6300                              // Add support for RDM6300 125kHz RFID Reader (+0k8)
+#define USE_IBEACON                                // Add support for bluetooth LE passive scan of ibeacon devices (uses HM17 module)  
 
 // Power monitoring sensors -----------------------
 #define USE_ENERGY_MARGIN_DETECTION              // Add support for Energy Margin detection (+1k6 code)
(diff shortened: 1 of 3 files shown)


Jason2866:
@gemu2015  Which chip is supported for the HM-17 module?
Only CYBL10162-56LQXI or TI CC2640 chip too?

gemu2015:
is used this module => https://www.ebay.de/itm/HM-17-BLE-4-1-Bluetooth-CC2541-UART-Modul-Wireless-neuer-als-HM10-HM11-HM17/223189729091?hash=item33f724ff43:g:yxoAAOSweTlbw8l7

the module is driven via AT cmds over serial so the chip may not be relevant if the AT CMDs are supported.

the older HM10 does not support master mode, so hm17 or eventually hm19 should work.

itProfi:
Is  Xiaomi Mijia Hygrometer Bluetooth possible in the future ? A very popular device, it seems.:))
It's a library on JS for it.. https://github.com/AlCalzone/ioBroker.ble

gemu2015:
@itProfi 
unfortunately the hm17 is very limited. it can not explore services as would be needed.
i also have a very sophisticated rn4020 module from microchip which can do  all that. but at the moment i have other focuses.

PR #6297, opened 2019-08-27
Fix Would not compile if USE_LIGHT not defined

s-hadinger:
## Description:

Looking to reduce flash space, I realized Tasmota would not compile when USE_LIGHT is not defined, because of missing #ifdefs.

By the way, I'm not sure why `#define USE_LIGHT` is in `sonoff.h` rather than `my_user_config.h`.

## Checklist:


arendst:
The define is in sonoff.h as the user is not supposed to use/need it. It's just there to help make the minimal version even smaller.

PR #6330, opened 2019-08-31
Move '#include user_config_override.h' in 'my_user_config.h'.

s-hadinger:
## Description:

Moved `#include user_config_override.h` at the end of `my_user_config.h` instead of inside `sonoff.ino`.
Benefit: allows user_config_override.h to change parameters like USE_MQTT_TLS.
Downside: generates 3 warnings instead of 1 when 

Theo, I don't see any wrong side effect for this, but I would like your advice on this one.

## Checklist:

### DIFF
--- a/sonoff/my_user_config.h
+++ b/sonoff/my_user_config.h
@@ -549,6 +549,10 @@
  * No user configurable items below
 \*********************************************************************************************/
 
+#ifdef USE_CONFIG_OVERRIDE
+  #include "user_config_override.h"         // Configuration overrides for my_user_config.h
+#endif
+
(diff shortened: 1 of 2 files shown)


arendst:
Thx.

I don't see side effects too but I will do some tests first

arendst:
I don't like the triple warning (in my case six warnings)

I tried to move the USE_DISCOVERY and USE_MQTT_AWS_IOT test from my_user_config.h to sonoff.ino but then it fails with all kinds of bearssl error messages.

So for now let's merge this and see how we fare.

PR #6342, opened 2019-09-02
Add ina226 driver to Tasmota

hwstar:
## Description:

This change adds support for the Texas Instruments INA226 current sensor module.
The additional sensor driver module is named xsns_76_ina226.ino.

The reason this was developed is because the INA219 driver can't be configured to work over a wide range of shunt values and full scale currents, and I was looking for something I could use with my solar backup system to monitor currents and voltages.

This driver supports up to 4 INA226 devices. No I2C slave addresses are probed until the driver is configured with the shunt resistance in ohms, and the the full scale current reading. All 4 INA226 devices can have independent shunt resistances, and full scale current values.

At this point, I'm just looking for comments and feedback on the sensor driver. Provided suggestions are reasonable, I will make any necessary changes.

Still unsure as to what driver number should be assigned to this. For testing I used 76.

## Checklist:
  - [ Y] The pull request is done against the latest dev branch
  - [Y ] Only relevant files were touched
  - [ Y] Only one feature/fix was added per PR.
  - [Y ] The code change is tested and works on core 2.3.0, 2.4.2 and 2.5.2
  - [ N] The code change pass travis tests. **Your PR cannot be merged unless tests pass**
  - [ N] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).


arendst:
Thx for the implementation.

As it stands the Settings area wastes 4 x 2 bytes because of the struct used. 
Pls reconsider a change from uint32_t to uint16_t for the shunt resistor range.

If it can replace the current ina219 driver just give it the same sensor number

hwstar:
I have a plan for the first two concerns: 2 ea. uint16_t[4] arrays will replace the struct which will address the wasted EEPROM space, and also reduce the space used from 32 bytes to 16 bytes. The r_shunt value will use one bit of the uint16_t to select a range of milliohms or microhms.

Regarding the sensor number, I have question:  Are you wanting me to add support for the INA219 in my driver, or is it OK use the same sensor number, as long as both drivers as mutually exclusive to each other?

arendst:
As I'm under the impression that your driver can replace the INA219 driver it makes sense to give it the same sensor number (as is done with the DS18B20 drivers).

If your driver does not replace the INA functionality then select sensor number 54

hwstar:
Thanks for getting back to me. There is no support currently for the INA219 in this driver currently.  I will go with sensor number 54. With the different command interface between the INA219 and INA226 drivers , the two drivers won't be  form fit and function interchangeable.

I haven't looked at the impact of adding the INA219 other than detecting an INA219 vs. INA226. There may or may not be limitations. To me, it makes sense to get the driver into Tasmota, then look at the INA219 support as a separate follow-on project.

arendst:
OK.

hwstar:
Wasted bytes removed. Retested with uint16_t arrays. Please review.

PR #6352, opened 2019-09-04
sendmail support

gemu2015:
## Description:

email support example

sendmail [smtp.gmail.com:465:user:password:<source.mail.de>:<destination.mail.de>:subject] Hallo Tasmota

or 

sendmail [\*:\*:\*:\*:\*:<destination.mail.de>:subject] Hallo Tasmota

the stars are replaced by these defines in user_config_overwrite.h

#define EMAIL_SERVER "smtp.gmail.com"
#define EMAIL_PORT 465
#define EMAIL_USER "user"
#define EMAIL_PASSWORD "passwd"
#define EMAIL_FROM "<mr.x@gmail.com>"

only SSL port 465 (or 587 ???) is supported (gmail,gmx,yahoo etc) because STARTTLS is not supported on ESP8266 TLS

works on core 2.42, and pre 2.6

core 2.3 has memory restrictions and no TLS
core 2.52 has bearssl problems with SMTP

## Checklist:

### DIFF
--- a/lib/SENDMAIL/sendemail.cpp
+++ b/lib/SENDMAIL/sendemail.cpp
@@ -40,7 +40,8 @@ bool SendEmail::send(const String& from, const String& to, const String& subject
   DEBUG_EMAIL_PORT.println(port);
 #endif
 
-#ifndef ARDUINO_ESP8266_RELEASE_2_4_2
+#if defined(ARDUINO_ESP8266_RELEASE_2_3_0) || defined(ARDUINO_ESP8266_RELEASE_2_4_2)
+#else
   client->setInsecure();
   bool mfln = client->probeMaxFragmentLength(host.c_str(), port, 512);
 #ifdef DEBUG_EMAIL_PORT
@@ -50,6 +51,7 @@ bool SendEmail::send(const String& from, const String& to, const String& subject
       client->setBufferSizes(512, 512);
   }
 #endif
+
 
   if (!client->connect(host.c_str(), port))
   {
--- a/lib/SENDMAIL/sendemail.cpp
+++ b/lib/SENDMAIL/sendemail.cpp
@@ -1,0 +1,302 @@
+#include "sendemail.h"
+
+// enable serial debugging
+//#define DEBUG_EMAIL_PORT Serial
+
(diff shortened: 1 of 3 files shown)


s-hadinger:
From my experience TLS eats a lot of RAM and Flash code. I needed to optimize very aggressively to make it reasonable for AWS IoT support.

Did you measure the RAM/Flash impact of activating TLS for SMTP?

gemu2015:
@s-hadinger 
here are the numbers for sendmail
AXTLS (core 2.42)   Flash 64k5 min ram about 19k
BEARTLS (core pre 2.6) Flash 64k min ram about 20k

would be interesting if it would also work with your "lite" version.

Gerhard

s-hadinger:
The main optimizations I used in BearSSL were:

- limit the cipher supported to strict minimum (saves a lot of flash code)
- remove full validation of the certificate against the CA (including domain name...) - saves a lot of RAM during handshake
- use StackThunk (secondary stack) only during handshake instead of always allocated (around 5k of RAM)

What may make it difficult to use for sendmail is the limited number of ciphers supported and the partial CA validation. For ciphers, I hope that all major providers support either `BR_TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` or `BR_TLS_RSA_WITH_AES_128_GCM_SHA256`. For CA validation, we could add some CA beyond Letsencrypt or use fingerprint validation.

Memory/Flash impact are detailed here: https://github.com/arendst/Sonoff-Tasmota/wiki/TLS#memory-usage

killjoy2505:
I'm having great difficulties trying to use this. I've succesfully compiled anf flashed a version with usemail ony my Sonoff RF Bridge. However when I try to send an email through the console it always comes back with "false parameters". I've tried countless of times, with the gmail settings and with some other smtp mail settings from another account I have.
Can someone give me any pointers as to what I might be doing wrong?

I was hoping to use the sendmail so I can have an 433mhz SOS button trigger a rule to send a mail to my family whenever my mom falls and cannot get back up :(

Sorry if this is not the right place.

gemu2015:
it is essential that your gmail account is set to "allow less secure apps" 
did you check this ?

killjoy2505:
TYVM! Got it to work, I was using my 2FA account which doesn't allow that so I just made a new account.

Also it is important to include <> in the e-mail from and e-mail to field.

Issue #6357, opened 2019-09-05
How do add the EPT Font when building Tasmota?


DeviousPenguin:
<GUIDE>

<This troubleshooting issue template is meant to help Tasmota users with difficult problems. It is aimed to be opened if using the wiki and the support chat could not solve the issue. The Github Issue tracker is NOT a general discussion forum>
  
<Please DO NOT OPEN AN ISSUE:>
  <If you have general questions or you need help on Tasmota usage, go to the Tasmota support chat>
  <If your Tasmota version is not the latest, please update before posting. Your issue might be already solved. Latest precompiled bins of Tasmota can be downloaded from http://thehackbox.org/tasmota/>
  <If your issue is a new device, please use the Tasmota Template Feature. See wiki for that>
  <If your issue is a flashing issue, please address that to the Tasmota Support Chat>
  <If your issue is compilation problem, please address that to the Tasmota Support Chat>
  <If your issue has been addresed before (duplicated issue), please ask in the original issue>
  <If your issue is wifi problem or mqtt problem, please try first the steps provided in troubleshooting of the wiki>

Please take a few minutes to complete the requested information below. Our ability to provide assistance is greatly hampered without it. The details requested potentially affect which options to pursue. The small amount of time you spend completing the template will also help the volunteers providing the assistance to you to reduce the time required to help you.

### ISSUE DESCRIPTION - TROUBLESHOOTING

Hi, I'm using a SSD1306 OLED screen with Tasmota Stable 6.6 and it works well with the default GFXFont.

I would quite like to add the EPDFont as detailed in the wiki:

https://github.com/arendst/Sonoff-Tasmota/wiki/Displays  

 >The EPD fonts use about 9k space, which can be selected at compile time using #ifdef directives.
> SSD1306 - 1.15k

I've searched the `my_user_config.h` file for how to add EPD fonts but there is nothing there.

### REQUESTED INFORMATION
_Make sure these boxes are checked before submitting your issue. Thank you_

**FAILURE TO COMPLETE THE REQUESTED INFORMATION WILL RESULT IN YOUR ISSUE BEING CLOSED**

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Sonoff-Tasmota/blob/development/CODE_OF_CONDUCT.md)
[...]

PR #6360, opened 2019-09-06
touch button update, fix display settings font and size

gemu2015:
## Description:

**Related issue (if applicable):** fixes #<#6357>

displayfont, displaysize now immediately set font and size
touch button interface rework.

remark:
on pre core 2.6 the touch i2c driver is unreliable. until now i could not figure out what is wrong
obviously there must be changes in the i2c driver.

## Checklist:

### DIFF
--- a/sonoff/xdrv_13_display.ino
+++ b/sonoff/xdrv_13_display.ino
@@ -1247,6 +1247,12 @@ void DisplayInitDriver(void)
 {
   XdspCall(FUNC_DISPLAY_INIT_DRIVER);
 
+  if (renderer) {
+    renderer->setTextFont(Settings.display_font);
+    renderer->setTextSize(Settings.display_size);
+  }
+
+
 //  AddLog_P2(LOG_LEVEL_DEBUG, PSTR(D_LOG_DEBUG "Display model %d"), Settings.display_model);
 
   if (Settings.display_model) {
@@ -1376,14 +1382,18 @@ void CmndDisplaySize(void)
 {
   if ((XdrvMailbox.payload > 0) && (XdrvMailbox.payload <= 4)) {
     Settings.display_size = XdrvMailbox.payload;
+    if (renderer) renderer->setTextSize(Settings.display_size);
+    else DisplaySetSize(Settings.display_size);
   }
   ResponseCmndNumber(Settings.display_size);
 }
 
 void CmndDisplayFont(void)
 {
-  if ((XdrvMailbox.payload > 0) && (XdrvMailbox.payload <= 4)) {
+  if ((XdrvMailbox.payload >=0) && (XdrvMailbox.payload <= 4)) {
     Settings.display_font = XdrvMailbox.payload;
+    if (renderer) renderer->setTextFont(Settings.display_font);
+    else DisplaySetFont(Settings.display_font);
   }
   ResponseCmndNumber(Settings.display_font);
 }
@@ -1804,7 +1814,7 @@ void Restore_graph(uint8_t num, char *path) {
   if (!fp) return;
   char vbuff[32];
   char *cp=vbuff;
-  char buf[2];
+  uint8_t buf[2];
   uint8_t findex=0;
 
   for (uint32_t count=0;count<=gp->xs+4;count++) {
--- a/sonoff/xdsp_07_sh1106.ino
+++ b/sonoff/xdsp_07_sh1106.ino
@@ -145,13 +145,13 @@ void SH1106Refresh(void)  // Every second
   if (Settings.display_mode) {  // Mode 0 is User text
     switch (Settings.display_mode) {
       case 1:  // Time
-        Ssd1306Time();
+        SH1106Time();
         break;
       case 2:  // Local
       case 3:  // Local
       case 4:  // Mqtt
       case 5:  // Mqtt
-        Ssd1306PrintLog();
+        SH1106PrintLog();
         break;
     }
   }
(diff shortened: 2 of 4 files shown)


ascillato:
>on pre core 2.6 the touch i2c driver is unreliable. until now i could not figure out what is wrong
obviously there must be changes in the i2c driver.

Please, can you test if with the actual arduino STAGE core from today this I2C issue is still there?
Thanks.

Jason2866:
@gemu2015  in my build core pre2.6 i use a different i2c (wip) driver which saves about 600 bytes of iram. PR https://github.com/esp8266/Arduino/pull/6326
This driver works fine for the standard Tasmota builds, but it seems it does make problems with the touch driver...
So as @ascillato suggested try Arduino stage core version. The still use the old i2c driver.
But you have to disable features using iram...

ascillato:
>But you have to disable features using iram...

Yes, that is right. I could compile STAGE by disabling:

- IR Reception/Send
- Tuya Support
- Emulation Support
- mDNS Support

gemu2015:
as it turns out it is not related to pre 2.6
the bug also appears in 2.52.
it is extremely strange as i can not see any reason why this happens.

this routine returns wrong values on 2.52 and above 
and works well in 2.42
(it falsely returns no zero values when there is actually no hit)

has anybody  a hint what might be wrong ?

original code called with reg 2 (number of touch points)

uint8_t FT6236readTouchRegister(uint8_t reg)
{
  Wire.beginTransmission(FT6236_i2c_addr);
  Wire.write(reg);  // register 0
  uint8_t retVal = Wire.endTransmission();
  uint8_t returned = Wire.requestFrom(FT6236_i2c_addr,uint8_t(1)); 
  if (Wire.available())
  {
    retVal = Wire.read();
  }
  return retVal;
}

gemu2015:
i checked the i2c changes between 2.42 and 2.52 but couldn't figure out what could be the reason for the touch controller failure.

a simple workaround fixed the problem however => reading the status register twice 

while the 1. reading delivers wrong readings every few seconds, the 2.reading is always correct

Strange!

Jason2866:
Could this be a clock strech issue?

joba-1:
There are a few things wrong with FT6236readTouchRegister().

1. return code does not indicate failure (have you read button touched/not touched or error?)
2. return value of endTransmission not used (cannot be sure you now really read reg with requestFrom() or some other register)
3. return code from requestFrom() is not checked (might be ok since available() probably implies returncode was ok.
4. available() seems redundant, since requestFrom() just told you in returned, if there is a byte available or not.

Basically, if all works ok, you get the correct button touch state, but in case of error you cannot detect it. I2C is a shared bus, meaning things will go wrong at times.

How about sth like this?:

https://github.com/joba-1/HumiControl/blob/master/src/main.cpp#L162

With that you could see what the "real" error is. Maybe a timing issue. Maybe bus congestion, Maybe the datasheet demands an unusual delay somewhere, Whatever...

HTH

gemu2015:
@Jason2866 
the touch controller doesn't use clock stretching.

@joba-1 
all you arguments are coherent but
core 2.3 and 2.42 NEVER show an error in this application
while above core 2.42 there are frequent errors every few seconds.
and that with 2 different hardware circuits and 2 different (but compatible) touch controllers 
from Focaltech.

joba-1:
not saying my proposals will fix the error, just that modifying the code to check and report errors instead of guessing could help finding the actual cause :)

gemu2015:
you are right!

  Wire.beginTransmission(FT6236_i2c_addr);
  Wire.write(reg);  // register 2
  uint8_t retVal = Wire.endTransmission();

Wire.endTransmission returns error 3 (received NACK on transmit of data)

but what difference in the core causes this ???
the second call never fails.

since the workaround fixes the problem i will not further investigate

but this may affect other i2c devices also

joba-1:
ok, maybe the error has nothing to do with this call.

It could well be, the previous i2c access (from what ever) leaves the bus in an inconsistent state and the first stop at the end of endTransmission() sorts this out.

joba-1:
me again, lol. Interesting how much can happen in so few code lines.

What about the default stop parameter of endTransmission(bool stop = true)?
Default is to release the bus which usually is not what you want if you set a register address intending to read from it. I would set stop to false here. 
Ok, I'm sure some i2c devices out there might actually need a stop=true, but maybe here it is the other way round and with the new core a false is now not only better, but required.

Or maybe it is just another core bug...

Whatever, I don't have the device and can't test, so I guess I'm out of here now, too.

gemu2015:
ok, i was curious what is going on and took my logic analyzer.
strange extra clocks in cores above 2.42 !!!
what was changed between 2.42 and 2.52 was the twi slave interface added. probably this interferes here? in the twi master i could not figure out a difference that would cause this.
 (i have only one device on the i2c bus)
someone with a better knowledge of i2c secrets should look at this problem.

core pre2.6
core_pre2 6
core 2.42
core_2 42

gemu2015:
found the bug!
as i expected the slave mode has a bug. it enables CHANGE interrupts on scl and sda.
commenting this out solves the problem.

i don't know however why this doesn't affect other sensors.

very strange


void Twi::init(unsigned char sda, unsigned char scl)
{
    // set timer function
    ets_timer_setfn(&timer, onTimer, NULL);

    // create event task
    ets_task(eventTask, EVENTTASK_QUEUE_PRIO, eventTaskQueue, EVENTTASK_QUEUE_SIZE);

    twi_sda = sda;
    twi_scl = scl;
    pinMode(twi_sda, INPUT_PULLUP);
    pinMode(twi_scl, INPUT_PULLUP);
    twi_setClock(preferred_si2c_clock);
    twi_setClockStretchLimit(230); // default value is 230 uS

    if (twi_addr != 0)
    {
     //   attachInterrupt(scl, onSclChange, CHANGE);
     //   attachInterrupt(sda, onSdaChange, CHANGE);
    }
}

Jason2866:
Have you addressed your finding at Arduino ESP8266 github?
Would help ;-)

PR #6415, opened 2019-09-13
Create an objdump from firmware.elf automatically

Staars:
## Description:

Simple convenience script to create firmware.asm from firmware.elf in the build process as a light weight debugging tool.
Will create a file of roughly 20 MByte and should probably be deactivated by default.

## Checklist:

### DIFF
--- a/pio/obj-dump.py
+++ b/pio/obj-dump.py
@@ -1,0 +1,9 @@
+# Little convenience script to get an object dump
+
+Import('env')
+
+def obj_dump_after_elf(source, target, env):
+    print("Create firmware.asm")
+    env.Execute("xtensa-lx106-elf-objdump "+ "-D " + str(target[0]) + " > "+ "${PROGNAME}.asm")
+    
+env.AddPostAction("$BUILD_DIR/${PROGNAME}.elf", [obj_dump_after_elf])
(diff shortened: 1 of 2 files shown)


meingraham:
@Staars 

`as a light weight debugging tool`
Could you point me to some reading on how this file would be used to track down problems?

Thanks.

Mike

Staars:
For me it is handy from time to time to take a quick look at the compiled code. For instance for a small function like this:

`40240378 <_Z20MhzCalculateChecksumPh>:
40240378:	130c                	movi.n	a3, 1
4024037a:	040c                	movi.n	a4, 0
4024037c:	523a                	add.n	a5, a2, a3
4024037e:	000552               	l8ui	a5, a5, 0
40240381:	331b                	addi.n	a3, a3, 1
40240383:	445a                	add.n	a4, a4, a5
40240385:	744040               	extui	a4, a4, 0, 8
40240388:	f08366               	bnei	a3, 8, 4024037c <_Z20MhzCalculateChecksumPh+0x4>
4024038b:	602040               	neg	a2, a4
4024038e:	742020               	extui	a2, a2, 0, 8
40240391:	f00d                	ret.n`

(nothing special with this function)

The output is not as good as IDA, ScratchABit, ...  but sometimes this is enough. 
Just for a quick estimation, if it is useful to choose uint8_t over uint32_t or something like that.
I can simply click on the generated  .asm-file and search for function names, even if they are not nicely post-processed.

To be honest, this should be part of the the IDE, but at least this is the simplest solution that I have found so far.

meingraham:
@Staars 

Thanks. So just gives you a bit more granular insight into what the code is actually doing. Understood.

Mike

P.S. use _triple_ backticks (\`\`\`) to enclose blocks of code.

PR #6424, opened 2019-09-15
Need feedback: new sensors for Home Assistant discovery

effelle:
## Description:
This is not really a PR, but just a basic idea I had for adding some other things to HA MQTT discovery. 
Needing an example, I created a ready-to-use sensor that shows a combined SSId and RSSI value to keep on eye on low signals, issue I had in the past with core 2.5.x. 
The sensor was created directly using a `value_template`.

Screenshot_2019-09-13-15-50-11-984_com android chrome

Since is one of the sensor I use most, I thought could be interesting for others too. 
The code is not perfect and probably needs a revamp, but is good enough to explain what I'm doing. There could be other useful sensors to implement , like IP Address, a reset button or an uptime monitor. 

Now, I would like to know what you think: can we integrate this type of sensors into the firmware?

Post scriptum: the wifi sensor doesn't show up on discovery page under HA but is perfectly selectable as entity and I don't have errors in logs. No idea why. 

**Related issue (if applicable):** None

## Checklist:

### DIFF
--- a/sonoff/xdrv_12_home_assistant.ino
+++ b/sonoff/xdrv_12_home_assistant.ino
@@ -152,6 +152,11 @@ const char HASS_DISCOVER_DEVICE_INFO_SHORT[] PROGMEM =
 const char HASS_DISCOVER_TOPIC_PREFIX[] PROGMEM =
   ",\"~\":\"%s\"";
 
+const char HASS_DISCOVER_INFO_WIFI[] PROGMEM =
+  // Combined WiFi sensor, not usable for graph in Hass //
+  ",\"val_tpl\":\"{{[value_json.Wifi." D_JSON_SSID ", value_json.Wifi." D_JSON_RSSI "] | join (' (') + '%%)' }}\","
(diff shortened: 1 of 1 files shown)


effelle:
Since I fear it could be merged without a check from @arendst I prefer close it for now. Awaiting feedback from devs.

effelle:
@arendst Theo do I have your bless? :D

arendst:
You might know my HASS feelings ;-)

effelle:
Indeed I know! I just need to know if it worth the time, since every sensor will add at least 20-30 lines of code.

arendst:
It's not just the code. It's also the firmware image size.

Isn't there another possibility to get sensors in Hass?

effelle:
Sure there is, in fact I don't understand why you merge this one since was just something to show you. 
I think I'll add some new YAML examples on our wiki.

arendst:
I'll unmerge now ;-)

I was just in a merge flow.

effelle:
better. I was not happy with the code format

PR #6477, opened 2019-09-23
Energy: Add "EnergySaveTime" command to save the energy data every x minutes set by the command

pablozg:
@arendst As you know, These days I have a lot of grid failures, so my domoticz charts are not valid, because I lose all data measured by the meters.

I added a new command to save the energy data every x minutes, the command needs a minimum of 30 minutes an a maximum of 1440 (24 hours), to avoid excessive writes to the eeprom, by default is disabled (value 0), but the user has the choice to set a value to store the data.

Thanks.


arendst:
OK. I get the idea. 

It's already possible using a rule where you execute commands ``SaveData`` on a regular basis. I prefer it that way.

pablozg:
Hi, I do it with a rule like this:

````
ON Time#Minute|30 DO SaveData ENDON
````

If you want close this PR, thanks.

PR #6515, opened 2019-09-29
support for i2c-moisture-sensor AND Chirp!-sensor

Staars:
## Description:

Due to a misconception on my side, I initially did not recognise the differences between the 2 sensors. This should now be solved.
Basically the driver slows down and splits up the read process for the Chirp!-sensor, which also works for the faster i2c-moisture-sensor.

It can also be confirmed now, that multiple sensors are possible, if the hardware setup allows this.

Minor changes:
The false naming of the light value will become a unit-less DARKNESS, because ILLUMINANCE already proved to be misleading.
Only publish temperature-values, when they are existent.
Move most of the DEBUG-messages to DEBUG_SENSOR_LOG.


Staars:
Related issue:
https://github.com/arendst/Sonoff-Tasmota/issues/6430

PR #6539, opened 2019-10-01
Fix TasmotaSerial: move serial send to IRAM for high speed baud rates

s-hadinger:
## Description:

Occasionally some bytes are wrongly sent at 115200 baud when using software serial. I realized that serial send needs to be in IRAM to ensure consistent timing, otherwise a memory cache miss can occur in the middle of the function, which results in corrupt byte sent.

I tried to minimize impact on IRAM: 136-148 bytes

Bumped to TasmotaSerial version 2.4.1.

## Checklist:
  - [x] The code change is tested and works on core 2.3.0, 2.4.2, 2.5.2, and pre-2.6


arendst:
If I see use of IRAM I get reluctant for implementing it.

As every use of TasmotaSerial library in a driver increments the IRAm size it soon becomes to large leading to abandon another driver from the standard set.

Has this update been tested with te current selected firmware images?

s-hadinger:
Thanks Theo. I am as reluctant as you are to burn more IRAM than necessary.

I've been thinking more. As this is mainly a concern for Zigbee, we can avoid spending IRAM from everyone and I can instead overload the TasmotaSerial class to change the `write()` function. This would use the IRAM only if Zigbee is activated.

I didn't think of it at the first place, so I'm good if you revert this PR.

arendst:
I didn't check iram usage increase but at least the current selected firmwares still compile fine.

PR #6562, opened 2019-10-04
process Backlog commands without minimum delay

ljakob:
## Description:

improve latency processing rules with multiple commands via Backlog

## Checklist:
  - [x] tested on v6.6.0

### DIFF
--- a/sonoff/sonoff.h
+++ b/sonoff/sonoff.h
@@ -124,7 +124,7 @@ const uint16_t MIN_MESSZ = 893;             // Min number of characters in MQTT 
 const uint8_t SENSOR_MAX_MISS = 5;          // Max number of missed sensor reads before deciding it's offline
 
 const uint8_t MAX_BACKLOG = 30;             // Max number of commands in backlog
-const uint32_t MIN_BACKLOG_DELAY = 0;       // Minimal backlog delay in 0.1 seconds
+const uint32_t MIN_BACKLOG_DELAY = 2;       // Minimal backlog delay in 0.1 seconds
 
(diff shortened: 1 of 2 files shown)


Jason2866:
This brakes backward compability and for some commands a delay is necessary (rf bridge)

ljakob:
> This brakes backward compability and for some commands a delay is necessary (rf bridge)

what do you propose? I've no problem keeping that patch locally. I needed a solution where rules with a couple of commands are processed without noticeable delay.

Jason2866:
The idea for the possibility to have no delay in between backlogs commands is fine.
For a 100% solution it needs to behave like before and with a (new) Setoption to change to minimum delay.
Anyway, It is up to Theo to decide!

andrethomas:
Perhaps this can be set to a default of 2 and user modified from `my_user_config.h`

I don't think it warrants wasting a setoption location for something that would not necessarily need to be changed during runtime - I think being able to set this prior to compile would be sufficient.

ascillato:
Agree

laurentdong:
I like the idea that have an adjustable Setoption for backlog delay time. It would be better to have the setting by ms instead of 100ms. In order to keep the backward compatibility, the default value could be 200.

arendst:
Phase 1

arendst:
And phase 2.

Everybody happy with command SetOption34 0..255 to set backlog delay. Default stayed at 200 mSec

andrethomas:
@arendst You're the only one that **_HAS_** to be happy but I am sure others are also :)

Jason2866:
Happiness all around 😀
Thx!

Issue #6561, opened 2019-10-04
Cannot reset Energy Today/Total for external meter


rt400:
> **GUIDE**
>
> This BUG issue template is meant to REPORT Tasmota software BUGS ONLY>
>  
> Please DO NOT OPEN AN ISSUE:
>  - If your Tasmota version is not the latest from the development branch, please update your device before submitting your issue. Your problem might already be solved. The latest precompiled binaries of Tasmota can be downloaded from http://thehackbox.org/tasmota/  
>  - If your issue is a flashing issue, please address it to the [Tasmota Support Chat](https://discord.gg/Ks2Kzd4)  
>  - If your issue is compilation problem, please address it to the [Tasmota Support Chat](https://discord.gg/Ks2Kzd4)  
>  - If your issue has been addressed before (i.e., duplicated issue), please ask in the original issue  
>  - If your issue is a Wi-Fi problem or MQTT problem, please try the steps provided in the [FAQ](https://github.com/arendst/Sonoff-Tasmota/wiki/FAQ) and troubleshooting wiki articles  
>
> Please take a few minutes to complete the requested information below. Our ability to provide assistance is greatly hampered without it. The details requested potentially affect which options to pursue. The small amount of time you spend completing the template will also help the volunteers providing the assistance to you to reduce the time required to help you.

### BUG DESCRIPTION
I try to reset Energy Total/Today unsuccessfully, after I run the EnergyReset3 0 command, I get the same values but do not get 0. If I run the EnergyReset1 0 command then TODAY gets 0 and after a few seconds suddenly it gets a value that was previously.
Also run Reset5 , still got the same values , flash again with "Erase All Sketch" got the same results.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

**FAILURE TO COMPLETE THE REQUESTED INFORMATION WILL RESULT IN YOUR ISSUE BEING CLOSED**

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Sonoff-Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [X] Searched the problem in [issues](https://github.com/arendst/Sonoff-Tasmota/issues)
- [X] Searched the problem in the [wiki](https://github.com/arendst/Sonoff-Tasmota/wiki/Troubleshooting)
[...]

PR #6564, opened 2019-10-04
Fix energyReset command issue #6561

pablozg:
Now we only use the hardware energy total counter as reference if the "setOption72" is enabled.


pablozg:
Hi, when setoption is set to 1 u

andrethomas:
Is there a specific reason why this would need to be changed during runtime?

Assuming it would be either enabled or disabled permanently it may be beneficial to set this in `my_user_config.h`

Just a thought...

meingraham:
even better - `user_config_override.h` :wink:

pablozg:
> Assuming it would be either enabled or disabled permanently it may be beneficial to set this in `my_user_config.h`

As the "sonoff-sensors" firmware has activated all supported meters, I think we have to support a precompiled option to enable or disable the hardware sync, them the users can choose what they needs without to have to compile all firmware every time they want to change it, and the change don't have a code impact because only use a existing bit.

arendst:
Agree with @pablozg and will merge. The default will be the long standing firmware counter (0), the hardware counter should be (1).

Thx all for input and considerations

meingraham:
Is this applicable only to the DDS238?

arendst:
No. It's applicable to all sensors with hardware energy total support. Currently they are
- All PZEMs
- SDM120
- SDM630
- DDS238
- DDSU666

PR #6583, opened 2019-10-07
Development Branch is also Stable

meingraham:


arendst:
Well it's still tricky as being a development branch. I would still recommend not to use it in production environment as sometimes settings are changed.

Jason2866:
Mhh, it is a bit of a Gordian Knot. Development version does have bug fixes since last release...
So it is often better than release. Many users ask for help in Discord for features which are NOT in release (introduced after).
Maybe not calling anymore development version. -> agile Tasmota version

PR #6616, opened 2019-10-11
Initial IPv6 support (SLAAC). Builds against 2.5.2 and 2.6.0 staging.…

altelch:
## Description:
Initial IPv6 support for SLAAC (automatic address configuration). Webfrontend and MQTT works using IPv6. Problems with `.text1' will not fit in region `iram1_0_seg' for all builds besides sonoff-minimal, sonoff-basic, sonoff-classic, sonoff-ir. IPv6 support seems to use some iram what causes the larger environments to fail. 

## Checklist:
Only 2.5.2 and pre-2.6 support IPv6
  - [ x] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

Greetings,

      Heiko.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -95,6 +95,32 @@ build_flags               = ${esp82xx_defaults.build_flags}
                             -fno-exceptions
                             -lstdc++
 
+[core_2_5_2_ipv6]
+; *** Esp8266 core for Arduino version 2.5.2
+platform                  = espressif8266@~2.2.2
+build_flags               = ${esp82xx_defaults.build_flags}
+                            -Wl,-Teagle.flash.1m.ld
+; Code optimization see https://github.com/esp8266/Arduino/issues/5790#issuecomment-475672473
+                            -O2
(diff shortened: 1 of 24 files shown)


arendst:
Heiko, 

thx for your PR.

As the message is only a logging message there is no need to use tranlations for it; it makes management a hell and I/We have abandoned translations for logging some time ago.

So pls update your PR by removing the changes to the language files.

Regarding the iram issue. Let's decide to only support IPv6 on core pre-2.6.0 and up. This makes management easier. We make a map file an see where the iram increase is happening an how much it is increasing (we have the tools nowadays). If we manage to bring iram down by disabling a small amount of features it makes sense to do that that way.

altelch:
Ok. Thanks for your comment. I'll remove the logging and PR again.

joba-1:
just in case you aren't aware:

you can commit to/change existing pull requests.
-- 
iphone so i typo

PR #6618, opened 2019-10-11
Initial IPv6 support for SLAAC (automatic address configuration).

altelch:
## Description:
Initial IPv6 support for SLAAC (automatic address configuration). Webfrontend and MQTT works using IPv6. Problems with .text1' will not fit in region iram1_0_seg' for all builds besides sonoff-minimal, sonoff-basic, sonoff-classic, sonoff-ir. IPv6 support seems to use some iram what causes the larger environments to fail.

## Checklist:
  - [ X] The pull request is done against the latest dev branch
Only 2.5.2 and pre-2.6 support IPv6
No, see iram error
  - [X ] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

Greetings,

  Heiko.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -216,12 +216,12 @@ build_flags               = ${esp82xx_defaults.build_flags}
 ;build_flags               = ${core_2_4_2.build_flags}
 ;platform                  = ${core_2_5_2.platform}
 ;build_flags               = ${core_2_5_2.build_flags}
-platform                  = ${core_stage.platform}
-build_flags               = ${core_stage.build_flags}
+;platform                  = ${core_stage.platform}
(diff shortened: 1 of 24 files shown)


arendst:
While compiling on stage I see the iram issue too BUT it is not caused by the IPv6 code. Analysing the .map files shows iram usage is the same for both with and without ipv6.

Now testing the pre-2.6.0 (which uses less iram anyway)...

arendst:
And indeed, on pre-2.6.0 there is no additional iram usage too.

My only concern is the increase in code size of 32k....

So I suggest I'll change platformio.ini ipv6 selection from stage to pre-2.6.0

PR #6634, opened 2019-10-13
Allow IPv6 only setup. Still dualstack but works without configured …

altelch:
…IPv4 address. WiFi.Status() returns WL_DISCONNECTED even when IPv6 global address is set. Minimal rework of setup code. This PR allows usage with IPv6 only. DualStack configures a automatic IPv4 after a while (like Windows does). DHCPv6 in LWIP2 is disabled by default and if enabled is used for DHCPv6 full config and not stateless config. Currently I have no way to configure NTP or DNS Server. Work in progress...

## Checklist:
Only pre-2.6


altelch:
Strange, some of the builds fail in travis but the code is always the same and everything builds fine local.

localhost61:
@altelch  I guess you compile locally with core_pre_ipv6

PR #6638, opened 2019-10-14
Added deepsleep function as a new xDrv

stefanbode:
## Description:
Some users battery power their ESP devices to measure outside humidity or temperature. Deepsleep is essential not to drain the battery.

Please check seriously. There are some changes to RTCMemory and Settings. a new hook FUNC_AFTER_TELEPERIOD was initiated o give the deepsleep command a time to operate. Also on wakeup if you rely on the FIRST Teleperiod (e.g. 10 sec). You have to ensure that all preparations are successfully finished. Some checks here. The deepsleep uses UTCTIME to sync wakeup with clock. e.g. deepsleep 3600 wakes up every FULL hour. e.g. 2am, 3am, 4am...
Like in ESPeasy there can be a switch defined. If the GPIO is pulled to ground the device stay awake for configuration without changing settings.

## Checklist:
  - [ x] The code change is tested and works on core 2.3.0, 2.4.2, 2.5.2, and pre-2.6
  - [ x] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).


stefanbode:
Please change the definition of max cycle. It was 1 minute for testing purposes. Travis integration fail not because of my changes :-)

#define MAX_DEEPSLEEP_CYCLE    60 --> #define MAX_DEEPSLEEP_CYCLE    3600

sonoff/xdrv_90_deepsleep.ino
```diff
@@ -0,0 +1,170 @@
+/*
+  xdrv_90_deepsleep.ino - Shutter/Blind support for Sonoff-Tasmota
```
shantur: "Deep sleep support for Sonoff-Tasmota" ?

stefanbode: ”Shutter / Blind is wrong”. Copy & Paste error. Yes deepsleep for battery based esp8266 like mentioned in the push request.

sonoff/i18n.h
```diff
@@ -193,6 +193,8 @@
 
 #define D_LOG_SOME_SETTINGS_RESET "Some settings have been reset"
 
+#define D_SENSOR_DEEPSLEEP "DeepSleep Switch"
+
```
arendst: Text is too long. It must fit in the sensor drop down list which is also language specific. 

So D_SENSOR_DEEPSLEEP needs to move to every langugage file and have a default name like "DeepSleep"

arendst: Stefan, I'll make some changes and will merge your code if I've done it.

There are some RTCMem issues too.

No need for you to do anything now.

stefanbode: "DeepSleep" for the GPIO is fine. Let's wait how this gets translated into other languages by the community. RTCMem needs recheck for the "free" bytes. You will do it right. Besides this I got no problems here with the "pre"-core.

stefanbode: Maybe you just comment out all the MORE_DEBUG messages to save space. No need for it.

digiblur:
Great work!  I'd love to test this on my outdoor weather station that is solar/battery powered.

arendst:
Is there any nice way to get it out of deepsleep? I may have put it to sleep for the next century using command ``deepsleeptime 1571060000`` 

```
15:34:01 MQT: tele/wemos4/LWT = Offline (retained)
15:34:01 MQT: tele/wemos4/UPDATE TIMER = -1152847296

press device reset
00:00:00 CFG: Loaded from flash at F7, Count 1032
00:00:00 DSL: Remain DeepSleep 1571056359

press device reset
00:00:00 CFG: Loaded from flash at F7, Count 1032
00:00:00 DSL: Remain DeepSleep 1571052759

press device reset
00:00:00 CFG: Loaded from flash at F7, Count 1032
00:00:00 DSL: Remain DeepSleep 1571049159
```

stefanbode:
power off power on. remaining is only in RSTmem

stefanbode:
Anything beyond 24x3600 = 86.400 might be a bit crazy... but who knows...

stefanbode:
if you define a deepsleep GPIO, like D0 (very common). just connect it to ground and it will stay online

stefanbode:
I think it is fine to limit it for a day. Anyhow I'm a bit nervous about the negative number:
`15:46:41 MQT: tele/wemos4/UPDATE TIMER = -1152847296`
This is not as it should be. We can define for deepsleep>0 and no deepsleep pin assigned to set it do D0 = GPIO16 by default. YOu hast to connect this to RST anyway. Therefore is is free for sure. in this case just switch from RST to GROUND and you can get control again.

arendst:
Regarding the deepsleep disable GPIO I think it makes (more) sense to disable deepsleep if any configured button or switch is pressed too. Makes that sense to you too?

digiblur:
Would we be able to set a topic/payload in MQTT as retained so that when the device wakes up it gets this message that would disable deep sleep?  

Thinking of this flow.  You have a sensor outside waking every hour or so but want to make some changes to it.  Flip the switch in your home automation software that kills the sleep.  When the wake comes around you make your changes and then kill the switch and put the device back into its sleeping loop.

stefanbode:
Yes and no. The GPIO is quite handy because it is a small hardware 3pin switch that connect GPIO16 with Ground or RST. Therefore no reconfiguration, because on every TELEPERIOD it tries to get into deepsleep. And on these devices TELEPERIOD is often 10 seconds. Therefore the button-solution is only something to help in the wakeup process. if the button changes deepsleep=0 then the wakeup works but you have to send a command to bring it back to deepsleep. As said. pros and cons.

stefanbode:
you can do this already with /cmd/<mydevice>/deepsleeptime 0

> Would we be able to set a topic/payload in MQTT as retained so that when the device wakes up it gets this message that would disable deep sleep?
> 
> Thinking of this flow. You have a sensor outside waking every hour or so but want to make some changes to it. Flip the switch in your home automation software that kills the sleep. When the wake comes around you make your changes and then kill the switch and put the device back into its sleeping loop.

arendst:
OK. I leave it for now and will release current code. Brace yourselves....

arendst:
Done.

stefanbode:
@arendst : Power off only deletes the "remainingtime" in RST. It does NOT change the deepsleeptime itself. This is in Settings.deepsleep. Therefore on 6h deepsleep. you need to reset it 6 times before it wakes-up normally. And then for sure, it takes only some seconds before going again for 6h into deepsleep. During this time you can send commands. The faster it goes to deepsleep the less battery is consumed.

arendst:
I guess it's now time you write a nice wiki page with your use case and how the different features work together.

stefanbode:
Shutter will not work with deepsleep :-) Different use-case.
Here the wiki:
https://github.com/arendst/Sonoff-Tasmota/wiki/Support-for-Deepsleep-and-Wake-Up-PIN

stefanbode:
Sorry, one mistake. You can delete this one. not needed anymore in RTCmem
`  unsigned long uptime_old;                // 2CC`

jziolkowski:
Can the LWT state be changed to "sleep" instead of offline? I mean, at the moment the device is offline, but it's not the same offline like a "disconnected" offline or during reboot. Just an idea.

andrethomas:
Thanks @stefanbode and @arendst 

Now I have to figure out which one of my 200 devices is in need of some deep sleep :)

stefanbode:
hi @jziolkowski, possible but not easy, because this is the Last-Will and the Offline is already set with the start of the device and the initial connect. Additionally "offline" is maybe the correct state because you never know if it wakes up...

stefanbode:
@andrethomas : Question should be easy to answer. As long as you have power, I would not touch them. The energy consumption is nothing (e.g. 200x 5V x 0.1A = 100W) Not worth to change it. Anyhow for battery-powered devices, that might have a small solar panel this makes absolute sense. I have ~10 of them.

jziolkowski:
Yes, that makes sense. So maybe a payload published to /RESULT topic each time the device goes to sleep, similar to when cmd Upgrade does its job and reboots the machine? Maybe with current deepsleep settings? So the user can know that the device indeed went to sleep?

digiblur:
Rule trigger maybe? On sleep and on wake?

jziolkowski:
@digiblur what would be the purpose of rule triggering on sleep, when the device should stop all actions instead of potentially starting new ones?

stefanbode:
Gents please search The wiki for deepsleep. There is a help page with a lot of use cases and how to. Rules work fine. The new wake up time is send as utc into Result, if someone is interested.

stefanbode:
Rules are great to power devices up on wake up and power them down securely before deepsleep

digiblur:
@jziolkowski custom rule to fire off a topic in mqtt

jziolkowski:
you can do it if the sleep is announced in /RESULT or /SLEEP or sth like that

jziolkowski:
@stefanbode guilty as charged, I didn't read it very thoroughly :)

andrethomas:
> Question should be easy to answer.

@stefanbode That was actually a joke :)

I was looking at the weather station which @digiblur did a video on a few days ago (https://www.youtube.com/watch?v=VUqOIPVbeF0) and thought I would build one but I'm not up to change the firmware so I thought I'd park the idea for a bit but now I can actually go ahead and start printing the plastic parts :)

digiblur:
Testing some other i2c temperature sensors in this muggy last few days here with all the rain and throwing some deep sleeping Tasmota on it is just icing on the cake for this week.

qingz2004:
Which device you guys use with deep sleep? D1 Mini or NodeMCU or ESP-01? I think ESP-01 uses less power since it has less components.

joba-1:
with deep sleep you might need pin D0/GPIO16 for external wakeup. This is not easily available on ESP-01

qingz2004:
If the external components consume too much power comparing to esp8266 itself, it totally destroys the idea.

joba-1:
so true :( 

all boards with onboard ldo or usb or power led are basically bad candidates - all consume milli amps while you want micro amps.

Cut off power for everything but the esp on your own pcb with a mosfet or use trigBoard as a readymade solution (or tell me of a third option, PLEASE :) )

arendst:
I bought the LSC Smart Door Sensor (https://www.action.com/nl-nl/p/lsc-smart-connect-deursensor/) and Smart Movement Sensor (https://www.action.com/nl-nl/p/lsc-smart-connect-bewegingsmelder/) from the Dutch Action shop. Both are  using batteries so could be a good candidate for DeepSleep.

qingz2004:
These are compitible with Tuya. Good chance to have ESP8266 inside.
https://www.aliexpress.com/item/32960523646.html
https://www.aliexpress.com/item/33047480276.html

stefanbode:
> If the external components consume too much power comparing to esp8266 itself, it totally destroys the idea.
> Anybody compared the current between the device in and out of deep sleep?
There are two options. I have some bare ESP8266 boards with a low current LDO (HT7333A https://de.aliexpress.com/item/32531141419.html) and a small transistor BC373 to cut of external consumers (use rule System#Save fo power off). Or you can use a normal wemos board and just hardwire a 5V solar cell. Only use protected 18650!!!. No more additional hardware required or useful.

stefanbode:
> with deep sleep you might need pin D0/GPIO16 for external wakeup. This is not easily available on ESP-01

Using these boards and bare ESP8266 makes it easy also for the non-experts in soldering. https://de.aliexpress.com/item/32962787646.html

stefanbode:
> Which device you guys use with deep sleep? D1 Mini or NodeMCU or ESP-01? I think ESP-01 uses less power since it has less components. And I would remove the on board LED.

Please see my posts above. With a 5V solar cell it works in most cases. Without buy the white board (link above) and a bare ESP8266 with 4MB

arendst:
Regarding the LSC hardware, see this https://github.com/arendst/Sonoff-Tasmota/issues/6624#issuecomment-542661318

qingz2004:
If you want a bare esp8266 for DIY, this one should be good for battery powered.
https://www.aliexpress.com/item/32910642969.html

PR #6639, opened 2019-10-14
Reset everything if rebooted 7 times within Bootloop period.

shantur:
## Description:

This helps to reset devices that don't have buttons or are hard to reach once installed.

## Checklist:

### DIFF
--- a/sonoff/sonoff.ino
+++ b/sonoff/sonoff.ino
@@ -1510,6 +1510,7 @@ void setup(void)
   if (Settings.param[P_BOOT_LOOP_OFFSET]) {
     // Disable functionality as possible cause of fast restart within BOOT_LOOP_TIME seconds (Exception, WDT or restarts)
     if (RtcReboot.fast_reboot_count > Settings.param[P_BOOT_LOOP_OFFSET]) {       // Restart twice
+      AddLog_P2(LOG_LEVEL_ERROR, PSTR(D_LOG_APPLICATION D_LOG_SOME_SETTINGS_RESET " (%d)"), RtcReboot.fast_reboot_count);
       Settings.flag3.user_esp8285_enable = 0;       // Disable ESP8285 Generic GPIOs interfering with flash SPI
       if (RtcReboot.fast_reboot_count > Settings.param[P_BOOT_LOOP_OFFSET] +1) {  // Restart 3 times
         for (uint32_t i = 0; i < MAX_RULE_SETS; i++) {
@@ -1531,7 +1532,14 @@ void setup(void)
         Settings.module = SONOFF_BASIC;             // Reset module to Sonoff Basic
   //      Settings.last_module = SONOFF_BASIC;
       }
-      AddLog_P2(LOG_LEVEL_DEBUG, PSTR(D_LOG_APPLICATION D_LOG_SOME_SETTINGS_RESET " (%d)"), RtcReboot.fast_reboot_count);
+      if (RtcReboot.fast_reboot_count > Settings.param[P_BOOT_LOOP_OFFSET] +5) {  // Restarted 7 times
+        SettingsErase(0);                             // Reset all settings
+        SettingsDefault();
+        SettingsSaveAll();
+        RtcReboot.fast_reboot_count = 0;
+        RtcRebootSave();
+        EspRestart();
+      }
     }
   }
 


andrethomas:
I guess the intention of this is to allow users with lights to reset their devices by causing the fast boot counter to increment but I'm just curious how the fast counter will increment to 7 when it is power cycled 7 times - remember that the RTC is not battery-backed.

shantur:
@andrethomas : TBH, I didn't know RTC isn't persisted in flash.

I tried Resetting the NodeMCU again and again and it incremented the counter. I thought the current implementation also handled power cycling, so just added a case into it.
In this case, is there any other way of implementing this?

andrethomas:
No sweat.

> In this case, is there any other way of implementing this?

Still thinking...

arendst:
Keep thinking but I'm afraid the only solution would be flash writes which you do not want during an unstable period...

On the other hand tuya uses three switch off/on to start re-init of device so it must be possible in some way.

shantur:
How about just writing onto an independent sector not related to settings?

andrethomas:
> How about just writing onto an independent sector not related to settings?

The problem is still that you will be writing to flash at a time when power may not be stable...

Yes in the cases where you actually want this feature to work you have complete control over the wall switch but if you have a few dips in your power supply from the provider and you get some funky spi transaction half written it could make things irrecoverable anyway so by fixing something else you may be creating a new kind of complication.

shantur:
Consider me 100% noob.

Can power failure while writing to a sector lead to data corruption anywhere else ?
If not then it could be a simple check the value of sector, if its anything other than valid values would lead to reset too

andrethomas:
> Can power failure while writing to a sector lead to data corruption anywhere else?

Sure, maybe a sector (4096 bytes) of flash gets erased on the wrong location... maybe remove some code from the firmware portion?

> If not then it could be a simple check the value of sector, if its anything other than valid values would lead to reset too

Its not quite that simple... what if the code that's supposed to check the sector was previously inadvertently overwritten or erased by the same mechanism designed to protect it from just this.

> My understanding is power loss can happen anytime even when doing other writes.

Yes and mitigated as far as currently possible for the Settings structure.

I recommend moving this discussion to Discord until a better solution comes to mind. Between now and then study the startup code and see how the configuration settings are handled... maybe it will inspire a new idea that can be tested.

andrethomas:
I think the general theory is that devices which are flashed for the first time with specific reference to those which may have been using other firmware such as original sonoff or tuya etc one could use this to detect a first boot and if 0xF9000 is high up enough it should remain intact during OTA also.

Theory:

1. When device has started read a uint32_t from address 0xF9000 and see if it has the known value in it

2. If known value is there then jump over the rest

3. backup current config to sram

4. perform erase of unused flash

5. erase sector 0xF9

6. write known unit32_t value to 0xF9000

7. restore the settings.

8. restart

This way you'll only use one flash cycle on that specific address on first boot of firmware or if it was somehow overwritten by other means. Either way writes to this location should be pretty negligible if its close to the top of the spi flash (i.e. out of the way of an incoming OTA)

shantur:
@andrethomas 

Below can be used for reset purposes

1. When device has started read a uint32_t from address 0xF9000 and read the value.
2. If value is between 0 - 3 , increment the value and write it.
3. If value is unknown or greater than 3 then perform erase of unused flash and restart
4. After 3 seconds, reset the value to 0 and write it to 0xF9000

andrethomas:
@shantur I think we've already deduced that steps 2-4 is not an option at this point.

Think harder :)

andrethomas:
Thinking outside the bulb now.

During startup scan for available wifi access points and if one named Tasmota is found go into AP mode which will allow the user to connect and issue whatever commands they want.

This will only be valid for that very short instance when the firmware starts so if the Tasmota SSID is not seen on the scan list then continue working as normal.

Just throwing ideas around, of course.

arendst:
Suppose the RTC memory stays valid for a few seconds AFTER power down. Has anyone tested this scenario.

While working on a Wemos D1 I had to remove power for over more than ten seconds before the RTC became invalid (tested with deepsleep and re-applying power too soon).

ascillato:
>During startup scan for available wifi access points and if one named Tasmota is found go into AP mode which will allow the user to connect and issue whatever commands they want.

It is a good idea, but I see a security hole on that. Someone can take control of your device using that.

I think the original idea is Ok:

If the device is energized and de-energized in less than 2 seconds and doing that 7 times in a row, all settings goes to default. If in the middle, the device is powered for more than 2 seconds (normal boot), the counter is reseted.

andrethomas:
> Someone can take control of your device using that.

If they know exactly when you are going to switch the device on since it will only behave this way for a very short time - it can be made more interesting by only allowing it if that particular Tasmota AP is the strongest signal in the range of the device.

arendst:
> Suppose the RTC memory stays valid for a few seconds AFTER power down. Has anyone tested this scenario.

Just did a test and alas won't work. Using a Sonoff B1 bulb the RTC is invalid immediatly.

So forget this one as you already did...

arendst:
> If the device is energized and de-energized in less than 2 seconds and doing that 7 times in a row, all settings goes to default. If in the middle, the device is powered for more than 2 seconds (normal boot), the counter is reseted.

As shown above won't work. Even a within 2 seconds the RTC data is invalid and the counter will never reach 7.

s-hadinger:
Theo, if I remember well, flash erase resets all bits to 1s. Then you can freely switch bits from 1 to 0 with SPI writes. Only resetting bits from 0 to 1 requires a full erase of the flash memory area.

Is it possible to use this property, let's say to start a specific area in flash with value 0xFF. Then turn one bit to zero at each boot and count up to 7 bits, and do a full flash write when it is stabilized (after 10 seconds).

Maybe this would avoid corrupting the flash area with fast power off/on.

Does this make sense?

ascillato2:
Closing this PR as this has been added on https://github.com/arendst/Sonoff-Tasmota/commit/782563b3ae1301b91759bcdc8f7cfcea221f3be0

PR #6659, opened 2019-10-16
Revert changes to add IPV6

Jason2866:
there are several issues rised in Discord after upgrading to latest development version.
Devices are not reachable. Only after activating IPV6 in router this devices are working normal again.
Implementation of IPV6 needs further investigations

## Description:

After OTA devices are not working anymore

## Checklist:
  - [x] The code change is tested and works on core 2.3.0, 2.4.2, 2.5.2, and pre-2.6

### DIFF
--- a/sonoff/support_wifi.ino
+++ b/sonoff/support_wifi.ino
@@ -33,9 +33,6 @@ const uint8_t WIFI_CHECK_SEC = 20;         // seconds
 const uint8_t WIFI_RETRY_OFFSET_SEC = 20;  // seconds
 
 #include <ESP8266WiFi.h>                   // Wifi, MQTT, Ota, WifiManager
-#if LWIP_IPV6
-#include <AddrList.h>                      // IPv6 DualStack
-#endif  // LWIP_IPV6=1
 
 struct WIFI {
   uint32_t last_event = 0;                 // Last wifi connection event
@@ -238,20 +235,6 @@ void WifiBegin(uint8_t flag, uint8_t channel)
   }
   AddLog_P2(LOG_LEVEL_INFO, PSTR(D_LOG_WIFI D_CONNECTING_TO_AP "%d %s " D_IN_MODE " 11%c " D_AS " %s..."),
     Settings.sta_active +1, Settings.sta_ssid[Settings.sta_active], kWifiPhyMode[WiFi.getPhyMode() & 0x3], my_hostname);
-
-#if LWIP_IPV6
-  for (bool configured = false; !configured;) {
-    uint16_t cfgcnt = 0;
-    for (auto addr : addrList) {
-      if ((configured = !addr.isLocal() && addr.isV6()) || cfgcnt==30) {
-        AddLog_P2(LOG_LEVEL_INFO, PSTR(D_LOG_WIFI "Got IPv6 global address %s"), addr.toString().c_str());
-        break;  // IPv6 is mandatory but stop after 15 seconds
-      }
-      delay(500);  // Loop until real IPv6 address is aquired or too many tries failed
-      cfgcnt++;
-    }
-  }
-#endif  // LWIP_IPV6=1
 }
 
 void WifiBeginAfterScan()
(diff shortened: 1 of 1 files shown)


arendst:
People complaining about this issue are they using the pre-compiled binaries or are they compiling it themselves? What platformio option do they use?

I cannot understand why this update causes issues and why ipv6 needs to be activated in a router as long as the default platformio compile options are used; ipv6 is not even enabled in the default compile set for pre-2.6.0

andrethomas:
Its a bug introduced by a change in platformio.ini when [core_pre_ipv6] was added and more specifically this

```
platform                  = ${core_pre.platform}
build_flags               = ${core_pre.build_flags}
platform                  = ${core_pre_ipv6.platform}
build_flags               = ${core_pre_ipv6.build_flags}
```

So it is impacting on dev bins pre-2.6

Will advise when fixed.

altelch:
Ok, as I wrote on discord I can understand why it probably causes problems with v4 when v6 is enabled. I don't check for v4 if v6 is ok. This could be changed if one needs dual-stack. But if it's not caused by my code but by enabling pre_2_6 the the revert won't solve this.

arendst:
This should be the correct platformio.ini settings:
```
[core_active]
; Select one core set for platform and build_flags
;platform                  = ${core_2_3_0.platform}
;build_flags               = ${core_2_3_0.build_flags}
;platform                  = ${core_2_4_2.platform}
;build_flags               = ${core_2_4_2.build_flags}
;platform                  = ${core_2_5_2.platform}
;build_flags               = ${core_2_5_2.build_flags}
platform                  = ${core_pre.platform}
build_flags               = ${core_pre.build_flags}
;platform                  = ${core_pre_ipv6.platform}
;build_flags               = ${core_pre_ipv6.build_flags}
;platform                  = ${core_stage.platform}
;build_flags               = ${core_stage.build_flags}
```

andrethomas:
Correct - the build script was searching for `core_pre` to select the core, so now to make it unique it has to search for `core_pre.`

Jason2866:
My 2 cents. The IPv6 implementation is not well tested. It is too early to include and risk peoples devices.

andrethomas:
> My 2 cents. The IPv6 implementation is not well tested. It is too early to include and risk peoples devices.

It is selected by user during self-compile so just wait a little so that I can fix the build script because I did not anticipate having another core_pre in platformio.ini

Jason2866:
Closing everyone who enables IPv6 should know what he/she does.

andrethomas:
@altelch I think dual-stack would be preferred if you want to continue support for IPv6 - I find it very unlikely that users would be using IPv6 exclusively... perhaps even less likely that someone actually knowingly and willingly using IPv6 on a LAN.

@Jason2866 The issue with pre-2.6 dev bins should be solved now - if you have time please test to confirm.

Why do I feel like I played a part in a Loriet cartoon strip...

altelch:
@andrethomas I'll look into probable issues regarding dual stack usage but my reason to support v6 only is that I really need this use case as I already have several tasmota-devices in my work environment and we plan to roll out more. All our mgmt. and sensor networks are IPv6 only without configured IPv4. In home use I don't really see the need to have dualstack. Mostly we see NATed legacy environments. It's a nice to have that I'm able to run dualstack and access my devices from remote if I own a static v6 prefix but in a private environment I normaly don't have constrains with the number of legacy ip addresses. Outside this scenario it's nice to not burn your legacy addresses on IOT Devices if you don't run a NATed environment. So if the PR will remain in Tasmota we'll start to continue a deeper v6 integration into some missing parts to make v6 more usefull. Currently the only part that will not work with v6 is the emulation part but there the cause is amazon echo not supporting IPv6.

Regarding Loriot cartoon strip:
https://tmpfile.uni-kl.de/feb6de58-40c4-4313-a898-2e25ac8387d6/+inline?

Greetings,
    
             Altelch.

andrethomas:
@altelch Thanks would be appreciated if you can look into it because without dual-stack I'm not sure it will ever be useful to 99% of the users... even those that want to "play" with IPv6 would want and probably need dual-stack support.

Regarding Loriot check this https://www.youtube.com/watch?v=XGzf_ENs0YU (It has English subtitles if you don't understand german)

joba-1:
how about starting with what one contributor currently needs and once that works add dual stack?

> Why do I feel like I played a part in a Loriet cartoon strip...

because life _is_ a Loriot cartoon :)

PR #6664, opened 2019-10-16
Remove build entry for core 2.4.2....

Jason2866:
because there are many unsolved issues (bugs) in this core.
The core 2.3.0 and core pre 2.6 are better. 
Users should not use core 2.4.2 for Tasmota anymore.

Edit since 2.5.2 is already removed

## Description:
Fixes weird wifi errors

## Checklist:
  - [x] The pull request is done against the latest dev branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR.
  - [x] The code change is tested and works on core 2.3.0, 2.4.2, 2.5.2, and pre-2.6
  - [x] The code change pass travis tests. **Your PR cannot be merged unless tests pass**
  - [x] I accept the [CLA](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

PR #6667, opened 2019-10-16
Drop old cores

Jason2866:
new PR for #6664


andrethomas:
I think its fine to leave them in platformio.ini for users who really want to compile on those cores for whatever reason they choose.

ascillato:
I agree to leave them. Anyway, we should add a comment in platformio.ini file with the issues of those old cores.

arendst:
Agre with @ascillato Adrian.

PR #6673, opened 2019-10-17
Nova SDS polling enhancement 

baschdello:
## Description:  

Changed the type for getting sensor values to polling. Before we used the "working period" method (the sensor starts/stops periodically on it's own and reports the data if Tasmota asks for values). Because of that I've detected a drift after some time which means that you don't get current values if "tele_period" is reached. Another point was that the "working period" method only allows a frequency above 1 minute. The new code starts the sensor directly (with commands sleep and work) if we reached a time we set in define "STARTING_OFFSET". 5 seconds before tele_period, we are doing 4 measurements, calculate the average and current values are present if tele_period is reached. The new method allows us also to get a polling frequency below 1 minute (usable for early fire detection). For precise measurements values of STARTING_OFFSET above 30 seconds are recommended. If using STARTING_OFFSET of 30 seconds and tele_period of 30 seconds the sensor now runs permanent.

Fixed also wrong values at defines NOVA_SDS_SLEEP and NOVA_SDS_WORK.

## Checklist:


arendst:
This change removes the functionality of command Sensor20.

What are the chances that a user wants to change the new STARTING_OFFSET on the fly?

altelch:
Suggestion @baschdello: Use Sensor20 Settings.novasds_period instead of STARTING_OFFSET.

PR #6704, opened 2019-10-20
Add support for Arduno Uno/Mini/Nano on Serial

andrethomas:
## Description:

This PR adds initial support to OTA flash and use a serial connected Arduino Uno/Mini/Nano.

Uploading a hex file which was compiled for an Uno/Mini/Nano is done through the same OTA menu page as would be done for an RF Bridge.

Support is added for both normal reset and inverted reset.

A normal reset is for when the reset pin would be connected directly to the reset pin of the atmega328 chip of the arduino board and inverted reset operation allows using the Green/RTS pin on the edge of the pro mini boards.

I have not extended the communication protocol so currently, only JSON support is possible in a sensor like implementation but after getting some feedback from how users want to use this functionality as an extension of their Tasmota wemos/nodemcu boards I will extend that functionality to include bidirectional control where possible (i.e. act as a sensor or a driver)

## Checklist:

### DIFF
--- a/lib/ArduinoHexParse/README.md
+++ b/lib/ArduinoHexParse/README.md
@@ -1,0 +1,3 @@
+# ArduinoHexParse
+
+Parse hex files created by Arduino for Uno/Mini/Nano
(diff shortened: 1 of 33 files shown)


arendst:
Andre thx.

While scanning the code I see some tricky assumptions (flash start location) which I will need to change. So expect an updated commit soon before you start expanding this code.

PR #6712, opened 2019-10-21
Rules: Add TuyaMcuReceived#Data=<HexBytesReceived> rule and fix others

shantur:
## Description:
Rules: Add TuyaMcuReceived#Data=<HexBytesReceived> rule and fix SerialReceived and SSerialRecieved rules

## Checklist:

### DIFF
--- a/sonoff/i18n.h
+++ b/sonoff/i18n.h
@@ -456,6 +456,7 @@
 // Commands xdrv_16_tuyadimmer.ino
 
 #define D_CMND_TUYA_MCU "TuyaMCU"
+#define D_JSON_TUYA_MCU_RECEIVED "TuyaMcuReceived"
 
 // Commands xdrv_23_zigbee.ino
 #define D_CMND_ZIGBEE_PERMITJOIN "ZigbeePermitJoin"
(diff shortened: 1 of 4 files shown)


arendst:
How often does tuya receive data? Is it only on user interaction or more often? If more than once a second do not do this. Too much traffic and what’s the use of knowing the massage size without the message?

Why do you remove the time from the serialsend receive data. It is there on user request. If rule triggers don't work then that has to be investigated.

arendst:
Ah I see. It’s not the tuya message count but it's the message. But still, how often?

shantur:
@arendst : Thanks for merging.
Tuya messages are received every 10 seconds on heartbeat and on user interaction.
I had to remove the time from SerialSend as rules don't work with that. 
rules expect JSON `{"SerialReceived":"<BYTES>"}` which is converted to `{"SerialReceived":{"Data":"<BYTES>"}}` but with Time this fails.

arendst:
Thx for the feedback.

I noticed too that serialreceived would fail. Reverting the time option is the quickest solution and I leave it be that way.

10 seconds for tuya shouldn't be a problem either.

Thx again.

PR #6730, opened 2019-10-24
Update de-DE.h

Jason2866:
## Description:

## Checklist:

### DIFF
--- a/sonoff/language/de-DE.h
+++ b/sonoff/language/de-DE.h
@@ -157,7 +157,7 @@
 #define D_TO "zu"
 #define D_TOGGLE "An/Aus"
 #define D_TOPIC "topic"
-#define D_TOTAL_USAGE "Gesamtnutzung"
+#define D_TOTAL_USAGE "Gesamtverbrauch"
 #define D_TRANSMIT "Übertragen"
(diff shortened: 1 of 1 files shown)


curzon01:
Suggest "Gesamtverbrauch" instead of "Gesamtnutzung" might be more common for "Total Usage".

When still using special char like for Ω unit "Ohm" for string constant like D_UNIT_KILOOHM it might be transformed in modern browsers, not on older on. If decission is to keep that can also be used for µ, m³... as they are all double byte UFT?

Jason2866:
Hallo Norbert, i changed Ohm to Ω, because µ is already used in the language file.
Will change "Gesamtnutzung" to "Gesamtverbrauch"

ascillato:
I prefer also the Ω instead of Ohm, but for non-technical people would be better the ohm letters approach, because using the symbol they won't be able to read that (mmm, it says 3.5 little arcs :rofl: )

Jason2866:
Non technical people will never see the Ω in Tasmota ;-)

meingraham:
> ( mmm, it says 150 little arcs 🤣 )

150 little horseshoes :wink:

PR #6772, opened 2019-10-28
Send a RAW command to the Tuya Device

Bettman66:
I flashed Tasmota on a Beca BHT-6000 thermostat. To set the thermostat, I have to send Raw commands.
Set temperature example:
TuyaSend5 0,020200040000002C
set the temperature to 20 ° C.
It would be nice if you could add that.

### DIFF
--- a/tasmota/xdrv_16_tuyamcu.ino
+++ b/tasmota/xdrv_16_tuyamcu.ino
@@ -124,7 +124,7 @@ TuyaSend4 11,1 -> Sends enum (Type 4) data 0/1/2/3/4/5 to dpId 11 (Max data leng
 
 
 void CmndTuyaSend(void) {
-  if ((XdrvMailbox.index > 0) && (XdrvMailbox.index <= 4)) {
+  if ((XdrvMailbox.index > 0) && (XdrvMailbox.index <= 5)) {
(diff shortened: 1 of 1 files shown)


shantur:
@Bettman66 : 

Data Type 5 is actually a fault message. What you are trying to implement is already implemented. 
If i am not wrong you are trying to send `55 AA 00 06 00 08 02 02 00 04 00 00 00 2C` which says
`55 AA` -> Tuya protocol identifier
`00` -> Version idetifier
`06` -> Send command
`00 08` -> Length of command packet
`02` -> DpID where to send the command
`02` -> Data type 2 = Send a value
`00 04` -> Length of Data
`00 00 00 2C` -> Value of Data. In this case = 0x2C = 44 = 22Deg C * 2 (More info here https://github.com/klausahrenberg/ThermostatBecaWifi/blob/master/src/BecaMcu.cpp#L55)

Long story short, you need to do `TuyaSend2 2,44`

Bettman66:
Thank you, that works. ;-)

arendst:
Thx @shantur.

I just finished decreasing the code increase from 1.2k to only 130 bytes by NOT using std::string

Anyway, nice excercise.

PR #6838, opened 2019-11-04
support for 4096 bits certificates

ahoiahoi:
## Description:
Adds support for user config to support 4096 bits certificates

## Checklist:


arendst:
@s-hadinger Stefan will this work with respect of (flash) memory?

s-hadinger:
The only impact is Heap memory and compute time. I don't remember the exact impact but it's still reasonable. Having it as a non-default option is the way to go.

I'm curious what is the increased time for the handshake phase.

ahoiahoi:
The handshake on 6.6.0.18 with a letsencrypt 4k certificate, ECDHE_RSA_WITH_AES_128_GCM_SHA256 cipher and chain validation is 1700ish

s-hadinger:
80MHz I presume?

This would mean it went from 0.7s to 1.7s. 

You can consider changing to 160MHz.

ahoiahoi:
Yes, 80mhz shelly 1 v3. I don't mind the duration of the handshake, it only occurs on boot up and when the connection is restarted which does not happen often

PR #6866, opened 2019-11-07
Disable support for mDNS...

Jason2866:
since mDNS needs a lot of space.
Let us test how many users really use. If there is a massiv traffic in Discord it can be enabled again...
We could use the free space for enabling KNX in standard build. No need for extra KNX version anymore.
@ascillato  did test and tasmota.bin + KNX. Does work fine

## Description:

## Checklist:


localhost61:
I do use it and don't need KNX.
That said, I don't feel really concerned as long as I just have to revert the define in `my_user_config.h`

arendst:
Let's do this once 7.1.0 is released.

ascillato:
@localhost61 

KNX and mDNS are not conflicting to each other, no problem with them.

The tests I was performing were to add to standard tasmota builds:
* All the sensors of tasmota_sensors.bin but without IR.
* KNX
* Hue Emulations.

This was a test in order to eliminate the firmware types of KNX and SENSORS. So far, this full version is 617K leaving enough room for minimal.bin for OTA update. (Also it has enough RAM and iRAM) And as now all firmwares types requires minimal for doing an OTA (except for tasmota_basic.bin), it seems to be useful to include more sensors and features in all language tasmota firmware types.

In this tests, mDNS was included too. Jason's idea is to free more space in order to add even more sensors to standard files.

ViToni:
Is this PR about removing discovery or about removing the ability to announce itself to the world?
On the former I not sure about the benefit since discovery is mainly doing DNS queries on a different address/port and doing DNS queries is basic operation for regular networking.
On the latter I would vote against removing the ability to announce itself using mDNS since it is one of the most lightweight discovery protocols.
I have many devices in my (home) network and being able to "find" them using mDNS is really helpful.

scargill:
Alright, maybe I had this wrong, I don't know, I assumed mDNS was about the device announcing itself on the network so you can call it by name - but am I right in saying this works regardless - please do not remove that - who wants to go back to talking to devices by IP number.. I'm no idea what KNX is. Can anyone clarify? If this is about the devices finding for example the local MQTT broker by name, that's anoter matter. Either way, could someone please clarify?

ascillato:
KNX is another protocol for communicating devices, like MQTT.

KNX and mDNS are not related and they are not conflicting to each other. This PR has nothing to do with KNX. It was just mentioned as an idea.

This PR aims to free more flash, RAM and iRAM if mDNS was not required or used, in order to add even more features to standard builds. 

As can be seen by the comments and emoji reactions, mDNS is used. So, it should remain in the standard builds. 

As, my comment before, my tests were not related to mDNS. My tests were related to eliminate the 2 builds called Tasmota_knx.bin and Tasmota_sensors.bin due to it is feasible now to put those features inside the standard builds.

marius:
I would like to see some actual numbers how much this saves and how that relates to the build size and maybe sizes of other features.

arendst:
Read my_user_config.h regarding size and iram usage

arendst:
We'll leave it as it is.

PR #6946, opened 2019-11-16
Added save MCU state and dump to MQTT for Tuya devices

lucamot:
## Description:

Added save/dump capability to TuyaMCU driver: this means that every dpId is saved in a memory structure and dumped to MQTT every minute.
This option is quite useful because on devices with many dpId it is of interest to monitor and record them (for example: a humidifier might report its parameters each with own dpId and also temperature/humidity with other dpIds).
Since doing the same with rules is quite complicated due to the fact that each dpId sends a specific TuyaCmd string, with this changes the status is report without doing anything.

It might also be useful to add and option to enable/disable this but I'm not sure if something is already reserved here: it might be a twin option of Option66, where instead of enabling all MQTT messages it enables only the retained one sent with TuyaDumpState. I can add this with no effort if only I know which Option it should be associated with.

## Checklist:
  - [ X] The pull request is done against the latest dev branch
  - [ X] The code change is tested and works on core 2.6

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -456,6 +456,7 @@
 #define D_CMND_TUYA_MCU "TuyaMCU"
 #define D_CMND_TUYA_MCU_SEND_STATE "TuyaSend"
 #define D_JSON_TUYA_MCU_RECEIVED "TuyaReceived"
+#define D_JSON_TUYA_MCU_STATUS "TuyaStatus"
 
 // Commands xdrv_23_zigbee.ino
 #define D_ZIGBEE_NOT_STARTED "Zigbee not started (yet)"
--- a/tasmota/xdrv_16_tuyamcu.ino
+++ b/tasmota/xdrv_16_tuyamcu.ino
@@ -49,6 +49,9 @@
 
 #define TUYA_BUFFER_SIZE       256
 
+#define TUYA_MCU_STATUS_UPDATE_INTERVAL 60
+#define TUYA_MCU_STATUS_UPDATE_DUMP_DELAY 10
+
 #include <TasmotaSerial.h>
 
 TasmotaSerial *TuyaSerial = nullptr;
@@ -68,8 +71,17 @@ struct TUYA {
   int byte_counter = 0;                  // Index in serial receive buffer
   bool low_power_mode = false;           // Normal or Low power mode protocol
   bool send_success_next_second = false; // Second command success in low power mode
+  uint8_t mcustatus_timer = 0;           // timer for MCU status update
 } Tuya;
 
+typedef struct TUYA_STATE {
+  uint8_t dpId = 0;
+  uint8_t dpIdType = 0;
+  char *dpIdData = nullptr;
+  int dpDataLen = 0;
+  struct TUYA_STATE *next;
+} TuyaState;
+TuyaState *TuyaMCUState = nullptr;
 
 enum TuyaSupportedFunctions {
   TUYA_MCU_FUNC_NONE,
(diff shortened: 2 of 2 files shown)


blakadder:
Can you please paste an example of the payload thats sent via MQTT?

lucamot:
Sure, here it is (taken from a humidifier I have with TYWE3S):

```{
  "TuyaStatus": [
    {
      "DpId": 1,
      "DpIdType": 1,
      "DpIdData": 1
    },
    {
      "DpId": 2,
      "DpIdType": 4,
      "DpIdData": "02"
    },
    {
      "DpId": 3,
      "DpIdType": 2,
      "DpIdData": 51
    },
    {
      "DpId": 6,
      "DpIdType": 4,
      "DpIdData": "00"
    },
    {
      "DpId": 7,
      "DpIdType": 1,
      "DpIdData": 0
    },
    {
      "DpId": 12,
      "DpIdType": 5,
      "DpIdData": "00"
    },
    {
      "DpId": 13,
      "DpIdType": 4,
      "DpIdData": "00"
    },
    {
      "DpId": 14,
      "DpIdType": 2,
      "DpIdData": 0
    },
    {
      "DpId": 101,
      "DpIdType": 1,
      "DpIdData": 0
    },
    {
      "DpId": 102,
      "DpIdType": 1,
      "DpIdData": 1
    },
    {
      "DpId": 103,
      "DpIdType": 4,
      "DpIdData": "04"
    },
    {
      "DpId": 104,
      "DpIdType": 2,
      "DpIdData": 23
    },
    {
      "DpId": 105,
      "DpIdType": 1,
      "DpIdData": 0
    },
    {
      "DpId": 106,
      "DpIdType": 1,
      "DpIdData": 0
    }
  ]
}```

blakadder:
Is this done due to a limitation of an external data collector (smart home solution or something else)?

I don't see anything in that payload that is not already available in TuyaReceived. 

The only plus I see is reducing the mqtt traffic but then you have the issue of states being reported only every 60 seconds instead of immediately.

Just trying to understand this feature...

lucamot:
The main difference in the usage I found useful for this kind of devices is that the different dpIds are reported each with their own command: the only way to receive them all at once is by sending command `55aa0001000000`, which results though in multiple TuyaReceived messages. If, for example, you want temperature and humidity in the same JSON payload you'll be forced to use a couple of variables and rules, which is ok but reduces this scarce resources. In this particular example what I want to accomplish is have in the same message both temperature, humidity, target humidity and the state of a led: 4 separate "TuyaReceived" messages which can be collapsed into this single big MQTT periodic status update.

This behavior of receiving multiple messages is a bit annoying for me, because what I want is to have a periodic report of all the states of the machine. The 60 seconds is because I didn't wanted to flood the MQTT queue, but might be stepped down to every second, if needed.

Also: TuyaReceived doesn't distinguish between received type of values and this is a bit tricky when parsing MQTT messages, because you always receive a hex value that has to be parsed accordingly. I found much simpler to let the ESP8266 do this kind of work and the use MQTT as a bus where "all the data is ready to be used".

Does it make sense to you? Is there another viable option you'll use for the above mentioned scenarios? If so, I'll be more than happy to discard these changes, right now I'm stuck because I used all three available rules to report only temperature & humidity!

blakadder:
I was just questioning this from my use case standpoint. My Home Assistant monitors the MQTT topic and filters out the dpId I need from JSON of TuyaReceived. Your solution would not improve anything in my system (see https://blakadder.github.io/templates/GD-30W.html and https://blakadder.github.io/templates/oil_diffuser_550ml.html)

We haven't had any reports from users of other platforms and we are not aware of their limitations so your input is valuable. 
The design of the TuyaMCU component started with switches and dimmers and we didn't even have devices like yours to predict that kind of use. Later on I helped on designing the TuyaReceived feature and TuyaSend so there is more modular control and timely status reports that can be parsed outside of Tasmota (mostly due to limited resources). 
I see you're looking at this from another perspective, relative to the tools you use, and I understand what you're trying to achieve.

The best course of action, since we're talking temperature and humidity and Tasmota already supports those, is expanding the available TuyaMCU functions to be able to map Temp and Hum fnId to their dpId and use the built in telemetry system for reporting the values (and even displaying them in the gui like for any other temp sensor)

lucamot:
I see your point, here are a few more considerations: my initial thought was to integrate the humidifier as a new supported component but then I considered a few things.

- The component has temperature/humidity but also a lot of other functions (change target humidity, set night/day mode, set humidifier mode etc.) which are too specific to be adapted to other device (even other humidifiers)
- I must admit I have a solution that is similar to the one you mentioned and 99% of the time I'm not using the Tasmota GUI
- Even if I added support for this specific humidifier, I wouldn't be able to fully control it through the UI, because it has button combinations to enter/exit program modes etc. (see here for a full detailed explanation of all the functions: https://www.proscenic.com/fr/mallManage/files/imgFile/view/6b1e10c4-3195-4c75-bf70-68859626931b.pdf)
- There would be a lot of work and code to write to support this specific device without the possibility to extend most of the functionalities to other similar devices
- Finally, this solution is to me quite useful because it virtually supports all devices without knowing how many dpIDs they will support. If tomorrow someone will integrate a TYWES3 into, say, a fridge, the current Tasmota version is perfectly adaptable to that too and this is to me quite convenient because I can concentrate on integrating it on my system (I'm not using HomeAssistant, I have a custom HomeBridge instance which bridges all the devices to HomeKit)

Anyway, I shared this just because I thought it might be useful and, to be honest, because I wanted it to be on the main project so that I can get all the updates 'for free' without following the main branch from my fork.
I see a lot of options here, starting from integrating the enabled code for all builds, from having a dedicated define that enables it (like the scripting part) to not integrating it at all. The choice is of course up to you, my device is currently working as I need and hopefully shouldn't need any updates in the future!

shantur:
How is it different from TuyaReceived ?

lucamot:
> How is it different from TuyaReceived ?
> SetOption66

In short: are you able with TuyaReceived/SetOption66 and any other combination of existing functionalities/rules/etc. to periodically check for every dpID, convert it from hex to dec and publish it in a single summary message to an MQTT broker?

I'm not, so I wrote this extension: if you are I'll be more than happy to use your solution and delete this extension...

shantur:
With SetOption66, you get notified state of every dpid on each heartbeat response, converted from hex to dec and published to MQTT. I agree it won't be one message but as series of messages for each DPid.
But then every message will have latest DPid value

lucamot:
> With SetOption66, you get notified state of every dpid on each heartbeat response, converted from hex to dec and published to MQTT. I agree it won't be one message but as series of messages for each DPid.
> But then every message will have latest DPid value

I'm not sure about the hex to dec conversion, anyway it's quite handy to have all the information in a single place, be it because I'm lazy or because not all dpID are reported periodically or predictably. It simplifies adding new unknown devices, especially when you don't have skills (or time) to write a specific adapter for it: you simply put the stock firmware on and let the real software-oriented hardware (Pc? Raspberry Pi?) to do the rest. That's a kind of an abstraction I love to have in firmwares, but again the only reason to add this for me it's to have all Tasmota updates 'for free' on the main branch, I really have no problem adding the patch to new releases by hand, if ever I'll need it again...

meingraham:
My opinion is that adding code that duplicates information already available is not worth the flash memory.

lucamot:
> My opinion is that adding code that duplicates information already available is not worth the flash memory.

#ifdef the code and enable it just as the scripting thing.

meingraham:
Duplicate information, not duplicate function. One can obtain a summary through "assembling" the information from the individual commands.

Yes, the code could be disabled/enabled by a compiler directive... but I think this introduces other issues/questions.

The first is documenting this so that it both differentiates one approach vs. the other plus making it very clear which option is there by default and which requires a user to compile their own firmware... users will inevitably try to use the feature and then likely reach out for support to find out why it isn't working.

This latter point (requiring self-compilation) also means that the majority of users won't use it anyway and "settle" for getting the information through the feature that is in the firmware by default.

With this in mind, I'd still vote for this PR to be declined. Of course, your modification is still available to you in your fork.

PR #6958, opened 2019-11-18
#enhancement: Adding option to compensate on a global level temperature

stefanbode:
## Description:
nearly all sensors have some deviation on temperature based on different reasons. To compensate for all different kinds o temperature sensors would result in a heavy add of code. Therefore I use the "small edition" with only ONE compensation factor. If there are more sensors this compensation will work for all temp-sensors. Comp can be between -12.6 °C/F and +12.6°C/F. It should suitable for most cases. Just digital +- independent if it is C or F. If changing the UNIT, the compensation will be wrong.

## Checklist:
  - [ x] The code change is tested and works on core 2.6
  - [ x] The code change pass travis tests. **Your PR cannot be merged unless tests pass**

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -290,6 +290,7 @@
   #define D_JSON_GPIO "GPIO"
   #define D_JSON_FLAG "FLAG"
   #define D_JSON_BASE "BASE"
+#define D_CMND_TEMPOFFSET "TempOffset"
 
 // Commands xdrv_01_mqtt.ino
 #define D_CMND_MQTTLOG "MqttLog"
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -434,8 +434,8 @@ struct SYSCFG {
   uint32_t      deepsleep;                 // E94
   uint16_t      energy_power_delta;        // E98
   uint8_t       shutter_motordelay[MAX_SHUTTERS];    // E9A
-
-  uint8_t       free_e9e[2];               // E9E
+  int8_t        temp_comp;
+  uint8_t       free_e9e[1];               // E9E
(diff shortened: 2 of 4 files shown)


blakadder:
A feature that got requested quite a few times in discord. Definitely more practical than using rules and var writing

arendst:
Simple and yet so useful!

stefanbode:
Just someone needs to have a good idea, where to put it in the wiki, so it can be found

blakadder:
say no more :D

meingraham:
@stefanbode 
> If changing the UNIT, the compensation will be wrong

Does this mean that if you first set the offset and then change the temperature unit (`SetOption8`), the offset value will be "converted" and therefore not what the user originally entered (i.e., wrong)?

blakadder:
the offset number will stay the same and thus changing the measurement unit will change the offset value aka temperature be wrong

meingraham:
OK... this explanation should be included - the compensation will be wrong... and why.

stefanbode:
yes. The offset will be added at the end. Maybe we should check, if we report in F there is a double call the ConvertTemp/ConvertCelciusToFahrenheit function. It works fine with the DS18b20 and DHT in °C. I made the change in The ConvertTemp and in the ConvertCelciusToFahrenheit. Maybe we need a IF clause in ConvertTemp to avoid problems, if temp is reported in F afterwards.

blakadder:
seems you've done the wiki job yourself :D

meingraham:
The command needs to be added to the Commands article instead... and referenced from each of those sensor pages. I was going to wait for Stefan to make the change he proposed above.

PR #6993, opened 2019-11-22
include light_type 3 and up for colorpicker

Staars:
Small bugfix for the colorpicker.

All light_devices above 2 should be considered as color lights (I guess):
 *  0          -                 no         (Sonoff Basic)
 *  1          PWM1       W      no         (Sonoff BN-SZ)
 *  2          PWM2       CW     yes        (Sonoff Led)
 *  3          PWM3       RGB    no         (H801, MagicHome and Arilux LC01)
 *  4          PWM4       RGBW   no         (H801, MagicHome and Arilux)
 *  5          PWM5       RGBCW  yes        (H801, Arilux LC11)
 *  9          reserved          no
 * 10          reserved          yes
 * 11          +WS2812    RGB    no         (One WS2812 RGB or RGBW ledstrip)
 * 12          AiLight    RGBW   no
 * 13          Sonoff B1  RGBCW  yes

### DIFF
--- a/tasmota/xdrv_01_webserver.ino
+++ b/tasmota/xdrv_01_webserver.ino
@@ -1022,7 +1022,7 @@ void HandleRoot(void)
             changeUIntScale(Settings.light_color[i], 0, 255, 0, 100), 'd', i+1);
         }
       }  // Settings.flag3.pwm_multi_channels
-        if (light_type > 3) {
+        if (light_type > 2) {
           char hexColor[65];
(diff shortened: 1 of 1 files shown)


s-hadinger:
@Staars The right way to check the number of channels is to use `Light.subtype` instead of `light_type`.

Since you don't have visibility on the `Light` struct in `xdrv_01_webserver.ino', you can use the following instead:

```
        if ((light_type & 7) > 3) {
```

arendst:
I'll fix now

PR #7008, opened 2019-11-23
Support 'relay' outputs triggered by 'low' values where the relay circuit is triggered by driven 3.3v being 'low'

btsimonh:
## Description:
if Relay<N>oc or Relay<N>oci are selected, then when the port value is 1, set the port to INPUT (basically, high impedance).

This is to allow relays to be attached which have 5v pullup, and are triggered by low.  If the pin is left at OUTPUT, then the 3.3v driven 'high' value sinks enough current to trigger the relay.

So; this basically allows you to use '5v' relays triggered by grounding an input.

## Checklist:
  - [ x ] The pull request is done against the latest dev branch
  - [ x ] Only relevant files were touched
  - [ x ] Only one feature/fix was added per PR.
  - [ x ] The code change is tested and works on core 2.6.1
We await the test....?
  - [ x ] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -550,6 +550,9 @@
   #define USE_TASMOTA_SLAVE_FLASH_SPEED 57600      // Usually 57600 for 3.3V variants and 115200 for 5V variants
   #define USE_TASMOTA_SLAVE_SERIAL_SPEED 57600     // Depends on the sketch that is running on the Uno/Pro Mini
 
+//#define USE_RELAYS_OC                            // Allow RelayNoc and RelayNoci in gui selection
+
+
 // -- End of general directives -------------------
 
 /*********************************************************************************************\
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -1092,6 +1092,12 @@ bool GetUsedInModule(uint32_t val, uint8_t *arr)
   if ((val >= GPIO_REL1_INV) && (val < GPIO_REL1_INV + MAX_RELAYS)) {
     offset = -(GPIO_REL1_INV - GPIO_REL1);
   }
+  if ((val >= GPIO_REL1_OC) && (val < GPIO_REL1_OC + MAX_RELAYS)) {
+    offset = -(GPIO_REL1_OC - GPIO_REL1);
+  }
+  if ((val >= GPIO_REL1_OC_INV) && (val < GPIO_REL1_OC_INV + MAX_RELAYS)) {
+    offset = -(GPIO_REL1_OC_INV - GPIO_REL1);
+  }
 
   if ((val >= GPIO_LED1) && (val < GPIO_LED1 + MAX_LEDS)) {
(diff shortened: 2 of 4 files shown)


arendst:
Why don't you use a decent driver for your 5V relays? I see no reason to implement this PR as it is bound to blow the inputs switching 5V relays.

btsimonh:
Note: Providing 5v to a pin on the ESP8266 is not necessarily a GOOD thing, but multiple places state it should not be an issue.
My WemoD1 is still alive....

Network upgrade from a version based on master did not retain configuration, but after config over serial, all works with normal 'tasmota.bin' (flashed from vscode).

btsimonh:
sorry, comments crossed over.
You have a fair point; but from what I have read previously the esp8266, although not specified as such, is 5v capable on GPIO pins.  If i can find the references would you consider?

arendst:
No I still wouldn't as the GPIO's are unable to power a relay directly. It may work for a week but fail in due time. Connecting relays directly to esp8266 gpios is bad design.

btsimonh:
I'm not going to claim this is correct :) :
https://hackaday.io/page/2024-esp8266-is-5v-tolerant-after-all

It's driving an opt-isolator in my case.  So it's basically an LED with 5v on the other end.
There may be other reasons to drive low, and go high impedance for 'high'?

s

p.s. love the changes since my custom build.  Are we about to get 4M parts recognised without changing the LD?

btsimonh:
I have added USE_RELAYS_OC around all functionality, tested with and without it defined, defaulted to excluded.
(web OTA is broken for me on development branch?).
5v tolerance was officially confirmed by the Expressiv CEO here:
https://www.facebook.com/groups/1499045113679103/permalink/1731855033731442/?hc_location=ufi
Penny for your thoughts?

Jason2866:
5v tolerance does not mean the circuit design should be use 5V (TTL levels)...
The project owner did already wrote he will not merge. It is a bad design choice
to connect a relais directly to a IO port of a logic chip. This is a general design rule.
Has nothing to do which voltage levels are used.

mike2nl:
esp8266 datasheet tells you the things you must know
- I/O Voltage:  3.6V (max)
- I/O current: pin/12mA

- Voltage and current restrictions
The ESP8266 is a 3.3V microcontroller, so its I/O operates at 3.3V as well. 
The pins are not 5V tolerant, applying more than 3.6V on any pin will kill the chip.

The maximum current that can be drawn from a single GPIO pin is 12mA.

ascillato2:
Thanks for sharing your ideas.

Closing this PR as is not going to be merged. Sorry.

btsimonh:
thanks for all the comments guys.
New 8 way relay module came today, still driven by low-going signal, but not held on by the ESP sinking current when outputting 3.3v.

PR #7039, opened 2019-11-26
domoticz blinds -> tasmota support

ahoiahoi:
## Description:

This adds support for Blinds devices in domoticz to send updates to tasmota

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -494,7 +494,6 @@
     #define D_JSON_MOTOR_MIS "setMIS"
   #endif
 
-  // Commands xdrv_27_Shutter.ino
   #ifdef USE_SHUTTER
     #define D_PRFX_SHUTTER "Shutter"
     #define D_CMND_SHUTTER_OPEN "Open"
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -494,6 +494,7 @@
     #define D_JSON_MOTOR_MIS "setMIS"
   #endif
 
+  // Commands xdrv_27_Shutter.ino
   #ifdef USE_SHUTTER
     #define D_PRFX_SHUTTER "Shutter"
     #define D_CMND_SHUTTER_OPEN "Open"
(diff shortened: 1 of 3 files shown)


roondar:
Hi @stefanbode, can you add your review ?

stefanbode:
I would not move the definitions to main. Maybe we can integrate the additional code into shutter. In general i support the Integration

roondar:
> I would not move the definitions to main. Maybe we can integrate the additional code into shutter. In general i support the Integration

Ok , So may be @arendst can approve the changes ?

arendst:
I'm in favour of Stefan's integration keeping it as confined as possible.

stefanbode:
Give me 24h and will provide a new version with same functionality hopefully. Will keep you posted here

stefanbode:
The crossreference of these two modules are only:

#define D_CMND_SHUTTER_POSITION "Position"
#define D_SHUTTER "SHUTTER"

besides this this xdrv7 could be committed without the change to the shutter driver xdrv27. the proposed changes to xdrv27 do not make it better. In general I think the main topic: D_SHUTTER is ok for the i18n. But there need be a function that you can call to get the other commands. Like the  FUNC_SHOW_SENSOR something that give you in a static order the keys for the additional functions this add-on provide. This would enable that any module can call another module without getting source code reference. Anyhow after some thinking not an easy to solve problem.

stefanbode:
If we move all shutter definitions to i18 and make them conditional with
```
#ifdef USE_SHUTTER

#endif //USE_SHUTTER
```
we can at least avoid impact on the compilation code. I will prepare a push for this change.

stefanbode:
Submitted push #7195 to make your life easier.

roondar:
> Submitted push #7195 to make your life easier.

Well done :)
Thanks you.

ascillato:
@ahoiahoi 

Hi, can you update your PR. Now, you should use the definitions made on #7195 to simplify your changes. Thanks.

PR #7048, opened 2019-11-28
Make DNS resolution order switchable by define (only IPv6 related)

altelch:
## Description:
PIO uses prebuilt liblwip6-1460-feat.a which is built with DNS resolution order IPV4 before IPv6. This causes problems in IPv6 only environments if there is a DNS record with A and AAAA. lwip then tries to connect to the resolved v4 even if there is no global IPv4 address. Didn't find a way to get pio rebuild the complete library. That's why I patched ESP8266WifiGeneric.cpp to use dns_gethostbyname_addrtype instead of dns_gethostbyname (which is only a wrapper of _addrtype and the define to prefer v4). With this patch it's possible to select resolution beaviour during compile time. The patch-script checks for IPv6 and does nothing if not enabled.

## Checklist:

### DIFF
--- a/patches/1-framework-arduinoespressif8266-change_dns_gethostbyname.patch
+++ b/patches/1-framework-arduinoespressif8266-change_dns_gethostbyname.patch
@@ -1,0 +1,6 @@
+619a620,622
+> #ifdef LWIP_IPV6
+>     err_t err = dns_gethostbyname_addrtype(aHostname, &addr, &wifi_dns_found_callback, &aResult,LWIP_DNS_ADDRTYPE_DEFAULT);
+> #else
+620a624
+> #endif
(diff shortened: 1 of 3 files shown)


altelch:
Travis fails during setup of built environment (platformio update) what is not related to the code changes.

ascillato:
Hi,

Why adding this patch at compiling time and not making a PR directly to arduino core repository?

altelch:
You are right. For current development on v6 we stumbled over this problem and searched a quick fix. Don't know how long it takes to get this PR into Arduino Core and then into the tasmota built env. On the long run it makes more sense to PR this to Arduino. I'll do that.

ascillato2:
As a new ESP8266 Arduino core (2.6.2) has been released yesterday, and as they are releasing cores more often, it makes more sense to address core related patches to them, in order to not produce future issues with the patch script in new core releases.

Thanks a lot for making the PR to them :+1:

ascillato:
Thread continues on https://github.com/esp8266/Arduino/pull/ 6854

Issue #7092, opened 2019-12-02
Return scaled peak to peak range of analog input as an option


DillonJ:
**Have you looked for this feature in other issues and in the wiki?**  
Yes
**Is your feature request related to a problem? Please describe.**  
_A clear and concise description of what the problem is._
No problem :-)
**Describe the solution you'd like**  
_A clear and concise description of what you want to happen._
I have configured a NodeMCU as a simple AC current meter using a 20A/1V current transformer and a simple circuit to add a dc bias to the ac voltage (similar to here: https://learn.openenergymonitor.org/electricity-monitoring/ct-sensors/interface-with-arduino)

The current code returns the average of 32 samples with a 1ms delay. I would like instead to receive the peak to peak voltage over this range. 32ms is plenty of time to get the peak to peak range of a 50Hz or 60Hz signal.  I have modified the code to do this myself as it is quite simple, but I am sure my code is inefficient compared to what someone else might write. This is how I have rewritten xsns_02_analog.ino, just so you know what I'm on about.

This works quite nicely and when scaled properly gives a pretty accurate current measurement and an apparent power measurement using the voltage from a PZEM based monitor I have elsewhere in the house.

Additionally it would be nice to be able to scale the value.

I can imagine this working by having an additional option in the analog dropdown like AC range or something like that, or even, scaled AC range. It would be nice to be able to send the scaling factor via MQTT.

```
uint16_t AdcRead(uint8_t factor)
{
  // factor 1 = 2 samples
  // factor 2 = 4 samples
  // factor 3 = 8 samples
  // factor 4 = 16 samples
  // factor 5 = 32 samples
  uint8_t samples = 1 << factor;
  uint16_t analog = 0;
  uint16_t analog_min = 1023;
  uint16_t analog_max = 0;
  for (uint32_t i = 0; i < samples; i++) {
    //analog += analogRead(A0);
    analog = analogRead(A0);
    if (analog < analog_min) {
      analog_min=analog;
    }
    if (analog > analog_max) {
      analog_max=analog;
    }

    delay(1);
  }
//  analog >>= factor;
  analog=analog_max-analog_min;
  return analog;
}
```

**Describe alternatives you've considered**  
_A clear and concise description of any alternative solutions or features you've considered._
Doing the scaling etc via node red - but it would be nice to have the scaled value at source and see it displayed on the webserver

**Additional context**  
_Add any other context or screenshots about the feature request here._

**(Please, remember to close the issue when the problem has been addressed)**

[...]

PR #7100, opened 2019-12-04
Support for apparent power calculation with current transformer connected at A0

DillonJ:
## Description:
This implements support for a simple AC apparent power monitor using a current transformer connected to the analog input A0. The full circuit details can be found [here](https://learn.openenergymonitor.org/electricity-monitoring/ct-sensors/interface-with-arduino) but in summary, you must have a CT with an applied DC bias connected to A0 so the raw DC reading is near 512 (for zero input) and an appropriate burden resistor such that the peak to peak voltage does not exceed 1V or 3.3V depending on whether or not your board has a voltage divider at A0.

The code measures the peak to peak range of the ADC input signal over 32 samples at 1ms intervals which is plenty to calculate the peak to peak range of a 50Hz or 60Hz signal.

You supply the appropriate scaling factor via adc_param2 to convert the raw adc range (0..1023) to the RMS current. Details of the relevant calculations are [here](https://learn.openenergymonitor.org/electricity-monitoring/ct-sensors/introduction). For a current transformer and burden resistor that reads 1V RMS for 20A RMS input current, the calculations are as follows:
1V RMS = sqrt(2)x1=1.41V peak to peak. Apply DC bias of 3.3/2 = 1.65 => max input voltage of 3.06V for 20A RMS input current. Therefore, the appropriate scaling factor is (20*(3.3/3.06))/1024=0.021063.

The adc_params must be integers, so this is multiplied by 100000 to give 2106 and is supplied via adc_param2.

Using a voltage supplied via adc_param3 in tenths of volts (so 2205 for 220.5 V) this is converted to apparent power. 

A running energy total is calculated and displayed. At the moment, this can be reset by sending a value of 1 for adc_param1 which then gets reset back to zero.

The code

Implements feature request: https://github.com/arendst/Tasmota/issues/7092

## Checklist:


digiblur:
Nice!  Looking forward to seeing this one.

arendst:
For my understanding. 

You measure the voltage once and store it fixed in a adc parameter. Then you measure the current using the ADC  input and apply a formule to calculate the apparent power which is equal to true power.

It does not expect or calculate reactive or capacitive loads as it cannot measure phase shifts.

So only AC resistive loads will provide valid power values.

Correct?

DillonJ:
> For my understanding.
> 
> You measure the voltage once and store it fixed in a adc parameter. Then you measure the current using the ADC input and apply a formule to calculate the apparent power which is equal to true power.
> 
> It does not expect or calculate reactive or capacitive loads as it cannot measure phase shifts.
> 
> So only AC resistive loads will provide valid power values.
> 
> Correct?

Yes, correct. It's not intended to be a high quality precision measurement. In my case, I have a PZEM connected elsewhere from which I can send the voltage measurement - and in any event, from measurements I have taken over long periods of time, the voltage doesn't vary that much.

Using the scaling parameter, you could correct for power factor if the nature of the load is known.

The intention here is to provide an estimate of power where precision isn't that important.

digiblur:
Perfect for those "is the clothes dryer running or not?"  I did a build of the CT clamp on ESPHome a while back when they started supported it, always good to have options and do it in Tasmota too.

arendst:
Can we decide that this solution will be used as the only power monitoring feature on a device?

If yes, then I want it to integrate with the energy monitoring code to leverage the energy code and shrink code size too.

meingraham:
@arendst Seems fair to me. If one is choosing the "cheap" route for monitoring energy, I would not think they would invest in a separate energy monitor to add to the device. And, if the device is already a power monitoring device, Tasmota already discouraging adding any external peripherals.

DillonJ:
I actually have a PZEM 004T connected to a sonoff communicating over the serial connection that works nicely... and in fact I am sending the voltage from this device over to the device with the CT to use in the power calculation.

So it wuld seem a pity to limit people to this solution, particularly since it is just calculating apparent power and doesn't calculate the power factor.

meingraham:
@DillonJ 

What I understood Theo to be suggesting is to "limit" a particular Tasmota device to have connected to it only one energy monitoring peripheral at most. In your case, one Tasmota device has the PZEM attached and reads the voltage and the current. On another Tasmota device you have this A0 current monitor and it receives the voltage measurement from the other device (the one with the PZEM). I think this fits what Theo suggests as each Tasmota device only has one energy monitor attached.

arendst:
It also allows for code re-use. The display of values could be done by the energy code already present. No need for another Voltage, Current or Power translation

DillonJ:
Ah, ok, I get you now. I suppose you could conceive a use case where someone has a PZEM connected and also uses a CT connected to A0 on the same device as a second power channel and rides on the back of the PZEM to get power factor and voltage.... but that is quite specific.

DillonJ:
If we went that route - how would the interfacing with adc_params work?

arendst:
Don't worry. I'll integrade your PR one of these days.

arendst:
Thx for your input.

I've merged it in the latest dev branch my way.

Pls give it a try as I do not have the CT handy ;-)

stackner:
u can not send calibration commands like powerset i get unknown command responce.

PR #7114, opened 2019-12-05
Add Crash Recorder to help investigate rare crashes

s-hadinger:
### CrashRecorder:

This tool is designed to help in cases of random and rare crashes. In case of a crash, the stack_trace is dumped to Serial by the Arduino Core. In case of random crashes, there is often no device recording the Serial output.

Moreover, some crashes only show `epc1:0x4000df64` as crash location. Unfortynately `0x4000df64` is actually inside the `memcpy` function is ROM, which does not tell us where this faulty calls comes from.

CrashRecorder, when enabled, stores in Flash the Stack_trace for later analysis.

Note: the Arduino Core first dumps the stack trace on Serial, then gives control to CrashRecorder. In case of crash, the state of the system is unknown so it's possible that Flash writes fail.

### How to use:

To avoid any interference with stable systems, CrashRecorder is always disabled at boot. You can enable it manually or through a rule.

Automatically enable CrashRecorder:
`Backlog Rule1 On Power1#Boot Do CrashRecord Endon; Rule1 1`

#### `CrashRecord`: will enable the crash recorder only if no crash dump is already present.

Parameters:
`(no parameter)`: Enables CrashRecorder until next boot, except if a CrashDump is alread in memory. Use without a parameter if you use a rule to enable it. This is a protection to avoid hammering the Flash in case of a bootloop.
`1`: Erases a previous CrashDump and enables CrashRecorder until next boot.
`0`: Disables CrashRecorder.

#### `CrashDump`: outputs data from the last recorded crash.
It outputs two JSON message. The first one contains register information of the crash, and the uptime in seconds when the crash appeared. The second JSON outputs a `call_chain`: the stack is filtered to show only values that are within the code range addresses (between `0x40000000` and `0x40300000`) and that are likely actual function calls. The output is limited to 64 values to fit within the MQTT buffer size.

#### `Crash`: when used with parameter `1`, generates a crash and allows to try the crash recorder.

```
00:04:00 CMD: Crash 1

Exception (28):
epc1=0x40205283 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffc50 end: 3fffffc0 offset: 01a0
3ffffdf0:  3ffffeb4 40262390 3fff1260 4025f88c  
3ffffe00:  3ffffeb4 4025f898 3fff1260 4020297c  
...
```

Full example:

```
xx:xx:xx CMD: CrashRecord
xx:xx:xx RSL: stat/tasmota/RESULT = {"CrashRecord":"Aborted: Crashrecord already present, use 1 to erase"}

xx:xx:xx CMD: CrashRecord 1
xx:xx:xx RSL: stat/tasmota/RESULT = {"CrashRecord":"Crashrecord erased, Enabled"}

xx:xx:xx CMD: Crash 1

Exception (28):
epc1=0x40205283 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

>>>stack>>>

ctx: cont
sp: 3ffffc50 end: 3fffffc0 offset: 01a0
3ffffdf0:  3ffffeb4 40262390 3fff1260 4025f88c  
3ffffe00:  3ffffeb4 4025f898 3fff1260 4020297c  
...


 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 1384, room 16 
tail 8
chksum 0x2d
csum 0x2d
v482516e3
~ld

00:00:00 CFG: Loaded from flash at FA, Count xx

xx:xx:xx CMD: CrashDump
xx:xx:xx RSL: tele/tasmota/RESULT = {"epc1":"0x40205283","call_chain":"40262390 4025f88c 4025f898 4020297c 402029ae 4020c4dd 402169c6 40219f15 402306ec 40213a85 402147b1 40216ae0 402147ca 40216a20 402181bf 4020dbc5 4021d8b7 4022b78c 401013e9 401000c0 4010f000 4010f15c 4000044c 40000000 400000f3 4000e328 "}
xx:xx:xx RSL: stat/tasmota/RESULT = {"CrashDump":"Ok"}
```

Flash impact: +1.0KB
Ram impact: +28 bytes

## Checklist:

### DIFF
--- a/tasmota/CHANGELOG.md
+++ b/tasmota/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix Zigbee uses Hardware Serial if GPIO 1/3 or GPIO 13/15 and SerialLog 0 (#7071)
 - Fix WS2812 power control (#7090)
 - Change light color schemes 2, 3 and 4 from color wheel to Hue driven
+- Add Crash Recorder to help investigate rare crashes
 
 ## Released
 
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -490,6 +490,17 @@
     #define D_JSON_MOTOR_MIS "setMIS"
   #endif
 
+// Commands xdrv_32_crash_recorder.ino
+#define D_CMND_CRASHRECORD "CrashRecord"
+#define D_CMND_CRASHDUMP "CrashDump"
+#define D_CMND_CRASH "Crash"
+  #define D_JSON_ONE_TO_CRASH "1 to crash"
+  #define D_JSON_CRASH_RECORD "Crashrecord"
+  #define D_JSON_CRASH_RECORD_ERASED "Crashrecord erased"
+  #define D_JSON_CRASH_RECORD_PRESENT D_CMND_CRASHDUMP " already present, use 1 to erase"
+  #define D_JSON_ERR_ERASE_SECTOR "unable to erase sector"
+  #define D_JSON_NO_DUMP "no dump found"
+
 /********************************************************************************************/
 
 // Log message prefix
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -268,6 +268,9 @@
 //#define MY_LANGUAGE            zh-CN           // Chinese (Simplified) in China
 //#define MY_LANGUAGE            zh-TW           // Chinese (Traditional) in Taiwan
 
+// -- Crash recorder ---------------------------
+#define USE_CRASH_RECORDER                       // record the crash stack trace in Flash for further investigation (+1.0k code, +0.03k mem)
+
 // -- Wifi Config tools ---------------------------
 #define WIFI_SOFT_AP_CHANNEL   1                 // Soft Access Point Channel number between 1 and 13 as used by Wifi Manager web GUI
 
--- a/tasmota/support_command.ino
+++ b/tasmota/support_command.ino
@@ -74,6 +74,13 @@ void ResponseCmndChar(const char* value)
   Response_P(S_JSON_COMMAND_SVALUE, XdrvMailbox.command, value);
 }
 
+void ResponseCmndChar_P(const char* value)
+{
+  char tmp[strlen_P(value) + 1];
+  strcpy_P(tmp, value);
+  ResponseCmndChar(tmp);
+}
+
 void ResponseCmndStateText(uint32_t value)
 {
   ResponseCmndChar(GetStateText(value));
--- a/tasmota/xdrv_32_crash_recorder.ino
+++ b/tasmota/xdrv_32_crash_recorder.ino
@@ -1,0 +1,238 @@
+/*
+  xdrv_23_crash_recorder.ino - record a complete stacktrace in Flash in case of crash
+
+  Copyright (C) 2019  Stephan Hadinger, Theo Arends, 
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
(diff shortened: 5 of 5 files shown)


andrethomas:
Can we enable this by default until at least the next release?

s-hadinger:
@andrethomas Sure. It is now enabled by default, but will not override a previous crash dump.

After a first crash, you still need to do `CrashRecord 1` to clear the crash dump and make a new capture.

arendst:
@s-hadinger how big will a crashdump be? In my opinion the most useful data is at the end of the stack     so I guess 128 bytes at most.

If so, why not store it in the RTC memory?

User (Tasmota) RTC memory (512 bytes) lay-out currently is a mess caused by wrong assumptions by me. This can easily be changed where I move current Tasmota used RTC memory to the beginning of user RTC memory. What is left can be used for the stack trace. 

This way possible flash corruption is not possible while still having easy access to the data.

Is this feasable?

s-hadinger:
Theo, this would be far, far better!

I discarded RTC at first because I thought it was too small. But if you free 128 bytes, I can filter out only relevant data at crash time. This leaves us with 32 deep call stack, which is hopefully enough to know where the crash comes from.

Additionally, it would avoid adding new commands: it would just be always activated, and we could just add a new STATUS option for the stack dump.

arendst:
Been looking over the RTC memory map tasmota is using. 

The RTC byte address range is from 0x000 to 0x2FF where 0x000 to 0x0FF is used by the SDK and core. User RTC memory is available from 0x100 to 0x2FF. 
Accessing RTC memory is by functions ``ESP.rtcUserMemoryWrite`` and ``ESP.rtcUserMemoryRead`` both using a byte offset of 0x100 so user RTC memory is accessed from 0x000 to 0x1FF.

KEEP IN MIND THAT THE ADDRESS USED IN THESE FUNCTIONS IS A 4 BYTE ADDRESS so using function ``ESP.rtcUserMemoryWrite(100, (uint32_t*)&RtcSettings, sizeof(RTCMEM));`` is actually accessing RTC memory at byte address 100 * 4 = 400 decimal = 0x190 hexadecimal in user space and 0x290 absolute RTC memory space. Here I went wrong in the beginning.

Currently it looks like this:
```
  // |<--SDK data (256 bytes)-->|<--User data (512 bytes)-->|
  // 000 - 0FF: SDK
  //  000 - 01B: SDK rst_info
  // 100 - 2FF: User
  //  280 - 283: Tasmota RtcReboot   (Offset 100 (x 4bytes) - sizeof(RTCRBT) (x 4bytes))
  //  290 - 2EB: Tasmota RtcSettings (Offset 100 (x 4bytes))
```

Changing Tasmota's use of RTC memory will break functionality of minimal binaries as old firmware expects to set a bit at location 0x293 (ota_loader).

So my first remark of changing tasmota's RTC lay out has to be dropped.

As I also use RTC memory 0x280 to 0x283 for reboot loop detection this still leaves you with RTC memory location 0x100 until 0x280.

So for now I suggest you start using RTC memory from 0x100 to max 0x280 leaving for me the possibility to lower my watermark of 0x280 if need be. For now I have no plans to occupy though.

s-hadinger:
Closing, will be working on a RTC version instead

s-hadinger:
Let's make it simple, I'll take exactly 256 bytes, from 0x100 to 0x1FF, that's leaving us 64 call addresses, so more than enough.

arendst:
You're my man ;-)

PR #7141, opened 2019-12-07
Fix crashdump bad JSON

s-hadinger:
## Description:

Fix to avoid generating a bad JSON from crash recorder. The test for valid value is not needed anymore in CrashDump(), it never dumps more than 63 values.

## Checklist:

### DIFF
--- a/tasmota/support_crash_recorder.ino
+++ b/tasmota/support_crash_recorder.ino
@@ -69,10 +69,8 @@ void CrashDump(void)
 
   for (uint32_t i = 0; i < count; i++) {
     ESP.rtcUserMemoryRead(i, (uint32_t*)&value, sizeof(value));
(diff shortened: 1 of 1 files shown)


arendst:
Agree.

On a side note I'll add the exception text to this status 12 too AND if no exception has occured I won't show status 12.

Do we really need up to 63 call return addresses? I think 32 is more than enough to have some idea where the cause lies. This way we solve too long status 12 json string.

s-hadinger:
From ratsputin dump of this morning, let's keep 64 values. We may need them, and we can reduce eventually.

The fix was to avoid results like `15:34:36 MQT: LSLight3/stat/STATUS12 = {"StatusSTK":{"call_chain":[,"40227c70"]}}` which is definitely invalid JSON.

PR #7148, opened 2019-12-08
Timers: Use visibility instead of disabled

ChristianTacke:
## Description:

The "+/-" chooser is only needed for sunrise/sunset.
Otherwise it's disabled currently.
This works.

But it's a little confusing. Especially, if you first had
"-" there (for sunset) and then switched to normal "time",
then the "-" is still there, but you can't change it,
because it's disabled.

It looks better, if one uses .style.visibility to hide the
element. It doesn't change the layout, just the element
isn't shown.

## Checklist:


arendst:
Thx

PR #7167, opened 2019-12-10
Allow 4 DHT sensors

localhost61:
## Description:
*** The smallest PR ever ;-) ***

Why limit DHT driver to 3 sensors only when a Wemos D1 can easily handle 4 of them?
I know one can't go further due to MQTT message size limit (I already reached this limit when trying to read 10x DS18B20).

Currently it works for more than a week (precisely since the 29th as the chart says ;-) ) at a `Teleperiod 60` with `Heap~27` but with a reduced Tasmota footprint. Now I replaced that firmware with the full size tasmota of this commit.  

chart_4x AM231_Monthly

For me the purpose is to check new and old DHT devices and calibrate, I already have one AM2301 that stick to 100% RH that I have to replace and here an other old one that shows a 10% RH deviation, so now I have some spare devices.

## Checklist:

### DIFF
--- a/tasmota/xsns_06_dht.ino
+++ b/tasmota/xsns_06_dht.ino
@@ -28,7 +28,7 @@
 
 #define XSNS_06          6
 
-#define DHT_MAX_SENSORS  3
+#define DHT_MAX_SENSORS  4
 #define DHT_MAX_RETRY    8
 
 uint32_t dht_max_cycles;


arendst:
Isn't that the beauty of Tasmota!

localhost61:
> 
> 
> Isn't that the beauty of Tasmota!

Yesss! It's so cool to adapt one's hardware to our own will... and then share it!
Love it !!!

Issue #649, opened 2017-07-27
h801 all leds flash on when applying power


jormus:
hi,

I take away the voltage of the h801 controller when i'm not using it, but when I apply voltage, the leds flash on for like 50 ms. This looks like somebody is flashing with a camera, because my led strips are RGBWA all in one. Can this be disabled somehow?

thx

arendst:
I guess not as it's the result of initial power up of the esp8266 having set the pwm ports to a default stae before the firmware kicks in.

Same happens on my BN-SZ ceiling led. Just leave the power on and turn the led off.

ramafun:
I have the same issue. I don't think it is due to a HW problem, because when I leave the original firmware and configure with ewelink, it doesn't blink on powering up.

andriej:
Exactly the same issue

bernhardhauffen:
I have the same issue. With the orignial FM the leds are not flashing while powering up.

PR #7177, opened 2019-12-10
Fix flashing H801 led at boot (#7165, #649)

### LINKED ISSUE #7165: H801 controller startup issue
### PROBLEM DESCRIPTION
In Tasmota the H801 controller flahes LEDS attached, briefly at full brilliance on powerup.

### REQUESTED INFORMATION

- [ ] Searched the problem in the [wiki](https://github.com/arendst/Tasmota/wiki/Troubleshooting)
- [ ] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
  - [ ] Pre-compiled
  - [ ] Self-compiled
```
  Configuration output here:


```
```
No rules involved


```
```
 Status is fine.


```
```
 Output is fine.


```

### TO REPRODUCE
Flash Tasmota 7.1.2 onto an H801 controller

### EXPECTED BEHAVIOUR
Power up with levels as set last time the unit wsa powered up

### SCREENSHOTS

### ADDITIONAL CONTEXT

Hi guys

Simple one here, maybe I just need enlightening if this is not new. I just received an H801 controller from AliExpress and with a suitably powerful supply and 5m of 12v RGB LED STRIP, I went into the deep end and installed Tasmota (7.1.2 release). It works, no problem but opens 2 questions.

On powerup the LEDs flash brightly for an instant - is that within Tasmota's remit or an issue with the controller?

TDM (I know that's another subject but I'll mention it while I'm here)... RGB + brilliance is great - and easy (top left of TDM screen) - but annoying that when you change colour in TDM - the brilliance goes to hell and you have to set the latter again. Wouldn't it be nice to be able to mess with the hue and leave the brightness intact?

Just trying to get the best out of already brilliant software.

Regards

Pete

s-hadinger:
## Description:

In light mode, PWM GPIO were not initialized until `LightAnimate()` is called 50ms after the boot. It seems that leaving GPIOs in undefined state sets light on H801, but has no effect on other led controllers. Maybe a MOSFET effect (I don't know enough about MOSFETs).

Now PWM GPIOs are set to low value (Off) right at start and may change to the desired value 50ms later when `LightAnimate()` is called.

**Related issue (if applicable):** fixes #7165, #649

## Checklist:

### DIFF
--- a/tasmota/CHANGELOG.md
+++ b/tasmota/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add Wifi Signal Strength in dBm in addition to RSSI Wifi Experience by Andreas Schultz (#7145)
 - Add Yaw, Pitch and Roll support for MPU6050 by Philip Barclay (#7058)
 - Add reporting of raw weight to JSON from HX711 to overcome auto-tare functionality by @tobox (#7171)
+- Fix flashing H801 led at boot (#7165, #649)
 
 ### 7.1.2.3 20191208
 
--- a/tasmota/support_tasmota.ino
+++ b/tasmota/support_tasmota.ino
@@ -1286,15 +1286,19 @@ void GpioInit(void)
   }
 #endif  // USE_SONOFF_SC
 
-  if (!light_type) {
-    for (uint32_t i = 0; i < MAX_PWMS; i++) {     // Basic PWM control only
-      if (pin[GPIO_PWM1 +i] < 99) {
+  for (uint32_t i = 0; i < MAX_PWMS; i++) {     // Basic PWM control only
+    if (pin[GPIO_PWM1 +i] < 99) {
+      pinMode(pin[GPIO_PWM1 +i], OUTPUT);
+      if (light_type) {
+        // force PWM GPIOs to low or high mode, see #7165
+        analogWrite(pin[GPIO_PWM1 +i], bitRead(pwm_inverted, i) ? Settings.pwm_range : 0);
+      } else {
         pwm_present = true;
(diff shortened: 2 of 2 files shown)


joba-1:
Don't you now switch on LEDs on controllers where a low GPIO means lights on?
If a controller switches on light before the cpu can take over to do an informed decision, that is a hardware design issue. Pretty sure it is not possible to fix that in software for everyone in all cases.

s-hadinger:
Any GPIO set to PWM is set to low and PMWi is set to high, only in light mode. This should be harmless to set lights to off at start and then change their states. 

For pure non-light PWMs, the behavior is unchanged.

bernhardfrenking:
It's still flickering with my H801 and 7.2.0. Do I have to change some settings?

FulvioSpelta:
I confirm that all the channels flash at power-on (powering on the power supply) on H801 flashed with 9.1.0 version

kugelkopf123:
As far as I know, its a HW Problem. 
I've rebuild the schematics by myself and had the same problem. Doesn't matter if there is an ESP or not. 
Its the 74HC245D on this board. It converts the 3,3V from the PWM to 5V for the Mosfets. When its powerd up it is in an undefined state. This causes the flash. I fixed this with a delay circuit on the OE pin. So the 74HC245 has a defined State when Starting up (OFF) and goes on after some 500ms. Sure this causes also a delay of 500ms when powering up with Lights set to 100%. 
Bildschirmfoto 2020-12-12 um 15 42 00
Its hard to fix something like this on the H801. Perhaps its possible to desolder just the OE pin and add a delay circuit.

PR #7192, opened 2019-12-12
Reset 0 to clear down stats

localhost61:
## Description:
When `reset 99` feature appeared in v6.6.0.10, I smiled because I had had the same need and already implemented this locally with some extends.
Then I propose you `reset 0` to reset boot count but also Wifi disconnect and MQTT diconnect counts, it's coupled with a change in the WebUI Info page to gather in the same place those 3 counters and shows the offline duration too.
It's really handy to monitor the health of tasmota devices, issue a global MQTT `reset 0` and compare counts between devices. 
Here is the result in the WebUI info page , a reset 0 will clear those counters :  

   

I compiled all Tasmota flavors successfully so no language was missed.

The code size cost is 144 bytes that may be optimized if merged wit reset 99.

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -272,6 +272,7 @@
   #define D_JSON_ONE_TO_RESTART "1 to restart"
 #define D_CMND_RESET "Reset"
   #define D_JSON_RESET_AND_RESTARTING "Reset and Restarting"
+  #define D_JSON_RESET_DOWN_STATS "Counts reset"
   #define D_JSON_ONE_TO_RESET "1 to reset"
 #define D_CMND_TIME "Time"
 #define D_CMND_TIMEZONE "Timezone"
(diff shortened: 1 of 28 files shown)


meingraham:
Was this "reset link counts" code ever merged with `Reset 99`?

localhost61:
No, I withdrew my proposal after 4 days because it did not seem to arouse much interest. :-/

meingraham:
@arendst - I think it makes sense as part of a Reset 99 global boot/link/mqtt cont reset.

localhost61:
I would love it, that will make some files I won't have to monitor for changes anymore  ;-)
And in that eventuality I see a duplicate in he-HE.h... it's strange it didn't trigger a warning.

arendst:
Did you ever thought why these counters are present? 

The uptime is important after a restart. Certain events are only allowed after restart and within some seconds and should not happen again in a running system. Resetting uptime on the fly may lead to badly initialized drivers. It's called uptime, Uptime, UPTIME and not sometime.

The other counters can be cleared with a restart so if for any reason you want to reset counters, other than the bootcount which can be reset with a command, execute a restart.

meingraham:
Uptime - agree.

I was referring to Link/MQTT counts - without a restart (i.e., a reboot). But I get it... `Reset 99` and live with a boot and reset the other counts. Not a huge hassle for sure.

PR #7193, opened 2019-12-12
separate relay and switch of a device 

RemiDing:
## Description:
Minor changes to  to use the switch of a device separate from the relay.
The relay can be adressed separate from the switch

To use this feature create a template where switch and relay have different index numbers.

Tested on shelly1 and shelly2.5

## Checklist:


ascillato:
Hi, This PR breaks backwards compatibility and the proposed behaviour is going to be different as explained in the wiki.

The default for Tasmota was single switch or 2-way switches (like for a hallway) if adding more switches.

As explained in the wiki, to modify this default, you just need to add a rule with the behaviour you want.

arendst:
Closing as this can be supported using rules.

PR #7231, opened 2019-12-15
PubSubClient: Allow clean disconnection as optional.

### LINKED ISSUE #7189: MQTT: Send Offline message to LWT on clean disconnect
Yes

On clean disconnects of Tasmota the LWT topic of the devices retains its message "Online". When for example reconfiguring the device with wrong parameters the device disconnects and can't come up again but the LWT still says "Online".

Send "Offline" message to LWT on clean disconnect. This should probably be configurable with SetOption as for example devices with DeepSleepTime set disconnect regularly but their last sent sensor values are still valid.

Ignore it as it's not such a common use case.

**Additional context**  

ascillato:
## Description:

**PubSubClient Library**: Allow clean disconnection as optional.

This avoid the automatic deletion of the LWT message in the MQTT broker for clean disconnections in order to let LWT to work on all disconnections cases (Wifi disconnection, wifi bad comms, Normal Restart, Config Changes, unintended or intentional crash, OTA updates, ArduinoOTA, etc) and showing the Tasmota device as OFF-LINE in the home automation software until it reconnects and the LWT with ONLINE is sent again.

**Related issue (if applicable):** fixes #7189

## Checklist:

### DIFF
--- a/lib/PubSubClient-EspEasy-2.7.12/src/PubSubClient.cpp
+++ b/lib/PubSubClient-EspEasy-2.7.12/src/PubSubClient.cpp
@@ -610,11 +610,13 @@ boolean PubSubClient::unsubscribe(const char* topic) {
     return false;
 }
 
-void PubSubClient::disconnect() {
+void PubSubClient::disconnect(bool disconnect_package) {
     buffer[0] = MQTTDISCONNECT;
     buffer[1] = 0;
     if (_client != nullptr) {
-      _client->write(buffer,2);
+      if (disconnect_package) {
+        _client->write(buffer,2);
+      }
       _client->flush();
       _client->stop();
     }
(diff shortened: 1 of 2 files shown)


arendst:
I still can't see why we have to deviate from the standard. A clean disconnect tells the broker we're done for now.

Pls convince me.

ascillato:
Hi Theo,

We wouldn't be deviating from the standard. The issue is that Tasmota is sending a close package to the broker on clean disconnections (restarts, ota flash process, etc.) That close package is not needed because that tells the broker to dismiss the LWT. So, when the LWT is dismissed by the broker, Tasmota device will keep been showed, in the home automation software, as ONLINE while in reality is OFFLINE. This happens because as the LWT was dismissed by the broker (due to the closing package) the broker never publish the LWT as OFFLINE.

So, an easy test is to change the topic in Tasmota of a fully configured and working device. After the Tasmota's restart you will still see as ONLINE the old topic as long as the new topic.

That is why, the use of closing package for clean disconnections is not useful for Tasmota.

arendst:
Makes sense. Thx.

PR #7239, opened 2019-12-17
Add GPS and local NTP Server Support

ascillato:
## Description:

This PR adds GPS and Local NTP Server Support for using serial GPS NMEA chips as STRATUM 1 Home Timer Server.

**Related issue (if applicable):** Improves local network time sync with a local precise time source to be used as primary or as backup source. More information at: https://www.instructables.com/id/GPS-Real-Time-Clock-Stratum-1-Time-Server/

In smart homes, the availability of a time source have become very important for automations. If smart devices don't have a correct time, some automations won't work. Most of the actual setups uses SNTP service that is syncing the time with NTP servers on the internet. Some other setups use a local NTP server in a router, but that router also requires to sync its time with a NTP server on the internet. So, in the case of an prolongated internet shortage, or that any concentrator or hub of the ISP have an incorrect time, or that the ISP have blocked the port 123; the time sync with NTP servers on internet won't be possible. In those cases, a local precise time server as a backup is very recommendable.

This PR, adds an optional and lightweight support (+3.1kb flash, +132 bytes RAM) for a local NTP server that syncs time with GPS system to make a GPS Real Time Clock Stratum 1 Time Server, to be used as backup or as a primary time source for your smart home.

Instructions on how to use it:

* You need a GPS chip, connected to any free gpio.
* In Tasmota WebUI, at module configuration, you need to set GPS_RX and GPS_TX pins, and just wait your GPS to sync to satellites.
* Then you can start the NTP server by `sensor60 9` command
* You can auto-sync Tasmota's time with the GPS time using `sensor60 11` command.
* Then, configure the rest of your devices (other tasmotas, cameras, raspberrys, routers, etc) to sync time with your new STRATUM 1 server IP.

More explanations in _xsns_60_GPS.ino_ file.

All Credits goes to **Christian Baars** ( @Staars ). Thanks Christian!!! Code taken from [Tasmota Playground](https://github.com/arendst/Sonoff-Tasmota-Playground) (Experimental drivers). Original [PR on Tasmota Playground](https://github.com/arendst/Sonoff-Tasmota-Playground/pull/2). I have just added the Gpio selections for module config and some small tweaks.

In my case, this driver solved all my network time sync issues due to my ISP can't provide a reliable time and sync is not possible most of the time.

_ToDo:_
* _NTP Server more verbose on debug log_
* _More compact GPS debug log_
* _More tweaks_
* _Add this driver commands to the settings in order to retain config_ (so far a rule can be used like: 
`rule1 on system#boot do backlog sensor60 9; sensor60 11 endon`)
* _Add information and commands to the Docs._

Tested ok:
* gps chip without fix
* gps chip with fix
* gps chip wires not connected and tasmota starts
* gps chip wires connected after tasmota starts (hot plug)
(in all these cases, this driver works fine and the time is synced as soon as gps has a fix)

## Checklist:


mike2nl:
@ascillato 
finally we get somewhere with all that. Staars has given me the possibility to test the early stages some time ago of his driver tests and it amazing to see how it works. Still we must describe which GPS chips are supported so that users have the information. Otherwise we get extra calls/questions about these sensor or that one. 
That NTP Server idea is a really good one and Staars and me we had talked about it a lot. Finally we have it. Very well done !!!

meingraham:
So... a "remote" Tasmota device with no Wi-Fi can now maintain a realtime time sync via GPS (whereas an RTC is set and can drift)... meaning the Tasmota device can now use Timers & RuleTimers. Yes?

ascillato:
Yes :) and also all the devices on the same network that point their NTP to the Tasmota device with GPS

digiblur:
nice work everyone!  This looks like a very cool project indeed!  Perfect for a simple local NTP server to point some cameras at that I have firewalled off from the world.

nagyrobi:
Would be cool to have PPS support, for under-second precision. A GPS receiver with PPS output is all that's needed, and simply an extra GPIO input in tasmota, to process the pulse-per-second impulses.

Staars:
> Would be cool to have PPS support

That should be possible to implement and would indeed be the most precise version. I already thought about it, but maybe we first start without it and add this later.

nagyrobi:
As i've noticed running through the code, this only supports proprietary Ublox protocol. I think at least the naming of the driver should reflect this in the UI to avoid confusion on the user side. Average user generally would expect that hooking up any standard GPS receiver would work - this is not the case.

The open/universal NMEA protocol is not vendor specific - perhaps implementing that would be nice too for people with existing, non-ublox receivers. Note that even ublox is able to output NMEA if configured properly.

Staars:
I understand your opinion and added the missing links to the already existing WIKI-article, where this was already written down.

IMHO NMEA is not a good choice on small MCU's as it creates a lot of overhead, but it is of course possible to implement somehow. The driver is quite protocol agnostic in its core and it should be relatively easy to add NMEA with the TinyGPS(++)-library. But this reminds me a tiny bit of the "we need Adobe Flash on smartphones"-discussion" from the past. (a very tiny bit :grim

Chaot1:
Thank you for your perfect work.
I have a small problem.
When i start the NTP Server with sensor60 9 it works for about 5 Minutes. Then i need to start it again.
How can i solve this to work permanently?

Staars:
It is impossible to guess this without further info.
Besides insufficient power supply or cabling, the number one reason for weird problems with the GPS-sensor in my internal testing was my old WIFI-router. After I got my new one, the accuracy and stability of the NTP-server improved a lot and - slightly off-topic - a seemingly defective Sonoff basic (connected temp/hum-sensor and LED-strip on the relay) began to work rock stable. It had a boot count of ~17000 (over a year) and suddenly the the crashes stopped.

Feel free to open an issue on GitHub and provide more info there. Maybe there is a bug to find and fix.

Chaot1:
Meanwhile my Problem seems to be solved.
I use a rule: 
rule1 on system#boot do sensor60 9 endon

Since that time the NTP is running permanently.
I don`t understand why, because the System don`t start new. Maybe it have some crossactions with sleepmode.

PR #7269, opened 2019-12-20
Added support DS1624 and DS1621 termometer sensor (multiple and hotplug)

### LINKED ISSUE #7227: DS16xx temperature sensors

Yes

I can't use DS1624 and other DS16xx sensors.

I can add support for this, if no one does it and does not mind.

Buy another sensors? booooo

**Additional context**  
I think to get the LM75AD as example, or give me a better example

BASM:
## Description:

**Related issue:**  #7227

## Checklist:
* [x]  The pull request is done against the latest dev branch
* [x]  Only relevant files were touched
* [x]  Only one feature/fix was added per PR.
* [x]  The code change is tested and works
* [x]  The code change pass travis tests. **Your PR cannot be merged unless tests pass**
* [x]  I accept the [CLA]

### DIFF
--- a/BUILDS.md
+++ b/BUILDS.md
@@ -107,6 +107,7 @@
 | USE_PCF8574           | - | - | - | - | - | - | - |
 | USE_HIH6              | - | - | - | - | x | - | - |
 | USE_DHT12             | - | - | - | - | x | - | - |
+| USE_DS1624            | - | - | - | - | x | - | - |
 |                       |   |   |   |   |   |   |   |
 | Feature or Sensor     | minimal | lite | tasmota | knx | sensors | ir | display | Remarks
 | USE_SPI               | - | - | - | - | - | - | x |
--- a/I2CDEVICES.md
+++ b/I2CDEVICES.md
@@ -62,4 +62,6 @@ Index | Define              | Driver  | Device   | Address(es) | Description
   39  | USE_DISPLAY_RA8876  | xdsp_10 | FT5316   | 0x38        | Touch panel controller
   40  | USE_TSL2591         | xsns_57 | TLS2591  | 0x29        | Light intensity sensor
   41  | USE_DHT12           | xsns_58 | DHT12    | 0x5C        | Temperature and humidity sensor
+  42  | USE_DS1624          | xsns_59 | DS1621   | 0x48 - 0x4F | Temperature sensor
+  42  | USE_DS1624          | xsns_59 | DS1624   | 0x48 - 0x4F | Temperature sensor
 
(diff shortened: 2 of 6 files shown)


BASM:
Other then 1624 have another interface. I rename DS16xx to DS1624.

ascillato:
Hi, please fix the conflicting files. Thanks.

BASM:
Ok, I fixed conflicts. Conflicts from CHANGELOG only. Additionally I added support for multiple sensors on one i2c bus and hot plug support.

ascillato:
Great!!! Thanks!!!

BASM:
I hurried. Made a mistake in initialization. Corrected.

arendst:
Thx for your PR.

I have some reservations:
- Current implementation does not check if the I2C address is already reserved by another sensor.
- I hate hot-plug. It's nice for a user. It's a pain for other sensors. Current implementation will always go out on the I2C bus looking for something on one of the 8 addresses. Nice on an arduino. A pain on Tasmota.

So in order to merge
- the I2C address reservation has to be implemented.
- Get rid of hot-plug as it takes too many resources every second. Just use detection once and if nothing found make sure the other function are not called as they take up valuable time.

BASM:
Hello.

> * Current implementation does not check if the I2C address is already reserved by another sensor.

Yes, my fault. I will add check before I2cValidRead16 on the DS1624GetTemp();

> * I hate hot-plug. It's nice for a user. It's a pain for other sensors. Current implementation will always go out on the I2C bus looking for something on one of the 8 addresses. Nice on an arduino. A pain on Tasmota.

I agree, this is additional processor time. But one read on the one second is very low costs. The missing hotplug makes really more pain for the user than for processor time and firmware. User have to reset device (or push driver manually) if his sensors change or there are any problems with cable.

Tasmota is not ideal. Tasmota already have the "one second" problems with RULES if the firmware needs to make action when some sensor value less or greater threshold. It makes it cycle every time.

> * Get rid of hot-plug as it takes too many resources every second. Just use detection once and if nothing found make sure the other function are not called as they take up valuable time.

Sorry, but my ideals do not allow me to do this. Let's find a compromise.
For example:
* HotPlug every 10 seconds, not one
* HotPlug mode on the driver switch on by options
* HotPlug Drivers off by default
* Or else?

Usual using expects the use of 0, or 1, maximum 2 temperature sensors of different models. There is no need to scan the entire bus every second. Only 8 addresses.

I thnink, that developement without a hot plug doesnt make a sense. Sorry.

BASM:
Support for switch off hotplug

PR #7309, opened 2019-12-24
Analog Moisture Sensor

### LINKED ISSUE #7182: Set range for analog value with AdcParam

Yes

I would like the ability to have a meaningful value returned from an analogue input.  This is generic for many applications, but in my case I am using an analogue capacitive soil moisture sensor which outputs a voltage.  Reading using the Analog input on A0 I get a value between 720 (dry) and 200 (wet).  That's small subsection of the entire range 0 -> 1023, but also it is very difficult for me to represent in my head (or in a graph) how "wet" the soil is when I e.g. get the value 250 (quite wet).  

What I would like is to be able to output the equivalent 0->100 reading.
I imagine this being configured with `AdcParam` something like:
```
AdcParam 4 200 720 100 0
```
Where 4 = Remap analogue values.
`<param1>` = lower limit
`<param2>` = upper limit
`<param3>` = value for lower limit
`<param4>` = value for upper limit

In the example I have deliberately set `<param3>` (value for lower limit) to a number greater than `<param4>` (value for upper limit) because in my example I want 0 to represent "no water" (dry) and 100 to be "full of water" (wet), so while `<param1>` should always be less than `<param2>`, `<param3>` and `<param4>` should just be a range where it is not important which is greater.
So going back to the example above, the value `250` would be:
```
int( ( ( (<param2> - <analogue-value>) / ( <param2> - <param1> ) ) * ( <param3> - <param4> ) ) + <param4> )
or
int( ( ( 720 − 250 ) ÷ ( 720 − 200 ) ) × (  100 − 0 ) ) + 0
= 90
```
So the result `90` (of 100) is much more meaningful than 250 as a measure of "wetness".

It is possible to do the manipulation later on in the stack - e.g. using templating in home assistant, but this still makes looking at the tasmota web interface and seeing a number there quite meaningless.

**Additional context**  
_none_

effelle:
## Description:
This PR adds ADC0 Moisture sensor on analog selection. 
Since those sensors have a quite limited reading range when configured as pure analog  (220-550 instead of 0-1023) I've also added a calibration using `AdcParam 6, Low, High`.
Example of calibration using a glass of water: 
```
22:34:47 MQT: stat/tasmota/RESULT = {"AdcParam":[6,0,1023]}
22:34:47 MQT: stat/tasmota/STATUS8 = {"StatusSNS":{"Time":"2019-12-23T22:34:47","ANALOG":{"Moisture":76}}}
22:34:59 CMD: backlog adcparam 6,243,632; status 8 
22:34:59 MQT: stat/tasmota/RESULT = {"AdcParam":[6,243,632]}
22:34:59 MQT: stat/tasmota/STATUS8 = {"StatusSNS":{"Time":"2019-12-23T22:34:59","ANALOG":{"Moisture":100}}}
```
The calibrated value is visible on webui too:
analog_moist

**Related issue (if applicable):** fixes #7182

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -99,6 +99,7 @@
 #define D_JSON_MEMORY_ERROR "Memory error"
 #define D_JSON_MINIMAL "minimal"
 #define D_JSON_MODEL "Model"
+#define D_JSON_MOISTURE "Moisture"
 #define D_JSON_MQTT_COUNT "MqttCount"
 #define D_JSON_NO "No"
 #define D_JSON_NOISE "Noise"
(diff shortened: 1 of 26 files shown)


arendst:
Thx.

It breaks the chirp driver but I will fix this now.

effelle:
I feared something would go south when I found  `D_JSON_MOISTURE` was already on code but not on  i18n...  
Glad you catch the issue "on the fly". :)

Issue #2333, opened 2018-04-02
Make WiFI/WLAN static IP address/gateway/DNS configurable


knopserl:
Not sure if this was already requested (I have at least searched for it).
Would it be possible to extend the WLAN config dialog to include a static IP address/gateway/DNS to store it as part of the configuration?
I dont like DHCP and flashing for a new IP address is also not convinient.

BTW: The is a branch from Stefan Bode which includes a port PCF8574 port extender, which I'm requireing like Stefan and others. Would be nice if that could be included in one of the next original Tasmota project version.

reloxx13:
how about use commands?...

knopserl:
you mean the console in the Web-UI? Can I change it with the conole commands and does it persist the WLAN settings (IP/GW/DNS)?
It would be anyhow nice to have this three parameters in the Web-UI (at least for the primary AP1 for the alternative would be even more convinient)

mizrachiran:
+1

Dickesplumpel:
Yes it persist in the WLAN settings.

Frogmore42:
Has been discussed recently.
https://github.com/arendst/Sonoff-Tasmota/issues/2219

The Web UI is there for simple things that most people need/would find useful/not confusing.  Static IP is not something most people should be using.  If you are one of the few who NEED it, it is supported directly from the console (web and serial).

What do you have against DHCP?

Aileron79:
There is absolutely nothing confusing about static IP addresses. In a proper home automation environment it makes sense to assign static ip addresses to various switches to make other hardware easily communicate with that switch. I agree that this is a required setting for any GUI.

Frogmore42:
If you feel it is required, feel free to fork the code and add it.  While there are some people who use static IP addresses, most do not, nor is there an actual need to do so in most cases.  Perhaps you have a case where you feel it is indispensable.  That is the beauty of open source, you are free to make it do what you feel is important.  I don't control what happens to Tasmota, but not even Theo (who created it and maintains it) can stop you from doing what you feel is required.

Now, if your goal is to convince Theo that static IP has to be in the GUI and you believe it is his requirement to fix, I am not sure you have done that so far (but don't let me stop you from trying :-)

As far as your last statement goes 

 `In` a proper home automation environment it makes sense to assign static ip addresses to various switches to make other hardware easily communicate with that switch

dikodahan:
I 100% agree!

Frogmore42:
[...]

Issue #2219, opened 2018-03-21
Static ip configuration through the web interface


mizrachiran:
Hi Tasmota folks,

This is a feature request for adding option to set static IP using the web interface.
I would love to have such feature.

Cheers.

mizrachiran:
I'm aware of this command and used it before.
However, I think that for such a common configuration it would be much easier to be able to set it through the UI.

Jason2866:
I agree with @Frogmore42 fixed IPs are not needed by almost everyone.
This job is best handled by a a DHCP Server (in most cases does this a router)
And there is the possibility to assign always the same IP to a MAC adress.
If not trash the so called router in the bin

akasma74:
There is Static IP field in Espurna Web interface, and I filled it in to make ip address of my RF bridge static.
However, my Virgin Media Hub 3.0 assigned it a different address.. 
Don't know who to blame, but I just decided to stick with the hub's reservation rules to be on the safe side.

Jason2866:
@akasma74  This is Tasmota not Espurna!

akasma74:
@Jason2866 I know. My post is just an illustration of what @Frogmore42 and you explained to the TS.

knopserl:
You're all right to a ctain extent :-)
However, I like to have an overview of my home network and each device especially server or home devices have it's fixed static address only my mobile devives (phones and tablets) get dynamic addresses. 
I have even reserved ranges for certain classes (server, TV, PC, Home Automation) and from .0-.99 the DHCP devices are assigned.
It also does not allways work in every case thet you can connect via hostnam (but via IP-Adress always works) and static IP devices connect faster.
But in gernal, I have nothing against DHCP.

peteeee:
Yes here too I am interested in static IP configuration.  Today I do utilize DHCP reservation for a few devices but most of my network devices are using static IP.  (subnet mask here is full now at /25).
The Sonoff modes have introduced me to using Mosquitto and Node Red.  Current automation is using X10, UPB, ZWave and Zigbee.  I like what I can do with Mosquitton/Node Red.  

![image](https://user-images.githubusercontent.com/8072890/41185968-24fe6564-6b54-11e8-866b-d621128c1e39.png)

andrethomas:
@peteeee 

>  (subnet mask here is full now at /25)

You use live internet IP addresses for Tasmota devices? Just trying to understand why you have a /25 limitation?

andrethomas:
[...]

PR #7314, opened 2019-12-24
AP MODE: Added Button to Configuration

### LINKED ISSUE #7306: Restore Configuration on first boot.

yes

I can't fast deploy/restore configuration if reset. Resetting the configuration has become very easy.

Add "Restore Configuration" on first boot menu.

I don't know. Do recovery in two steps? Not user friendly.

**Additional context**  

It became extremely important because of the default reset after 3 power cycles. 
You may reset  besauce of poor AC/DC. For example, I added 4 relay to Sonoff 4ch,
switch on 8 relays and got reset configuration after 2-5 minuts.

Of course.

ascillato:
## Description:

Added easy access to all configurations while on AP mode. This helps when the user wants to restore a previously saved configuration, config MQTT, set fixed IP by commands, etc, etc, making the initial setup faster.

**Related issue (if applicable):** fixes #7306, #2219 and #2333

BEFORE and AFTER this PR:

image

The configuration button goes to the standard configuration MENU of Tasmota, where the user can reset the actual configuration, restore a previously saved configuration, config MQTT, etc. On the Configuration MENU, if you click in MAIN MENU, it returns to the main AP menu showed in the picture above. The same goes to the console. Fixed IP and other settings can be defined by commands in the console.

## Checklist:

### DIFF
--- a/tasmota/xdrv_01_webserver.ino
+++ b/tasmota/xdrv_01_webserver.ino
@@ -1693,10 +1693,11 @@ void HandleWifiConfiguration(void)
   }
 
   if (WifiIsInManagerMode()) {
-    WSContentSpaceButton(BUTTON_RESTART);
 #ifndef FIRMWARE_MINIMAL
-    WSContentSpaceButton(BUTTON_RESET_CONFIGURATION);
+    WSContentSpaceButton(BUTTON_CONFIGURATION);
+    WSContentButton(BUTTON_CONSOLE);
 #endif  // FIRMWARE_MINIMAL
+    WSContentSpaceButton(BUTTON_RESTART);
   } else {
     WSContentSpaceButton(BUTTON_CONFIGURATION);
   }


meingraham:
Console button :100: :100: :100: :100:

arendst:
What happens if the user selects configuration then mqtt configuration then save? The wifi credentials are still empty. The mqtt save will trigger a restart. The Ap might not come up again. 

Makes this a better initial config experience?

I think you do one step too many. Config restore is fine but providing individual config pages with their save buttons will make initial config a hell.

Jason2866:
Mhh, is there a easy way (not much code) to check if wiifi setting is done?
I agree wifi has to be set first, everything else can be corrected (if done wrong) 
without "the chance" to brick the device.
Having all the configuration options gives the possibility to set up the device with "one stop"
No more searching for IP address the device has got from DHCP to do the config in the needed second step with the actual way it is done now.

ascillato:
I agree. Too many fixes at once.

Let's simplify.

By now, just add the Restore button asked on #7306 

Closing this PR in favour of #7329

PR #7355, opened 2019-12-28
Counter - SetOption79 to enable reset of counters after telemetry sent

andrethomas:
## Description:

This PR adds functionality to enable the resetting of counters after telemetry data is sent.

This is useful for counting the number of triggers from any given external signal for the entire duration of teleperiod (usually 300 seconds / 5 minutes) from for example the impulses generated by the optically isolated output of energy meters or the number of impulses received in any given telemetry period from a wind speed sensor.

It should default to SetOption79 0 which means this is not enabled by default.

## Checklist:


arendst:
You'll need to move your code down as currently the Domoticz users won't see any change ;-)

PR #7366, opened 2019-12-29
support for hm17 (v110) and hm10 (V609)

gemu2015:
## Description:

support for hm17 firmware V110 (V120 worked already)
and for hm10 firmware V609 (beware of hm10 clones!)

## Checklist:

### DIFF
--- a/tasmota/xsns_52_ibeacon.ino
+++ b/tasmota/xsns_52_ibeacon.ino
@@ -27,7 +27,11 @@
 
 #define HM17_BAUDRATE 9600
 
-//#define IBEACON_DEBUG
+#define IBEACON_DEBUG
+
+// use this for Version 110
+#define HM17_V110
+
 
 // keyfob expires after N seconds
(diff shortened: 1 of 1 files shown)


Jason2866:
@gemu2015  THX! Which chip has to be used for the HM-10?

gemu2015:
i bought a (claimed) compatible clone which did not work (only peripheral role)
@lacrimosa2k2 has an hm10 reflashed module (V609) which worked 
however all versions have slight differences in AT commands in how ibeacon mode is handled (which is stupid)
as long as it supports central role and ibeacon mode it should work.

as we use AT mode the chip itself should not matter.

(newer modules may be reflashed with a working version)

lacrimosa2k2:
@Jason2866 yes the cc2541 is used in the HM-10 module

i have updated the doc with flash instruction for the hm-10
https://tasmota.github.io/docs/#/peripherals/iBeacon-driver

PR #7430, opened 2020-01-04
Allow overrides of USE_I2C and USE_SPI

kabongsteve:
## Description:

The current my_user_config.h has some #IFDEF's based on the USE_SPI and USE_I2C

these #ifdefs occur prior to the include of user_config_override.h and thus do not react to any attempt to override these USE_'s.

The #IFDEF's sections have been moved to the end of the file after the override include so that the override can support these USE_ definitions

## Checklist:
  - [ X] The code change is tested and works on core 2.6.1

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -342,8 +342,8 @@
   #define TUYA_DIMMER_ID       0                 // Default dimmer Id
 #define USE_ARMTRONIX_DIMMERS                    // Add support for Armtronix Dimmers (+1k4 code)
 #define USE_PS_16_DZ                             // Add support for PS-16-DZ Dimmer (+2k code)
-#define USE_SONOFF_IFAN                          // Add support for Sonoff iFan02 and iFan03 (+2k code)
-#define USE_BUZZER                               // Add support for a buzzer (+0k6 code)
+//#define USE_SONOFF_IFAN                          // Add support for Sonoff iFan02 and iFan03 (+2k code)
+//#define USE_BUZZER                               // Add support for a buzzer (+0k6 code)
 #define USE_ARILUX_RF                            // Add support for Arilux RF remote controller (+0k8 code, 252 iram (non 2.3.0))
 //#define USE_SHUTTER                              // Add Shutter support for up to 4 shutter with different motortypes (+6k code)
 #define USE_DEEPSLEEP                            // Add support for deepsleep (+1k code)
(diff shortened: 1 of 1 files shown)


ascillato:
I have tested your issue and couldn't reproduce it. If I do #undef USE_I2C in user_config_override.h it got undefined and not used.

Exactly what is your issue?

kabongsteve:
My issue is about all the additional #defines that occur in the my_user_config.h immediately after the #define USE_I2C
ie
#ifdef USE_I2C
  #define USE_SHT                                // [I2cDriver8] Enable SHT1X sensor (+1k4 code)
  #define USE_HTU                                // [I2cDriver9] Enable HTU21/SI7013/SI7020/SI7021 sensor (I2C address 0x40) (+1k5 code)
  #define USE_BMP           
   ...

So we have an issue where there are a set of #defines, that are dependent on another #define.

Currently I'd have to #undef USE_I2C and ALL dependent defines as well in the override.

I would suggest that perhaps the following change would improve the situation (and simplify) for all

Leave my changes (leave it after), but update the #ifdef USE_I2C to

#if defined(USE_I2C) && ! defined(USE_CONFIG_OVERRIDE)

Thus, in the override you can either 
    #undef USE_I2C
or
  #define USE_xxx for whichever I2C sensors you wish

Thus the override is either turn ALL off, or add whichever you actually want

And the same for USE_SPI

ascillato:
The idea of the user_config_override.h is no modify the defaults. So if you want I2C and don't want the other sensors, you need to Undef all the sensors that are in the default but you don't need/want. The change in this PR will define all sensors that are in the default when you enable I2C in user_config_override.h.

That is why, seems to be better to leave as it was. The file user_config_override.h has the final word (so it must be at the end of the my_user_config.h file) and is to OVERRIDE the default configuration, by #define and by #undef

ascillato:
What I meant is, with this PR, for example, if I undef USE_SHT in user_config_override.h, it got defined anyway because the define is after the inclusion of user_config_override.h

So, it is better to leave these files as they are right now, and that you undef the sensors that you don't need in your own user_config_override.h file

ascillato2:
Thanks for sharing your ideas, but my_user_config.h file is better to be left as it is right now.

kabongsteve:
I guess my logic and guiding principals in allowing for differing configurations is to only turn on or define what you want when overriding a default, rather than have to have a convoluted set of on and offs to get what you want.

The additional issue here is that potentially every time the standard my_user_config.h is changed, then users have to see if any existing defines have changed and then update the override to handle this change, otherwise unintentional changes can occur without a user making changes.

Having a default configuration with half define'd half undefined (or commented out) and other inconsistencies makes it much more difficult to customize, as well as more complex than a simple #define here or there.

Perhaps a better implementation would be to take advantage of PlatformIO functionality, something like:-

move my_user_config.h into root include folder and rename tasmota.h
clone to tasmota-template.h but comment most settings out
clone for each supported env name

update platformio.in with 
[common]
build_flags               = ${core_active.build_flags}
                                 -include "${PIOENV}.h"

Thus for anyone to have a different setup
1) add a tasmota-XXXX to default_envs (or rename one
2) add a [env:tasmota-XXXX] section (needs nothing more)
3) copy tasmota-template.h to tasmota-XXXX.h
4) edit tasmota-XXXX.h and make your changes

Build away (quick test worked flawlessly)

Also almost means platformio_tasmota_env.ini would be simplier and override not needed.

It would then also be easier to split up tasmota.h into other includes for different reasons and then share.

ascillato:
That is another way of doing the override, but in the end is going to be the same as the actual way.

Also, remember that Tasmota was written for platformio and for Arduino. It should compile in both IDEs. Moving the my_user_config.h to root won't allow Tasmota to be compiled under Arduino.

Besides, take into account that changes in my_user_config.h are for features (that are added or deleted), config parameters and sensors (some I2C and some other serials, etc etc) 

So, it can't be warranted that your user_config_override.h will never be outdated and can always be compiled. We encourage to read the changelog everytime you update your codebase. Some things can change, can get deprecated, can be added, etc. We try to leave everything backwards compatible as far as we can, but sometimes some things can't be.

PR #7431, opened 2020-01-04
Improve Shutter operation for HASS, MQTT and Internationalization

kabongsteve:
## Description:
1) Update ShutterPosition(X) command so that UP/OPEN DOWN/CLOSE and STOP can be used (as well as percentage).  This negates need for using Backlog command for Home-Assistant or other controllers.

This also uses the Internationalization so they can be updates for other languages.

2) Update MQTT Reporting so that ALL Shutters are reported every time (regardless of current movement)
This removes parsing errors in Home-Assistant when updating status of MQTT Covers and controller supports more than 1 Shutter.

## Checklist:
  - [2*] Only one feature/fix was added per PR.
  - [ X] The code change is tested and works on core 2.6.1

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -500,6 +500,8 @@
 #define D_PRFX_SHUTTER "Shutter"
 #define D_CMND_SHUTTER_OPEN "Open"
 #define D_CMND_SHUTTER_CLOSE "Close"
+#define D_CMND_SHUTTER_UP "Up"
+#define D_CMND_SHUTTER_DOWN "Down"
 #define D_CMND_SHUTTER_STOP "Stop"
 #define D_CMND_SHUTTER_POSITION "Position"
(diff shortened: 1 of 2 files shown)


meingraham:
Should modifications intended primarily for Home Assistant be enclosed within `#ifdef USE_HOME_ASSISTANT` compiler directive blocks?

kabongsteve:
The change is about reporting from the device.  Since the device is, to some degree, pretending to be multiple devices, there is the need to make sure it operates effectively.

Thus, we either need different messages for different shutters, or the messages need to include all shutters in the data for the messages.

Using SetOption4 would make the messaging unique per Shutter, and that's a possible alternate to this solution.

But its also about consistency, and it was the interaction with Hass that showed the inconsistencies in the current messages.  From my analysis the HOME_ASSISTANT #define is about the discovery functionality and not the MQTT messaging.

PR #7487, opened 2020-01-11
Light palette

pcdiem:
## Description:

Add support for a light Palette. The Palette command set the palette colors. Schemes 2, 3 and 4 use the palette instead of the full color spectrum. The Color7 command sets the color based on the palette index or next (+)/previous (-) palette color.

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -378,6 +378,7 @@
 #define D_CMND_LED "Led"
 #define D_CMND_LEDTABLE "LedTable"
 #define D_CMND_FADE "Fade"
+#define D_CMND_PALETTE "Palette"
 #define D_CMND_PIXELS "Pixels"
 #define D_CMND_RGBWWTABLE "RGBWWTable"
 #define D_CMND_ROTATION "Rotation"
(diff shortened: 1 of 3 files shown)


arendst:
I have to get acquanted with the idea of losing 80 bytes of Settings space without knowing what I should do with it. 

I guess I have to add colors for 16 steps but have no idea what they should look like.

mike2nl:
Why not add the driver to the Tasmota-Playground, as an idea.

DavinKD:
While you guys are pondering this, One issue I have with scheme 2 is that it always starts at the last color it left off on.  This makes it impossible to have 2 different devices loop in unison.  I made a custom bin that added Light.wheel = 0; in the CmndScheme() subroutine to fix this.  I didn't want to do a pull request, because I'm guessing others may want it that way which would then require a setting.  Sounds like Color7 addresses this?

pcdiem:
Color7 in Light palette doesn't address solve the issues of schemes 
starting with the last color. Color7 just uses the palette as the set up 
colors to use instead of the full color palette.

Seems like the ability to make schemes always start with wheel at zero 
would be a good flag option in settings.

PR #7521, opened 2020-01-15
Make the B or brightness part of the HSBColor command optional

dgwynne:
This makes it easier (possible?) to use the `HSBColor` with `hs_command_topic`
in Home-Assistant, which in turn allows completely independent control
of the color part of an RGBW or RGBWW bulb. HA setting RGB colors via the Color command means the white channel gets reset to 0, even in split mode where it is particularly annoying.

Hopefully the generated code is a bit shorter too.

I tested this on a bulb that was configured with RGB and W split. The change to using `light_state.getBriRGB()` instead of `light_state.getBri()` as the default for the brightness will probably change behaviour when the bulb isn't split.

## Checklist:

### DIFF
--- a/tasmota/xdrv_04_light.ino
+++ b/tasmota/xdrv_04_light.ino
@@ -2249,31 +2249,31 @@ void CmndHsbColor(void)
     bool validHSB = (XdrvMailbox.data_len > 0);
     if (validHSB) {
       uint16_t HSB[3];
-      if (strstr(XdrvMailbox.data, ",") != nullptr) {  // Command with 3 comma separated parameters, Hue (0<H<360), Saturation (0<S<100) AND Brightness (0<B<100)
-        for (uint32_t i = 0; i < 3; i++) {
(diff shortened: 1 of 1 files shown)


Jason2866:
> The change to using `light_state.getBriRGB()` instead of `light_state.getBri()` as the default for the brightness will probably change behaviour when the bulb isn't split.
Does this mean that you have NOT tested if this PR brakes other things?
If not please do test. Thanks

dgwynne:
I'll unsplit RGB and W (ie, `SetOption37 0`) when I get home and see what the effect is when switching between white and RGB lights. You're mostly interested in what happens with `HSBColor1`, `HSBColor2`, `HSBColor H,S` and `White` are set, right?

dgwynne:
Here's the state of the bulb according to MQTT while trying to commands I described above: [mqtt.txt](https://github.com/arendst/Tasmota/files/4070066/mqtt.txt)

PR #7581, opened 2020-01-22
make kCodeImage/my_image more customizable

patrickdk77:
## Description:
Add ability to define more subversions in user overrides and remove ram/flash usage to store all versions.

## Checklist:

### DIFF
--- a/tasmota/tasmota.h
+++ b/tasmota/tasmota.h
@@ -31,6 +31,7 @@
 \*********************************************************************************************/
 
 #define CODE_IMAGE 0
+#define CODE_IMAGE_STR "tasmota"
 
 #define USE_LIGHT                           // Enable light control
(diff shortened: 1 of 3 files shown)


Jason2866:
Tbh, i dont get the benefit and i like to know ram/flash usage

patrickdk77:
Sorry thought I out that in but got distracted.
Total firmware savings of 80bytes.
Ram savings is atleast 22bytes.

The real benefit is letting the user in user overrides is able to put on  custom tag for their custom builds. Like tls enabled, shutters, zigbee, so you know you had that custom version on it. Currently you can only select one of the defined flavors only.

arendst:
This functionality is also used by the hackbox compiler to add commit numbers to the code. It will need more research...

PR #7584, opened 2020-01-23
Fif le01mr

saper-2:
## Description:

Added driver for F&F LE-01MR Single Phase Modbus Energy meter.
I based the code on SDM-120 driver.

## Checklist:

Need to translate one added define to other languages from English to that language (I did copy-paste it from en-GB; my language pl-PL & sv-SE should be ok).
I did update some comments too (from SDM220 to "SDM120, SDM220 and LE-01MR" 😄  )

I'm also willing to update docs too 😄


arendst:
Thx.

What's the difference between the tasmota ``Total`` and your ``Total Active``? Shouldn't they be the same? If not, did you use ``SetOption 72 1``

saper-2:
Total Active is a Energy counted by LE-01MR (from it's register - it can not be reset by any means). While Tasmota energy counters can be reset (e.g.: zero-memory and flash new tasmota to ESP).

ascillato:
So, TOTAL ACTIVE label should be ENERGY COUNTER TOTAL?

arendst:
Yes and No. TOTAL ACTIVE is the same as the hardware counter as will be shown when ``SetOption72 1`` is set.

With this enabled you can tell Tasmota to reset the counters to any value you like as it will substract that value from the total value provided by your device. See  command ``EnergyReset4``.

The point here is that I do not want to add another field when there is already a field for that. This way every new energy monitoring device adds it's own counters duplicating functionality that is already present.

I'm still in limbo if I want this to add.

saper-2:
Hmmmm....
The one thing that bugging me it I would like to have both 😄 (yes, I'm very GREEDY person 😈  ): the value that I can preset in tasmota (energyReset4), and the meter counter value.

But if this bugging you that much 😁 , then I'll settle to Option72 , and let the user to decide by leaving disabled "#define" to enable this additional value in mqtt json , would be that fine?

@ascillato  - yes, you can name it that, I was using name from registers description so I might not give this value best name.

saper-2:
Then should I revert all those changes to language files? Then, I would move this additional string to driver file.

Issue #7402, opened 2020-01-02
Add decimal precision for moisture sensor


robin13:
**Have you looked for this feature in other issues and in the docs?**  
Yes

**Is your feature request related to a problem? Please describe.**  

This is related to https://github.com/arendst/Tasmota/issues/7182 and https://github.com/arendst/Tasmota/pull/7309
The moisture is given as a percentage, but the percentage is rounded to an integer.  In many situations (e.g. if the analogue range is wet -> dry 720 -> 200), a mere 100 steps is loosing a lot of granularity.  As the total possible range for the analogue input is 0->1023, it would be good to have one decimal point after the value to give the full possible granularity of the range.
The issue is compounded that with a moisture sensor which has 200 for wet and 720 for dry, the effective useful range for a pot plant may be only 300 -> 400.  Without further precision this results in only 19 steps in the range where there could be 100.

**Describe the solution you'd like**  

I think about here (https://github.com/arendst/Tasmota/pull/7309/commits/c1f30da81ee71fcfb9b293997fb4d4b1dc75fe7f#diff-346bd9707370e3b0529b922b797ef43bR122) it would be necessary to use a float rounded to one decimal place

robin13:
@effelle Maybe you are also interested in this one? :)

effelle:
I'm on vacancy :P
Anyway, all `Adcparam` commands work with integer result, even if we use double precision as input values. I'll have a look but is not on my priority list.

joba-1:
I would definitely try to avoid floats. How about using promille instead?

GrendelWWU:
As far, as I understand this, we're talking about a moisture sensor connected to A0? So, I guess, the best practice will be using the raw data 0-1023. Or you can try this sensor via I2C: https://www.tindie.com/products/miceuz/i2c-soil-moisture-sensor/
(Yes, I know, you'll probably need to compile the stuff yourself....)

effelle:
@joba-1 yes, if I'll change it I will use a per mille. 

@GrendelWWU I prefer use the pure Analog value, but some users from time to time ask for a direct conversion/calibration similar to what we have added for analog LDR and TEMP. Normally is a better implementation for HAB. 

`Aadcparam 6` is simple and even if I understand Robin's request I'm not sure I'll add a more precise version. I'll think on it.
For now if you want a better calibration `scalex` and a `var` on a rule is still the best choice. 
What do you think @ascillato?

ascillato:
[...]

PR #7589, opened 2020-01-23
ADC Range (replaces adc Moisture)

### LINKED ISSUE #7578: Change Moisture to AdcRange
yes

Currently the ADC range feature is named as a "Moisture" sensor.  In the mean time I have used this type for
- Moisture sensor
- Air quality sensor
- Voltage sensor
- Audio volume sensor

I see that it is a very useful feature, BUT it is a bit bothersome that each of these values are displayed (and passed on to Home Assistant) as a moisture sensor.  

I think it would be more useful for all if this feature was something generic such as "ADCRange" - this would allow more people to correctly find the right type and then adjust the display appropriately in Hassio etc. rather than have to document somewhere "Use the 'Moisture sensor' - it's a nice hack to be able to display measured XYZ"

Currently everything is "Moisture" ... it works, but it's a bit annoying...

**Additional context**  
 
See original issue which resulted in the Moisture sensor: https://github.com/arendst/Tasmota/issues/7182

CC @effelle @ascillato as you were involved in the previous case too.

effelle:
## Description:
New definitions and translations for both console and webui added to update `ADC Moisture` to a more generic `ADC Range`. 
As discussed with the team leader the new formula will returns just int values.
Usage for calibration:
```
AdcParam 6, fromLow, fromHigh, toLow, toHigh
```
reset to default configuration [6,0,1023,0,100]:
```
AdcParam 6
```

**Related issue (if applicable):** fixes #7578 and solve definitely #7402 .

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -114,6 +114,7 @@
 #define D_JSON_ACTIVE_POWERUSAGE "ActivePower"
 #define D_JSON_APPARENT_POWERUSAGE "ApparentPower"
 #define D_JSON_REACTIVE_POWERUSAGE "ReactivePower"
+#define D_JSON_RANGE "Range"
 #define D_JSON_PRESSURE "Pressure"
 #define D_JSON_PRESSUREATSEALEVEL "SeaPressure"
 #define D_JSON_PRESSURE_UNIT "PressureUnit"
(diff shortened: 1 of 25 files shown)


robin13:
@effelle I have just tested and this looks great - especially as I can now get higher precision by doing e.g. `toHigh=1000`.
One issue overlooked: in the UI it still shows a `%` character which is now no longer appropriate.
screenshot-2020-01-24_14-42-21

effelle:
ahhhh, forgot to remove it.
I'll do on next pass.

robin13:
Correct - it's only a UI blemish.  Technically it's working great! :)

arendst:
I'll fix it now.

arendst:
I fixed the GUI by removing the ``%``. 

For HASS it's a different cookie as it tests for MOISTURE and there are still other sensors providing moisture values so changing this doesn't make sense right now. 

There probably has to be an entry added regarding RANGE but I'm lucky I have no idea if HASS would accept this. 

Over to @effelle Frederico.

meingraham:
> I'm lucky I have no idea if HASS would accept this

😹

effelle:
Thanks @arendst, I was doing it today.
We don't need a new PROGMEM for that Theo. 
Some time ago I added a special fallback to manage generic sensors when there is not a specific configuration without any kind of UM but with graph function enabled. 
`Range` is detected correctly

range
All the sensor created using the fallback have the generic icon but surely we can live with that. 
`Moisture` sensor stay in place for i2c Chirp (and I will test one soon).

arendst:
Thank you too!

PR #7671, opened 2020-02-05
Added functionality when in WIFI_RETRY mode, the device will reboot after a prescribed period of failed reconnection attempts.

wir3z:
## Description:

**Added functionality when in WIFI_RETRY mode, the device can reboot after a prescribed period of failed reconnection attempts by switching to WIFI_RESTART mode.**

## Checklist:

### DIFF
--- a/BUILDS.md
+++ b/BUILDS.md
@@ -128,6 +128,7 @@
 |                       |   |   |   |   |   |   |   |
 | USE_IR_REMOTE         | - | - | x | x | x | x | x |
 | USE_IR_RECEIVE        | - | - | x | x | x | x | x |
+| USE_IR_REMOTE_FULL    | - | - | - | - | - | x | - | Enable ALL protocols
 |                       |   |   |   |   |   |   |   |
 | USE_SR04              | - | - | x | x | x | - | x |
 | USE_TM1638            | - | - | - | - | x | - | - |
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -52,11 +52,13 @@ The following binary downloads have been compiled with ESP8266/Arduino library c
 
 ## Changelog
 
-### Version 8.1.0.4
+### Version 8.1.0.5
 
 - Change Lights: simplified gamma correction and 10 bits internal computation
 - Change commands ``Prefix``, ``Ssid``, ``StateText``, ``NTPServer``, and ``FriendlyName`` displaying all items
 - Change IRremoteESP8266 library updated to v2.7.2
+- Change Zigbee command prefix from ``Zigbee*`` to ``Zb*``
+- Change wifi connectivity stability (#7602)
 - Fix Sonoff Bridge, Sc, L1, iFan03 and CSE7766 serial interface to forced speed, config and disable logging
 - Fix commands ``Display`` and ``Counter`` from overruling command processing (#7322)
 - Fix ``White`` added to light status (#7142)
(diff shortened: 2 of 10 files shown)


ascillato:
Hi,

Thanks for sharing.

Why is this needed?

If you set in the Tasmota console `wificonfig` to 0, it will make Tasmota to restart if the wifi connection is lost.
and if you set `wificonfig` to 4 or 5, will make Tasmota to retry the wifi connection without restarting.

Wificonfig mode 4 and 5 were added to NOT restart Tasmota because several users setups require that. So, why we need to make wificonfig 4 and 5 be an hybrid with the mode 0? why you can't use just wificonfig 0?

wir3z:
Hi,


If you use the device to control shutters or shades, then on every restart the motor/servo's engage causing the blinds to open or close slightly on each restart.  If it's an outlet, the controlled device turns off and back on during the restart.


Most of the time, retry mode works fine.  But under test, I have encountered a state where the only fix for 20 out of 27 devices was a physical power cycle.  The hybrid mode works as a watchdog to catch this corner case, while not perpetually toggling the outputs.


Lyle



________________________________

ascillato2:
> If you use the device to control shutters or shades, then on every restart the motor/servo's engage causing the blinds to open or close slightly on each restart.  If it's an outlet, the controlled device turns off and back on during the restart.

So, as many others, you also need wificonfig 4 or 5

> But under test, I have encountered a state where the only fix for 20 out of 27 devices was a physical power cycle. 

exactly which state? I'm interested on the why because seems to be another type of issue that I never saw.

wir3z:
I'm using a SonicWALL corporate firewall with 3 SonicPOINT AP's.  The 2.4GHz band where I am is highly interfered, which is resulting in occasional hangups.  During testing on the weekend, I found if I changed the preample length from long to short on the AP's to improve throughput, the Tasmota network slowly "rotted" where devices would disconnect from the AP's, and would not reconnect.


Even after changing the preamble back, and rebooting the AP's, the devices remained disconnected.  Only a power cycle would force a reboot so they could reconnect.  I suspect that there is an issue in the Arduino WiFi SDK, and this commit is basically a bandaid to address that issue.


Lyle



________________________________

ascillato:
> Even after changing the preamble back, and rebooting the AP's, the devices remained disconnected.  Only a power cycle would force a reboot so they could reconnect.

Ok, but this deep change would make many other devices to not be able to reconnect because your SSID is being saw as a new wifi network different than the one connected. If you try that change with IP cameras, or TVs, etc, you will saw the same behaviour. That on the fly-change is the same as changing the security encryption of a network. For example you have a SSID/password on a wifi PSK network and you change to another type of encryption, but using the same SSID/password, a lot of devices will refuse (on purpose) to reconnect.

> I suspect that there is an issue in the Arduino WiFi SDK, and this commit is basically a bandaid to address that issue.

What you see is not a bug, it is a security protection of the protocol to avoid Man in the Middle attack.

ascillato2:
@wir3z 

What you have explained is very interesting, but it is not going to be added soon due to it is changing the behaviour of wificonfig 4 / 5 as expected for many users that don't want a restart.

If you need that hybrid restarting behaviour for your network, you can use the following rules instead:

```yaml
rule1 1
rule1 on wifi#disconnected do add1 1 endon on var1#state=9 do restart 1 endon on wifi#connected do var1 0 endon
```

ascillato2:
If you want, this information and the rules, can be added to the docs.

wir3z:
Thanks, and fully understand.  It would change standard behavior and was a one of the concerns I had.


Lyle



________________________________

PR #7683, opened 2020-02-06
support LYWSD03-BT-sensor via HM-10(-bridge)

Staars:
## Description:

Support for the HM-10-BLE-module. First supported sensor is the (quite new) LYWSD03. Some other Mijia-sensors will be shown and support is planned for the future.

## Checklist:
  - [x ] The code change is tested and works on core 2.6.1
  - [x ] The code change pass travis tests. **Your PR cannot be merged unless tests pass**


Staars:
The HM-10 is a 3.3v Bluetooth 4.0  module based on the TI CC2540/1-Chip. It is made by Jinan Huamao and one of a whole family of Bluetooth devices.
In reality the chance to buy a clone on the „usual“ distribution channels (BG, AE, Ebay, …) is sky-high. In order to get them to work in Tasmota it is mandatory to flash the original firmware on it, which is described elsewhere. Example: https://circuitdigest.com/microcontroller-projects/how-to-flash-the-firmware-on-cloned-hm-10-ble-module-using-arduino-uno
(Note: it is even possible to do this without soldering by directly sticking male jumper wire connectors through the plastic foil of some boards).

The communication with the ESP8266 works over serial connection using AT-commands and the driver expects the default baud rate of firmware>700 which is 115200.  
The expected firmware version is 707. Other versions may work too, but before raising an issue on GitHub a test with this version is mandatory.  


### Getting data from BT-Xiaomi-Devices:

Different vendors offer BT-solutions with different accessibilities under the MIJIA-brand. A common solution is the use of so-called „MiBeacons“ which are BLE advertisement packets with a certain data structure, which are broadcasted by the devices automatically. These packets already contain the sensor data and can be passively received by other devices. 

Most of the „older“ BT-sensor-devices use unencrypted messages, which can be read by all kinds of BLE-devices or even a NRF24L01. The big advantage is the power efficiency as no active bi-directional connection has to be established. This is therefore the preferred option, if technically possible.

With the arrival of the (cheap) LYWSD03 came the problem of encrypted data in MiBeacons, which to date has not been successfully decrypted in open source projects.
At least the device allows the use of a simple BLE connection without any encrypted authentication and the reading of the sensor data using normal subscription methods to GATT-services. This is more power hungry than the passive reading of BLE advertisements.

### Tasmota-HM10-driver

#### prerequisites:
-firmware 707 (other versions may work, but this is undefined behavior)  
-simple serial cable connection  (TODO: create new GPIO-pin-names)  
-HM-10 is set to default baud rate of 115200  
-uncomment #ifdef USE_HM10 in my_user_config.h
-select GPIO-pins HM10_RX and _TX

#### expected behavior:
1. The driver will set a few options of the HM-10
2. A discovery scan will search for known sensors (Mi Flora, MJ_HT_V1, LYWSD02, LYWSD03)
3. Only the LYWSD03-sensors will be connected at a given interval, a subscription is established for 5 seconds and temperature/humidity/battery will be read.
4. After deconnection return to point 3 after the interval.

#### command interface:  
+ hm10scan  
start new discovery scan  
+ hm10period  
set or show interval in seconds between sensor read cycles  
+ hm10baud  
set or show the speed of the serial interface of the esp8266, not of the hm10  
+ hm10at  
sends AT-commands,e.g. hm10at verr? results in AT+VERR?

Staars:
2F3E161D-8AD3-4A4D-8D81-C5448B8E17A9

LYWSD03 with sensor data
Other sensors with BT-MAC (not yet supported)

arendst:
Many thx!

markusb68:
Hello Christian,
any plans to support the MJ_HT_V1 also with the HM-10?

Tried it and see:

HM10 Firmware 	707
MJ_HT_V1 MAC Address	58:2d:34:32:e8:5b

Maybe an alternative for the buggy nRF code .

PR #7788, opened 2020-02-24
Jarolift shutter working with keeloq

he-so:
## Description:

This fork adds the [Keeloq](https://en.wikipedia.org/wiki/KeeLoq) algorithm to control rf-shutter/rollers. Especially it works like a Jarolift  TDRC-01W. You need to connect a CC1101 rf-transmitter via the SPI pins. I have tested this with a Wemos D1 R2, but also a Wemos mini should work. For wiring see [Jarolift_MQTT](https://github.com/madmartin/Jarolift_MQTT).
It is based on the Jarolift-Keeloq project found here:
http://www.bastelbudenbuben.de/2017/04/25/protokollanalyse-von-jarolift-tdef-motoren/

and here

https://github.com/madmartin/Jarolift_MQTT

Integrating the Keeloq remote with tasmota and the shutter/roller feature offers the advantage to use the homekit slider to move the shutter to a specific position, because the tasmota shutter driver  makes some sophisticated calculation regarding the timings to move the shutter to absolute positions. Previous implementations of the Jarolift Keeloq software did not offer this advantage, but were limited in sending the button codes manually by using the GUI or MQTT.

This driver only interfaces with tasmota Commands. There is no GUI component for it.
The implemented commands are:

Command | Parameter(s)
---|---
KeeloqSet |  msb, lsb, serial, counter: Sets the Keeloq master key MSB, LSB, serial and counter given either in hex notation prefixed with 0x or as decimal notation. The master key has to be retrieved from your original remote. The serial can be choosen individually. The counter should start at 0. To start using these values, you have to "learn" or "pair" the serial+counter value with your receiver by bringing your receiver to learning mode and sending the command for the learn button.
KeeloqSendButton | button: Sends a Keeloq command for the given button code. This command increments the counter and stores it persistently in the settings. It calculates the relevant encrypted value which is part of the data to be sent. Usual codes are: UP=8, STOP=4, DOWN=2, LEARN=1. i.e.: KeeloqButtonSend 2
KeeloqSendRaw | string: Send a bit string represented as a character of "0" or "1" via the rf transmitter. Used only for debugging. i.e: KeeloqSendRaw 01010101011001100

When using the USE_SHUTTER mode in Tasmota, it can be triggered by Rules to control the shutter motor by RF:

Rule1 On Power1#state=0 DO KeeloqSendButton 4 endon On Power2#state=0 DO KeeloqSendButton 4 endon on Power1#state=1 DO KeeloqSendButton 8 endon on Power2#State=1 DO KeeloqSendButton 2 endon

## Checklist:
  - [x ] The code change is tested and works on core 2.6.1
  - [x ] The code change pass travis tests. **Your PR cannot be merged unless tests pass**


Jason2866:
Needs the masterkey still to be extracted from a firmware dump?

he-so:
Yes, based on the master key, the device key will be generated when invoking the Set command.

he-so:
Two notes regarding the CLA:

- The Keeloq library is originally from here: https://github.com/franksmicro/Arduino/tree/master/libraries/Keeloq
That repo is published unter the Apache Licence 2.0 as stated here https://github.com/franksmicro/Arduino/blob/master/README
So, that is compatible with GPLv3.

- Regarding the CC1101 library: It's published under the LGPL v3 as stated here: https://github.com/panStamp/arduino_avr/blob/master/cores/panstamp/cc1101.cpp

arendst (changes requested):
Looks fine but...

As there are many AddLog_P2 calls taking up precious code space I suggest you use the special macro DEBUG_DRIVER_LOG. This functionality can be enabled during debugging by enabling define DEBUG_TASMOTA_DRIVER.

Search for DEBUG_SENSOR_LOG to see numerous examples.

Important AddLog_P2 can always exist.

arendst (changes requested):
You're doing great but...

- As time goes on you may have noticed driver id 35 is now occupied by PWM_DIMMER. So pls update to driver id 36
- Also revisit the Settings map as button_devices (F08) has been released.

Sorry, almost there.

ascillato:
The TX and RX pins are hardcoded and can be easily turned to user selection. Please check for example the GPS driver for that. Thanks.

arendst:
Agree with @ascillato Adrian. Pls revisit the GPIO selection.

he-so:
Ok, is see the point. But actually it is not possible to configure the pins, since the CC1101 driver uses hardware SPI only, and the ESP has only one set of MISO, MOSI, SCK etc. as far as I understand. SPI is used for initialisation and reading the status register for receiption (OK, receiving data is not implemented now, but might be soon).

But, maybe there is a already defined value in the arduino/ESP headers that can be reused? I will have a look.

Jason2866:
Talking about this GPIOs Tx / Rx
```
#define TX_PORT            4              // Outputport for transmission
#define RX_PORT            5              // Inputport for reception
```

he-so:
Sure, however the CC1101 has at least two operation modes. 1.: sending data received via SPI buffered in its registers, based on its internal timing and encoding settings, and 2.: sending data instaneously as received on the SPI miso pin (bit-banging) using simple amplitude-shift keying. The later one is used here, since the CC1101 does not allow the synchronized bit encoding used by Keeloq.
But, for setting up the CC1101 for 433MHz and the above mentioned mode, the SPI must be working. So one can not use another pin for the later bit-banging, but the SPI hardware pin.
However one could first initialize the CC1101 and then rewire the pin to any other GPIO, no.

TheChatty:
Hello, can I use `KeeloqSendRaw` to send arbitrary commands? In detail I want to send [this command](https://github.com/RFD-FHEM/RFFHEM/issues/906#issuecomment-731840351).

he-so:
> Hello, can I use `KeeloqSendRaw` to send arbitrary commands? In detail I want to send [this command](https://github.com/RFD-FHEM/RFFHEM/issues/906#issuecomment-731840351).

Hi, the command can send arbitrary keeloq commands.

gusman80:
hey there i made Jarolift_MQTT work fine in my home, sadly the software sometimes crashes.
Now i wanna move to tasmota with keeloq and have the following questions:
max # of controllable shutters 4? (is the limit coming from the maximum number of relays in tasmota?)
how do i change/switch the shutterX to whom i want to send the KeeloqSendButton X command?

thank you

he-so:
@gusman80 , the number of shutters is currently one, because only one serial, counter and key can be saved in the configuration in the way it’s done.

PR #7790, opened 2020-02-25
Device groups

pcdiem:
## Description:

The device groups module provides a framework to allow multiple devices to be in a group with values such as power, light color/temperature/brightness, PWM values, sensor values, etc. shared with other devices in the group. For example, with multiple light modules in a device group, the light settings can be changed on one module and the settings will automatically be changed on the other light modules. Dimmer switch modules could be in a device group with light modules and the dimmer switch could control the power, brightness and colors of all the lights in the group. Multiple dimmer switches could be in a device group to form a 3-way/4-way dimmer switch.

UDP multicasts, followed by UDP unicasts if necessary, are used to send updates to all devices so updates are fast. There is no need for an MQTT server but all the devices in a group must be on the same IP network.

To include device groups support in the build, define USE_DEVICE_GROUPS in your user_config_override. This adds 3.5K to the code size. All devices in a group must be running firmware with device group support and have device groups enabled.

## Checklist:


arendst:
Currently compilation fails when no emulation is selected.

Investigating...

pcdiem:
Thank you for your work.

pcdiem:
I wasn't sure where the doc (Device_Groups.md) should live. Probably not the root directory. Maybe the wiki?

arendst:
Yes it should be on the new wiki to be found here https://tasmota.github.io/docs/#/Home

pcdiem:
I could use some guidance on where in the wiki you'd like the device groups documentation. Also, under what category should I put the DevGroupShare command? Maybe Control or Management?

arendst:
@blakadder could you guide Paul in the right direction?

blakadder:
good question.... Off the cuff, I'd say Control.

pcdiem:
Thoughts on where to include Device_Groups.md? It documents the concept and usage. I also added a section on the API for other that want to add support to other modules.

blakadder:
just leave it in the repo and i'll integrate it in the menu once its in a stable release

PR #7797, opened 2020-02-25
change AHT1X support from initial to full

device111:
- both sensor address where supported
- dual operation support
- timing optimized
- supported commands added

## Description:

## Checklist:


arendst:
Pls revisit the code changes I made to integrate it in the tasmota i2c environment

Fix the use of global temperature and humidity

Fix the use of system provided celsius/fahrenheit

Remove unneeded warnings valid for all i2c drivers when more than one is in the compile set as is the case in tasmota-sensors.

arendst:
Pls revisit the code changes I made to integrate it in the tasmota i2c environment and supporting both global temperature / humidity and celsius/fahrenheit

tasmota/xsns_63_aht1x.ino
```diff
@@ -22,137 +22,160 @@
...
-      WSContentSend_PD(HTTP_SNS_HUM, AHT10.name, humidity);
-#endif // USE_WEBSERVER
+  for (uint8_t i = 0; i < aht1x_count; i++)
+  {
+     float tem,hum;
+     if (AHT1XRead(aht1x_sensors[i].address, tem, hum))
+     {
+       float aht_temp = ConvertTemp(tem);
+       float aht_hum = ConvertPressure(hum);
+
```
arendst: Almost there but ConvertPressure should be ConvertHumidity.

Why don't you do it during AHT1XRead as I did. That way global info is updated every 2 seconds independant of when data is presented.

device111:
uuups...sorry. Shame on me...I've don't have seen the pressure function. I'm blinded...:-)

tasmota/xsns_63_aht1x.ino
device111: ok, i rechange code complete to read the sensor every Second and only show the values if there are needed.
Thanks!

PR #7803, opened 2020-02-26
Improved INA219 driver supports custom defined shunt resistor

barbudor:
## Description:

This PR adds support for custom defined shunt resistor with INA219 sensor.
It also takes the opportunity to refactor code and improve efficiency.
It was planned after deciding to help Discord user @Massivenrg who needed a shunt resistor of 20 milli-ohm for measuring currents up to 16A.

Original code was limited to support the standard 0.1 ohm shunt resistor which is mounted on the INA219 breakout module available from Adafruit or Chinese providers. This provides allows a single current measurement range up to 3.2A. However some use case needs a different current range which can only be achieved using lower of higher shunt resistor.
This PR allows specifying a custom shunt resistor value ranging from 1 milli-ohm (320A range) up more than 10 kilo-ohm (sub µA).
Compliance with legacy `Sensor13` modes 0, 1 and 2 is preserved and mode values of 10 and above are used to specify the customer shunt resistor value.
Original code was also making an unnecessary complex use of the so-called "Calibration register". This PR uses a more efficient method to calculate the current by using Ohms' law and the shunt-voltage.
Some INA219 register constants have been fixed and completed. This allows the PR to configure the INA219 to average 16 samples for both Bus Voltage and Shunt Voltage to increase accuracy of the measurement. Because this is performed in hardware and still takes much less time (<20ms) than Tasmota sensor period (1 sec) it has no impact on the performance.

**Specifying custom shunt resistor**

In order to be able to specify a large range of shunt resistor values without breaking settings compatibility (`Settings.ina219_mode` is `uint8_t`) it has been chosen to encode the resistor value as a decimal value `RRM = [10.255]` which represent:
`Shunt_Resistor_in_milliohm = RR * 10^M`
This means that the following commands define the following shunt resistors:
`Sensor13 10` define Shunt_Resistor as 1 * 10^0 = 1 milli-ohm for a maximum current of 320A
`Sensor13 11` define Shunt_Resistor as 1 * 10^1 = 10 milli-ohm for a maximum current of 32A
`Sensor13 21` define Shunt_Resistor as 2 * 10^1 = 20 milli-ohm for a maximum current of 16A
`Sensor13 12` define Shunt_Resistor as 2 * 10^2 = 100 milli-ohm for a maximum current of 3.2A (same as default)
`Sensor13 13` define Shunt_Resistor as 1 * 10^3 = 1 ohm for a maximum current of 0.32A
This may seem clunky but allows large range of values while not impacting settings storage.

**Code refactoring and removal of calibration**

Usage of calibration register seems to have been misunderstood as this code can be found in many derivative including ports to MicroPython and CircuitPython. On those platforms I already did my own driver based on the same principle.
The calibration register on INA219 is not intended to provide an effective calibration as we are expecting it (ie providing a means to finely adjust a measurement mechanism to provide accurate results).
It is intended to drive the hardware-based calculation of current and power so that a MCU with low capabilityy can avoid doing multiplications and divisions and retrieve current and power from dedicated registers. While this seems a good idea, in this case it adds complexity to the code to adjust the full-scale of the current to avoid overflow in power calculation.
As Tasmota already calculates power by performing floating point multiplication of the voltage and current, as well as some other floating points multiplication and divisions, refactoring the code in order to calculate directly the current from the shunt voltage makes the code smaller and more efficient.
While this may seems independent of the goal of this PR, the refactoring is key to implementing it efficiently.

**PR Impact on code and data**

`.irom0.text     0x8b628	=>	0x8b5d8		(-80)`
`.data           0x764	=>	  0x764		(==)`
`.bss            0xabb0	=>	 0xaba8		(-8)`

Note: As this is my 1st PR for Tasmota and I believe some changes may be requested, I left some debug code (currently disabled). I will remove it for the final submission.

## Checklist:

### DIFF
--- a/tasmota/xsns_13_ina219.ino
+++ b/tasmota/xsns_13_ina219.ino
@@ -97,10 +97,11 @@
 uint8_t ina219_type[4] = {0,0,0,0};
 uint8_t ina219_addresses[] = { INA219_ADDRESS1, INA219_ADDRESS2, INA219_ADDRESS3, INA219_ADDRESS4 };
 
-#define __DEBUG__(a)
-
-__DEBUG__(char __dbg1[20];)
-__DEBUG__(char __dbg2[20];)
+#ifdef DEBUG_TASMOTA_SENSOR
+// temporary strings for floating point in debug messages
+char __ina219_dbg1[10];
+char __ina219_dbg2[10];
+#endif
 
 // The following multiplier is used to convert shunt voltage (in mV) to current (in A)
 // Current_A = ShuntVoltage_mV / ShuntResistor_milliOhms = ShuntVoltage_mV * ina219_current_multiplier
@@ -132,13 +133,15 @@ bool Ina219SetCalibration(uint8_t mode, uint16_t addr)
 {
   uint16_t config = 0;
 
-  __DEBUG__(printf("mode = %d\n",mode);)
+  DEBUG_SENSOR_LOG("Ina219SetCalibration: mode=%d",mode);
   if (mode < 5)
   {
     // All legacy modes 0..2 are handled the same and consider default 0.1 shunt resistor
     ina219_current_multiplier = 1.0 / INA219_DEFAULT_SHUNT_RESISTOR_MILLIOHMS;
-    __DEBUG__(dtostrfd(ina219_current_multiplier,5,__dbg1);)
-    __DEBUG__(printf("cur_mul=%s\n",__dbg1);)
+    #ifdef DEBUG_TASMOTA_SENSOR
+    dtostrfd(ina219_current_multiplier,5,__ina219_dbg1);
+    DEBUG_SENSOR_LOG("Ina219SetCalibration: cur_mul=%s",__ina219_dbg1);
+    #endif
   }
   else if (mode >= 10)
   {
@@ -147,8 +150,10 @@ bool Ina219SetCalibration(uint8_t mode, uint16_t addr)
     for ( ; mult > 0 ; mult-- )
       shunt_milliOhms *= 10;
     ina219_current_multiplier = 1.0 / shunt_milliOhms;
-    __DEBUG__(dtostrfd(ina219_current_multiplier,5,__dbg1);)
-    __DEBUG__(printf("shunt = %dmO => cur_mul=%s\n",shunt_milliOhms,__dbg1);)
+    #ifdef DEBUG_TASMOTA_SENSOR
+    dtostrfd(ina219_current_multiplier,5,__ina219_dbg1);
+    DEBUG_SENSOR_LOG("Ina219SetCalibration: shunt=%dmO => cur_mul=%s",shunt_milliOhms,__ina219_dbg1);
+    #endif
   }
   config = INA219_CONFIG_BVOLTAGERANGE_32V
          | INA219_CONFIG_GAIN_8_320MV               // Use max scale
@@ -163,7 +168,7 @@ float Ina219GetShuntVoltage_mV(uint16_t addr)
 {
   // raw shunt voltage (16-bit signed integer, so +-32767)
   int16_t value = I2cReadS16(addr, INA219_REG_SHUNTVOLTAGE);
-  __DEBUG__(printf("ShR = 0x%04X\n",value);)
+  DEBUG_SENSOR_LOG("Ina219GetShuntVoltage_mV: ShReg = 0x%04X",value);
   // convert to shunt voltage in mV (so +-327mV) (LSB=10µV=0.01mV)
   return value * 0.01;
 }
@@ -172,12 +177,12 @@ float Ina219GetBusVoltage_V(uint16_t addr)
 {
   // Shift 3 to the right to drop CNVR and OVF as unsigned
   uint16_t value = I2cRead16(addr, INA219_REG_BUSVOLTAGE) >> 3;
-  __DEBUG__(printf("BusR = 0x%04X\n",value);)
+  DEBUG_SENSOR_LOG("Ina219GetBusVoltage_V: BusReg = 0x%04X",value);
   // and multiply by LSB raw bus voltage to return bus voltage in volts (LSB=4mV=0.004V)
   return value * 0.004;
 }
 
-/*
+/* Not used any more
 float Ina219GetCurrent_mA(uint16_t addr)
 {
   // Sometimes a sharp load will reset the INA219, which will reset the cal register,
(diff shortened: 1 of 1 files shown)


barbudor:
I would like to invite @mike2nl to comment as we had a brief discussion on that addition.

It could also be interesting to support a different shunt resistor per INA219 device but I don't see how to make it without impacting `Settings`

tasmota/xsns_13_ina219.ino
```diff
@@ -85,67 +92,92 @@
...
 
+#define INA219_DEFAULT_SHUNT_RESISTOR_MILLIOHMS (100.0) // 0.1 Ohm
+
 uint8_t ina219_type[4] = {0,0,0,0};
 uint8_t ina219_addresses[] = { INA219_ADDRESS1, INA219_ADDRESS2, INA219_ADDRESS3, INA219_ADDRESS4 };
 
-uint32_t ina219_cal_value = 0;
-// The following multiplier is used to convert raw current values to mA, taking into account the current config settings
-uint32_t ina219_current_divider_ma = 0;
+#define __DEBUG__(a)
```
effelle: I see you used a lot of debug point, is there a reason to add  `__DEBUG__(a)` instead of using the internal logger?
``` json
AddLog_P2(LOG_LEVEL_DEBUG, PSTR("any text here: %s or val here %d"), textstr, val);
```

tasmota/xsns_13_ina219.ino
```diff
@@ -175,15 +218,11 @@ bool Ina219Read(void)
...
- *
- * 0 - Max 32V 2A range
- * 1 - Max 32V 1A range
- * 2 - Max 16V 0.4A range
 \*********************************************************************************************/
 
 bool Ina219CommandSensor(void)
 {
-  if ((XdrvMailbox.payload >= 0) && (XdrvMailbox.payload <= 2)) {
+  if ((XdrvMailbox.payload >= 0) && (XdrvMailbox.payload <= 255)) {
```
effelle: Is the `255` payload intended for retrieving data (like for `gpio 255` command )? 
If yes, why restart the device?

tasmota/xsns_13_ina219.ino
effelle: I see you used a lot of debug points... Is there a reason to add  `__DEBUG__(a)` instead of using the internal logger?
``` json
AddLog_P2(LOG_LEVEL_DEBUG, PSTR("any text here: %s or val here %d"), textstr, val);
```

arendst: Or even better use the MACRO DEBUG_SENSOR_LOG as defined in tasmota_post.h

It can be controlled by a global define DEBUG_TASMOTA_SENSOR saving a lot of code memory

tasmota/xsns_13_ina219.ino
barbudor: Hi @effelle,
No. In this case 255 is a considered a valid payload which declare a resistor of 25 * 10^5 = 2.5 Mohm. As explained in the description above, shunt resistor value is encoded by a decimal number in the form RRM (from 10 to 255) which means a value of MM * 10^M
This allows a large range of values while avoiding to change `Settings.ina219_mode` from a `uint8_t` to a large integer. 

The restart is here from the original code and I didn't change it because I don't know what would be the impact on Tasmota. On the INA219 driver itself, there is no reason to restart.
May be this is to force storage of the value in the flash ?
Any advice will be appreciated

tasmota/xsns_13_ina219.ino
barbudor: @effelle and @arendst 
I apologies for not searching further what would be the best solution for debug log. printf() seems to be quick and efficient for this small update.
Definitively these logs were useful for a short time and I will follow Theo advice using the macro. I'll keep reference to AddLog_P2() for future development.
I will submit a PR update by the end of the week-end.

tasmota/xsns_13_ina219.ino
effelle: If 255 is a payload to define the value ok, a restart is needed to update the value in flash.

effelle (commented):
Nice job, the code is also more readable now.
I have one ina incoming for the next week, I will do some tests for sure.

barbudor:
One feature that could have been interesting is the ability to use specific shunt resistor per INA219 for someone who need to measure different current ranges. But there is no way of doing it without expanding `Settings.ino21ç`
Let's see if this is needed. @TrDA-hab is planning a board with 3x INA219.

effelle:
One step at the time.
From my point of view have a  specific shunt doesn't seems a much requested feature, perhaps I'll change my mind when I start to play with it.

barbudor:
Thanks

mike2nl:
@effelle 
@barbudor and me we were talking about different shunt values because different currents. To get more out of the INA219. I was thinking about a high current switch or2x change type. But for that the shunt resistor has to be changed in runtime. That goes fra, far at it is now. But we won't forget it, haha.

djpankdo:
By convention, most shunts are designed to drop 50 mV, 75 mV or 100 mV when operating at their full rated current and most ammeters consist of a shunt and a voltmeter with full-scale deflections of 50, 75, or 100 mV.
If we consider the PGA value equal 2 we can go up to 80 mv making the perfect match with the 75mv commercial shunt.
the comercial shunt capacity (max rated current granted by manufacturer before heat dissipation became a problem) is usually 10A, 20A, 50A, 100A, 150A, 200A all with max 75mv drop.

The formula RMM if fantastic (Shunt_Resistor_in_milliohm = RR * 10^M) but for some correncial shunt resistor it became a problem like 10A(75mV) will result in 7.5m Ohm; 20A(75mV) will result in 3.75m Ohm; 50A(75mV) will result in 1.5m Ohm, etc..
the point is the market key voltage value for shunt is 75mv.

If you check the datasheet (ti.com/lit/ds/symlink/ina219.pdf?&ts=1590022205348) item 8.5.1 is possible to calculate the Calibration Register(05h) with more versatile way.

my suggestion is a way to insert the value directly in the register pointer (05h) like 45E7 for a commercial shunt 50A(75mv).

barbudor:
Hi @djpankdo 
Let me explain why your suggestion will not bring any value would even reduce the accuracy of the measure.
In the INA 219, there are 2 parts :
- The 2 ADCs which measures the bus voltage and the shunt voltage
- The current & power calculation unit

The ADCs have a fixed, given, unchangeable accuracy. This is true for both.
Regarding the shunt voltage ADC, it is extremely clear when you look at figures 20 to 23 : whatever is the gain, the LSB is always 10µV. Which means that while changing the gain reduce the full scale (from 320mV down to 40mV), this doesn't actually improve the measurement. On the opposite, you are reducing the number of meaningful bits (from 16 bits down to 12 bits).
Same for the bus voltage (section 8.6.3.2), whether you are selecting 32V or 16V, the LSB is always 4mV.

So what is all about PGA and calibration register ?
This is about the 2nd part of the device : the current and power calculation.
This part is intended to provide low end MCU, typically 8 bits with no support for large integer and floating point, a simple mean of conversion of shunt-voltage to current as well as computing power. This would avoid for such low end MCU a lot of painful calculations.
Because this section of the device work with limited number of bits, the goal of PGA and calibration is to scale bus voltage and shunt voltage in an acceptable range for the conversion and multiplication to remain in the chip capacity.

Considering that we have a 32 Bit MCU with support for floating point (limited in Tasmota to single precision which still provide 24 bits of mantissa), the 2nd part of the chip is useless just bring confusion to users.
As you can check in the driver's code, since my PR, current is computed directly from the ADC shunt voltage register with a single floating point divide **without a loss in accuracy**. 
Simple, efficient and accurate.

So if your point is essentially to be able to use resistors values (7.5MOhms, 3.75mOhms) that do not fit with the RR * 10^M formula, understand that I came to this because the size of flash setting assigned to Sensor13 is a single byte. The change was meant not to break existing settings which is always challenge with Tasmota.

I would be happy to work on a change to support a larger range of shunt values but we would need to negotiate with Theo to get assigned a larger flash setting (still a challenge for upgrade compatibility). This would surely be more efficient than any tweaking of the calibration register.
Beside, You can still use Tasmota as-is with standard resistors close to the values you are looking for : 10 mOhms (32A range) , 5 mOhms (64A range) or 2 mOhms (128A range) that are also current available resistors in the E12 series.

Let me know your feedback.

djpankdo:
Mr. @barbudor,
First thanks for your attention.
I read the datasheet 3 times to be able to understand and you are completely right.
The calibration reg is there for nothing! Why the chip developer add this functionality if the ADC LSB is always the same? 
So my suggestion  is to keep your method but changing the formula dividing RR by 4.
Shunt_Resistor_in_milliohm = (RR/4) * 10^M

You can keep the same Sensor13 value range but now the Shunt can be incremented in .25 steps.
it will change the max value from 24M ohms to 6M and the minimum value from 1m to 0.25m ohms.

the current range will be from 53.3nA to 1280A much more practical values.

what do you think about it?

barbudor:
Hi @djpankdo 

As I explained, this calculation stuff is intended for small MCU that do not have the capacity to make calculations themselves.

Now, regarding your suggestion, it raise a problem because it changes the way `Sensor13` behaves which means that it will create problems to someone that would upgrade from a previous version to such version. I need to think a little bit if a compatible way can be used.

But why do you absolutely want to use those resistors values ? Why using 3.75 mOhms when 4 or 5 mOhms does the job ?
Many values are available that are close to what you are looking for https://eu.mouser.com/Passive-Components/Resistors/Current-Sense-Resistors/_/N-7fjcf

oriolism:
Hi @barbudor, 
Is possible to increase the max current to 500A?

barbudor:
Hi @oriolism
This implementation would not support more than 320A with a 1milliohm resistor. 
However I would not be very confident to use such technical solution (INA219) for anything above 50A.
Inductive sensors are probably safer.
Best regards

PR #7822, opened 2020-02-29
refactor TX2x sensor driver

curzon01:
## Description:

enhanced statistic calculation can be disabled by defining USE_TX2X_WIND_SENSOR_NOSTATISTICS

Existing callback FUNC_AFTER_TELEPERIOD now also called for sensors 'cause 
callback FUNC_JSON_APPEND is useless to distinguish whether sensor data was sent via the SENSOR topic.

## Checklist:


kugelkopf123:
Isn't it usually the case with tasmota that functions that are added are defined instead of defining things that should be left out?
Therefore I would suggest deactivating statistics by default and activating them via define.

Just asking...

curzon01:
you are right in principle and I originally did it with reverse logic, but
1. this sensor is optional at all and not included in any default package
2. for wind sensor it's benefical to enable the additonal statistic funcitonality as default because the data changes too quickly to allow the calculation to be carried out by an external system - so default enable makes sense.

Consider USE_TX2X_WIND_SENSOR_NOSTATISTICS simply as a simple way to reduce the sensor to the bare minimum (for whatever reason). I didn't want to define this externally (with reverse logic) - or consider it a debugging aid

PR #7850, opened 2020-03-03
Replace strstr() with strchr() when appropriate.

rudo-thomas:
When appropriate, i.e. when the string we're looking for is 1 character long.

## Description:

Tiny and trivial optimisation. No need for strstr() when all we're looking for is a single character.

## Checklist:
  - [~] The code change is tested and works on core 2.6.1
    - I built it on 2.6.3, tested the IR (non full) module.

### DIFF
--- a/tasmota/support_command.ino
+++ b/tasmota/support_command.ino
@@ -1021,7 +1021,7 @@ void CmndTemplate(void)
   // {"NAME":"Generic","GPIO":[17,254,29,254,7,254,254,254,138,254,139,254,254],"FLAG":1,"BASE":255}
   bool error = false;
 
-  if (strstr(XdrvMailbox.data, "{") == nullptr) {  // If no JSON it must be parameter
+  if (strchr(XdrvMailbox.data, '{') == nullptr) {  // If no JSON it must be parameter
     if ((XdrvMailbox.payload > 0) && (XdrvMailbox.payload <= MAXMODULE)) {
       XdrvMailbox.payload--;
       if (ValidTemplateModule(XdrvMailbox.payload)) {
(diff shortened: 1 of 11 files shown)


eku:
@rudo-thomas  does it increase or decrease code size? How much?

rudo-thomas:
@eku I tried building tasmota, -minimal, -ir, -sensors using platformio and the reported used flash size was unchanged in all 4 cases.

I ran `platformio run -e X` for the 4 values of X, then checked out the other branch, removed .pio* and .cache and re-ran. I'll be happy to do something more rigorous - please let me know how. Thanks.

eku:
@rudo-thomas so what's the benefit?

arendst:
So no benefit, no merge. Thx for the exercise

rudo-thomas:
This should save some precious CPU cycles, though I admit the benefit might not be obvious, so I'm working on some measurements now.

I'll get some numbers later today.

rudo-thomas:
According to my measurements, strchr() is faster in all cases except when the haystack is an empty string. As expected, the savings depend on how far the searched-for character is within the haystack; or on how long the haystack is when the searched-for character is nowhere in the haystack. In the empty haystack case strchr() is slower by a mere 2 cpu cycles.

For the code, data, versions used see https://gist.github.com/rudo-thomas/8e6a973af4ff62e7ac18ca3428d0c720

sfromis:
What's precious on Tasmota is not so much CPU cycles (which seldom is a limitation), but bytes used (which always is a limitation).

Jason2866:
I compared the binary files. The dont differ. Please close this PR since it has zereo effect.
The only difference is the compile time stored in the binary
image

rudo-thomas:
@Jason2866 you're right. I should have looked at the assembly myself, my bad. Sorry.

It turns out gcc treats strstr() and other string functions as built-in and optimizes a strstr() call with literal one-char string into a strchr() call. Which is exactly what's happening here, and also what my measurement *did not* measure.

This PR does indeed nothing. Closing.

PR #7868, opened 2020-03-07
Optimized function 'mp'

maxela:
Leaner syntax of function `mp` and simplified implementation.
New documentation:

Mapping function `mp`

It addresses a standard task with less code and much flexibility: Mapping of an arbitrary incoming numeric value into a defined schema of allowed/reasonable ranges.

The numeric value `x` passed as the first parameter is compared to the mprules in the order they are provided as subsequent parameters. If the value matches the criteria, the defined value is returned. Subsequent mprules are skipped. If `x` matches none of the mprules, `x` is returned unchanged. 
```
mp(x mprule1 mprule2 ... mprule<n>)
```

An mprule starts with one of the comparison operators `<`, `>` or `=`, followed by a numeric value `v1`, optionally followed by a second numeric value `v2`:
```
<|>|=v1[ v2]
```

Each of the numeric values `x`, `v1` and `v2` can be a literal, an expression or a variable.

Example 1: `<8 0` - this mprule reads: If x is less than 8, return 0.
Example 2: `>100` - this mprule reads: If x is greater than 100, return 100. 

Example 3:
```
y=mp(x <8 0 >100)
```
Assigns 0 to y if x is less than 8.
Assigns 100 to y if x is greater than 100.
Assigns x to y for all values of x that do not meet the above criteria (8 to 100).

The above example 3 does the same as the following code - with just one line insead of 7 and 20 characters less:
```
y=x
if x<8 {
y=0
}
if x>100 {
y=100
}
```

## Description:

## Checklist:

### DIFF
--- a/tasmota/xdrv_10_scripter.ino
+++ b/tasmota/xdrv_10_scripter.ino
@@ -1512,7 +1512,7 @@ chknext:
                 (*opp=='>' && fvar1>fvar2) || 
                 (*opp=='=' && fvar1==fvar2)) 
             {
-              if (*lp!='<' && *lp!='>' && *lp!='=') {
+              if (*lp!='<' && *lp!='>' && *lp!='=' && *lp!=')' && *lp!=SCRIPT_EOL) {
                 float fvar3;
                 lp=GetNumericResult(lp,OPER_EQU,&fvar3,0);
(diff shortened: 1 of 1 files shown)


maxela (commented):
Fix works and is necessary for function 'mp' to work properly.

Issue #7848, opened 2020-03-03
Wrong units KWh -> kWh


McGiverGim:
### PROBLEM DESCRIPTION
When using MQTT discovery, the units in Home Assistant appear as KWh (first uppercase) and not kWh (first lowercase).

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [X] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [X] Searched the problem in the [docs](https://tasmota.github.io/docs/#/help/FAQ)
- [X] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [X] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [X] Device used (e.g., Sonoff Basic): Avatto
- [X] Tasmota binary firmware version number used: 8.1.0.10(a837a8a-tasmota)
  - [X] Pre-compiled
  - [ ] Self-compiled
    - [ ] IDE / Compiler used: _____
- [X] Flashing tools used: Tuya Convert
- [X] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```
08:44:04 MQT: enchufeavatto/stat/RESULT = {"NAME":"Avatto OT06","GPIO":[17,0,0,0,134,132,0,0,131,56,21,0,0],"FLAG":0,"BASE":49}
08:44:04 MQT: enchufeavatto/stat/RESULT = {"Module":{"0":"Avatto OT06"}}
[...]

Issue #7837, opened 2020-03-01
graphs not grouped properly in home assistant (missing units ?)


ptempier:
**Have you looked for this feature in other issues and in the docs?**
yes

**Is your feature request related to a problem? Please describe.**  
Graphs are mixed up in home assistant

**Describe the solution you'd like**  
These metrics are all appearing in the same graph, my understanding is there's no unit for theses metrics when the data is sent to HA (because HA groups graphs with the same unit together).

Please add units for the following metrics
Apparent power (should be VA ?)
Factor (should be cos phi ?)
Reactive power (should be VAR ?)
Total start time (should be h ?)
Status (should be % ?)

**Describe the solution you'd like**
Have the graphs grouping properly in home assistant, grouping by unit as done by HA while not perfect, is a very good start that works in a very generic way and give graphs some visibility.

**Describe alternatives you've considered**
Make and group all the graphs manually, or use grafana.

**Additional context**  
It s rather annoying because the units are différent, the graphs arent scalled properly and a lot of graphs are grouped together.

(Please, remember to close the issue when the problem has been addressed)

 
**(Please, remember to close the issue when the problem has been addressed)**

effelle:
What Tasmota version are you using?

ptempier:
Program Version	8.1.0(tasmota)
Build Date & Time	2019-12-25T12:33:25
Core/SDK Version	2_6_1/2.2.2-dev(38a443e)
Uptime	0T17:13:06
Flash write Count	144 at 0xF5000
Boot Count	37
Restart Reason	Software/System restart


Emulation	None
mDNS Discovery	Disabled
 
ESP Chip Id	12253746
Flash Chip Id	0x144051
Flash Size	1024kB
Program Flash Size	1024kB
Program Size	566kB
Free Program Space	436kB
Free Memory	27kB

This runs on the blitzwolf SHP6 high tension one, with the corresponding template

I trust that there are issues with discovery , i run HA on my computer because i didnt decided on a server/nas yet, so i restart it everyday. When i do this , i need to restart tasmota devices aferward

But i saw there were tickets for this kind of stuff so i didnt open one.

effelle:
Dev version starting from 8.1.0.6 (IIRC) got a lot of new sensors with additional UM. 
8.1.0 master is missing a lot of new stuff for HA discovery. Some UM will be changed soon. 
RSSI has not a `%` yet because in truth is not a % value(but yes, perhaps I'll add it). 
[...]

PR #7901, opened 2020-03-12
New HAss Discovery (3/4): INPUTS

### LINKED ISSUE #7599: MQTT Discovery: Switches & Multi-Press/Toggle
**Have you looked for this feature in other issues and in the docs?**  Yes. I have tried countless configurations and cannot get the functionality as I would like it.

I am trying to use a physical latching toggle switch on Tasmota as a input for HA to trigger events without triggering the local relay. Using Switch2 prevents the relay from triggering but the switch is not published to HA via discovery and HA does not know about it. Using Button2 solves the discovery but the button config does not seem to work well with latching switches.

Include switches (without relays) in the MQTT auto discovery as "read only" on/off sensors that cannot be controlled via MQTT. Toggeling the switch pushes the same events to HA as a switch with a relay. 

I've tried using Button2 but buttons seem to be hard wired for momentary contact and not latching switches.
I've setup manual binary sensors in HA but this is cumbersome and prone to problems when everything else is using auto discovery.

**Additional context**  

effelle:
## Description:
Add support for `triggers` under Home Assistant for automation control. Available starting from version 0.106.6. 

**Tasmota** is the first firmware using those new triggers.

More information about Triggers [here](https://www.home-assistant.io/docs/automation/trigger/) and [there](https://github.com/home-assistant/home-assistant/pull/31679).

What's new: 

**1. Buttons:**
  -  To have them available under Home Assistant `ButtonTopic` must be set to `1` or a custom name;
  - `SetOption1`, `SetOption11` and `SetOption13` are now supported;
  -  When used with `ButtonTopic 1`, a Button can only generate a `HOLD` trigger.

 **2. Switches:**
- To have them available under Home Assistant `SwitchTopic` must be set to a custom name;
- All `SwitchMode` type are available but 11 and 12 will generate just a `TOGGLE` trigger;
- switches with a known state (ON/OFF) will be created as `Binary_Sensor`;
- Switches with unknown state (TOGGLE) will be created as `Trigger`;
- `HOLD` and double click are now supported;
- Some `SwitchMode` can have a `binary_sensor` and a `trigger` at the same time (ex Switchmode 5).

Also:
**3.  New environmental sensors added:** `SGP30`, `CCS811` and `SCD30`.
**4. Updated some units of measure.**
**5. Code optimization.**

A special note: Thanks to @emontnemery for help in understanding how to add the new triggers and also to @arendst for the quick implementation of new functions necessary to optimize the code. 

**Related issue (if applicable):** fixes #7599 , #7837 and  #7848

## Checklist:

### DIFF
--- a/tasmota/xdrv_12_home_assistant.ino
+++ b/tasmota/xdrv_12_home_assistant.ino
@@ -26,96 +26,104 @@ const char kHAssJsonSensorTypes[] PROGMEM =
   D_JSON_TEMPERATURE "|" D_JSON_PRESSURE "|" D_JSON_PRESSUREATSEALEVEL "|"
   D_JSON_APPARENT_POWERUSAGE "|Battery|" D_JSON_CURRENT "|" D_JSON_DISTANCE "|" D_JSON_FREQUENCY "|" D_JSON_HUMIDITY "|" D_JSON_ILLUMINANCE "|"
   D_JSON_MOISTURE "|PB0.3|PB0.5|PB1|PB2.5|PB5|PB10|PM1|PM2.5|PM10|" D_JSON_POWERFACTOR "|" D_JSON_POWERUSAGE "|"
-  D_JSON_REACTIVE_POWERUSAGE "|" D_JSON_TODAY "|" D_JSON_TOTAL "|" D_JSON_VOLTAGE "|" D_JSON_WEIGHT "|" D_JSON_YESTERDAY;
+  D_JSON_REACTIVE_POWERUSAGE "|" D_JSON_TODAY "|" D_JSON_TOTAL "|" D_JSON_VOLTAGE "|" D_JSON_WEIGHT "|" D_JSON_YESTERDAY
+  D_JSON_CO2 "|" D_JSON_ECO2 "|" D_JSON_TVOC;
 const char kHAssJsonSensorUnits[] PROGMEM =
   "|||"
-  "W|%|A|Cm|Hz|%|LX|"
-  "%|ppd|ppd|ppd|ppd|ppd|ppd|µg/m³|µg/m³|µg/m³||W|"
-  "W|KWh|KWh|V|Kg|KWh";
+  "VA|%|A|Cm|Hz|%|LX|"
+  "%|ppd|ppd|ppd|ppd|ppd|ppd|µg/m³|µg/m³|µg/m³|Cos φ|W|"
+  "VAr|kWh|kWh|V|Kg|kWh|"
+  "ppm|ppm|ppb|";
 const char kHAssJsonSensorDevCla[] PROGMEM =
   "dev_cla\":\"temperature|dev_cla\":\"pressure|dev_cla\":\"pressure|"
   "dev_cla\":\"power|dev_cla\":\"battery|ic\":\"mdi:alpha-a-circle-outline|ic\":\"mdi:leak|ic\":\"mdi:current-ac|dev_cla\":\"humidity|dev_cla\":\"illuminance|"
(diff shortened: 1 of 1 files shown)


septillion-git:
@effelle What's the reason to force a switchtopic to have them discovered in HA? That topic isn't used. Because this change (from 8.1.0, which was not on the change list) made me search for quite some time... 😢

effelle:
@septillion-git if you look on `switchtopic` on our [buttons and switches](https://tasmota.github.io/docs/Buttons-and-Switches/) documentation you will find that:

 **SwitchTopic `<value>` sends an MQTT message command to the custom topic. This does not change the state of the devices power.** 

meaning the switches will be independent. Is not a new feature in Tasmota, on the contrary was and still is the official way to decouple switches from power devices (relays) from a long time, so definitely nothing new.  Having the switches/buttons created in Home Assistant without decoupling them was a bug, a not intended feature, and the change was decided in conjunction with HA devs as a preliminary support for the new mqtt triggers now completely implemented.  
Until I'll change the switches (no ETA, sorry) into something like i did for the buttons, now controlled with `SetOption73`, they will stay that way. 
If you don't want to set `switchtopic` to a custom name, remembering that is a setting born for this type of configuration, you only solution is to manually create the entities under Home Assistant.
More information about how the new Discovery works can be found on our [Home Assistant](https://tasmota.github.io/docs/Home-Assistant/) documentation.  

About the change list, perhaps you looked in the wrong version?
Since this PR was merged on dev version 8.1.0.11 it was correctly reported on the changelog for version 8.2.0 Elliot. You will not find this PR on 8.3.0 changelog for obvious reasons. 

If you need more assistance please address your requests to our  [Tasmota Support Discord Chat](https://discord.gg/Ks2Kzd4), thanks!

septillion-git:
@effelle thank you very much for this lengthy explanation! And when reading the next piece, don't get me wrong! I really appreciate all the hard work. But this is just my view as a user.

I didn't know it used to be a "bug". I run Tasmota on them since 7.something and I'm sure the documentation was in line back then. Aka, all the time the switches where visible in HA. And I simply decoupled them by making a rule for them as was explained in the documentation. And I toggled that rule based on MQTT#(dis)connect in order to regain local action as fall back.

So for me as a user is all looked like it was meant to be. For me as a user documentation and function where in line with switches discovered. #7901 was indeed mentioned in the change list of 8.2.0 but I would have never ever guessed (without reading the whole PR) that it would change the discovery of switches in HA. Aka, for me it came as a huge breaking change.

And you say " remembering that is a setting born for this type of configuration". But I don't get why. You now make a new topic to post the state to but that topic isn't used by HA discovery (luckily!) __yet__ that is the toggle to make switches discoverable. To me, switchtopic sounds like a way to directly control another Tasmota(/device) over MQTT and I don't see a link to HA. (Although losing rules when a switchtopic is set is weird from a user perspective.)

And I say luckily because the way I now fixed it is to set the same switchtopic on all Tasmota's and just let it publish there without anyone ever reading that topic. Bit "wasteful" but at least all Tasmota's can use the same topic. And I use MQTT#(dis)connect to set or reset that switchtopic in order to get local control as fall back support. 

And yeah, I would like to help more with the code itself but the whole project is pretty daunting! Getting an overview of where to find what and how things are dealt with (storing of settings, hooks etc) is kind of impossible. At least for a hobbyist like me. A page with general info like that would be great. But again, not a negative comment. I know it's one of the least exiting things to make and can be pretty hard and time-consuming. So from time to time I might take a new dive into the code.

effelle:
No problem Timo, I like to have feedbacks. 
I know in the old version switches/buttons were automatically created, in fact I'm the one who enabled (and wrote the docs for) them before starting to work on triggers and new inputs to solve an issue and then they stayed that way for a long time due to the lack of time from my side.

I understand your point of view about the Discovery and as I said the good news is I want to change switches to something similar of what I've done with buttons, mostly to re-add the ability to use rules in conjunction. The bad news is there is no ETA, because it involves a change on base code.

PR #7924, opened 2020-03-15
Zigbee move all strings to PMEM

s-hadinger:
## Description:

Reduces memory footpring of ZIgbee by 5.5KB.
I thought initially that string literals inside PROGMEM structures were automatically in PROGMEM too. But it's not the case.

This PR explicitly declares every string literal as PROGMEM, saving 5.5KB of previous RAM. Unfortunately this makes the code less readable.
The macros `ZF()` declares a string literal in PROGMEM, the macro `Z()` uses it in a struct.

## Checklist:


arendst:
Off the record, is there a way to compile Zigbee WITHOUT Hue Emulation? Currently I get all kinds of compiler errors in xdrv_23_zigbee_3_hue.ino when I disable Hue emulation.

I never used any emulation so this came as a surprise.

arendst:
You may be able to save some more code space by placing the text strings in a PROGMEM constant separated by "|" like the commands PROGMEM string and using an indexed uint8_t Z_status table to be used with function  GetTextIndexed.

A binary dumps now shows the Z_Status table uses 8 bytes per entry as the text string flash address is used instead of an index in the text string.

image

s-hadinger:
Good hint Theo, I will indeed refactor using a more compact structure.

s-hadinger:
@arendst Maybe you could also declare all modules names from `kModules` in `tasmota_template.h` to PROGMEM strings. It may also free up some hundreds of bytes of memory.

arendst:
Thx. I'll experiment.

PR #7930, opened 2020-03-16
Update AHT10 Driver

device111:
- Add support for both adresses 0x38, 0x39
- Add Reading without Delay
- Add support for DEW Point.
- only one Sensor allowed

## Description:

## Checklist:

### DIFF
--- a/tasmota/xsns_63_aht1x.ino
+++ b/tasmota/xsns_63_aht1x.ino
@@ -26,11 +26,11 @@
  * I2C Address: 0x38
  *
  * Attention: this Sensor is incompatible with other I2C devices on I2C bus.
- * 
+ *
  * The Datasheet write:
- * "Only a single AHT10 can be connected to the I2C bus and no other I2C 
+ * "Only a single AHT10 can be connected to the I2C bus and no other I2C
  *  devices can be connected".
- * 
+ *
  * after lot of search and tests, now is confirmed that works only reliable with one sensor
  * on I2C Bus
 \*********************************************************************************************/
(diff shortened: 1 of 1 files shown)


Jason2866:
First implementation of a Dew Point calculation with a good accuracy :-)
Considering using this values for a greater temp range (above 35°C) and still good accuracy?
`a = 6.112 mbar, b = 17.62, c = 243.12 °C; for −45 °C ≤ T ≤ 60 °C (error ±0.35 °C)`
See https://en.wikipedia.org/wiki/Dew_point

Jason2866:
For converting Celsius to Fahrenheit there is already a function in support.ino

device111:
Yes, i know the convert function but with the converting from Celsius to Fahrenheit you update also the "global" temperature. I don't know what happens when i update the global temperature with the dewpoint.

I think for a better accurate, we need a better Sensor. Most sensors themselves have such large deviations. But for the future implementaion i can build it in the 2 Sets. thats a good idea!
thanks!

tasmota/xsns_63_aht1x.ino
```diff
@@ -26,142 +26,198 @@
...
 
-void AHT10EverySecond(void)
+float CalcTemHumToDew(float h, float t)
 {
-  if (uptime &1) {
-    // AHT10: 55mS
-    if (!AHT10Read()) {
-      AddLogMissed(AHT10.name, AHT10.valid);
-    }
+  float gamma = log(h / 100) + 17.62 * t / (243.5 + t);
```
arendst: Could you try the below code if it reports the same result:
```
  float gamma = TaylorLog(h / 100) + 17.62 * t / (243.5 + t);
```
This would save 1k5 code space as Tasmota hates ``log()`` from the math library.

arendst:
As dewpoint calculation could be used by any Temp/hum sensor I will eventually move the calculation to the support file. For now pls have a look at the TaylorLog alternative which would save 1k5 code space as long as it provides the same output ofcourse.

arendst:
Merged part of this code and extracted dewpoint calculation for general use.

Thx.

tasmota/xsns_63_aht1x.ino
device111: Code is working with Taylorlog!

arendst: Great! Already released for all temp/hum sensors.

PR #8014, opened 2020-03-28
Add multiple GroupTopic<x> (x=1-4) (default disabled)

curzon01:
## Description:

`#define USE_GROUPTOPIC_SET` enables multiple grouptopics (1-4) subscription even when USE_DEVICE_GROUPS is not used.

Usefull for controlling devices in overlapping groups having [Device Group](https://github.com/arendst/Tasmota/blob/development/Device_Groups.md) feature disabled (e.g. using this MQTT based only for commands not implemented in device group feature).

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -344,6 +344,7 @@
 // -- MQTT ----------------------------------------
 #define MQTT_TELE_RETAIN     0                   // Tele messages may send retain flag (0 = off, 1 = on)
 #define MQTT_CLEAN_SESSION   1                   // Mqtt clean session connection (0 = No clean session, 1 = Clean session (default))
+//#define USE_GROUPTOPIC_SET                       // Enable multiple GROUPTOPIC<x>, x=1-4 (+0k1 code)
 
 // -- MQTT - Domoticz -----------------------------
 #define USE_DOMOTICZ                             // Enable Domoticz (+6k code, +0.3k mem)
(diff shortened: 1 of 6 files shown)


s-hadinger:
Nice. Why limit to 4 groups? Is it to limit to a single digit?

curzon01:
I had the idea independent of the group device implementation, during code view I saw that there were already 3 additonal GroupTopic strings defined (but only active if group device enabled and no subscription to it).
We could expand that, I just don't know where the limit for MQTT subscriber is without lib running out of memory (I think there would be a limitation, e.g. each subscribe needs an active socket etc...)

s-hadinger:
Ok thanks. Let's start with 4.

arendst:
I'll make it permanent for 4 group topics which will need more integration. Pls do not extent on this commit.

pcdiem:
I'd like to propose adding settings text values specifically for Device Groups. I used the MQTT topic when I wrote device groups support because it was logical that the device group name would match the MQTT topic. However, when a device supports controlling devices in multiple device groups, there is no logical tie between multiple MQTT topics and multiple device group names.

This PR is a great addition for all devices, not just those that don't use device groups. In order to enable this PR to work with devices that support multiple device groups, I believe we should split the device group name(s) off into separate settings rather than continuing to use the MQTT topic(s).

This would also support and upcoming PR I plan to submit that implements the DevGroupSend command which allows sending updates to a device group via a command. This command will likely be used in rules to send updates to device groups other than the one the device is a member of. This will require defining the device group names of any devices the user plans to send updates to because we can only reliably send to known device groups (since we need to not just send a packet but make sure the packet is received and acknowledged by the known members of the group).

Issue #8026, opened 2020-03-30
Rules & Scripts


prabbit237:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._

Compilation allows for mutually-exclusive defines of USE_RULES and USE_SCRIPTS and errors out with a message that's not clear as to what happened 
````
c:/users/mike/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: .pioenvs\tasmota\src\tasmota.ino.cpp.o:(.text._Z20MqttPublishTeleStatev+0x4): undefined reference to `RulesTeleperiod()'
````
Yes, I'll readily admit I can't read **sigh** I'm just suggesting an added 3 lines at the end of the my_user_config.h file so that if someone does define them both, the error will be clearer as to why it failed:
````
#if defined(USE_RULES) && defined(USE_SCRIPT)
  #error "Select either USE_RULES or USE_SCRIPT. They can't both be used at the same time"
#endif
````

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/#/help/FAQ)
- [x] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): NodeMCU
- [x] Tasmota binary firmware version number used: 8.2
  - [ ] Pre-compiled
  - [x] Self-compiled
    - [x] IDE / Compiler used: platformio
- [ ] Flashing tools used: _____
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```
  Configuration output here:


```
- [ ] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```
  Rules output here:


```
- [ ] Provide the output of this command: `Status 0`:
```
  STATUS 0 output here:


```
[...]

PR #8028, opened 2020-03-30
Rules OR Scripts enabled

Jason2866:
Solving #8026. It is documented but it seems not enough.

## Checklist:


arendst:
Thx but....

Wrong location as it doesn't trap user_config_override.

The correct location is at the end where you also find the Discovery and AWT test.
I'll add it now.

PR #8047, opened 2020-03-31
Change remove floating point libs from IRAM

s-hadinger:
## Description:

Changed linker script to move all floating point libs from libgcc.a out of IRAM. No interrupt handler is using float/double so we can safely move them back to IROM.

This is a quick and dirty patch, I'm sending a PR to Arduino Core in parallel to make it a compile flag.

IRAM size reduction:

- Tasmota (std): from 31660 to 30620 **(-1040 bytes)**
- Tasmota-sensors: from 31656 to 30616 **(-1040 bytes)**
- Tasmota-IR: from 29364 to 28324 **(-1040 bytes)**

The actual change is adding the following in `eagle.app.v6.common.ld` in section `.irom0.text:`:

```
    *libgcc.a:*f2.o(.literal .text)
    *libgcc.a:*f3.o(.literal .text)
    *libgcc.a:*fsi.o(.literal .text)
    *libgcc.a:*fdi.o(.literal .text)
    *libgcc.a:*ifs.o(.literal .text)
    *libgcc.a:*idf.o(.literal .text)
```

## Checklist:


Jason2866:
5 test devices with different sensors does work without a issue with this PR

arendst:
I store my source on "Google Drive". This fails the compiler as the linker path contains a space in the path. Moving the source to a non spaced folder works fine.

Working on renaming "Google Drive" to "GoogleDrive"...

s-hadinger:
This is probably not the best way to force a local linker script, but I didn't find a better way.

Anyways, I hope the PR https://github.com/esp8266/Arduino/pull/7180 will be quickly merged in Arduino Core. After that it will only be a compilation flag.

arendst:
Fixed it by removing and re-installing Google Drive pointing it to a renamed folder. Very (in)flexible this Google stuff.

Jason2866:
Another way would be patching the file on the fly.
https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html#override-package-files

EspEasy is doing this. Example [PUYA patch](https://github.com/letscontrolit/ESPEasy/blob/mega/patches/puya.patch)

s-hadinger:
That's very interesting, I didn't know this feature. I'm a bit reluctant to use patch files, they can very easily break whenever Arduino updates their files.

Issue #3555, opened 2018-08-22
Support 7 segment display


qingz2004:
**Have you look for this feature in other issues and in the wiki?**
Yes.

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Able to display temperature and humidity read from a sensor with a cheap 7 segment display.

**Describe alternatives you've considered**
N/A

**Additional context**
https://www.letscontrolit.com/wiki/index.php/7_segment_display

**(Please, remember to close the issue when the problem has been addressed)**

gemu2015:
why not use OLED?  (is already supported!)
128x64 Pixels from 2-5 Euros

or character LCDS (also supported)

qingz2004:
@gemu2015 Thanks! But I only need to display digits and 7 segment display is still much cheaper (less than half price).

Frogmore42:
There are oleds that are very small and currently supported.  It has taken a long time to get that support, so don't know when 7 segment will come.  But, will come sooner if you do the work to integrate it 😀.

qingz2004:
@Frogmore42 Got a link of a cheap OLED which is smaller than 7 segment and supported by Tasmota?
Thanks!

Frogmore42:
I haven't tried these yet, but I expect them to work with the new support.
https://www.amazon.com/MakerFocus-Display-SSD1306-3-3V-5V-Arduino/dp/B0761LV1SD

They are available cheaper directly from China, but I like 2-day shipping and have had good luck with Maker Focus products.  I also hate wasting time on stuff that doesn't work, so my definition of cheap includes how much time I need to spend getting it to work..

Frogmore42:
There are of course some very small 7-segment displays, but you didn't provide a reference for your definition of small.  I also like to be able to read my displays without needing to be closer than a few inches away.

qingz2004:
@Frogmore42 Thanks! I will try that. Hope it works out of box.
It is cheaper from Aliexpress, but I have to wait for around 2 months.

Jason2866:
@qingz2004 
I have ordered this some month ago and it is working with Tasmota.
https://www.aliexpress.com/item/1pcs-0-96-white-0-96-inch-OLED-module-New-128X64-OLED-LCD-LED-Display-Module/32640136921.html?spm=a2g0s.13010208.99999999.259.39c13c00rkUzgu

qingz2004:
[...]

PR #8116, opened 2020-04-10
New display driver for I2C HT16K33 based 7 segment LED

tthk:
## Description:
New display driver for I2C HT16K33 based 7 segment LED.

There exists a closed issue #3555 for this functionality, closed not implemented because the requester found a different hardware product which was supported.  I needed large 4 digits 7 segment LED display, driven by HT16K33 backpack which the Tasmota code already supported with #define USE_DISPLAY_MATRIX.  For the cleanest implementation, I used the xdsp_03 as a template and added xdsp_11 driver.  Supports only one 7 segment display (unlike matrix xdsp_03 which supports multiple matrices).  Scrolling is not supported.

Added driver files:
xdsp_11_sevenseg.ino (the next available unused xdsp number)
activated by:
#define USE_DISPLAY_SEVENSEG

Since the 7 segment display can only display numbers, I used some text prefix control codes to activate other segments:
: -> colon
x -> print hex
^ -> top left dot
v -> bottom left dot
. -> ampm dot
T -> 12 hour format
t -> 24 hour fomat
s -> given number is seconds
Examples:
787 -> 787
x47 -> 2F
st:241 -> 04:01
sT241 -> 4 01

## Checklist:


jziolkowski:
what about characters that can be shown on 7seg displays? If I wanted to display temperature as  24.2C or 155F?

tthk:
> what about characters that can be shown on 7seg displays? If I wanted to display temperature as  24.2C or 155F?

Since 24C (or 242C for 24.2C because the LED display doesn't have a segment for decimal point) and 155F are valid hex, the display can show it by giving it the decimal value and using 'x' to display as hex:
```
DisplayText x588 => 24C
DisplayText x9260 => 242C
DisplayText x5471 => 155F

```
My 7 particular segment display has an dot segment that's in the place of the degree symbol (between the right-most two digits).  So you can get the degree symbol as well by adding '.' like so:

```
DisplayText x.588 => 24°C
DisplayText x.5471 => 155°F

```
Here are the modifiers you can add anywhere before the number.
For reference, LED display with all segments on looks like:  :88:8°8

```
^ => top part of left :
v => bottom part of left :
. => AM/PM/degree dot
: => right colon
x => display given dec number as hex
s => given number is understood as duration in seconds, display as HH:MM or MM:SS as appropriate

```
DisplayMode:
```
0 => Print decimal number given, including any modifying prefixes (above)
1 => Display time HH:MM 12 hour format every second
2 => Display time HH:MM 24 hour format every second

```
For example, to display all segments:
```
DisplayText ^v:.8888 => :88:8°8
```

jziolkowski:
Thanks for a detailed response. However, back to my temperature display question:
There is no way in rules in tasmota to apply this conversion to simply display output of a thermometer on such display without involving an external tool. Maybe it would be worth adding a special prefix so that your code does this conversion internally. Ideally, it could take into account the current setoption for global tasmota temperature display. But manual prefix would be good enough I think (say, TC/TF?)

jziolkowski:
it's worth mentioning that the most often uses sensors in tasmota are indeed the temperature/humidity sensors, so I'm pretty sure this would be a welcome addition by many

tthk:
Temperature/humidity display on a seven segment LED seems like it would be very useful.
Perhaps an implementation of DisplayMode 2 that displays local sensors on a seven segment LED?
Also, on a different display from the seven segment LED, like a 2004A 20x4 LCD for example, how would a rule to display temperature work using DisplayText?

KnightsKingdom:
Which DisplayModel do I need to use for a 0.56" Led Backpack? When I use 3 I2C returns a 8X8Matrix starting address 0X73 to 0X76. The address of my backpack is 0X73 which I also define in DisplayAddress which returns DisplayAddress1 = 115.

tthk:
> Which DisplayModel do I need to use for a 0.56" Led Backpack? When I use 3 I2C returns a 8X8Matrix starting address 0X73 to 0X76. The address of my backpack is 0X73 which I also define in DisplayAddress which returns DisplayAddress1 = 115.
> Flashed 8.3.1 on a Wemos D1 mini, configured Tasmota as a generic 18 model and set D1 as SCL and D2 as SDA.
> Output from Display:
> 13:44:08 RSL: stat/tasmota_424054/RESULT = {"Display":"Model":1,"Width":16,"Height":2,"Mode":0,"Dimmer":15,"Size":1,"Font":1,"Rotate":0,"Refresh":2,"Cols":[16,8],"Rows":2}}
> Tested it also with an example sketch from Adafruit and that worked out of the box.

As per code and file ```I2CDEVICES.md```:
```
DisplayModel 11
```
You'll also need to compile with the following defines (usually in your ```my_user_config.h```), based on your LED backpack I2C address:
```
#define USE_DISPLAY
#undef SEVENSEG_ADDRESS1
#define SEVENSEG_ADDRESS1 0x73
```

KnightsKingdom:
Thanks, working fine now!

Any chance in the near future for multiple display support as it is I2C, adding 1 or 2 more is easy from a wiring perspective, don't know about code though.

tthk:
Multiple of same type DisplayModel of display is doable, kinda like Matrix display in a grid which is already in the code.  Multiple displays of different types is tough because of different DisplayModels driver loading functionality, which I would think is a major code change.

marksmanaz:
I have the Adafruit 7 segment .56" i2c backpack with the four dots "." in between each digit but I can't display a number with a decimal it just truncates the number to only the two digits to the left of the decimal point. Do you know what I am missing?

DisplayText 84.7 shows 84 in the last two digit positions

KnightsKingdom:
> Multiple of same type DisplayModel of display is doable, kinda like Matrix displays in a grid which is already in the code. Multiple displays of different types is tough because of different DisplayModels driver loading functionality, which I would think is a major code change.

Yes, I'm talking about the same DisplayModel. I want to use multiple DS18B20 temperature sensors for checking my pool heater water and sending the values to HomeAssistant, but also displaying them near the pool. Using one device is much more convenient than one device per sensor.

tthk:
> I have the Adafruit 7 segment .56" i2c backpack with the four dots "." in between each digit but I can't display a number with a decimal it just truncates the number to only the two digits to the left of the decimal point. Do you know what I am missing?
> DisplayText 84.7 shows 84 in the last two digit positions

Ah, I see.  I don't have this 7 segment model, so no code was written for these decimal points between the digits, and no tests were done on this model.  I'll check the code and post the best place to make this change soon.

tthk:
> Yes, I'm talking about the same DisplayModel. I want to use multiple DS18B20 temperature sensors for checking my pool heater water and sending the values to HomeAssistant, but also displaying them near the pool. Using one device is much more convenient than one device per sensor.

That sounds like a great use case: separately addressable, 7 segment displays.  This is definitely doable with some code changes.  I'll look into it.

tthk:
> > Yes, I'm talking about the same DisplayModel. I want to use multiple DS18B20 temperature sensors for checking my pool heater water and sending the values to HomeAssistant, but also displaying them near the pool. Using one device is much more convenient than one device per sensor.
> 
> That sounds like a great use case: separately addressable, 7 segment displays. This is definitely doable with some code changes. I'll look into it.

Added multiple seven segment display functionality.  Please see Pull Request #8886

tthk:
> > I have the Adafruit 7 segment .56" i2c backpack with the four dots "." in between each digit but I can't display a number with a decimal it just truncates the number to only the two digits to the left of the decimal point. Do you know what I am missing?
> > DisplayText 84.7 shows 84 in the last two digit positions
> 
> Ah, I see. I don't have this 7 segment model, so no code was written for these decimal points between the digits, and no tests were done on this model. I'll see if I can acquire one of these units and implement controlling these segments in the driver.

Added support for Adafruit 0.56" 7-Segment LED display, with floating point and raw segment addressing.  Please see Pull Request #8900

PR #8159, opened 2020-04-13
Fix ESP32 build env in PlatformIO and merge 

Jason2866:
override files. Move ESP32 env in extra file `platformio_tasmota_env32.ini`

## Checklist:
  - [x] The code change is tested and works on core Tasmota_core_stage and ESP32


arendst:
Do we need all those lines for each 32 binaries? The original ini used just two lines.

Jason2866:
yes (saidly), there can be only one [env] general definition. 
Since it is used for the ESP82xx setup, every ESP32 build have to include all

PR #8171, opened 2020-04-14
Better placing of ESP32 env...

Jason2866:
will be only activated when platformio_override is active

## Checklist:
  - [x] The code change is tested and works on core Tasmota_core_stage and ESP32

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,7 +12,6 @@ src_dir = tasmota
 build_dir = .pioenvs
 build_cache_dir = .cache
 extra_configs = platformio_tasmota_env.ini
-                platformio_tasmota_env32.ini
                 platformio_override.ini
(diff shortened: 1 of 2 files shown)


arendst:
That's much better as it is still Alpha functionality I want to keep hidden for casual users.

I self have just added the suggested env32 parameters at the end of my custom platform_override.ini file WITHOUT any changes to platform.ini and platform_tasmota_env.ini considered before your change of yesterday.

Woudln't that be a better solution to hide it even more?

Jason2866:
Doing like in this PR, a normal user will not recognize the ESP32 version. It is completely hidden.
Even after activating override file only the standard Tasmota is build. To compile a ESP32 variant user has to do changes in override file.

PR #8174, opened 2020-04-14
update scripter script size on esp32

gemu2015:
## Description:

scripter uses SPIFFS on esp32 for scripts
#define ESP32_SCRIPT_SIZE xxx  defaults to 8192 bytes

## Checklist:

### DIFF
--- a/tasmota/xdrv_10_scripter.ino
+++ b/tasmota/xdrv_10_scripter.ino
@@ -65,6 +65,31 @@ keywords if then else endif, or, and are better readable for beginners (others m
 uint32_t EncodeLightId(uint8_t relay_id);
 uint32_t DecodeLightId(uint32_t hue_id);
 
+#ifdef ESP32
+
+#include "FS.h"
+#include "SPIFFS.h"
+void SaveFile(char *name,const uint8_t *buf,uint32_t len) {
+  File file = SPIFFS.open(name, FILE_WRITE);
+  if (!file) return;
+  file.write(buf, len);
+  file.close();
+}
+
+#define FORMAT_SPIFFS_IF_FAILED true
+
+void LoadFile(char *name,uint8_t *buf,uint32_t len) {
+
+  if(!SPIFFS.begin(FORMAT_SPIFFS_IF_FAILED)){
+      return;
+  }
+  File file = SPIFFS.open(name);
+  if (!file) return;
+  file.read(buf, len);
+  file.close();
+}
+#endif
+
 // offsets epoch readings by 1.1.2019 00:00:00 to fit into float with second resolution
 #define EPOCH_OFFSET 1546300800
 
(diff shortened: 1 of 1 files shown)


Jason2866:
SPIFFS, really? Slow and it is deprecated. In ESP82XX it will be removed soon...
Theo avoided SPIFFS for good reasons

gemu2015:
@Jason2866 
you are right SPIFFS is slow and deprecated, but was 15 minutes to implement and since script is loaded only once on startup and saved only when modified by user speed is not a problem.
if it is deprecated it is extremely easy to update to a better driver.
nvs partition has only 20kb and the nvs driver is also not very stable.
one could also think about to adapt a fat system instead of the sdcard interface (optionally)
however spiffs partition is only 1,47 mB, so not very useful.
and i think we should shrink it in favor of larger app sections.
i want to implement the camera interface with AI which takes a lot of flash, so app size should be much larger.

arendst:
I see espeasy is using a 4M divide like ``esp32_partition_app1810k_spiffs316k``. 

What are the possiblities with regards to these parameters?
Would that divide be enough (preferred for easy switch between tools) although a lot is wasted in 316k SPIFFS?

gemu2015:
camera interface with ai uses 2 mb, so together with tasmota we get about 3 
so 2*1.8 should be enough.  anyhow we need 2 step update via OTA
 we actually do not need more SPIFFS then e.g. 32k
when using face recognition pictures should be stored on SDCARD anyhow.

Jason2866:
If i do not missunderstand the way partitions are defined for ESP32 it should be easy
to change the size for the SPIFFS partition.

arendst:
I think there are no changes needed in PlatformIO. 

It's just a matter of defining ``board_build.partitions  = min_spiffs.csv
`` in ``[common32]`` and ``board_build.partitions  = ${common32.board_build.partitions}
`` in every ``[env:tasmota32]``.

This results in the below layout:
```
# Name,   Type, SubType, Offset,  Size, Flags
nvs,      data, nvs,     0x9000,  0x5000,
otadata,  data, ota,     0xe000,  0x2000,
app0,     app,  ota_0,   0x10000, 0x1E0000,
app1,     app,  ota_1,   0x1F0000,0x1E0000,
spiffs,   data, spiffs,  0x3D0000,0x30000,
```

Jason2866:
Yes, that are the needed changes :-) No eeprom area?
image

arendst:
No. Tasmota does not use eeprom:

image

gemu2015:
perfect!

Jason2866:
okay than 32 k SPIFFS?

arendst:
And from what I gather from the Arduino implementation they moved the eeprom partition within the nvs partition.

Yes only 32k SPIFFS. Who needs SPIFFS?

Jason2866:
Gemu :-)

arendst:
I noticed you'll need a serial upload to change the partition layout so the sooner the better we implement this change.

@Jason2866 can you push a PR with these parameters? (I'm offline within a minute ;-)

Jason2866:
I try to do in a minute

gemu2015:
@Jason2866 
strange that your table did not work

this one works, gives 64k SPIFFS

# Name,   Type, SubType, Offset,  Size, Flags
nvs,      data, nvs,     0x9000,  0x5000,
otadata,  data, ota,     0xe000,  0x2000,
app0,     app,  ota_0,   0x10000, 0x1F0000,
app1,     app,  ota_1,   0x200000, 0x1F0000,
spiffs,   data, spiffs,  0x3F0000,0x10000,

Jason2866:
Thx @gemu2015   Not my day...
#8181

PR #8185, opened 2020-04-15
Scripter update

gemu2015:
## Description:

expands scripter text buffer to up to 30kb on ESP32 if defined
#define ESP32_SCRIPT_SIZE

normal scripter storage:
1536 bytes for text and 50 bytes for permanent vars
with ESP32_SCRIPT_SIZE eg set to 8192:
8192 bytes for text and 1536 bytes for permanent vars

comment on SPIFFS:
spiffs is deprecated on ESP8266 but the successor is not yet ported to ESP32
since the interface api is compatible it is easy to switch when the new filesystem is ported
 
## Checklist:

### DIFF
--- a/tasmota/xdrv_10_scripter.ino
+++ b/tasmota/xdrv_10_scripter.ino
@@ -4854,11 +4854,13 @@ bool Xdrv10(uint8_t function)
 
   switch (function) {
     case FUNC_PRE_INIT:
+    /*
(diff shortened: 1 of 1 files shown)


GrendelWWU:
Oh, is someone really working on an ESP32 port of Tasmota? I'd like it!

PR #8189, opened 2020-04-15
enable single PIN mode for SR04 sensor

VanillaFord:
## Description:

Purpose is to enable single PIN mode for SR04 sensor as depicted in newping library, see:  [https://bitbucket.org/teckel12/arduino-new-ping/wiki/Home#!single-pin-sketch](url)
If in tasmota pin GPIO_SR04_TRIG is not configured, use sensor interface in single PIN mode with GPIO_SR04_ECHO only

...verified this change on my Wemos D1 mini and SR04 sensor in single/dual PIN Mode and  wiring/configuration.

## Checklist:
  - [ x] The code change is tested and works on core Tasmota_core_stage
  - [ x] The code change pass travis tests. **Your PR cannot be merged unless tests pass**


effelle:
Tested the change with one HC-SR04P configured with 5V and 3V, with and without level shifter. 
The range seems not affected (and depends more by the quality of the sensor) and the oscillations are virtually the same of a normal setup.

effelle (approved):

arendst:
This change also influences the serial comms to a Mode 2 sensor. Was this tested verified too?

effelle:
If you mean one with additional OUT like the SRF05, no. 
I just tested the SR04 I had in my magic drawer in both 1 and 2 wires configuration.

arendst:
The issue is that Mode 2 devices seem to use the standard serial library. Defining both serial input and output to the same GPIO doesn't make sense.

What does your sensor report during startup? Should be something like ``SR04: Mode 2``

effelle:
In both cases shows SR04: Mode 1.

arendst:
Great. So they use the ping library. Now I wonder what a Mode 2 device would do with this chance.

effelle:
Perhaps @andrethomas has one to test? I remember he did a series of deep tests with the osci and perhaps he tested a mode2 too.

arendst:
As the change is only needed for mode <2 devices I'll make the change myself and upload commit.

arendst:
Pls try the latest commit. It should have this PR implemented correctly.

PR #8211, opened 2020-04-17
Update Italian and English language

bovirus:
@arendst 

- Italian

Revise line 560/561 to align to other strings similar scheme

- English/Italian

Change "Main menu" to "Back to Main Menu" (its the command from Configuration to go back to main menu)

## Description:
Fix language issues

## Checklist:
  - [ ] The code change is tested and works on core Tasmota_core_stage


Jason2866:
Why sacrifing bytes? It is understandable and logic as it is.

bovirus:
What so big sacrifice are 10 bytes that can help to understanding well the interface?.
We are talking about a very small about of bytes.

arendst:
As discussed before. Won't merge. Just use your own compiled languages.

bovirus:
@arendst 

Someone have to explain me because the benefit for this GUI improvements should only for me and not for all Italiain (and English) user?

Are you so at end of life of space to save 10 bytes ?

**_I beleieve that IS NOT TRUE!!!!!!

Is just your fixeld old point of view of your mind!!!!!!_**

meingraham:
@bovirus 

This does not restrict you from making changes you desire. You can fork Tasmota, modify your language file, and compile the firmware and you have the UI update you desire. Thus this is not keeping you from having the feature you want. Nor from using Tasmota or not.

Yes, it's "just 10 bytes". But your 10 bytes and then some other "just 10 bytes", and then another. Eventually these add up. When there is not a __substantial__ benefit to the entire (5 million plus) Tasmota installed base, then those 10 bytes have greater benefit when adding __future__ functionality or new features.

Also, recall that in order for OTA flash to continue to work, there must be "unused" memory to allow the intermediate "minimal" upload. So, having several Kbytes of "free" memory is not really "unused".

bovirus:
@arendst 

To have free Kb of memory doens't mean KB unused.
My meaning that with 42KB of free memory 10 bytes (for GUI changes) less are not a big small percentage of memory reduction (0,025%).

meingraham:
@bovirus 

To be clear when using the term "memory"... There is program memory, RAM, and iRAM. The firmware binary is built (by default) for 1MB total flash program memory devices. From this 1MB, there needs to be roughly 300KB free program memory in order to accommodate OTA updates. So, these bytes need to be first accounted for making the max firmware binary size roughly 650KB.

Then, RAM and iRAM are fixed and depending on the Tasmota variant (IR, sensors, display, etc.) those features use varying amounts of those limited resources. Having "42KB" free RAM in the variant you are compiling does not mean that the same amount of free RAM would be available for other variants. iRAM is even more limited.

These bytes are precious! So, "frivolous" use of 10 bytes here and 10 bytes there needs to be very critically evaluated.

I'm just trying to explain Theo's position on your request... and it is one many others agree with.

AND, if __you__ want to use those bytes, you are welcome to do what you have done and create your own version and compile it. In other words, you can make your own decision. But for the master branch of Tasmota, the decision is to not accept your feature request.

Cheers!

Issue #481, opened 2017-05-25
Hard Thermostat


off14:
Is it possible to include functionality of HARD thermostat - where no MQTT(or similar) server is available or there are no commands received.
Of course this mode will require temp sensor connected.
You need to include logic to support COOL or HEAT modes,  Hystersis setup  (0.1 - 5) together with TargetTemperature

This will make sonoff working as a thermostat with web interface, even without any server. Very useful when sonoff is installed on remote location and there are big chances that internet is not stable there

WDYT?

citrusfizz:
i like this idea.   i recently built a thermostat with a raspberry pi, some relays and Node-Red and mqtt all built in.   Trusting the PI for 24/7 operation is a bit sketchy. and if it reboots it takes a while to boot up and get all its gears turning.   Working well so far,  if you have a PI i can share my node red flows with you.

off14:
Thanks, i have mine also in node-red and in openhab rules
the main idea is to have **independent** thermostat, which not requires any server (mqtt or whatever)

CommodoreWhite:
Why not for any sensor that returns an analogue value (that is, not binary). So, for instance, a set point for humidity to control the bathroom extractor fan, a light sensor could turn security lights on and off, a proximity detector could control stair well lights. Hmm, the last one is a bad example but you get the picture. 

The underlying logic is generic, if a sensor value is exceeds some thresholds, switch a relay and or send a message.

The implementation might not be as responsive as you like because, if memory serves, sensors are polled rather than continuously monitored and why having a proximity sensor manage stair lights could have undesirable consequences.

So it's a +1 from me.

Peter

CommodoreWhite:
Oh! I forgot. It would be nice if the feature could be enabled or disabled depending on the presence or absence of an Mqtt connection. So, if the Mqtt service dies, lights still turn off and on. If Mqtt is working the control reverts to a higher authority.

AlexTransit:
I made a simple thermostat at my place. (On fast)
He has no settings

I wanted to make PID and settings.
Do not do it. :)

sonoff.ino
void every_second()
{
  char log[LOGSZ], stopic[TOPSZ], svalue[MESSZ], stime[21];
  float ctemp;

#ifdef TEMPERATURE_CONTROL_DS18x20
if (ds18x20_read(0, TEMP_CONVERSION, ctemp, true)) 
{
  if ((ctemp == 85.0) || (ctemp >= sysCfg.temp_target)){
[...]

PR #8212, opened 2020-04-17
Thermostat Controller

arijav:
## Description:

New driver created for thermostat control. I have created initially this as a LUA script running in Domoticz on a Raspberry Pi to control floor heating valves using Qubino Z-wave relays. I have ported this to a Tasmota driver embedding the functionality in the relays. This driver has been successfully tested with a Shelly 1PM, a Shelly 2.5 and a Sonoff 4CH PRO R2. Summary of the features:
- Implementation of multi independent controllers in the same device (number selected by define to save memory, in-depth testing started with Sonoff 4CH PRO R2, optimizations pending).
- The controller offers three operation modes, Off, Automatic (controller active) and Manual (controller following the input from a serial connected wall thermostat). The transition to manual happens once the input switch is active, allowing seamless integration with existing wall thermostats.
- The controller offers 2 climate modes, heating as well as cooling (for cooling the testing is pending as well as further optimizations).
- The controller offers 3 controlling strategies (Hybrid, predictive rampup control and PI). 
- The temperature can be received via MQTT or a local temperature sensor (name set by define, default DS18B20). 
- The default input format for the temperature is degrees Celsius but it can be switched to Fahrenheit via MQTT command.
- The controller offers a diagnostic functionality based on inconsistent states between output, command and input (Diagnostic for maximum power at output already covered by the energy driver).

Future features (TODO´s):
- Configuration over the Web Timer's GUI (Selector for Thermostat and box for Setpoint).
- Rework of global_temperature concept in Tasmota for using several temp. sensors
- Humidity control (to be defined).

## Checklist:


arijav:
I have found many time variables that can be reduced to uint16_t (resolution in minutes instead of seconds). I will commit the changes today in the evening.

device111:
issue, address not correct:
uint8_t       free_f36[70]; 
check:
 tasmota/tasmota.h

do you think, you can integrate like tasmota timers? i don't now if this works, but will be easier for set heating plan.
at the moment you are loose all settings from heating plan?

s-hadinger:
From what I can see you are not using Flash Settings anymore, but only compile time options. Right?

arijav:
> issue, address not correct:
> uint8_t free_f36[70];
> check:
> tasmota/tasmota.h
> 
> do you think, you can integrate like tasmota timers? i don't now if this works, but will be easier for set heating plan.
> at the moment you are loose all settings from heating p

arijav:
Thanks for the comments:

1. I will correct Tasmota.h and recommit the change. Regarding the timers I can investigate, which ones do you mean that are existing in Tasmota?

2. Yes, I do no longer use the flash settings. The only small drawback is that for specific configurations not fitting the default values. after a reset the device should be reconfigured via MQTT (which is possible for each of the parameters).

arijav:
> issue, address not correct:
> uint8_t free_f36[70];
> check:
> tasmota/tasmota.h
> 
> do you think, you can integrate like tasmota timers? i don't now if this works, but will be easier for set heating plan.
> at the moment you are loose all settings from heating plan?

I have added heating to the kCommandSource array. I did not fully get your comment about the uint8_t free_f36[70], I did not change anything in settings.h. Can you explain?, thanks

effelle:
uint8_t free_f36 is  already assigned on setting.h. 
Next in line: `uint8_t       free_f37[129]`

device111:
Timers from Tasmota. i.e. a checkbox with heater. the advantage, you can use the existing settings. I don't no how much work it is, but i think, you can save more memory.

grafik

arijav:
> Timers from Tasmota. i.e. a checkbox with heater. the advantage, you can use the existing settings. I don't no how much work it is, but i think, you can save more memory.
> 
> grafik

Seems interesting, I will investigate but for sure this would be a big change, which will require intensive testing. There should be a way to block the timer for reconfiguration from a different source if used by heating, or the heating driver to identify if the timer has been changed externally to turn heating off.

I am pretty new to Tasmota, how is the process here to get a driver integrated?. Is it possible to integrate it as is or with minor changes coming from reviews and afterwards evaluate further optimization with more intensive changes?

device111:
i don't no what Theo said, but i think, your are need to much settings bytes at the moment.
And have a look in the settings.h. All things must be correct.
free_f37[129]...correct index and address-name.
Timers: evtl. somebody can help you with more details.

arijav:
> i don't no what Theo said, but i think, your are need to much settings bytes at the moment.
> And have a look in the settings.h. All things must be correct.
> free_f37[129]...correct index and address-name.
> Timers: evtl. somebody can help you with more details.

Ok, in the meantime I will further optimize the memory requirements and understand better the timers driver.

I did not understood the issue you mean with settings.h. In this PR I have not changed settings.h (it stays untouched as in the development branch)

arijav:
I have checked how to use the rules and timer drivers. I think I can easily them remove all the heating scheduling from my driver and replace it with a combination of rules and timers.

Example:
For setting a target temperature of 22.5 degrees celsius from Monday to Friday at 15:00
Config timer:
Timer1 {"Arm":1,"Time“:“15:00“,“Window":0,"Days“:“SMTWT00“,“Repeat":1,"Output":1,"Action“:3}
Config rules:
Rule1 ON Rules#Timer=1 DO TempTargetSet 22.5

I would appreciate a short feedback if this is the way to go and then I will remove the unneeded code as well as variables (84 bytes).

Thanks

ascillato:
@arijav The whole Thermostat Routine can be achieved by rules and timers as the Thermostat example of the documentation: https://tasmota.github.io/docs/Rules/#thermostat-example

But, if you want to make it easy for users, so they don't need to "program" some rules, one idea is to add a new option in the dropdown menu of Timers like the actual RULE option. And/Or add an optional webmenu for the whole Thermostat.

Also, this can be extended as a climate controller (heating and cooling). Just an idea taken from old issues.

arijav:
> @arijav The whole Thermostat Routine can be achieved by rules and timers as the Thermostat example of the documentation: https://tasmota.github.io/docs/Rules/#thermostat-example
> 
> But, if you want to make it easy for users, so they don't need to "program" some rules, one idea is to add a new option in the dropdown menu of Timers like the actual RULE option. And/Or add an optional webmenu for the whole Thermostat.
> 
> Also, this can be extended as a climate controller (heating and cooling). Just an idea taken from an old issue (#481), an old PR (#2486) and from the @IceStuff Tasmota's Fork (https://github.com/IceStuff/Sonoff-Tasmota)

Thanks for the info. I have seen the example made by using rules and its a two point thermostat, which is a simple one for applications where the temperature precision is not so important. The current state of the art thermostats from the main manufacturers use PI control for accurate heating and even predictive algorithms which is what I have implemented in the driver.

The idea of cooling is a good one, I will add it for the future. For testing however it would be tricky and I will need somebody who has a house with cooling system. I am testing the driver in my own house and I have a floor heating system that does not allow cooling. Furthermore there is something mentioned in the old issue about support for external DS18B20 temperature sensors as alternative to MQTT. This is something I will add soon, as I have ordered one of them for the Shelly I use for testing.

device111:
now, i think, your concept is on the right way! Great...

ascillato:
@arijav 

The defines you have added in **_i18n.h_** should go inside your driver.

arijav:
> @arijav
> 
> The defines you have added in **_i18n.h_** should go inside your driver.

The defines that I have created in i18n.h are for each of the command callbacks, I have seen all drivers define them there...

device111:
for testing and the first try, its better you put it in your driver. you can expand it later to "system" files.
At the moment, there are a lot of changes with esp32 implementation. Then it's better all of your things are in your driver.

arijav:
> for testing and the first try, its better you put it in your driver. you can expand it later to "system" files.
> At the moment, there are a lot of changes with esp32 implementation. Then it's better all of your things are in your driver.

Understood, will move them then soon

arijav:
Done, every define is into the driver. I have renamed the driver to thermostat which is more suitable for future updates to allow cooling control

arijav:
Hello, maybe a question to Theo: am I allowed as suggested by ascillato to modify the Web interface of the timers and add an option to trigger a temperature setpoint?

In general I would appreciate a short feedback if you see the possibility of this driver being integrated into Tasmota or if it does not fit into your strategy.

Thanks

arendst:
Yes. You may add to the drop down list

image

but use a small name. It is probably easier that I add it to the current timer code as it is intertwined with both javascript and C. Just let me know what the name needs to be and what function in your driver to call.

arijav:
> Yes. You may add to the drop down list
> 
> image
> 
> but use a small name. It is probably easier that I add it to the current timer code as it is intertwined with both javascript and C. Just let me know what the name needs to be and what function in your driver to call.

Thanks,  I will. I have however experimented a little bit to learn. Do you imagine something like this?:

image

I will prepare 2 functions, one to set the temperature and activate the thermostat at the same time to be triggered once the timer is activated, the other to shut the thermostat. The reason is that it has to be shut-off if no longer active to avoid the thermostat to drive the relay further.

arijav:
> > Yes. You may add to the drop down list
> > image
> > but use a small name. It is probably easier that I add it to the current timer code as it is intertwined with both javascript and C. Just let me know what the name needs to be and what function in your driver to call.
> 
> Thanks, I will. I have however experimented a little bit to learn. Do you imagine something like this?:
> 
> image
> 
> I will prepare 2 functions, one to set the temperature and activate the thermostat at the same time to be triggered once the timer is activated, the other to shut down the thermostat. The reason is that it has to be shut-off if no longer active to avoid the thermostat driver to further actuate the relay.

Theo I have added the two functions:

- bool ThermostatTimerArm(int16_t tempVal) -> Expects the temperature in tenths of degrees celsius (f.i. 225 = 22.5). Returns true if temp plausible and thermostat armed, false if not

- void ThermostatTimerDisarm() -> Disarms the thermostat

I am experimenting with the web interface but I have not committed the changes, I think it is best if you change it as it suits you as you stated. The important thing will be to call ThermostatTimerDisarm to disarm the thermostat if the timer is disabled or changed for a different function, to prevent it to mess up with the relay.

arijav:
Support for local DS18B20 temperature sensor added further to MQTT

arendst:
There is no need to scan JSON messages to find a locally connected temperature sensor. Tasmota provides an internal variable called ``global_temperature`` which contains the temperature read by a locally attached temperature sensor. Could be any type of hardware.

So I suggest you rewrite your latest addition using the knowledge from ``xdrv_03_energy.ino`` lines 442-446 demonstrating the use of this local variable.

I suggest too to search for ``global_temperature`` to find more examples in the code.

And pls remember, keep the code as small as possible. I see a lot of logging which luckily can be disabled by a define which looks good.

arijav:
> There is no need to scan JSON messages to find a locally connected temperature sensor. Tasmota provides an internal variable called `global_temperature` which contains the temperature read by a locally attached temperature sensor. Could be any type of hardware.
> 
> So I suggest you rewrite your latest addition using the knowledge from `xdrv_03_energy.ino` lines 442-446 demonstrating the use of this local variable.
> 
> I suggest too to search for `global_temperature` to find more examples in the code.
> 
> And pls remember, keep the code as small as possible. I see a lot of logging which luckily can be disabled by a define which looks good.

Thanks for the tip, I have modified the code to use global_temperature but I have issues. In the shelly PM I have attached a DS18B20 and adapted the template according to your modifications (old issue from February I think) to read both temperatures the internal one of the module and the DS18B20. The issue is that the variable global_temperature is being update with both of them sequentially:

Example of my reads:
13:21:10 MQT: stat/tasmota_BA0DC9/RESULT = {"GlobalTemp":21.9}
13:21:13 MQT: stat/tasmota_BA0DC9/RESULT = {"GlobalTemp":33.5}
13:21:16 MQT: stat/tasmota_BA0DC9/RESULT = {"GlobalTemp":33.5}
13:21:17 MQT: stat/tasmota_BA0DC9/RESULT = {"GlobalTemp":33.5}
13:21:18 MQT: stat/tasmota_BA0DC9/RESULT = {"GlobalTemp":21.9}
Infos of the main Tasmota webpage of my Shelly:
Temperature: 33.5°
DS18B20 Temperature: 21.9°

I can have a look to solve the issue, but currently I foresee that all modules with an internal temperature sensor and an attached external one would face this problem.

Regarding Debug, yes, I will leave everything into the define and disable it by default, furthermore it is my intention to reduce it drastically soon, since I see that the controller reacts exactly as the one I have in Domoticz based on LUA.

arendst:
You're right. I had hoped only one sensor was connected but if more are used it becomes "messy".

Adding and "external" global_temperature is indeed a fix as long as there is only one external sensor connected. It will need a kind of large rework task.

For now, let's leave your solution in place and only think about a possible global solution.

kretu:
Would it be hard to implement control by humidity rather than temperature?

arijav:
> Would it be hard to implement control by humidity rather than temperature?

It means some rework but it is feasible. I can make it active via define to save some memory and will need to think about a better name for the driver. If you describe the system a little bit I can add it to the lists of TODO´s and evaluate

arijav:
I do not get why the TW build failed. Is there any way to run the check again?

yury-sannikov:
hi @arijav, awesome PR! Do you think it makes sense to add telemetry output enabling tasmota Rules tapping into the output value and do other things rather than MQTT and Relay control?

arijav:
> hi @arijav, awesome PR! Do you think it makes sense to add telemetry output enabling tasmota Rules tapping into the output value and do other things rather than MQTT and Relay control?

Thanks, what it is foreseen is to be able to configure the timers to arm the thermostat to reach a given set point and disarm them and therefore rules could be used. Do you mean this or something else (maybe an example can help)?

3ricj:
First of all, I am really excited about this capability.  I don't like the idea of thermostats running off external devices, so having it integrated on the controller is perfect.   Thank you.  

Some general feedback: 

 - there are lots of application specific control methods.   

For example, compressor based heat pumps (AC/cooling/heating) need a minimum cycle time because you don't want to restart a compressor quickly; this can lead to stalling. Inversely, a sousvide cooker would want a fast cycle time.   I feel this is pretty important. 

There are also TECs that may need proportional control (not just on or off but on percentages) using PIDs. Its unlikely many will use this.  Some really complicated systems use variable speed heat pumps.  You see these marketed as "inverters". 

It would be great if this code could also manage error states.  For example, if the heater is "ON" but it's not getting warmer, it's an error.  Same for AC on but not cooling.  In some cases it should error on overrun: if it's set to cool down to 1c, but we are below that (say -3c) that could be worthy of an error flag.  If this was used to control a fridge you don't want to go below freezing. 

A related, but harder to compute is a flag for "stable temp".  Eg: it's not oscillating, overrun, underrun, etc. 

- it would likely be important to have a concept of deadband.   Cold tolerance and hot tolerance. 

Anyway, none of this needs to be implemented, this is not a feature request, but just some food for thought.

yury-sannikov:
> Thanks, what it is foreseen is to be able to configure the timers to arm the thermostat to reach a given set point and disarm them and therefore rules could be used. Do you mean this or something else (maybe an example can help)?

My point is pretty much the same as @3ricj 
>  There are also TECs that may need proportional control (not just on or off but on percentages) using PIDs. Its unlikely many will use this. Some really complicated systems use variable speed heat pumps. You see these marketed as "inverters".

I can see 2 types of the telemetry output:
- Power, in a range of 0-100%
- The desired temperature in celsius.

For example, to control the duty cycle of some inverter one can use:
```
Rule
  on tele-pid#output#duty_cycle do event pwm=(%value%*1024) endon
  on event#pwm do  WebSend [...] Pwm %value% endon
```

For example, to control the boiler with PID:
```
Rule
  on tele-pid#output#temperature do ot_tboiler %value% endon
```

In those examples telemetry JSON might look like:
```
{...., "pid":{"output": {"duty_cycle": 0.85, "temperature": 78}}, ....}
```

arijav:
> > Thanks, what it is foreseen is to be able to configure the timers to arm the thermostat to reach a given set point and disarm them and therefore rules could be used. Do you mean this or something else (maybe an example can help)?
> 
> My point is pretty much the same as @3ricj
> 
> > There are also TECs that may need proportional control (not just on or off but on percentages) using PIDs. Its unlikely many will use this. Some really complicated systems use variable speed heat pumps. You see these marketed as "inverters".
> 
> I can see 2 types of the telemetry output:
> 
> * Power, in a range of 0-100%
> * The desired temperature in celsius.
> 
> For example, to control the duty cycle of some inverter one can use:
> 
> ```
> Rule
>   on tele-pid#output#duty_cycle do event pwm=(%value%*1024) endon
>   on event#pwm do  WebSend [...] Pwm %value% endon
> ```
> 
> For example, to control the boiler with PID:
> 
> ```
> Rule
>   on tele-pid#output#temperature do ot_tboiler %value% endon
> ```
> 
> In those examples telemetry JSON might look like:
> 
> ```
> {...., "pid":{"output": {"duty_cycle": 0.85, "temperature": 78}}, ....}
> ```

Understood, will evaluate for sure. Thanks for the input

PR #8235, opened 2020-04-20
BUTTON V2

### LINKED ISSUE #8018: Additional triggers for buttons in rules
Yes.

In rules there are only triggers for the following button state changes: ON, OFF, TOGGLE, HOLD. There is no trigger for double press. My goal is to react to single press, double press and hold of a button which is not possible  at the moment. By default double press will trigger the relay instead. It is possible to influence this behavior with e.g. SetOption11, but it is not possible to react to all 3 events.

Introduce a separate trigger for double press in rules. Additionally perhaps a further SwitchMode would be possible which would support all three events (single, double, hold).

The desired functionality can be achieved with scripts, but it would make the usage much easier if rules would support further triggers.

effelle:
## Description:

Changelog:
- `BUTTON1` can now control up to 5 Relays if they are present;
- Reverted changes made for inverted buttons on commit 7594108;
- Minor optimizations.

**Related issue (if applicable):** fixes #8018

## Checklist:
  - [X] The code change is tested and works on core Tasmota_core_stage

### DIFF
--- a/tasmota/support_button_v2.ino
+++ b/tasmota/support_button_v2.ino
@@ -23,7 +23,7 @@
  * Button support
 \*********************************************************************************************/
 
-#define MAX_RELAY_BUTTON1       4  // Max number of relay controlled by button1
+#define MAX_RELAY_BUTTON1       5  // Max number of relay controlled by BUTTON1
 
 const char kMultiPress[] PROGMEM =
   "|SINGLE|DOUBLE|TRIPLE|QUAD|PENTA|";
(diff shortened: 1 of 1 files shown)


timaseth:
Ciao @effelle 
This is so neat! Very well designed and thought-out feature.
I tried this on a 4ch Vanilla device. Button1, Switch2, 3 and 4.
With SetOption73 high, only Relay1 gets decoupled. Relay2, 3 and 4 continue to work well under SwitchMode.

Annotation 2020-04-20 224251

Quick Q; In the future, will it be possible to utilise this feature on, say, **Button3** (instead of Button1) of a 3 Ch device and leave the others channels in SwitchMode? This brings flexibility to those who wish not to decouple all relays but are happy to 'waste' one 'last' relay and enjoy the benefits of 6 operations from a 'Wifi Button'.
@digiblur Many thanks for compiling and making the bin available. Much appreciate!

effelle:
> With SetOption73 high, only Relay1 gets decoupled. Relay2, 3 and 4 continue to work well under SwitchMode.

A button can't intercept a switch and vice-versa. 
For now if you want a decoupled relay from a switch you must use a rule. Perhaps in the future.

> Quick Q; In the future, will it be possible to utilise this feature on, say, Button3 (instead of Button1) 

Button1 is the default on the 99% of the prebuilt device out there, that's why the special setup is on Button1. I'll think on it.

PR #8240, opened 2020-04-21
Mijia-BLE-Bridge for ESP32

Staars:
## Description:

Use internal BLE of the ESP32 to read out various BLE-sensors of the XIAOMI-universe.

## Checklist:


Staars:
some notes:

As it is not really useful to connect a HM-1x to the ESP32, this driver will use the sensor number of the HM-1x-driver (for the ESP8266). So this is `sensor62` too.

Many Thanks to @H2Zero for creating NimBLE-Arduino, which made it a lot easier to port this driver to the ESP32.

h2zero:
Glad to see it’s being put to good use. Keep an eye out for an important update coming soon.

joerg555:
Hi Stars
sag you give a quick example how to use it? 
I have tasmota32 Version from latest Development brach.
73 Jörg

Staars:
@joerg555 I tried to explain it in the DOCS as good as I could:
https://tasmota.github.io/docs/Bluetooth/#ble-sensors-using-built-in-bluetooth-on-the-esp32

Feel free to contact me, if there is something to improve.

joerg555:
Hi,

thanks for your Answer.

I have solved little problems in the BLE-lib, and made small 
optimizations, pull request tasmota development branch ?

*Question for Theo:* you have a source code formatter, may be build in 
VS-CODE ?

I have used http://astyle.sourceforge.net , but the configuration is the 
problem.

I have seen, that you are different from me with your code format, i 
will try to change it.


Regards Jörg


23.04.2020  08:02  Christian Baars:

arendst:
I'm not using a source code formatter but try to keep as strict as possible to the Google Cpp code formatting guide (by hand). 

Recently there was a PR from a user which used CPPLint to format the code based on the google Cpp guide too. I couldn't get that working on my VSC environment (the extension needs a python installed tool and python on my windows PC is a mess. It's everywhere in many versions and the tool just fails to find the correct version).

I also made numerous changes to your initial PR where I try to move code different between ESP8266 and ESP32 to one file  called ``support_esp32.ino`` without your initial ``tasmota_compat.h``. It worked fine but this way I have more control over the different implemtations.

Staars:
Hi @joerg555 ,

> I have solved little problems in the BLE-lib, and made small optimizations, pull request tasmota development branch ?

The current concept is, that we keep a copy of Arduino-NimBLE, which itself tracks NimBLE. So I think library code changes should happen there. (I hope, I understood you correctly).

Best regards
Christian

grzegorzjj:
@Staars 
> @joerg555 I tried to explain it in the DOCS as good as I could:
> https://tasmota.github.io/docs/Bluetooth/#ble-sensors-using-built-in-bluetooth-on-the-esp32
> 
> Feel free to contact me, if there is something to improve.

Could You write some steps to configure BLE beacons receiving?
What I managed: ESP32-WROOM-32D+tasmota32-sensors.bin ( 2 days ago version ), configured basics of module ( got +/-20 tasmota.bin dev at home ). Tried Sensor52 1 and Sensor52 2. No BLE beacons ( from lywsd03mmc - it works fine as I read temperature with active BLE connection from Raspberry PI ).

The main goal is to send beacon redings to domoticz. But I see no beacons...
I tried all tips from above link.

Grzegorz.

Staars:
Do you want to passively read (this would need decryption) or via active connection?
For the latter you have to comment out in the driver code:
#define USE_MI_DECRYPTION

Sensor52 is completely unrelated to the ESP32-driver.

grzegorzjj:
I would like to read passive at first - battery lives longer - but active reading is also good solution as I plan to read another lywsd03mmc that is far from Raspberry.

Staars:
For passive reading you MUST have a „bind_key“ for decryption.
IMHO this is always the superior solution and I see the direct connection only as a fallback.

Example for getting the bind_key:

https://community.home-assistant.io/t/xiaomi-passive-ble-monitor-sensor-platform/177352/101?u=magalex

grzegorzjj:
This link describes adding integration to Home Assistant. My home automation system is domoticz. Do You think it is possible to use this method ( passive redings ) with domoticz? And another question - what about active connection done via ESP32?

Staars:
Please read more carefully, the link describes how to get the bind_key (which can be used with Home Assistant too) in the context of a HA-setup. The term "bind_key" has nothing to do with Home Assistant, it is used internally by Xiaomi & Co.
Did you notice this picture with the red line drawing?


I never used domoticz, but I am pretty sure, that it would be technically possible.

"Active connection" is the fallback, when you comment out -> `// #define USE_MI_DECRYPTION` in line 39 of the driver.

grzegorzjj:
I did read this description. This subject is pretty much like a sicret. No documentation at all. I mean BLE. I understand the method of getting this key and understand its purpose. In documentation You mentioned for joerg555 ( and I read many times ) there are list of tasmota command that does not work in sensor version of tasmota32 for ESP32. I thought ( wishfull thinking ) that it will work pretty much like tasmota for NodeMCU v2 with BLE ( which I did not tested because wanted BLE on board on one chip - ESP32 ). 
As I accept active BLE read - what should be configured to start asking my lywsd03mmc ? All the command HM10Scan, HM10Period , HM10* return "Command":"Unknown".
TelePeriod <value> returns <value> , so its accepted by the module.
I would like to avoid re-compilation of tasmota/tasmota32 and thaught that reading (in active conn ) can be done by default.
I see a rule "_rule1 on System#Init do backlog NRFkey 00112233445566778899AABBCCDDEEFF112233445566; NRFkey 00112233445566778899AABBCCDDEEFF112233445566; NRFPage 6; NRFUse 0; NRFUse 4 endon_"
Do I have to find this bind_key to start collecting beacon ( only way ) or can I start active readings somehow without recompilation - with out of the box tasmota32-sensors.bin flash ?
GrzegorzJJ

Staars:
I think at the moment you should be comfortable with self-compilation. The complexity of this topic is obviously not too low and the main reason, why I still see the BLE-part as an alpha-stage is, that there is no feature freeze in sight and thus the documentation is in no way finalized. With more sensors arriving there are new problems to solve and I am still not sure about the best possible design. 
What I'm going to take away from here is that some parts of the documentation should expect less previous technical knowledge and I will try to make the onboarding easier.
As this (hidden) talk is not helpful for other users, please continue with specific questions here:
https://github.com/tasmota/docs/issues

Any answers should result in an improved documentation.

PR #8252, opened 2020-04-22
Support for setting the time in the Tuya MCU

Bettman66:
Switch on with USE_TUYA_TIME

## Description:
With the new script language, much more Tuya devices can now be integrated without changing the Tasmota source code.
What is still missing is setting the time in the Tuya MCU. This change would make this possible. 
You can activate it in User_Config. Would that be an option?

## Checklist:
  - [*] The pull request is done against the latest dev branch
  - [*] Only one feature/fix was added per PR.
  - [*] The code change is tested and works on core Tasmota_core_stage
  - [*] The code change pass travis tests. **Your PR cannot be merged unless tests pass**

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -410,6 +410,7 @@
 #define USE_SONOFF_SC                            // Add support for Sonoff Sc (+1k1 code)
 #define USE_TUYA_MCU                             // Add support for Tuya Serial MCU
   #define TUYA_DIMMER_ID       0                 // Default dimmer Id
+  #define USE_TUYA_TIME                          // Add support for Set Time in Tuya MCU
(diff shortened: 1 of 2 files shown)


s-hadinger:
Should this option be activated by default? What happens if the MCU doesn't support this command?

Bettman66:
No, this option should be deactivated by default.

ascillato:
* Why should be disabled by default?

* What is the benefit of setting the time to the secondary MCU?

Bettman66:
By setting the MCU time, the time is displayed correctly when using Tuya devices with LCD.

diabl0w:
Is there a way to trigger this? its not working for me, but i can change time manually using SerialSend5 command

diabl0w:
Opened an issue here: https://github.com/arendst/Tasmota/issues/8866

PR #8255, opened 2020-04-22
Change default PWM Frequency to 223 Hz instead of 880 Hz

s-hadinger:
##  Description:

Change the default PWM Frequency to 223 Hz. The previous was 880 Hz.

For LEDs, any frequency above 100 Hz should not create any visible flickering. 223 Hz is high enough to look smooth, while not too close to a multiple of 50 Hz or 60 Hz. And it's a prime number (which is mostly irrelevant but cool).

Lowering frequency makes the interrupt load lower, and less flickering at reboot when wifi connects.

## Checklist:

### DIFF
--- a/tasmota/CHANGELOG.md
+++ b/tasmota/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add command ``SetOption73 1`` for button decoupling and send multi-press and hold MQTT messages by Federico Leoni (#8235)
 - Add command ``SO`` as shortcut for command ``SetOption``
 - Change PWM implementation to Arduino #7231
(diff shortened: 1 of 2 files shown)


arendst:
Perhaps users need to be told that their current pwm freq will NOT be changed by this PR. Only new installation will use the new frequency.

If we are absolutely sure users NEVER changed their default frequency we could do it once if I add it to settings.ino with a version change.

s-hadinger:
Sure, but I think it needs more testing and feedback from users to make sure it does not create new issues. Then we can evaluate whether we need to change from 880 to the new value when users update to the latest version.

Issue #8011, opened 2020-03-27
CT Power on analog input


selahattinerdal:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x ] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x ] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x ] Searched the problem in the [docs](https://tasmota.github.io/docs/#/help/FAQ)
- [x ] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [ ] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [ ] Device used (e.g., Sonoff Basic): Nodemcu
- [ ] Tasmota binary firmware version number used:8.2.0
  - [ ] Pre-compiled
  - [x ] Self-compiled
    - [x ] IDE / Compiler used: _____
- [ ] Flashing tools used: _____
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```
  Configuration output here:
17:44:10 CMD: PowerSet 60.0
17:44:11 MQT: stat/tasmota_BF5881/RESULT = {"Command":"Unknown"}
17:44:28 CMD: VoltageSet 225.0
17:44:29 MQT: stat/tasmota_BF5881/RESULT = {"Command":"Unknown"}
17:44:56 CMD: CurrentSet
17:44:56 MQT: stat/tasmota_BF5881/RESULT = {"Command":"Unknown"}

```
- [ ] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```
  Rules output here:


```
- [ ] Provide the output of this command: `Status 0`:
```
  STATUS 0 output here:


```
- [ ] Provide the output of the Console log output when you experience your issue; if applicable:
  _(Please use_ `weblog 4` _for more debug information)_
```
  Console output here:


```

### TO REPRODUCE
_Steps to reproduce the behavior:_

### EXPECTED BEHAVIOUR
_A clear and concise description of what you expected to happen._

### SCREENSHOTS
_If applicable, add screenshots to help explain your problem._

[...]

PR #8273, opened 2020-04-25
Add ADC OPAmp for CT_POWER

kugelkopf123:
## Description:
Extension of CT_POWER to use it with as few adaptations as possible also for current measurement with OP amps in differential circuit for DC Voltage.

Tested on a Shelly RGBW2 which obviously uses a LM321 OpAmp at the ADC input.

Adaptation of the values according to: #8011
ANALOG_CT_FLAGS (default 0) actually reserved for possible future use;
ANALOG_CT_MULTIPLIER (default 2146) multiplier*100000 to convert raw ADC peak to peak range 0..1023 to RMS current in Amps. Value of 100000 corresponds to 1
ANALOG_CT_VOLTAGE (default 2300) to convert current in Amps to apparent power in Watts using voltage in Volts*10. Value of 2200 corresponds to 220V

The first parameter is used to set the adcLOW value.
This is necessary because the OPAmp already has voltage at its output even if no current is consumed. So this sets the base.
For this ShellyRGBW2 the Values for 12V are: 
`adcparam 7, 406, 3282, 0.012`

It is my first PullRequest for Tasmota. Hope to have done everything correctly.

## Checklist:
  - [x] The code change is tested and works on core Tasmota_core
  - [ ] The code change is tested and works on core ESP32


ascillato:
Hi, please do not modify **_platformio_tasmota_env.ini_** in the same PR of **_Add ADC OPAmp for CT_POWER_**

If you want to provide other envs for platformio, do it in another PR. Mixing features in the same PR is not a good option.

Thanks.

kugelkopf123:
Oh sh**! How do I reverse this? Probably send in the P.R. again, right?

PR #8352, opened 2020-05-05
Webcam update

gemu2015:
## Description:

add face detector 

## Checklist:
  - [x] The code change is tested and works on core ESP8266 V.2.7.0

### DIFF
--- a/tasmota/xdrv_10_scripter.ino
+++ b/tasmota/xdrv_10_scripter.ino
@@ -73,7 +73,7 @@ uint32_t DecodeLightId(uint32_t hue_id);
 
 
 void SaveFile(const char *name,const uint8_t *buf,uint32_t len) {
-  File file = SPIFFS.open(name,FILE_WRITE);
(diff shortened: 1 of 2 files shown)


arendst:
Thx.

What tool do you use that adds and removes spaces at random code lines. Especially around = and arithmetic signs and even line indents. I try to use google cpp code book to keep the code readable and consistent. Having spaces around these items adds to readability.

Could you be so kind to have a look at readability and consistenty too?

gemu2015:
which tool do you recommend ? (i am on macOS)
i tried online tools and was not satisfied at all.

some formatting differences result from hand merging.
because we both worked at the same source i could not automatically merge sometimes

arendst:
I understand.

I just use VSC with extension ``Trailing Spaces`` and tried ``cpplint`` but that needs external python tools which is an issue on Windows where there are python version everywhere. It doesn't work for me so it's all by hand.

gemu2015:
ok, will be more thoroughly next time

PR #8365, opened 2020-05-06
Reduce code size of SettingsDefaultSet2()

s-hadinger:
## Description:

Small trick to decrease the size of `SettingsDefaultSet2()` by 400 bytes.

Instead of changing `Settings.flag[1-4]` bits one by one, I moved to temporary variables `flag[1-4]` and do the Settings assignment at the end. This allows GCC to optimize code and group the changes, since it knows their are no side effects.

Also, change the assignment to a bit-OR. We initialize all values to zero with `memset()` and we only need to OR the non-zero values. This removes a lot of code. For example flag3 and flag4 are all zeroes with the default `my_user_config.h`

Code size impact: **-400 bytes of Flash**

## Checklist:
  - [x] The code change is tested and works on core ESP8266 V.2.7.0

### DIFF
--- a/tasmota/settings.ino
+++ b/tasmota/settings.ino
@@ -715,6 +715,12 @@ void SettingsDefaultSet2(void)
 {
   memset((char*)&Settings +16, 0x00, sizeof(Settings) -16);
 
+  // this little trick allows GCC to optimize the assignment by grouping values and doing only ORs
+  SysBitfield   flag = { 0 };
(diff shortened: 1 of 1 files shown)


arendst:
My, my, my you're busy shrinking :+1:

s-hadinger:
I spotted this BIG function some time ago, which is a shame for code that is almost never run. It's smaller now.

PR #8397, opened 2020-05-08
Add automatic compression of Rules to achieve ~60% compression

s-hadinger:
##  Description:

Adds automatic compression of Rules when they exceed the 511 characters limit. The compression algorithm uses a highly optimized and customized version of Unishox compressor. Typical rules from the cookbook compress more than -60%, effectively extending the total sizes for Rules from 1533 bytes to more than 3500 characters.

Warning: compressed rules will be cleared if you downgrade to a firmware not supporting compressed rules.

With the standard buffer sizes, you can't enter directly a Rule of more than 1000 bytes. You need to use the append mode `Rule<x> +<rule>`

To enable Rules compression, use `#define USE_RULES_COMPRESSION` is `my_user_config.h` or `user_config_override.h`. This takes ~3.8KB.

Rules compression adds `SetOption93 1`. When rules are compressed, running the rule needs to uncompress in memory. By default, Tasmota keeps an uncompressed copy of the rule in RAM to avoid the CPU overload of uncompressing rules at each tick. However, if you run low in RAM, you can set `SetOption93 1` to avoid keeping a copy of Rules, hence trading memory for CPU.

Example:

```
Rule1 ON Switch1#State==1 DO Add1 1 ENDON ON Var1#State==0 DO ShutterStop1 ENDON ON Var1#State==1 DO ShutterClose1 ENDON ON Var1#State>=2 DO Var1 0 ENDON ON Shutter1#Close DO Var1 0 ENDON ON Switch2#State==1 DO Add2 1 ENDON ON Var2#State==0 DO ShutterStop1 ENDON ON Var2#State==1 DO ShutterOpen1 ENDON ON Var2#State>=2 DO Var2 0 ENDON ON Shutter1#Open DO Var2 0 ENDON

xx:xx:xx RUL: Stored uncompressed, would compress from 361 to 141 (-61%)

Rule1 +on event#chkSun do backlog var1 0; event chkSunrise=%time%; event chkSunset=%time%; event chkmorn=%time%; event chknight=%time%; event setPower endon   on event#chkSunrise<%sunrise% do var1 1 endon   on event#chkSunset>=%sunset% do var1 1 endon   on event#chkmorn<%mem1% do var1 0 endon   on event#chknight>=%mem2% do var1 0 endon   on event#setPower do power1 %var1% endon

xx:xx:xx RUL: Compressed from 735 to 289 (-61%)

Rule1 +on event#doorbell do var1 %value% endon on switch4#state=1 do publish domoticz/in {"idx":11,"nvalue":1} endon on switch4#state=1 do power1 %var1% endon on switch4#state=0 do publish domoticz/in {"idx":11,"nvalue":0} endon on switch4#state=0 do power1 0 endon

xx:xx:xx RUL: Compressed from 994 to 385 (-61%)

Rule1 +on Time#Initialized do event chkSun endon   on Time#Minute=%sunset% do event chkSun endon   on Time#Minute=%mem2% do event chkSun endon   on Time#Minute=%sunrise% do event chkSun endon   on Time#Minute=%mem1% do event chkSun endon

xx:xx:xx RUL: Compressed from 1225 to 440 (-64%)
```

## Checklist:


meingraham:
After one enters a rule and appends to the rule set over time... when one executes `Rule1` to view the rule set, as all the rules decompressed for viewing?

To clarify, I can enter an initial rule set of 1000 bytes or less as a single "load"? And you can continue to append until the 511 byte rule set fills?

If a compressed append causes the 511 allocation to be exceeded, is an error returned? I seem to recall that currently there is no error... it just doesn't update the rule set and the user has to fully inspect the rule set to confirm whether a rule set was successfully updated.

s-hadinger:
> After one enters a rule and appends to the rule set over time... when one executes `Rule1` to view the rule set, as all the rules decompressed for viewing?

Yes, but by default the rules are kept in RAM, it does not consume CPU. The buffer is currently limited to 1200 char, if the rule is longer, it will be truncated and you won't be able to see the end of it.

> To clarify, I can enter an initial rule set of 1000 bytes or less as a single "load"? And you can continue to append until the 511 byte rule set fills?

Yes you can enter 1000 bytes with a single Rule command, and then append until the storage space of 511 bytes if full.

> If a compressed append causes the 511 allocation to be exceeded, is an error returned? I seem to recall that currently there is no error... it just doesn't update the rule set and the user has to fully inspect the rule set to confirm whether a rule set was successfully updated.

Yes if the total size does not fit, an error is returned and the rule is left unchanged.

```
xx:xx:xx RUL: not enough space
```

PR #8445, opened 2020-05-14
SDM630 import export total energy

jkostorz:
## Description:

## Checklist:

### DIFF
--- a/tasmota/xdrv_03_energy.ino
+++ b/tasmota/xdrv_03_energy.ino
@@ -104,11 +104,9 @@ struct ENERGY {
   bool type_dc = false;
   bool power_on = true;
 
-#ifdef USE_SDM630
   float imp_active[3] = { NAN, NAN, NAN };      // kWh import active energy
   float exp_active[3] = { NAN, NAN, NAN };      // kWh export active energy
   float tot_active[3] = { NAN, NAN, NAN };      // kWh total active energy
-#endif
 
 #ifdef USE_ENERGY_MARGIN_DETECTION
   uint16_t power_history[3] = { 0 };
--- a/tasmota/xdrv_03_energy.ino
+++ b/tasmota/xdrv_03_energy.ino
@@ -104,6 +104,12 @@ struct ENERGY {
   bool type_dc = false;
   bool power_on = true;
 
+#ifdef USE_SDM630
+  float imp_active[3] = { NAN, NAN, NAN };      // kWh import active energy
+  float exp_active[3] = { NAN, NAN, NAN };      // kWh export active energy
+  float tot_active[3] = { NAN, NAN, NAN };      // kWh total active energy
+#endif
+
 #ifdef USE_ENERGY_MARGIN_DETECTION
   uint16_t power_history[3] = { 0 };
   uint8_t power_steady_counter = 8;  // Allow for power on stabilization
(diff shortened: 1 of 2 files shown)


arendst:
I'm weighing the pros and cons with regards to functionality and  code size. 

Implementing this will increase code size for all energy monitoring devices while most are not using these parameters. See SDM120 where a local fix was introduced. Also adding these modbus registers to read adds another 2.5 seconds before the first register is read again. So the total cycle will last 25 * 250mSec = 6.25 seconds which may result on strange values between the first and last register. It also maybe no problem at all.

Extracting the code from energy.ino will introduce redundancies which also not really wanted...

Let me think about it for a while.

kugelkopf123:
How about using the SML Driver that still exist for this?

jkostorz:
Maybe implement only part of this patch for import and export energy? Total energy we can calculate from import and export on computer if needed. This save 750mSec and 16bytes of RAM and... 814bytes of flash.

jkostorz:
While i moved all [code](https://github.com/jkostorz/Tasmota/blob/jkostorz-patch-1/tasmota/xnrg_10_sdm630.ino) for import and export energy as is in sdm120 to one file only, without touching xdrv_03_energy.ino, flash size have more 882 bytes (patch from PR -> test code).

arendst:
Now partly implemented (Only Export Active) with some re-writes in ``xdrv_03_energy.ino`` and additions to ``xnrg_10_sdm630.ino`` as you suggested.

Pls give it a try as I don't own a SDM630.

effelle:
Theo both will be in `/SENSOR` topic? 
Just checking if Discovery will grab them fine because will be a long topic to parse.

arendst:
Yes. As said, I don't have a three phase SDM630 but find below the new layout for a single phase SDM120:
```
17:53:46 MQT: tele/rf1/SENSOR = {"Time":"2020-05-22T17:53:46","ENERGY":{"TotalStartTime":"2018-11-23T15:33:47","Total":0.017,"TotalTariff":[0.000,0.017],"Yesterday":0.000,"Today":0.002,"ExportActive":0.000,"ExportActiveTariff":[0.000,0.000],"Period":0.00,"Power":0.00,"ApparentPower":7.84,"ReactivePower":-7.21,"Factor":0.39,"Frequency":50.0,"Voltage":234.31,"Current":0.039,"ImportActive":12.580,"ImportReactive":0.002,"ExportReactive":39.131,"PhaseAngle":290.45}}
```
I deliberatly changed the version number to 8.3.1.2 so you can check it with regards to layout change.

effelle:
The only issue is the name `ExportActiveTariff`, too long and rendered as `ExportActiveTar` . 
This lead to just one entity since the second one overwrite the first one. 
Do you want to shorten the name or do you prefer a bigger size for `unique_id` field? 
Actually is set to 29+1 and I will need at least 34+1.
tarif

arendst:
I'll rename to ``ExportTariff``

effelle:
OK, the parser will create the fields with no issue with the shorter name.

arendst:
Thx for the quick response!

jkostorz:
With my SDM630 work ok. Log from console...

<code>
18:00:15 MQT: tele/home/meter/power/SENSOR = {"Time":"2020-05-22T18:00:15","ENERGY":{"TotalStartTime":"2020-05-22T15:12:47","Total":118.560,"Yesterday":13.356,"Today":7.917,"ExportActive":[0.000,0.000,0.000],"Period":3,"Power":[213,0,7],"ApparentPower":[214,37,11],"ReactivePower":[-7,-37,-8],"Factor":[1.00,0.10,0.58],"Frequency":[49.91,0.00,0.00],"Voltage":[241.2,238.5,238.5],"Current":[0.89,0.16,0.05]}}
</code>

I do not export energy to my operator (but planed) - export energy is reported as 0.
Since the frequency on the three phases is the same, this value can be reported in one field but it requires corrections in all files from energy meters.

arendst:
Thx. I will revisit the Frequency.

arendst:
Update frequency display just now.

jkostorz:
Thx

PR #8448, opened 2020-05-14
Thermostat: Small bugfix and PI autotune feature added (disabled by define, as experimental and un…

arijav:
…tested)

## Description:

## Checklist:


prahjister:
Hello. I would like to test this out with sous vide cooking. Can you explain a little how to use.

arijav:
Hello I have not yet had the time to test the autotune myself, but I have already seen a bug therefore I am sure it will not work yet.

If you want anyway to check the behavior you would have to compile following the latest point in the thermostat documentation:
https://tasmota.github.io/docs/Thermostat/

You can enable the PI autotune by sending the following command (not yet documented in commands):
ControllerModeSet 3

You can play with the performance levels using this command (not yet documented either):
CmndPerfLevelAutotune value -> Value would be 0 (fast), 1 (normal), 2 (slow)

The behavior should be the follow (but again it is not there yet). The controller would work normally until it reaches the Setpoint for heating f.i. with positive gradient. Once there the autotune feature activates and actuates a PWM with a defined duty cycle. This will make the temperature to oscillate around a certain value and based on the oscillations (peaks) the parameters for the autotune can be calculated.

PR #8460, opened 2020-05-16
Allow DS18x20 sensor in lite version

nagyrobi:
## Description:

Allow DS18x20 sensor in lite version. It's so small. On many devices it's enough to use the lite version, with DS18x20 enabled it's more universal.

## Checklist:

### DIFF
--- a/tasmota/tasmota_configurations.h
+++ b/tasmota/tasmota_configurations.h
@@ -459,7 +459,7 @@
 
 #undef USE_COUNTER                               // Disable counters
 #define USE_ADC_VCC                              // Display Vcc in Power status. Disable for use as Analog input on selected devices
-#undef USE_DS18x20                               // Disable DS18x20 sensor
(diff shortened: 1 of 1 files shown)


ascillato:
The lite version is low energy consuming version intended for devices with a bad quality power regulator. Adding this will make Tasmota to use more energy and make those devices to fail. 

Sorry.

nagyrobi:
> If needed that sensor, Tasmota sensors bin should be used.

No problem. 
I thought it would be useful for others too with Sonoff Basics which are notoriously bad with standard builds (lock up randomly after a few days), but run the lite version very well, and they are extended with just a DS18x20 sensor.
I use a couple of those with the lite version like that and they run rock stable for months.

I'll keep using my own builds of lite with just DS18x20 enabled for these...

eku:
@nagyrobi you can always compile your own customized image.

PR #8486, opened 2020-05-19
scripter update

gemu2015:
script compression option (thanks to @s-hadinger )
more google charts (histogram, tables, gauges)
smal bug fixes

## Description:

## Checklist:

### DIFF
--- a/tasmota/xdrv_10_scripter.ino
+++ b/tasmota/xdrv_10_scripter.ino
@@ -66,8 +66,9 @@ keywords if then else endif, or, and are better readable for beginners (others m
 uint32_t EncodeLightId(uint8_t relay_id);
 uint32_t DecodeLightId(uint32_t hue_id);
 
-#if defined(ESP32) && defined(ESP32_SCRIPT_SIZE) && !defined(USE_24C256) && !defined(USE_SCRIPT_FATFS)
+#include <unishox.h>
 
+#if defined(ESP32) && defined(ESP32_SCRIPT_SIZE) && !defined(USE_24C256) && !defined(USE_SCRIPT_FATFS)
 #include "FS.h"
 #include "SPIFFS.h"
 
@@ -3035,7 +3036,7 @@ int16_t Run_Scripter(const char *type, int8_t tlen, char *js) {
                 uint8_t index=glob_script_mem.type[ind.index].index;
                 if ((vtype&STYPE)==0) {
                     // numeric result
-                  if (glob_script_mem.type[index].bits.is_filter) {
+                  if (glob_script_mem.type[ind.index].bits.is_filter) {
                     uint8_t len=0;
                     float *fa=Get_MFAddr(index,&len);
                     //Serial.printf(">> 2 %d\n",(uint32_t)*fa);
@@ -3965,6 +3966,25 @@ void ScriptSaveSettings(void) {
     glob_script_mem.script_mem_size=0;
   }
 
+
+#ifndef UNISHOXRSIZE
+#define UNISHOXRSIZE 2560
+#endif
+#ifdef USE_RULES_COMPRESSION
+#ifndef USE_24C256
+#ifndef USE_SCRIPT_FATFS
+#ifndef ESP32_SCRIPT_SIZE
+  uint32_t len_compressed = unishox_compress(glob_script_mem.script_ram, strlen(glob_script_mem.script_ram), Settings.rules[0], UNISHOXRSIZE);
+  if (len_compressed > 0) {
+    AddLog_P2(LOG_LEVEL_INFO,PSTR("compressed to %d"),len_compressed * 100 / strlen(glob_script_mem.script_ram));
+  } else {
+    AddLog_P2(LOG_LEVEL_INFO, PSTR("script compress error: %d"), len_compressed);
+  }
+#endif
+#endif
+#endif
+#endif // USE_RULES_COMPRESSION
+
   if (bitRead(Settings.rule_enabled, 0)) {
     int16_t res=Init_Scripter();
     if (res) {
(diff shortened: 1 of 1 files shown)


s-hadinger:
Nice. What compression ratio do you find on scripts?

gemu2015:
not more then about 40 Percent

gemu2015:
@s-hadinger 
i now updated to your optimized class version and it no longer works. the decompressor crashes with heap corruption. i reinstalled your previous version (renamed to old_ ...) and when calling the "old" version everything works again.

i use an extremely simple approach , just allocating a fixed buffer at start and decompressing to this buffer, and on changing script source compressing to rules buffer.

you may test with my fork. it contains both versions.

s-hadinger:
@gemu2015 Ouch. Let me have a look. Where is your branch/commit with the new class-based Unishox?

gemu2015:
https://github.com/gemu2015/Sonoff-Tasmota  (universal8)

s-hadinger:
Here are some suggestions, yet not a reason for memory corruption:

- Don't encode the last NULL of the string:

```uint32_t len_compressed = SCRIPT_COMPRESS(glob_script_mem.script_ram, strlen(glob_script_mem.script_ram)+1, Settings.rules[0], MAX_SCRIPT_SIZE);```

Why `strlen(glob_script_mem.script_ram)+1` and not `strlen(glob_script_mem.script_ram)`. You should try to avoid including the last NULL char because it's takes 2 bytes to encode and it's mostly untested.

I suggest you also set the output limit to `MAX_SCRIPT_SIZE-1` to leave space for the NULL terminator. If the buffer is full, strlen() will go crazy. Then force a NULL terminator.

My proposed modification:
```
uint32_t len_compressed = SCRIPT_COMPRESS(glob_script_mem.script_ram, strlen(glob_script_mem.script_ram), Settings.rules[0], MAX_SCRIPT_SIZE-1);
Settings.rules[0][len_compressed] = 0;  // string is NULL terminated
```

- (minor) you are allocating a fixed size UNISHOXRSIZE 2560 buffer for the uncompressed scripts in memory. This can be a waste since most of the time you won't need so large scripts, and it can happen that the script is greater than 2560 (unless you want a hard limit here).
In Rule, I'm storing in the first byte an "indication" of the size: first byte multiplied by 8, which limits each rule to 2040 bytes. You could multiply by 16 with a hard limit of 255*16= 4080 bytes, and allocate to the nearest 16-bytes chunk.

s-hadinger:
I found the bug. I will do the PR soon.

gemu2015:
thanks!

gemu2015:
according to  allocating full buffer size. i intentionally allocate this space at start of program to avoid memory fragmentation. If editing a script reallocating the buffer causes fragmentation. i run into this problem in older versions. 
i agree when not editing however it would save memory.

s-hadinger:
Here you are. The stop condition of the input buffer was broken, it was by luck that it never crashed with rules.

gemu2015:
@s-hadinger 
a remark to strlen(in)+1 on compress:
when i just take strlen(in) i get unwanted characters at the end of string on decompress, despite 
of terminating the compressed string with out[comp]=0 and
of terminating the decompressed string with out[decomp]=0

@arendst 
if would prefer to not have USE_RULES_COMPRESSION default on with scripts.
people needing larger scripts may easily switch this on in user_config

s-hadinger:
@gemu2015 Yes, the unwanted characters at the end of the string on decompress was a side effect of the bug. It should be fixed now.

I suggest you set your own `#define USE_SCRIPT_COMPRESSION` flag, the USE_RULES_COMPRESSION is now default for rules.

Also, I noticed there is a big difference between rules and scripts, as scripts use a lot of LF. They are encoded with 9 bits, and I could get it down to 3 bits at the expense of unicode characters. This would give you 5% more compression ratio. If you're interested I can do a customized version of the compressor.

gemu2015:
@s-hadinger 
according Linefeed:  don't know the impact on whole scripts but anyhow the more compression the better.

off topic question regarding optimized math.
the fastprecisepow(x y) function is not really precise as the name states!
fastprecisepow(2,2) return 3,88 which i would not tell precise.
a scripter user claimed about this as it introduced errors into his calculations
as this function is also used at various other places one should doubt about precision.

arendst:
Regarding pow. It's all about code size! It saves 5k being a bit less precise. Remember, tasmota is no calculator.

gemu2015:
i know its no calculator what bothers me is the functions name "precise" which it isn't,

s-hadinger:
The full explanation is here: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.9.4508&rep=rep1&type=pdf

2^2 ~= 3.88 is 3% error which is within expected range for this fast implementation. When used for gamma correction, 3% error is not noticeable.

s-hadinger:
Hmmm. My change for better LF encoding only brings 2%, which is not worth it.

I noticed an issue though. If you enter a rule without compression enabled, then enable compression, it will try to decompress raw text and will give you garbage. I have noticed a crash of `Run_Scripter()` when the script is full of garbage, and it will bootloop forever. The device is bricked. I suggest we provide a better path. Either detect that the script was not compressed, i.e. with the first two bytes like in rules, or clear the script in case of a bootloop.

gemu2015:
ok, thanks for trying,

all scripts must start >D
i implemented a filter for that sequence
 should no longer crash with invalid scripts

s-hadinger:
I found the bug that would put garbage characters at the end of the decompressed script. Everything should be fine now.

gemu2015:
@s-hadinger 
sorry to say unishox still buggy

// original script
[testscript.txt](https://github.com/arendst/Tasmota/files/4677516/testscript.txt)
// decoded script
[decode.txt](https://github.com/arendst/Tasmota/files/4677525/decode.txt)

s-hadinger:
Thanks. Investigating.

s-hadinger:
@gemu2015 It is fixed now. I also fixed a hole in my testing patterns. 4 uppercase letters (including 'J') require 13 bits of encoding and was wrongly encoded.

The bug was in the compressor, you need to re-enter the rule.

PR #8491, opened 2020-05-20
Ledpowerpwm

grob6000:
## Description:

Provides PWM control of status/power LEDs. Useful for when your lightswitch has touch buttons with leds, which indicate status, but also need to be found at night. With this feature, can set a pwm value for off and on state of power leds. A mode setting allows masking pwm mode for the status leds - in case you have a device with some button leds, and other non-button leds.

An alternative would be to configure the button LEDs as pwm lights, but this breaks all the default power led functions (no relay power state, etc) which then have to be added as rules.

I've used the existing analogWrite wrappers, and it respects the analogRange settings etc. used for LED lights.

Link led is not fiddled with too much - but in the legacy case where LEDLINK is not defined and LED1 is, LED1 is used as the link led, there is some interference. I've chosen for this case to respect the new pwm settings for LED1 for the most intuitive behaviour.

Defaults are set for identical function. This doesn't change anything unless the ledpwm_mode setting is changed by the user.

Developed this feature for my own use, but reasonably clean and proposing it for inclusion if others see it as handy.

Tested reasonably thoroughly on ESP8266 & a device using one, pending basic testing on a generic ESP32 (have one in a drawer somewhere).

Change summary:
* Adds settings: ledpwm_on, ledpwm_off, ledpwm_mode
* Adds commands: cmnd_setledpwmon, cmnd_setledpwmoff, cmnd_setledpwmmode
* Changes setledpoweridx to do an analogwrite instead of digitalwrite as required
* Changes setlinkled to use setledpoweridx when in the LED1 mode
* Adds a helper function UpdateLedPowerAll - sets the led i/o states to respect the led_power memory variable. Used for when user is updating the new settings, so the pwm changes propogate and they can see the changes take effect immediately.

## Checklist:
  - [X] The code change is tested and works on core ESP8266 V.2.7.1


tasmota/settings.ino
```diff
@@ -1055,6 +1055,11 @@ void SettingsDefaultSet2(void)
   Settings.flag2 = flag2;
   Settings.flag3 = flag3;
   Settings.flag4 = flag4;
+  
+  // Led PWM
+  Settings.ledpwm_off = 0;
+  Settings.ledpwm_on = 1023;
```
s-hadinger, line 1061: ledpwm_on = 255

With 8 bits ranges

tasmota/settings.ino
```diff
@@ -1411,6 +1416,13 @@ void SettingsDelta(void)
       if (Settings.rules[1][0] == 0) { Settings.rules[1][1] = 0; }
       if (Settings.rules[2][0] == 0) { Settings.rules[2][1] = 0; }
     }
+	
+	// ledpwm
+    if (Settings.version < 0x08030001) {
+      Settings.ledpwm_off = 0;
+      Settings.ledpwm_on = 1023;
```
s-hadinger, line 1423: Same 255

tasmota/support_command.ino
```diff
@@ -1890,3 +1890,57 @@ void CmndDriver(void)
...
+  ResponseCmndNumber(Settings.ledpwm_off);
+}
+
+void CmndSetLedPwmOn(void)
+{
+  if (XdrvMailbox.data_len > 0) {
+    if (XdrvMailbox.payload < 0) {
+	  Settings.ledpwm_on = 0;
+    } else if (XdrvMailbox.payload > Settings.pwm_range) {
+	  Settings.ledpwm_on = Settings.pwm_range;
```
s-hadinger, line 1915: 255 max

tasmota/support_tasmota.ino
```diff
@@ -351,7 +358,25 @@ void SetLedPowerIdx(uint32_t led, uint32_t state)
     } else {
       led_power &= (0xFF ^ mask);
     }
-    DigitalWrite(GPIO_LED1, led, bitRead(led_inverted, led) ? !state : state);
+	uint16_t led_pwm_set = 0;
+	if (bitRead(Settings.ledpwm_mask, led)) {
+		if (bitRead(led_inverted, led)) {
+		  if (state) {
```
s-hadinger, line 364: Use `ledGamma10()` to turn 0..255 range into 0..1023 with gamma correction applied.

tasmota/support_tasmota.ino
```diff
@@ -351,7 +358,25 @@ void SetLedPowerIdx(uint32_t led, uint32_t state)
...
+			led_pwm_set = Settings.pwm_range - Settings.ledpwm_off;
+		  }
+		} else {
+		  if (state) {
+			led_pwm_set = Settings.ledpwm_on;
+		  } else {
+			led_pwm_set = Settings.ledpwm_off;
+		  }
+		}
+		analogWrite(Pin(GPIO_LED1, led), led_pwm_set);
```
s-hadinger, line 376: You can mimic the full version of `xdrv_04_light.ino`:

```
cur_col = cur_col > 0 ? changeUIntScale(cur_col, 0, Settings.pwm_range, Light.pwm_min, Light.pwm_max) : 0;   // shrink to the range of pwm_min..pwm_max
analogWrite(Pin(GPIO_PWM1, i), bitRead(pwm_inverted, i) ? Settings.pwm_range - cur_col : cur_col);
```

with full support of range and pwm_min/pwm_max

s-hadinger (commented):
Very interesting feature, I love it.

I suggest to put ranges in 0..255 for consistency and apply Gamma Correction to get full range at low brightness and make it more natural.

arendst:
I'll wait for @s-hadinger suggested changes before merging

grob6000:
Having a go today. Thanks for the suggestions, I didn’t know there was an existing fn for gamma correction, neat!

arendst:
Thx. As tasmota is all about less code in tiny places I would suggest you reconsider the commands code too. All three commands will need to be reduced to just one being LedPwm as there is too many code duplication present.

I suggest to take a look at other command implementations having On/Off functionality.

Thx in advance.

arendst:
Ah. I will do the change regarding the command structure.

Just sit back and let me handle it for now.

grob6000:
Thanks, great news!

garci66:
Hi! Do you think it would be possible to expose the pwmled in homeassistant so that automating brightness changes is easier? (for example, tied them to a motion sensor?)

grob6000:
I can have a look. Are you using MQTT.publish as a workaround currently?

garci66:
from HA? yes, i am using the following:
```
data:
  payload_template: '{{states(''input_number.switch_led_brightness'')|int}}'
  topic: cmnd/tasmotas/LedPwmOff
service: mqtt.publish
```

inside an automation where I have a numeric input that I use to dim all my tasmota switchs.

Previously, I had defined the LED GPIO as PWM in the template thus I had dimmable backlight (for the sonoff touch) but I lost the wifi/mqtt driven activity flashing. Im now using this in one of the switches, but having the dimmer slider appear automagically in HA is much nice for noob users like me, and allows you to write automation using the HA gui without going crazy about the mqtt topics, etc..

basically the backlight can be advertised as an MQTT light with a `command_topic` as `cmnd/%topic%/LedPwmMode` and the `brightness_command_topic` as `cmnd/%topic%/LedPwmOn` (or `LedPwmOff`)

I guess we would need a flag to select which of the two (LedPwmOn or LedPwmOff) to use in the autodiscovery.  (In my case I want the LED on all the time as it makes the buttons easy to see in the dark but want to dim the max value when going to sleep cause otherwise its too bright). And in my particular case (I need to test further), I think using `Led1i` in the template didnt work properly, while using `Led1` and using `LedPwmOff 128` (and `LedPwmOn 0`) was enough to get what I wanted.

PR #8525, opened 2020-05-24
add private pow function in VEML7700 lib for future use

device111:
alternate pow  function in VEML7700 libary for future use in Tasmota to calculate normalized Lux and normalized white values.

## Description:

## Checklist:

### DIFF
--- a/lib/Adafruit_VEML7700/Adafruit_VEML7700.cpp
+++ b/lib/Adafruit_VEML7700/Adafruit_VEML7700.cpp
@@ -76,6 +76,29 @@ boolean Adafruit_VEML7700::begin(TwoWire *theWire) {
   return true;
 }
 
+float Adafruit_VEML7700::alternate_pow(float a, float b)
+{
+  // https://martin.ankerl.com/2012/01/25/optimized-approximative-pow-in-c-and-cpp/
+  // calculate approximation with fraction of the exponent
+  int e = abs((int)b);
+  union {
+    double d;
+    int x[2];
+  } u = { a };
+  u.x[1] = (int)((b - e) * (u.x[1] - 1072632447) + 1072632447);
+  u.x[0] = 0;
+  // exponentiation by squaring with the exponent's integer part
+  // double r = u.d makes everything much slower, not sure why
+  double r = 1.0;
+  while (e) {
+    if (e & 1) {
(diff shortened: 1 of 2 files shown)


s-hadinger:
@device111 It's the same implementation as `FastPrecisePow()` in `suppor_float.ino`. Could you please use this instead to avoid duplication of code?

arendst:
I think the problem is accessing the function from the library won't work.

device111:
yes, the problem is calling from the libary. My first thought was to delete the functions from lib and implement the affected pow functions in the driver, but that would break the idea of libarys.

device111:
@Theo
is it planed in the future, to transfer the related (pow, log, ..) functions in `support_float.ino` to a c++ lib?

s-hadinger:
Actually it works if you add the following in 'Adafruit_VEML7700.h`:

Before the class definitio:
```
extern float FastPrecisePowf(const float x, const float y);
```

Then replace the definition of the function with:
```
  static inline float alternate_pow(float a, float b) { return FastPrecisePowf(a, b); }
```

You can then completely remove the `Adafruit_VEML7700::alternate_pow()` implementation.

s-hadinger:
Anyways, I'm afraid the function `alternate_pow()` is actually not used, since Tasmota code never calls `readLuxNormalized()` nor `readWhiteNormalized()`.

I checked in the map file and it's not there. This optimization was unfortunately useless.

device111:
yes, the pow changes are only made for the future developement of the driver.
but the code from above are good. I will change it.
thanks!

s-hadinger:
By the way, if it does not hurt the precision, replace all floating points literal with floats. Otherwise calculation is done by the compiler using `double` and code goes huge.

For ex, replace 
```
    lux = 6.0135e-13*alternate_pow(lux,4) -  9.3924e-9*alternate_pow(lux,3) + 8.1488e-5*alternate_pow(lux,2) + 1.0023*lux;
```

with
```
    lux = 6.0135e-13f*alternate_pow(lux,4) -  9.3924e-9f*alternate_pow(lux,3) + 8.1488e-5f*alternate_pow(lux,2) + 1.0023f*lux;
```

device111:
ok, in the next pr i will change it.
thanks a lot!

PR #8581, opened 2020-05-30
Command interface for touch button driver and a serial plotter

Staars:
## Description:

Add commands for the touch button functionality on the ESP32.

TouchCal x - x=button 1 .. 4. This plots the sensor values to the console, to get information regarding the setting of the 2 following commands

TouchThres x - sets threshold to x, which will get smaller by touching the pin. In the driver a sensor value beneath that threshold, will trigger a button touch.

TouchNum x - sets number of ignored measurements below the threshold, because there will likely be spikes. The default value of 3 is very conservative and 1 should be fine most of the time. A higher value is safer in a noisy environment, but for obvious reasons you will have to touch the pin (or cable ...) longer to trigger the button press.

Serial plotter:
A little tool to visualize sensor data in real-time written in python using matplotlib. More info later.

## Checklist:

### DIFF
--- a/pio/serial-plotter.py
+++ b/pio/serial-plotter.py
@@ -1,0 +1,188 @@
+#!/usr/bin/env python
+
+"""
+  serial-plotter.py - for Tasmota
+
+  Copyright (C) 2020  Christian Baars
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


Staars:
plotter

Needs matplotlib (including numpy).

This little tool should be helpful to get a feel for the touch values.
Usage: 
./serial-plotter.py --port /dev/XXX --baud 115200

You can send commands to Tasmota via SEND-box.
Example for the touch button driver:
TouchCal 255 - turns on calibration mode for all buttons

Could be used for other drivers too by sending data to the log in the format:
PLOT: number ,  value
number will be graph 1..4
value is a integer value

In code it is something like:
AddLog_P2(LOG_LEVEL_INFO, PSTR("PLOT: %u, %u "), number, value);

PR #8592, opened 2020-06-01
fix start without stop condition

device111:
## Description:
If a driver call the function `I2cValidRead` while in init phase and the I2C device ist not present,
the `I2cValidRead` function makes a Start without Stop condition after 3 attempts and block the bus.
this fix makes a stop condition after 3 attempts.

background:
A view driver call only `I2cActive` at init phase and afterwards reads data with the  `I2cValidRead` function but the device is unavailable.

Normaly we must check both conditions.
1. Is address aktiv?
2. Is address available?

with `I2cActive` we only check the activity of device but not the physically presents.

In the second step, i will fix the affected driver with the function `I2cSetDevice` seperate PR (xsns_65_hdc1080.ino, xsns_67_as3935.ino, xsns_70_veml6075.ino )

## Checklist:

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -1459,6 +1459,7 @@ bool I2cValidRead(uint8_t addr, uint8_t reg, uint8_t size)
     }
     retry--;
   }
+  if (!retry) Wire.endTransmission();
   return status;
 }
 


device111:
@arendst 
should we change all driver from `I2cActive` to `I2cSetDevice`?

arendst:
No. I need to investigate as I just changed them a few months ago for some reason....

device111:
ok, thanks.

arendst:
Why would you want to change all I2cActive to I2cSetDevice?

The reason I do it this way is that I2cActive is NOT invasive to the device I2c protocol; some devices need a different startup then just performing the I2cSetDevice code.

device111:
Hello Theo,

so I2CActive checks only the device-addresses that are registerd in Tasmota.
If we init a new device, we should seek, if the address is responsible.
this is verry similar to I2Cscan. We open and close the connection and have a result if the address is responsible.

Wire.beginTransmission(address)
Wire.endTransmission()

 I2cSetDevice do both things, look if the address is already registerd in Tasmota and look if the address is responsible.

I know, a view devices need sometimes 2 start conditions to work, but wire.beginTrasnmission and Wire.endTransmission should work for all devices. (I'm not sure!)

some devices, i.e. the hdc1080 looks only if the address is registerd in tasmota. If not, the driver begin to read out some registers, but have not checked if the device is realy responsible.
here:
hdc_manufacturer_id = HdcReadManufacturerId();
hdc_device_id = HdcReadDeviceId();

I think, we do not have to read the registers since the device is not available anyway.

I'dont know, whats the best solution..
best regards,
Martin

arendst:
I suggest to leave it as it is. Changing might introduce new issues and as it stands all works as expected.

If some sensors need another approach it's no problem to change.

device111:
yes, that's how we do it.

PR #8593, opened 2020-06-01
fix serial buffer issues

gemu2015:
## Description:

fixes serial receive buffer to small in sml and ibeacon driver (ESP8266 and ESP32)

## Checklist:

### DIFF
--- a/lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
+++ b/lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
@@ -174,6 +174,7 @@ bool TasmotaSerial::begin(long speed, int stop_bits) {
       m_uart = tasmota_serial_index;
       tasmota_serial_index--;
       TSerial = new HardwareSerial(m_uart);
+      if (TM_SERIAL_BUFFER_SIZE != serial_buffer_size) TSerial->setRxBufferSize(serial_buffer_size);
       if (2 == m_stop_bits) {
         TSerial->begin(speed, SERIAL_8N2, m_rx_pin, m_tx_pin);
       } else {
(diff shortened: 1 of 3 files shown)


arendst:
I see the advantage of being able to resize the serial buffer but does this really solve an issue? The default ESP32 hardware serial buffer size is 256 as can be seen in HardwareSerial.cpp:
```
_uart = uartBegin(_uart_nr, baud ? baud : 9600, config, rxPin, txPin, 256, invert);
```

lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
```diff
@@ -174,6 +174,7 @@ bool TasmotaSerial::begin(long speed, int stop_bits) {
       m_uart = tasmota_serial_index;
       tasmota_serial_index--;
       TSerial = new HardwareSerial(m_uart);
+      if (TM_SERIAL_BUFFER_SIZE != serial_buffer_size) TSerial->setRxBufferSize(serial_buffer_size);
```
arendst, line 177: Considering the default size of 256 for hardware serial I think we better change this line into ```
if (256 != serial_buffer_size) TSerial->setRxBufferSize(serial_buffer_size);
```

**Comment by gemu2015** [comment]
i just wanted to be able to set the buffer like on esp8266
 ok, agree !

**Comment by gemu2015** [comment]
however what i had in mind was to change the buffer only when size was given, otherwise leave to default. this changes the size to 64 always when size is not given.

**Comment by arendst** [comment]
I know. That's why I did it this way:
```
      if (serial_buffer_size > 256) {
        TSerial->setRxBufferSize(serial_buffer_size);
      }
```
It only changes the default from 256 bytes if requested more.

PR #8610, opened 2020-06-03
Add MAX6675 sensor

aik:
This is basically a cutdown version of MAX31855 without reference
temperature reading and lower resolution (only positive, 12bit only).

Signed-off-by: Alexey Kardashevskiy aik@ozlabs.ru
Description:

The fix of #4764 claimed it might fix 6675 in the future but this never happened. I wonder if

    this is a desired feature (I need it for sure, not sure about the others)
    it is better to have a separate file rather than share it with 31855
    add support for multiple similar sensors per board (I'd like a couple but that's about it)

Thanks.

Changes:
v2:
* treat occasional 0xfff as an error
* fixed language packs


arendst:
I understand the ease in adding your required functionality to an already available driver.

The problem is that it adds a lot of redundant code (size) just to handle an almost identical sensor. A much better way is to try and implement the MAX6675 sensor using as small code change as possible. 

So refrain from adding redundant GPIO's but focus on a SetOption to select either MAX31855 or MAX6675. The current GPIO additions also break current supported GPIO configs as it inserts GPIO's instead of appending it.

So I suggest you remove the additional GPIO's and first implement using a define which in time can be replaced by a SetOption.

arendst:
Thx for the alternative!

PR #8628, opened 2020-06-05
fix: ADC temperature calculation

oori:
## Description:
the ADC value is already 10-bit normalised, not need to multiply it again.
I used a wemos d1 (with 3.3v analog input) and hit this bug. I've tested this code with NTC thermistor on wemos d1. Analog values like 512 would produce 72°C (instead of 25°C).    This with the correct `AdcParam` (47k Rs, 50k NTC).

## Checklist:

### DIFF
--- a/tasmota/xsns_02_analog.ino
+++ b/tasmota/xsns_02_analog.ino
@@ -201,7 +201,7 @@ void AdcEverySecond(void)
   if (ADC0_TEMP == my_adc0) {
     int adc = AdcRead(2);
     // Steinhart-Hart equation for thermistor as temperature sensor
-    double Rt = (adc * Settings.adc_param1) / (1024.0 * ANALOG_V33 - (double)adc);
+    double Rt = (adc * Settings.adc_param1) / (1024.0 - (double)adc);
     double BC = (double)Settings.adc_param3 / 10000;
     double T = BC / (BC / ANALOG_T0 + TaylorLog(Rt / (double)Settings.adc_param2));
     Adc.temperature = ConvertTemp(TO_CELSIUS(T));


arendst:
Alas, this doesn't seem to work on a Shelly 2.5

The valid temperature of 62 C drops to 20 C with this formula adjustement so something is fishy in your setup where you need this change.

oori:
@arendst  Thanks for your quick response.


The connections on the Wemos D1:
```
                        A0
                        |
                        |
                        |
 GND -----  NTC (50k) -----  Rs (47k) ----- 3.3v
```

- [x] Same result with two different 50k NTC sensors.
- [x] Check NTC resistance with multimeter
- [x] Check voltage divider with multimeter (over A0)

Tasmota:
- [x] Template --> ADC0 --> Analog:       get **correct** numbers (480-520, in room temperature)
- [x] Configure `AdcParam 2,47000,50000,3950`
- [ ] Template --> ADC0 --> Temperature:      get **wrong** result (over 70c).
- [x] Build with PR:   returns the expected temperature.

As the analog measurement is correct, where else could the issue be, if not in this line?     I don't know the Shelly 2.5 integration, does it use the same configuration I outlined here?   (ADC0 template set to temperature).

arendst:
Shelly 2.5 integration is outlined in file xsns_02_analog.ino lines 34..40

Your 3950 is the NTC B Coefficient. Are you sure you use the correct one? On Shelly I was told it is 3350 for their NTC used.

arendst:
For celsius you need to set ``SetOption8 0`` but I expect you checked that already

lalo-uy:
May be Wemos A0 330K impedance must be taken into account.

Gnd —/\/\/\/\———-|———/\/\/\/\/\——— input
           100k.            adc.        220k

lalo-uy:
It ended in something like this:
Rntc=Vcc*100/(377*Vadc)-47*330/377

oori:
@lalo-uy Actually, I originally tested it with a 330k Rs because I forgot the Wemos's analog input was 3.3v.  330k fits 1.0v A0.    Ofcourse, that ended up with the wrong reading, by far...

@arendst I saw Beta 3350 mostly on 10k NTC, and 3950 on 50k.   But, not fixed, nor is our issue (_minor affect on the end number_).   About `SetOption8 0`, yes of course..  was my first bet.  it's not that.

Again, switching the template to Analog, gives the expected number (around the 512 center).   So, I'm mostly looking at the formula [double Rt = (adc * Settings.adc_param1) / (1024.0 * ANALOG_V33 - (double)adc);](https://github.com/arendst/Tasmota/blob/54ab26439ef7c30daa392ce72b1efb78b83bf4ae/tasmota/xsns_02_analog.ino#L204).   works for the shelly's NTC (`32k,10k,3350`), but not on my case.  on the shelly, if you switch template to "Analog", you get the expected value?
I'm now looking into other SteinhartHart implementations, to see how the numbers work.

Untitled

oori:
I've looked at a couple of [references](https://www.allaboutcircuits.com/projects/measuring-temperature-with-an-ntc-thermistor/),  and did the math.   The patch seems mathematically correct. Maybe something's funky with the Shelly?

Examples:
`1/T=1/T0 + (1/β)*ln(R/R0)`
1 / (1/298.15 + (1/3950 * log(40000/50000)))   ==  `300.34`

`(BETA * ROOM_TEMP) / (BETA + (ROOM_TEMP * log(rThermistor / RESISTOR_ROOM_TEMP)))`
(3950 * 298.15) / (3950 + (298.15 * log(40000 / 50000)))   == `300.34`

`(this patch)`
3950 / (3950/298.15 + log(40000/50000)) == `300.34`

arendst:
I have both Shelly 2.5 and Shelly 1PM. On both the temperature displayed equals the expected temps and complies with the Shelly community information.

Shelly 1PM: 36C = Analog 580
SHelly 2.5: 69C = Analog 234

So for now I cannot explain other than my shelly supplied values are wrong (if I install your PR).

arendst:
You might want to try this:
```
    // Steinhart-Hart equation for thermistor as temperature sensor
    double Rt = (adc * Settings.adc_param1) / (1024.0 * ANALOG_V33 - (double)adc);
//    double BC = (double)Settings.adc_param3 / 10000;
    double BC = (double)Settings.adc_param3 / (double)Settings.adc_param2);
    double T = BC / (BC / ANALOG_T0 + TaylorLog(Rt / (double)Settings.adc_param2));
    Adc.temperature = ConvertTemp(TO_CELSIUS(T));
```
as I'm wondering where my 10000 comes from and it maybe the NTC resistance value...

oori:
>  I'm wondering where my 10000 comes from

The `/ 10000` is correct.  it's probably a "hack" due to [line 80](https://github.com/arendst/Tasmota/blob/61ea7535a80cf0e0f3283df550437865568e26f4/tasmota/xsns_02_analog.ino#L80)

I think this is issue is related to the Wemos A0 @ 3.3v (via internal voltage divider) .     I tried same setup with ESPHome, there it's resolved with a `filter multiply: 3.3` on the `adc` (see `NOTE` [here](https://esphome.io/components/sensor/adc.html), and [this link](https://esphome.io/components/sensor/ntc.html?highlight=ntc) as well).

Two things I don't get:
1. Why `analog0` value is **correct**, and only the temperature is off ?   [Here](https://github.com/arendst/Tasmota/blob/61ea7535a80cf0e0f3283df550437865568e26f4/tasmota/xsns_02_analog.ino#L217)'s the same `AdcRead()` right?
2. Why this multiplication in voltage `* ANALOG_V33`?   `Rt` is supposed to be in Ohms.

Anyway, I've ordered a shelly 2.5..

hallard:
Well, the ESP boards may have so much different wiring on ADC line, Looking at wemos schematic (below) they added R/R divider to be able to have up to 5V  going to A0 pin (not analog ADC input) so this can totally break any calculation.

An each board have it owns input model, so my guess, hard to have a formula that works with ALL boards. I already had headache with that.

So 330K impedance on A0 is correct, on ADC input it's different.So in your case as you use 3V3 only, I would remove the 100K and 220K resistor and put 0 Ohm on 220K to remove this mess.

image

kugelkopf123:
The esp8266 allows a max of 1V on ADC of the ESP itselfs. So with this Voltagedivider on the Pic. it is capable to run 3,3V on A0 not 5v! If you want to drive A0 with 5V you need to add 180k in series to the 220k to get 400k.

hallard:
Correct, my fault so this is worse than I thought.

lalo-uy:
The problem with Wemos is that the NTC is not in serie with 10K R2, but with R2 in paralel with 320K . That is 9697 ohms. Using this value in the NTC R divisor formula may be more acurate.

arendst:
So for now I think we can close this.

refael6:
As Oori say and I checked there is a mistake in the formula, no need to multiply by 3.3V.
double Rt = (adc * Adc[idx].param1) / (ANALOG_RANGE * ANALOG_V33 - (double)adc); 
Calculation of units
resistance != (bit*resistance)/(bit*Voltage-bit)
resistance = (bit*resistance)/(bit-bit)

the Right formula:
double Rt = (adc * Adc[idx].param1) / (ANALOG_RANGE - (double)adc); 

I'm new if github and tasmota hope I helped.

ascillato2:
https://github.com/arendst/Tasmota/issues/13668#issuecomment-968320441

Issue #8637, opened 2020-06-07
Using VL53L0X with MQTT and Domoticz


pakosel:
### PROBLEM DESCRIPTION
I am using a number of ESP8266 devices flashed with Tasmota. All of them are reporting the state using MQTT and the state is stored is Domoticz. Recently tried to add a new ESP8266 device with VL53L0X sensor. Since the support of this sensor is not **built-in** I've used GitPod (specifically a ***TasmoCompiler***) to turn a support of the sensor by defining USE_VL53L0X flag:
```
#define USE_VL53L0X
```

The sensor works fine but for unknown reason the values are not sent to MQTT broker. Neither the Domoticz input topic ("domoticz/in"). Anyone else having the same issue with this sensor?

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [ ] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [ ] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): Wemos D1 mini
- [x] Tasmota binary firmware version number used: 8.3.1.2
  - [ ] Pre-compiled
  - [x] Self-compiled
    - [ ] IDE / Compiler used: _____
- [x] Flashing tools used: Tasmote firmware update
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```
  Configuration output here:


```
- [x] Provide the output of this command: `Status 0`:
```
15:09:56 CMD: status 0
[...]

PR #8640, opened 2020-06-07
Fix for #8637

pakosel:
## Description:

Value change of **VL530L0X** sensor does not notify *Domoticz* (if configured). My implementation checks each second for a change of the **distance** value and notifies *Domoticz* when distance changes more than a minimal threshold (5mm)

## Checklist:


device111:
@pakosel 
the `sensor_index` is not defined an end in an compiler error.
you don't need this variabel:
`if (0 == tele_period) {` 
is right.

pakosel:
> the `sensor_index` is not defined an end in an compiler error.
> you don't need this variabel.

Fixed that one but it's odd - on my configuration this compiles without any errors...

> and check the correct idx. i don't think this is the right implementation.

What do you mean by checking the correct IDX? It's not a correct approach with storing previous value and updating Domoticz when changed or some other reason?

Issue #5494, opened 2019-03-18
Serial data reception using SERIAL_7E1 


Bry50:
Describe the solution you'd like

I would like to connect my connected energy counter (linky) to Tasmota. 
Linky Serial data is sent using: 1200 bps, 7 data bits, even parity, 1 stop bit

Describe alternatives you've considered

I have downloaded and flashed the latest Tasmota Master.
Program Version	6.3.0
Build Date & Time	2019-03-18T15:43:36
Core/SDK Version	2_3_0/1.5.3(aec24ac9)

I am trying to use Hardware Serial Bridge
Sonoff is configured as 
23 Sonoff 4CH PRO
All three GPIO are set to 00 None

Using protocol 8N1, I can send and receive data from termite ( baudrate 9600).

In issue [#3837](https://github.com/arendst/Sonoff-Tasmota/issues/3837), it is recommended :
You will have to change SERIAL_8N1 for SERIAL_7E1 in the TasmotaSerial.cpp

```
bool TasmotaSerial::begin(long speed) {
if (m_hardserial) {
Serial.flush();
Serial.begin(speed, SERIAL_8N1);
```

TasmotaSerial.cpp was slightly different, so I inserted
```
bool TasmotaSerial::begin(long speed, int stop_bits) {
  m_stop_bits = ((stop_bits -1) &1) +1;
  if (m_hardserial) {
    Serial.flush();
    if (2 == m_stop_bits) {
      Serial.begin(speed, SERIAL_7E2);
    } else {
      Serial.begin(speed, SERIAL_7E1);
    }
```

However, using 7E1 on termite
From Sonoff; Serialsend data is received by Termite OK
but
From Termite, data is not correctly received by the Sonoff
Often, no data is received, occasionally the last one or two characters are received.

Surprisingly, Using 8N1 on termite the serial data is still received correctly.

ascillato2:
Please, could you be so kind on completing the [troubleshooting template](https://github.com/arendst/Sonoff-Tasmota/issues/new/choose) in order to have more information so as to properly help you?

Also, please update to latest Tasmota (6.4.1.22). Your version 6.3 is very old and lots of changes related to serial were implemented, so there won't be support over an old version. 

Remember to read the [Contributing Guideline and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md). Thanks.

---------------------------

[...]

PR #8701, opened 2020-06-15
Added feature to set all possible mode for Parity, Stop, and Data bits

### LINKED ISSUE #3837: Support for generic Rx sensor
I would like to connect my connected energy counter (linky) to Tasmota. We can find on the web adaptative circuit to adapt signal outing from linky to a serial signal compatible with Rx of ESP2286  or Raspebery....

Here is a version you can order : https://hallard.me/pitinfov12-light/
Or here is DIY version : http://sarakha63-domotique.fr/nodemcu-teleinformation-wifi/

Is there already a way to connect such circuit to Tasmota, inside a Sonoff Basic, and configure it to the sensor input to automatically sending teleinfo TTL/UART message over MQTT ?

I assume we can create a specific sensor for this device, but as this is not of the shelf, it may not be accepted. A generic UART to Mqtt gateway would easily allow to get the data and perform the data functionnal parser over the home automation solution.

**Additional context**

hallard:
## Description:

This PR add new parameter `serial_config` to 
`bool TasmotaSerial::begin(long speed, int stop_bits, int serial_config)`

Leaving parameter `stop_bits` will not break the existing and leaving `serial_config` parameter empty will work as today.
`myTasmoSerial->begin(1200, SERIAL_8E1);`

Today calling with `myTasmoSerial->begin(1200, SERIAL_7E1);` leave to 8N1 so we needed to add another parameter to avoid breaking current implementation, so to do specific configuration for example we can now use  
`myTasmoSerial->begin(1200, 0, SERIAL_7E1);`

**Related issues** fixes #3837 and #5494

## Checklist:

### DIFF
--- a/lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
+++ b/lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
@@ -163,7 +163,6 @@ bool TasmotaSerial::begin(long speed, int stop_bits, int serial_config) {
     Serial.flush();
     if (serial_config >=0) {
       Serial.begin(speed, (SerialConfig) serial_config);
-
     // Keep old call begin compatibility
     } else {
       if (2 == m_stop_bits) {
@@ -186,7 +185,6 @@ bool TasmotaSerial::begin(long speed, int stop_bits, int serial_config) {
       }
       if (serial_config >=0) {
         TSerial->begin(speed, serial_config, m_rx_pin, m_tx_pin);
-
       // Keep old call begin compatibility
       } else {
         if (2 == m_stop_bits) {
--- a/lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
+++ b/lib/TasmotaSerial-3.0.0/src/TasmotaSerial.cpp
@@ -163,6 +163,7 @@ bool TasmotaSerial::begin(long speed, int stop_bits, int serial_config) {
     Serial.flush();
     if (serial_config >=0) {
       Serial.begin(speed, (SerialConfig) serial_config);
+      
     // Keep old call begin compatibility
     } else {
       if (2 == m_stop_bits) {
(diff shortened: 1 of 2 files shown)


arendst:
Thx for your PR but...

The current implementation of calling additional function ``SetSerial`` makes sure that the used serial config is known to Tasmota. It also sets some Tasmota persistent parameters used during power on and restart.

As a library cannot communicate easily it's config parameters back to the program without excessive code increase (which is a major Tasmota core feature) I will not implement this PR and want to keep the current functionality of calling ``SetSerial``.

hallard:
@arendst I have no problem with that so what's the best way for my energy device to setup 7E1 serial config ? If i understand correctly and I tested that, SetSerial works on ESP8266 (I tested it) but on ESP32 with 2 serials how SetSerial know wich one to setup ? Or may be we limit to only 1 serial ? Just to be sure

arendst:
Ah. ESP32. Good one. Let me think about it.

s-hadinger:
Charles, I strongly advise to handle this in software.

Receiving 7E1: just drop the MSB `val & 0x7F`

Sending 7E1: compute the parity bit and add it before sending.
Ex: https://stackoverflow.com/questions/29379006/calculate-parity-bit-from-string-in-c
```
char calcParityBit (unsigned char v)
{
    return (0x6996u >> ((v ^ (v >> 4)) & 0xf)) & 1;
}
```

hallard:
@arendst yeah ESP32 I have in mind using 2nd ESP32 Serial to drive LoRa Module :-)

image

@s-hadinger I was sure sender (teleinfo) was sending only 7bits data. So if Serial of ESP is configured to receive 8 bits data, even dropping MSB will drop a data bit and get all frame shifted 1 bits on each char received. So I just tried your solution, and guess what, it works, not sure to understand how serial is handled by hardware and how a sender can send 7 bits of data and receiver waits for 8 and both are in sync, anyway it works, still don't understand why.

PR #8702, opened 2020-06-15
Add serial to TCP bridge, ``TCPStart`` and ``TCPBaudRate``

s-hadinger:
## Description:

Add a new feature: Serial to TCP bridge (similar to esp-link). This allows to remotely communicate to a MCU through ESP8266. Needs `#define USE_TCP_BRIDGE`

It adds 2 GPIO types: `TCP Tx (208)` and `TCP Rx (209)` and can work with hardware or software serial.

Commands:

- `TCPBaudRate <x>`: sets the baudrate for serial (only 8N1 mode), min `1200`, max `115200` by 1200 increments.
- `TCPStart <port>`: listens to port `<port>`. This features supports 2 parallel TCP connexions, which can be useful if you need a terminal + a specific protocol (like XMODEM). The 3rd connection will disconnect an previous connection. The number of parallel connections is a compile-time option.
- `TCPStart 0` or `TCPStart`: shuts down the TCP server and disconnects any existing connection.

For security reasons, the TCP bridge is not started at boot, and requires an explicit `TCPStart` command (can be automated with Rules).

Size impact:
- Flash: +1.3kb
- Ram: +64 bytes

## Checklist:

### DIFF
--- a/tasmota/CHANGELOG.md
+++ b/tasmota/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Add initial support for Telegram bot (#8619)
 - Add support for HP303B Temperature and Pressure sensor by Robert Jaakke (#8638)
 - Add rule trigger ``System#Init`` to allow early rule execution without wifi and mqtt initialized yet
+- Add serial to TCP bridge, ``TCPStart`` and ``TCPBaudRate`` (needs #define USE_TCP_BRIDGE)
 
 ### 8.3.1.2 20200522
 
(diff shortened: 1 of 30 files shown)


s-hadinger:
@arendst I'm not sure why compilation fails (mDNS), but it seems unrelated to this PR.

TheChatty:
With no configured GPIO:
- issueing "tcpbaudrate <x>" causes an exception & reboot
- issueing "tcpstart <x>" causes a command error

How can I make it work with hardware serial (no configured GPIOs) thus allowing "serialsend" etc. simultaneously?

ayavilevich:
I couldn't find any mention of this feature in the docs. Is this intentional? To be added later?

s-hadinger:
You are right, it's missing in the doc. You are more than welcome to provide a doc Pull Request with the missing parts.

PR #8705, opened 2020-06-16
scripter call >BS after sensor init

gemu2015:
## Description:

extra call to >BS after sensor init

**Related issue (if applicable):** fixes 8685

## Checklist:


arendst:
Can't you trigger on the recently released (I think this week) new rules_flag.system_init ?

gemu2015:
since system_init is exactly set at this point it is easiest to call the >BS section directly from here instead of monitoring the flag in LOOP. I would also need an extra read once flag

PR #8771, opened 2020-06-23
Add support for sending GlobalCache codes via irsend

Dilbert66:
Small function to add support for GlobalCache formated ir commands.  Uses the same format as irsend<x> raw,xxx.
use irsend<x> gc,xxx,xxx 

## Checklist:

### DIFF
--- a/tasmota/xdrv_05_irremote_full.ino
+++ b/tasmota/xdrv_05_irremote_full.ino
@@ -470,9 +470,8 @@ uint32_t IrRemoteCmndIrSendRaw(void)
       return IE_INVALID_RAWDATA;
     }   // Parameters must be at least 3
 
-
 #ifdef IR_GC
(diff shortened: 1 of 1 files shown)


tasmota/xdrv_05_irremote_full.ino
```diff
@@ -470,6 +470,23 @@ uint32_t IrRemoteCmndIrSendRaw(void)
       return IE_INVALID_RAWDATA;
     }   // Parameters must be at least 3
 
+#ifdef IR_GC
+   if (strcmp(str, "gc") == 0 || strcmp(str, "GC") == 0) {  //if first parameter is gc then we process global cache data else it is raw
```
arendst, line 474: You won't catch Gc or gC. Usse strcasecmp instead.

Dilbert66: True and done. I also removed the ifdef ir_gc check since really this section is only enabled when irremote_full is defined which enables most protocols anyhow so will not really impact resources in any large extent at this point.  Up to you.

PR #8799, opened 2020-06-27
Add units to comments for temp & pressure

mhansen:
Reverse engineered what units these are from looking at the code.

## Checklist:

There's no change to test, this is comments-only.


arendst:
Alas. This is too generic.

Temperature can be Celsius or Fahrenheit depending on SetOption8 set. Same for pressure. It can be hPa or mmHg depending on SetOption24

It's better you scan the tele messages which inlude the correct unit type.

mhansen:
Thanks for the guidance - good to know as I was incorrectly assuming these were Celsius!

mhansen:
I did a bit of digging, and filed https://github.com/arendst/Tasmota/issues/8808, it looks like `global_temperature` is used as Celsius in all but one place.

mhansen:
Filed https://github.com/arendst/Tasmota/issues/8810 for figuring out the units of `global_pressure`.

Issue #8789, opened 2020-06-26
Next try of zero crossing PWM control after recent changes


stefanbode:
**Have you looked for this feature in other issues and in the docs?**  

**Is your feature request related to a problem? Please describe.**  
_A clear and concise description of what the problem is._
AC Dimmer need the PWM synced with the AC sinus. This was Impossible due to the PWM implementation until know. latest changes generate a very fine and reliable signal. The zero crossing therefore only need do Trimm after some time. Not on every sinus.
**Describe the solution you'd like**  
_A clear and concise description of what you want to happen._
Check if and how we can integrate the zero crossing as an input signal.

**Describe alternatives you've considered**  
_A clear and concise description of any alternative solutions or features you've considered._

**Additional context**  
_Add any other context or screenshots about the feature request here._

**(Please, remember to close the issue when the problem has been addressed)**

ascillato2:
AC Dimming in Tasmota is going to be a challenge because it is not a Real Time Software for making a precise ZCD.

Anyway, It can be checked also how it is implemented in ESPHome (https://github.com/esphome/esphome/pull/880)

stefanbode:
Here are my idea how we can make it possible:
- User PWM1 and PWMfrequency 100 or 120 depending on country
- Use COUNTER1 as input for the zero crossing
- Every 1000 steps (10 seconds) do some slight magic to keep the PWM start in Sync. There is now a PHASE that we might can use (@s-hadinger can you help how we can influence the phase synchronisation?)

Real Time Software is not required because be can use the PWM frequency. This is now after several fixes the last week very stable. We need "just" start the PWM with the zero cross and then check every 10 seconds if it is in phase and not shift it a few ns.

stefanbode:
I changed the counter driver with just some lines. After 1 second i realign the phase with the zero trigger. Works fine on the oscilloscope. Anyhow my first test with a 60w bulb did not work. Now it look like the triac is broken. The data sheet say 2A ok and max 5A. The bulb in operation only creates about 0.3A. Not heat detected. Any idea? Will share code changes for retest.

Jason2866:
@stefanbode  Probably harmonic oscillations in the high-frequency range were created.
That kills.
[...]

PR #8833, opened 2020-07-01
Add support for AC Dimmer #8789

stefanbode:
## Description:
The pull request add support for AC Dimmer through the normal DIMMER function. The resolution of PWMfrequency is only +- 1 Hz. This is much too high to avoid "pumping" of a dimmed light. There are two different options to fix this. One is to sync the PWM very often. e.g. 4 times a second. Then the pumping process is not that visible, but still there at low DIMMER values. The second approach is part of this pull request. The analogWrite procedure is enhanced to add the Frequency as a number of CPU cycles. Each second the current frequency is measured and the PM newly configured with the new frequency. Now there is no flickering at all. 

Configuration:
PWM1: PWM Pin AC Dimmer
COUNTER1: Zero-Cross AC Dimmer
COUNTERDEBOUNCELOW 1  (absolutely required to work on the rising edge of the Zero-Cross)
SETOPTION15 1 (default)
LEDTABLE 0
PWMfrequency 100 (Germany)

Known-issues:
Changes of the dimmer sometimes cause a full light for a fraction of a second
Counter can overflow?
Normal lamps need at least Dimmer 25 or more to start
No support for more than 1 Dimmer (e.g. SETOPTION68 1)

**Related issue (if applicable):** fixes #<Tasmota issue number goes here> #8789

## Checklist:
  - [x ] The code change is tested and works on core ESP8266 V.2.7.1


Jason2866:
Do you have made changes in core file Arduino.h ?
If so you have to add #ifdef ESP8266 because ESP32 compile will not work without.

stefanbode:
Here a video of the results. Yes let me make this change to he Arduino.h and see if it works.

https://www.youtube.com/watch?v=lcNu_36yafo

stefanbode:
Please wait for merging it. I found a way to avoid changing the Arduino.h and also changing the core_esp8266_wiring_pwm.cpp. This will make it much more future proven

PR #8848, opened 2020-07-04
Transparent default value for MqttFingerprint2

ryancdotorg:
## Description:

I was unable to find any information on where the existing default
MqttFingerprint2 value came from, and can't be sure that someone
doesn't have a key matching it.

For transparency, this patch replaces the default MqttFingerprint2 value
with the result of SHA1("Tasmota") as a "nothing-up-my-sleeve number".

## Checklist:

## Note:
I do not have an ESP32 device to test on, but the only thing changed is a `#define`'d constant.

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -97,8 +97,8 @@
 #define MQTT_USE               true              // [SetOption3] Select default MQTT use (false = Off, true = On)
 
 #define MQTT_HOST              ""                // [MqttHost]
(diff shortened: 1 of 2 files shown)


arendst:
It came from my cloudmqtt service.

ryancdotorg:
If there's an existing key that corresponds to that fingerprint, I think it would be best to replace it with something that doesn't have a valid key associated with it by default.

My understanding of the current code is that both fingerprints can be used, so the default key could be used as a skeleton key to impersonate the intended server. I don't mean to imply that anyone would do this, but there doesn't seem to be any reason to use that as a placeholder rather than something else.

PR #8849, opened 2020-07-04
Shutter reduce log output & reactivate movement rule trigger

curzon01:
## Description:

The loglevel change is the chatty shutter log: As shutter already output a lot of info via MQTT during movement, the log level must be 3+ to enable direct driver log outputs.

The reason why shutter triggers during move was basically disabled was the instability. As rule processing are now re-entrance protected, didn't observe any issues using the already implemented triggers during shutter move (checked 4 shutters simultaneously with all rule triggers enabled).

## Checklist:


arendst:
The re-entrance protection doesn't cater for trigger overrun; it just discards them.

PR #8874, opened 2020-07-09
repair float int problems for MI Flora moisture and fertility

d0m1n1qu3:
## Description:

fix an issue with float and int converts for ESP32 MiFlora

## Checklist:


Staars:
Looks good to me.

s-hadinger:
I thought `%f` was disabled in printf to reduce code size. It's not true anymore?

Staars:
I could have sworn, that I tested this a few days ago in another case, but as you mention this, I start to remember.
I am not able to test it on real hardware before tomorrow afternoon.

d0m1n1qu3:
hi. i also can change it to use only int and %d. this code is only for the ESP32 is there realy a problem with %f ? but if, good to know. :)

Staars:
>  only for the ESP32

Ahh, I guess that is the crucial point :)

arendst:
Indeed esp32 only

PR #8877, opened 2020-07-10
read firmware verion for FLORA devices

d0m1n1qu3:
## Description:

the FLORA sensors send the firmware version number while they send the
battery info so we parse it and send it via mqtt (json) and also
present it in the webinterface

only in files for ESP32

## Checklist:
  - [x] The code change is tested and works on core ESP8266 V.2.7.2

### DIFF
--- a/tasmota/xsns_62_MI_ESP32.ino
+++ b/tasmota/xsns_62_MI_ESP32.ino
@@ -1210,7 +1210,6 @@ bool MI32Cmd(void) {
 const char HTTP_MI32[] PROGMEM = "{s}MI ESP32 {m}%u%s / %u{e}";
 const char HTTP_MI32_SERIAL[] PROGMEM = "{s}%s %s{m}%02x:%02x:%02x:%02x:%02x:%02x%{e}";
 const char HTTP_BATTERY[] PROGMEM = "{s}%s" " Battery" "{m}%u %%{e}";
-const char HTTP_FIRMWARE[] PROGMEM = "{s}%s" " Firmware" "{m}%s{e}";
 const char HTTP_VOLTAGE[] PROGMEM = "{s}%s " D_VOLTAGE "{m}%s V{e}";
 const char HTTP_MI32_FLORA_DATA[] PROGMEM = "{s}%s" " Fertility" "{m}%u us/cm{e}";
 const char HTTP_MI32_HL[] PROGMEM = "{s}<hr>{m}<hr>{e}";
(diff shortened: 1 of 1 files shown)


d0m1n1qu3:
here a screenshot :-)

image

Staars:
There is nothing wrong with this approach and especially on the ESP32 we have some room left, but please ask yourself, if you really need a special feature. 
It is not so hard to read all kinds of info for every known device (I have 8 different types at home) but at some point in the future, we will have a memory problem.
I think the hardest challenge in the long run is to leave things out.
(Again: I have no problem with a merge)

d0m1n1qu3:
i found it really helpful to know which firmware the device have because i had problems with some older firmwares. 
and i like the idea to use tasmota to find out the firmware version without install apps or other tools. 
but i will keep it in my mind. :-)

arendst:
A better solution would be to drop static info from the gui and use mqtt only. This way you still have the info you need while the gui stays clean

d0m1n1qu3:
but this info is only displayed if you send the command MI32Battery
question: is it a display problem because of full ui or a memory problem because of the "const char HTTP_FIRMWARE[] PROGMEM"?

arendst:
On an esp32 it's a gui problem for now. I didn't see you did add a command too. In that case I'll merge.

d0m1n1qu3:
i use the command for the battery call because you get the frimware version in the same data packet from the sensor

tasmota/xsns_62_MI_ESP32.ino
```diff
@@ -1012,6 +1015,23 @@ bool MI32readBat(char *_buf){
...
+    char _firmware[5]; // FLORA send 5 byte for firmware version
+    strncpy(_firmware, _buf+2, 5);
+    AddLog_P2(LOG_LEVEL_DEBUG,PSTR("%s: Firmware: %s"),D_CMND_MI32,_firmware);
+
+    uint32_t _slot = MI32.state.sensor;
+    DEBUG_SENSOR_LOG(PSTR("MIBLE: Sensor slot: %u"), _slot);
+
+    memcpy(MIBLEsensors[_slot].firmware, _firmware, 5);
+    MIBLEsensors[_slot].firmware[5] = '\0';
+    return true;
```
arendst, line 1030: There is no firmware[5] on a defined char of [5].

arendst:
Battery level is fine as dynamic. Firmware being statis should not be on the gui

d0m1n1qu3:
but i and a mate from me like it on the gui :-) 
than i change it to mqtt only?

arendst:
Yes mqtt only. Thx in advance.

PR #8937, opened 2020-07-20
Serialsend6 - comma-separated decimal send as binary

grob6000:
## Description:

Alternative to #8920 (i.e. one or the other). Same use case - allowing sending of parameters using rules to binary serial devices e.g.  #5737. I've noted that the mqtt response formats in HEX when serial is received after a binary message is sent - I don't think this should change so it's treated the same.

example: `serialsend6 255,85,5,5,220,10`

An example rule that (should) allow the above device to work with vanilla tasmota + this PR:
`rule1`
`on dimmer#state=0 do event setdimmer=5 endon`
`on dimmer#state>0 do backlog scale1 %value%,1,100,16,255; event udimmer endon`
`on dimmer#boot do backlog baudrate 9600; scale1 %value%,1,100,16,255; event udimmer endon`
`on event#udimmer do event setdimmer=%var1% endon`
`on event#setdimmer do serialsend6 255,85,%value%,5,220,10 endon`
`on power1#state=0 do event setdimmer=5 endon`
`on power1#state=1 do event setdimmer=%var1% endon`

## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.2.1
> no device to test sorry

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -906,6 +906,17 @@ void SerialSendRaw(char *codes)
   }
 }
 
+// values is a comma-delimited string: e.g. "72,101,108,108,111,32,87,111,114,108,100,33,10"
+void SerialSendDecimal(char *values)
+{
+  char* &p;
+  uint8_t code;
+  for (str = strtok_r(values, ",", &p); str; str = strtok_r(nullptr, ",", &p)) {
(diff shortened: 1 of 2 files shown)


arendst:
Thx. Will evaluate

PR #8956, opened 2020-07-21
Added Generic Stepper Driver

holgerpieta:
This PR introduces a generic stepper driver.

First idea was to make a generic stepper driver that could drive UART controlled TMC2208 drivers as a drop in replacement for the existing A4988 stepper driver controller. But that didn't work out due to some of the design choices of the existing driver, let alone the fact that is uses the term A4988 at too many places.
So, here is a new driver that can, in the moment, control A4988-style pin configured drivers (movement and microstepping) and UART controlled TMC2208 drivers (also RMS current).

I know that this PR is not yet ready to be merged, but I'd like to use this place to discuss what's missing to make it merge-able.
So, all the experts, please tell me what's missing or point me to a documentation about what should be there.

Open points from my point of view:
- The driver uses the [TMCStepper ](https://github.com/teemuatlut/TMCStepper) library by @teemuatlut, which is also used by Marlin. It is currently not included in this PR, but from the license (MIT) I don't think there will be a problem. In PlatformIO it could also be included as an external library, but I don't know how that works in Arduino or other build systems. If you want to build the driver, you have to download it manually in the moment (put it in lib or libesp32).
- Everything has only be tested in PlatformIO with ESP32 and TMC2208 driver. I should be able to test it later with an A4988 driver, but I don't have any ESP8266 or other ESPs. And I don't know much about other build systems.
- All configuration is currently done only by configuring the proper components on GPIOs. If more stepper drivers should be supported in the future, more configuration options will be needed, e.g. for other TMCs, other sense resistors or for other pin-controlled drivers. Please point me to places where I can find information (or examples) about non-pin-based configuration. The new BLE interface for the ESP32 seems to be using something like that.
- No configuration (speed, acceleration, current, microsteps...) survives a reboot. Please point me to information about storing configuration data in the flash.
- Some commands should not only return Done or Failed but some value. Please point me to information about that, too.
- Speed and acceleration commands for different movement commands interact a bit unexpectedly if you mix them. If can think of two options to manage that:
  - A single set of configuration values as it is in the moment, but with some smartness to update everything if one is changed.
  - Have separate sets of configuration for each step based, angle based and turn based commands.

Please comment and let me know what else I might have been missing.


stefanbode:
Why don't you use the SHUTTER option already in TASMOTA to work with the STEPPER MOTORS. At least you should take a look how the SHUTTERS use the PWM_FREQUENCY to control the motor. Because TASMOTA does not allow realtime signals you cannot use a general approach for sending the TRIGGER signal.

holgerpieta:
Ah, well, given that I actually want to control a shutter, that's kind of the obvious solution. I just didn't find it.
Now that I tried it, I find that it doesn't work on ESP32. You just cannot configure the GPIOs to the right components. In hindsight in might have been easier to make that work than what I did. But anyway.

It uses PWM to generate the signals, which is kind of a good idea, but counting the steps and stopping at the correct position is a bit of a problem. I like a correct step position with a slight jitter in the speed (because Tasmota is not real-time) more than the other way round. But I will check how the shutter driver counts steps and maybe take some ideas.
Also it cannot go slower than 1000 Hz, which is still around 10 RPM for 32 microsteps and 300 RPM for full step mode.

Oh, and of course it cannot interact with UART controlled drivers like the TMCs. Nifty feature to set the current automatically instead of fiddling around with a screwdriver. Or 256 microsteps, if you want to be really precise. And all the other features.
Or set the microsteps on pin-controller drivers. In the end, it's a shutter driver, not a stepper driver.

Anyway, thanks for hint, I'll see what I can learn from it.

PR #8963, opened 2020-07-22
Add firmware variant ZigbeeBridge

Jason2866:
Since @s-hadinger did bring the adventures with the ZigbeeBridge to a sucess it is time for
a firmware variant for.  Surprise we call it `tasmota-zbbridge` 

## Checklist:
  - [x] The code change is tested and works on Tasmota core ESP8266 V.2.7.2.1

### DIFF
--- a/tasmota/tasmota_configurations.h
+++ b/tasmota/tasmota_configurations.h
@@ -472,7 +472,7 @@
 //#define USE_SONOFF_D1                            // Add support for Sonoff D1 Dimmer
 
 // -- Optional light modules ----------------------
-#undef USE_LIGHT                                    // DISABLES LIGHTS support
+#define USE_LIGHT                                    // Add LIGHTS support
 // #define USE_WS2812                               // WS2812 Led string using library NeoPixelBus (+5k code, +1k mem, 232 iram) - Disable by //
 // //  #define USE_WS2812_DMA                         // DMA supports only GPIO03 (= Serial RXD) (+1k mem). When USE_WS2812_DMA is enabled expect Exceptions on Pow
 //   #define USE_WS2812_HARDWARE  NEO_HW_WS2812     // Hardware type (NEO_HW_WS2812, NEO_HW_WS2812X, NEO_HW_WS2813, NEO_HW_SK6812, NEO_HW_LC8812, NEO_HW_APA106)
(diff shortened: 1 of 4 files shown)


tasmota/tasmota_configurations.h
```diff
@@ -406,6 +406,264 @@
...
+#define USE_RULES                                // Add support for rules (+4k4 code)
+//  #define USE_EXPRESSION                         // Add support for expression evaluation in rules (+3k2 code, +64 bytes mem)
+//    #define SUPPORT_IF_STATEMENT                 // Add support for IF statement in rules (+4k2 code, -332 bytes mem)
+//  #define SUPPORT_MQTT_EVENT                     // Support trigger event with MQTT subscriptions (+3k5 code)
+
+//#define USE_SCRIPT                               // Add support for script
+  #define USE_SCRIPT_FATFS 4                     // Add support for script storage on SD card (+12k code, +4k mem)
+
+// -- Optional modules ----------------------------
+//#define ROTARY_V1                                // Add support for MI Desk Lamp
```
arendst, line 454: As this is enabled in my_user_cfg.h shouldn't they all be undef?

meingraham:
Are there other device specific variants? For example, Sonoff RF Bridge doesn't have its own variant; does it? Is this the right path to take for this? Pandora's Box?

arendst:
As the zigbee code takes up so much of the total code space available it's imposible to have zigbee in the default tasmota.bin. Other devices still fit in tasmota.bin so I'm afraid we have to choose this way. I rather would see ALL zigbee code in one binary though...

meingraham:
I think that's the direction that the variant should take rather than device specific. Like there is an IR variant, there can be a Zigbee variant... which includes the Z2T and Sonoff ZBridge code.

Jason2866:
@arendst  you are right. I have the undef in my override file. Will add

tasmota/tasmota_configurations.h
Jason2866: right, changed

Jason2866:
@meingraham  Let us do in steps. Get a precompiled working version first. Adding a (maybe possible) universal zigbee version in second step. At the moment there is no precompiled zigbee version. So imho this is better than having nothing.

PR #8994, opened 2020-07-29
Adding frequency output mode for buzzer

hello-world-dot-c:
Allow to use passive buzzers (piezo) by having the buzzer module turn on/off a 50% duty cycle PWM signal on the buzzer pin. (Edited to reflect changed functionality after review)

## Description:

Right now the Buzzer command and pin configuration only allows to control active buzzers that generate the sound themselves, so it is a simple turn on/off logic. Some hardware has only passive buzzers based on piezo crystals that need a frequency input. The PWM logic has all capabilities to generate such a frequency. The proposed change causes the existing buzzer logic to turn on/off a 50% PWM frequency at its assigned pin. A new SetOptions flag turns on/off this functionality. (Edited to reflect changed functionality after review)

Code changes are minimal. While I haven't tested it against the ESP32 there is no hardware dependency that could make it fail IMO.

## Checklist:
  - [x] The code change is tested and works on Tasmota core ESP8266 V.2.7.2.1

### DIFF
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -129,7 +129,7 @@ typedef union {                            // Restricted by MISRA-C Rule 18.4 bu
     uint32_t virtual_ct_cw : 1;            // bit 25 (v8.4.0.1)  - SetOption107 - Virtual CT Channel - signals whether the hardware white is cold CW (true) or warm WW (false)
     uint32_t teleinfo_rawdata : 1;         // bit 26 (v8.4.0.2)  - SetOption108 - enable Teleinfo + Tasmota Energy device (0) or Teleinfo raw data only (1)
     uint32_t alexa_gen_1 : 1;              // bit 27 (v8.4.0.3)  - SetOption109 - Alexa gen1 mode - if you only have Echo Dot 2nd gen devices
-    uint32_t spare28 : 1;                  // bit 28
+    uint32_t buzzer_freq_mode : 1;         // bit 28 (v8.4.0.3)  - SetOption110 - SetOption110 - Use frequency output for buzzer pin instead of on/off signal
     uint32_t spare29 : 1;                  // bit 29
     uint32_t spare30 : 1;                  // bit 30
     uint32_t spare31 : 1;                  // bit 31
--- a/tasmota/xdrv_24_buzzer.ino
+++ b/tasmota/xdrv_24_buzzer.ino
@@ -32,6 +32,7 @@ struct BUZZER {
   uint8_t inverted = 0;            // Buzzer inverted flag (1 = (0 = On, 1 = Off))
   uint8_t count = 0;               // Number of buzzes
   uint8_t mode = 0;                // Buzzer mode (0 = regular, 1 = infinite, 2 = follow LED)
+  uint8_t freq_mode = 0;           // Output mode (0 = regular, 1 = using frequency output)
   uint8_t set[2];
   uint8_t duration;
   uint8_t state = 0;
@@ -39,9 +40,28 @@ struct BUZZER {
 
 /*********************************************************************************************/
 
-void BuzzerOff(void)
+void BuzzerSet(uint8_t state)
 {
-  DigitalWrite(GPIO_BUZZER, 0, Buzzer.inverted);  // Buzzer Off
+  if (Buzzer.inverted) {
+    state = !state;
+  }
+
+  if (Buzzer.freq_mode == 1) {
+    static uint8_t last_state = 0;
+    if (last_state != state) {
+      if (state) {
+        analogWrite(Pin(GPIO_BUZZER, 0), Settings.pwm_range / 2);  // set 50% duty cycle for frequency output
+      }
+      else {
+        analogWrite(Pin(GPIO_BUZZER, 0), 0); // set 0% (or 100% for inverted PWM) duty cycle which turns off frequency output either way 
+      }
+      last_state = state;
+    }
+  }
+  else {
+    DigitalWrite(GPIO_BUZZER, 0, state);  // Buzzer On/Off
+  }
+  
 }
 
 //void BuzzerBeep(uint32_t count = 1, uint32_t on = 1, uint32_t off = 1, uint32_t tune = 0, uint32_t mode = 0);
@@ -69,11 +89,19 @@ void BuzzerBeep(uint32_t count, uint32_t on, uint32_t off, uint32_t tune, uint32
   }
   Buzzer.count = count * 2;  // Start buzzer
 
-  AddLog_P2(LOG_LEVEL_DEBUG, PSTR("BUZ: %d(%d),%d,%d,0x%08X(0x%08X)"), count, Buzzer.count, on, off, tune, Buzzer.tune);
+  // We can use PWM mode for buzzer output if enabled.
+  if (Settings.flag4.buzzer_freq_mode) {     // SetOption110 - Enable frequency output mode for buzzer
+      Buzzer.freq_mode = 1;
+  }
+  else {
+    Buzzer.freq_mode = 0;
+  }
+
+  AddLog_P2(LOG_LEVEL_DEBUG, PSTR("BUZ: %d(%d),%d,%d,0x%08X(0x%08X),%d"), count, Buzzer.count, on, off, tune, Buzzer.tune, Buzzer.freq_mode);
 
   Buzzer.enable = (Buzzer.count > 0);
   if (!Buzzer.enable) {
-    BuzzerOff();
+    BuzzerSet(0);
   }
 }
 
@@ -81,7 +109,7 @@ void BuzzerSetStateToLed(uint32_t state)
 {
   if (Buzzer.enable && (2 == Buzzer.mode)) {
     Buzzer.state = (state != 0);
-    DigitalWrite(GPIO_BUZZER, 0, (Buzzer.inverted) ? !Buzzer.state : Buzzer.state);
+    BuzzerSet(Buzzer.state);
   }
 }
 
@@ -113,7 +141,7 @@ void BuzzerInit(void)
 {
   if (PinUsed(GPIO_BUZZER)) {
     pinMode(Pin(GPIO_BUZZER), OUTPUT);
-    BuzzerOff();
+    BuzzerSet(0);
   } else {
     Buzzer.active = false;
   }
@@ -140,7 +168,7 @@ void BuzzerEvery100mSec(void)
           Buzzer.duration = Buzzer.set[Buzzer.state];
         }
       }
-      DigitalWrite(GPIO_BUZZER, 0, (Buzzer.inverted) ? !Buzzer.state : Buzzer.state);
+      BuzzerSet(Buzzer.state);
     } else {
       Buzzer.enable = false;
     }


tasmota/xdrv_24_buzzer.ino
```diff
@@ -32,16 +32,39 @@ struct BUZZER {
...
-  DigitalWrite(GPIO_BUZZER, 0, Buzzer.inverted);  // Buzzer Off
+  if (Buzzer.inverted) {
+    state = !state;
+  }
+
+  if (Buzzer.pwm_mode == 1) {
+    static uint8_t last_state = 0;
+    static uint8_t current_color = 0;
+    if (last_state != state) {
+      if (state) {
```
s-hadinger, line 53: I don't see why you are messing with the lights structure for something that is not a light at all. Doing this way you're just destroying the Red channel of the light.

s-hadinger (commented):
Please don't hack into the Lights API. This will mess the light settings.
Also keep in mind that you don't control the PWMFrequency that can be 233Hz or 977Hz depending on the Tasmota version.

Edit: I believe you should definitely use the GPIO_BUZZER and leave lights alone, they are applying Gamma Correction so you won't get 50% ratio anyways. Use the low-level `analogWrite()` on GPIO_BUZZER and you're done. You will still have the PWM Frequency issue but I can't think of a simple solution here. The PWM Frequency is global to all PWMs.

tasmota/xdrv_24_buzzer.ino
hello-world-dot-c: Thanks for reviewing. Let me answer this one by one.

1. Messing with light control?
No, see xdrv_24_buzzer.ino line 95:
```
  // If we use multi-channel PWM and have a PWM1 output, we can use PWM mode for buzzer output if enabled.
  if ((Settings.flag4.buzzer_pwm_mode) &&     // SetOption102 - Enable PWM mode for buzzer, uses PWM1 if enabled
      (Settings.flag3.pwm_multi_channels) &&  // SetOption68 - Enable multi-channels PWM instead of Color PWM
      PinUsed(GPIO_PWM1)) {
      Buzzer.pwm_mode = 1;
  }
  else {
    Buzzer.pwm_mode = 0;
  }
```
The PWM buzzer mode depends on Buzzer.pwm_mode which can only be enabled if multi-channels PWM with SetOption68 is also enabled. IOW, you can't use light control and PWM buzzer at the same time.

I asked in the TasmotaUsers group how to best control on/off an individual PWM channel in multi-channel PWM mode directly, and using "light control" is the most high-level and recommended way to do it. analogWrite() did not work and was pointed out to potentially break a lot of things:

https://groups.google.com/g/sonoffusers/c/GxOUfRmHLSg/m/jR-2MawzAgAJ

If you enable multi-channel PWM, the web UI offers to turn on/off the individual channels interactively, the same way I do it here programmatically, so I wouldn't expect this to break anything and be quite save, actually.

2. Why PWM?
Piezoelectric (cheap) buzzers need an oscillating input. See
https://en.wikipedia.org/wiki/Buzzer#Piezoelectric_2

3. Can't control the PWMFrequency?
Not true - one can use PwmFrequency command: https://tasmota.github.io/docs/Commands/
The duty cycle setting for PWM1 should be left the default 50% but I found that changing it gives slightly different sound, so one can adjust it to make the buzzer sound best.

s-hadinger: Thanks for you answer. I don't understand why you decided that Buzzer and Lights should not be used at the same time. If we are to make this feature available to anyone, we shouldn't have this constraint.

Please just use `analogWrite()` on `GPIO_BUZZER`. This would make it much easier to understand for users, not needing to change SO68 and not conflicting with PWM lights. This way also you make sure you that Gamma correction will not interfere.

When you say `analogWrite()` did not work, I assume it is because you did it on `GPIO_PWM1` which was in direct conflict of the light management code.

As mentioned, the only caveat is that the `PWMFrequency` should be the same for lights and buzzer.

hello-world-dot-c: Thanks, I did not realize that one could use use analogWrite() on GPIO_BUZZER and the platform would then apply PWM to it, I thought that a PWM output could only work on a configured GPIO_PWMx pin. Of course if that works it is the much better option. I will try this and get back to you.

hello-world-dot-c: This is now changed and works well. Again, thanks for the suggestion.

Sorry it took a while because I noticed that the generated beeps are sometimes shorter, sometimes longer and I had to do testing to make sure it's not only happening when using the new frequency output but also the plain/static output for the beeper. Measuring the output, it looks like this:

image

Even visually, you see that the times when frequency/sound is created are not always equally long. When turning the frequency output off and using it in the original static mode, the same thing is happening, too, though:

image

This is with the command "Beep -1" which does the endless generation of beeps. I don't know how the timing for the beeper module is generated, and anyway it's not related directly to this change but should rather be a bug report. I'll post it to https://github.com/arendst/Tasmota/issues but if you already have some pointers where to look for the cause I can try to look into it.

tasmota/settings.h
```diff
@@ -121,7 +121,7 @@ typedef union {                            // Restricted by MISRA-C Rule 18.4 bu
     uint32_t zerocross_dimmer : 1;         // bit 17 (v8.3.1.4)  - SetOption99 - Enable zerocross dimmer on PWM DIMMER
     uint32_t remove_zbreceived : 1;        // bit 18 (v8.3.1.7)  - SetOption100 - Remove ZbReceived form JSON message
     uint32_t zb_index_ep : 1;              // bit 19 (v8.3.1.7)  - SetOption101 - Add the source endpoint as suffix to attributes, ex `Power3` instead of `Power` if sent from endpoint 3
-    uint32_t spare20 : 1;
+    uint32_t buzzer_freq_mode : 1;         // bit 20 (v8.3.1.8)  - SetOption102 - Use frequency output for buzzer pin instead of on/off signal
```
s-hadinger, line 124: You need to catch up with setoptions now.

s-hadinger (commented):
This looks better, thanks!

tasmota/settings.h
hello-world-dot-c: I've had to learn how to do that (merge with upstream) and moved the setting to SetOption110. I hope I didn't screw it up, if I did I am sorry. I am not completely fluent in git yet ...

hello-world-dot-c: OK, I made a mess in the meantime while trying to fix it. Did this cause the pull request to close? It should be in a good state now, though, with only the two relevant files changed compared to current upstream development.

Do you advise me to delete this pull request and open a new one, or can you re-open this one?

hello-world-dot-c: It is now using the free SetOption110, but I am unsure about the version note in the comment and the free_f*a[] change. Can you just change it as appropriate according to your conventions and merge?

hello-world-dot-c:
There are now failing checks in the pull request, but afaics they are coming from other code changes not mine.

tasmota/xdrv_24_buzzer.ino
stefanbode: hi, I have a lot of positive and negative experience with the PWM due to all of my shutter related drivers. And also the AC-Dimmer that resync the PWM signal to follow the slightly changes in the AC-Frequency. From my experience, the PWMfrequency and the PWM implementation in the SDK is very stable and accurate (since june 2020). Also with wifi transmission in the background. I will check the buzzer, but if this uses normal PWM functionality any out-of-sync is very unlikely.

hello-world-dot-c: The PWM frequency and generation itself is rock-solid, I agree. The timing issue lies solely in the buzzer implementation. As you can see in the bug report I filed, the exact same issue is present in the current firmware when using the Buzzer command, not using PWM/frequency output at all but static on/off output.

It seems more related to how often or when the buzzer module is called or what timing reference is used to call it. As far as I observed, the 'on' phase can become shorter and the 'off' phase can become longer, but never the other way around, at least with the non-inverted buzzer.

stefanbode: If it is the first I agree. There is no guarantee in TASMOTA to get every50ms control over something. Sometimes the step is just missed if there is too much load. But in this case also a PWM does not help. To shift the on/off into a RTC timer is a bit overengineered but would solve the missing 50ms loop.

hello-world-dot-c: With the Buzzer -1 command, the buzzer GPIO pin is turning on/off every 100 ms indefinitely, this is what I used. In your case using the current implementation, the output would look like this:

image

Every high and low phase is approx. 100 ms here. This is turning an active buzzer such as yours on and off. When using a passive piezoelectric buzzer, this won't do anything, though, because the buzzer doesn't generate the sound itself but is more like a loudspeaker so you need to feed it the frequency it should use to generate a beeping sound. That's what the new option does. In this diagram:

image

the solid bars are again 100 ms wide but in this case, during the 'on' phase the 50% PWM frequency signal is generated to generate the buzzer sound. You can't see the PWM signal because of the scaling.

hello-world-dot-c: PWM here is not meant to solve the timing issue, maybe a misunderstanding here. This pull request is only to add the support for passive buzzers.

tasmota/settings.h
hello-world-dot-c: Caught up with setoptions and is now using SetOption111.

hello-world-dot-c:
Thank you for merging! Please remove the 'on hold' label, thanks.

alfs:
Thanks, this works great with my buzzer. It also allows playing simple melodies, e.g. using

Backlog PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1760; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1976; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1318; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1046; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4;

Perhaps frequency can be included in the Buzzer command, or an extended Buzzer command if the syntax is not simple to extend?

hello-world-dot-c:
> Perhaps frequency can be included in the Buzzer command, or an extended Buzzer command if the syntax is not simple to extend?
The point was to keep all syntax as-is so no breaking change and require virtually no additional code space. But sure, there is the Tone.cpp module in the Arduino framework and libraries such as EasyBuzzer that would support generating melodies without using PWM and analogWrite(). But this would add to the code size with not too many people using it so this could push it into a build option that is excluded for regular releases. The buzzer command format is also not easy to extend without breaking it or make it unwieldy to use.

alfs:
Actually, tone() did not work for my buzzer (needs the analogWrite() method). I had a look at EasyBuzzer, and, while not that much code (basically a wrapper with timers), it also employs tone() for generation (https://github.com/evert-arias/EasyBuzzer/blob/master/src/EasyBuzzer.cpp#L149). Not sure if this is true for all passive buzzers or only the model I happen to have (https://www.kjell.com/se/produkter/el-verktyg/arduino/moduler/passiv-summer-for-arduino-p87887 )

For another project, I did a short function mytone() that incorporates the frequency setting and the analogWrite:

```
void mytone(unsigned int frequency, unsigned long duration) {
  analogWriteFreq(frequency);
  analogWrite(PIN_BUZZER,500);
  delay(duration);
  analogWrite(PIN_BUZZER,0);
}
```

However a problematic part is the delay() that blocks execution. It was fine for that simple project, but for tasmota it needs some other timing - e.g. using EasyBuzzer's update() method and checking millis().

And when playing a melody, it's quite evident when there are other things blocking execution and delaying update() for even a slight time :)

Anyway, adding a Melody command could be an option, accepting an array of notes and durations, calling
playMelody(char notes[], int duration[], int num_notes) {
  // loop over notes, e.g. in https://www.arduino.cc/en/tutorial/melody
}

and employing a millis() delay strategy. I'll have a stab at an implementation if I get some time in the near future.
The goal would be to play Axel-F using MQTT.

hello-world-dot-c:
Yes, you can't use delay() in Tasmota and even the millis() strategy won't give you enough accuracy, see issue https://github.com/arendst/Tasmota/issues/9265. Without using a timer-based mechanism you won't get accurate enough timing for melody playback I am afraid. And based on the shot caller's comments including arendst's, this ins't likely to be accepted in the main branch. Of course one can always fork...

ozett:
sorry to bother in, 
but is the `setoption111 1 ` now useable in the development branch?
on the commands-documentation is a gap, this option is not documented,
so how to use (if possible?)

[use it like here :](https://github.com/arendst/Tasmota/pull/8994/commits/cc6a8a1c126dce95388be5af084ed804702f1eb7#diff-97a7d8fe2452e95fc805ae942e3ffd04dc451d040d68504380f0ec18a1557b45R100): -> 
```buzzer 1,3,3,0xF54,1750``` ?

is that correct? i have the dev-branch installed , but can only hear some pwm-sounds, but mostly not that loud.. 


i can use the same buzzer if i connect it to a pwm-configured pin and get (loud) sounds with:
```
SetOption15 0
SetOption68 0
PwmFrequency 1000
PwmRange 1023
Backlog PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1760; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1976; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1318; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1046; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4; PWMFrequency 1396; Buzzer 1,4;
```
(and user the buzzer-cmd in the backlog-command only for shorter delay timings)

PR #9023, opened 2020-08-03
Added exposure of MCP Out Pins as Relays

stefanbode:
## Description:
MCP can define GPIO and output pin. Currently, these are only accessible through sensor29 command. The new functionality will add the OUT PINS as additional RELAYS after the already defined one. The MCP is defined as a sensor but is is also a driver, therefore a bit xsns and xdv. Need to add the hook for FUNC_PRE_INIT and SET_POWER to interact with relay state and ensure the last state of the relay is retained after reboot. FUNC_INIT does not allow to come up with the correct RELAY state after reboot.

## Checklist:
  - [x ] The code change is tested and works on Tasmota core ESP8266 V.2.7.3.2

### DIFF
--- a/tasmota/support_tasmota.ino
+++ b/tasmota/support_tasmota.ino
@@ -207,6 +207,7 @@ void SetDevicePower(power_t rpower, uint32_t source)
 
   XdrvMailbox.index = rpower;
   XdrvCall(FUNC_SET_POWER);               // Signal power state
+  XsnsCall(FUNC_SET_POWER);               // Signal power state
 
   XdrvMailbox.index = rpower;
   XdrvMailbox.payload = source;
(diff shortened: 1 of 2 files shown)


Jason2866:
Nice!

ionultd:
@stefanbode Nice work but how does one add another MCP23017 to the works. Can one connect Wemos SDA&SCL to two MCP's or does one daisy chain one to the other via a GPIO?

Jason2866:
@ionultd i2c is a bus system. Adding additional i2c is just connecting to the bus (sda / scl).
In praxis just connect all sda together and all scl. The needed pull up resistors are normally 
already on the test boards you can buy.
Every device needs a driver to work. If more than one of the same device on the bus it needs
a different unique address. The driver has to take care for multiple use of the same i2c device too

ionultd:
@Jason2866  so are you saying that if I connect another MCP23017 chip it will need to have a new unique I2C address?

Jason2866:
Correct!

ionultd:
Ok but now one would have to add code to Tasmota to get this to work?

Jason2866:
If there is a use case maybe someone will do.

PR #9079, opened 2020-08-12
Change triple-mode TLS via configuration in a single firmware (TLS AWS IoT, Letsencrypt and No-TLS)

s-hadinger:
## Description:

This PR allows to have a multi-mode TLS firmware and moves some compile time options to configuration.

If you compile with:
```
#define USE_MQTT_TLS
  #define USE_MQTT_TLS_CA_CERT
  #define USE_MQTT_TLS_FORCE_EC_CIPHER
  #define USE_MQTT_AWS_IOT
```

You will get the 3 modes via `SetOption`
- AWS IoT with private device certificate and Amazon CA
- Mosquitto with or without device certificate, and Letsencrypt CA
- No-TLS

Options:
- `SetOption103 1` to enable TLS, `SetOption103 0` to disable TLS. Note that if the port if `8883` or `8884` TLS is always enabled (auto-config)
- `SetOption104 1` to disables retained messages (required by AWS IoT or Losant), `SetOption104 0` (default) to allow retained message. If the MQTT host is in the form `*.iot.*.amaconaws.com` then this option is automatically set (auto-config).
- If the MQTT host is from AWS IoT or Mosquitto using port `8884`, then TLS will wait to have a private key and certificate configured before trying to connect. This is to avoid failed connection.

Note: if you don't compile with TLS enabled, there is no change. If you compile with TLS and disable it, it does not take the RAM that TLS would require.
Note: you can still custom compile with only what you need.
## Checklist:


xhess:
Hi @s-hadinger ,
how did you manage to compile esp32 with USE_MQTT_TLS enabled? When I enable it I get a compile error: 
> In file included from git/Tasmota/tasmota/xdrv_02_mqtt.ino:25:0:
> tasmota/WiFiClientSecureLightBearSSL.h:58:12: error: 'size_t BearSSL::WiFiClientSecure_light::peekBytes(uint8_t*, size_t)' marked 'override', but does not override
>      size_t peekBytes(uint8_t *buffer, size_t length) override;
Since you added code for tls I thought you would test ist since you mention it under your checklist. But if you just used the default user_config then the code you added did not get tested?

s-hadinger:
This tls code is only for esp8266. Esp32 uses mbedtls and it's a completely different library.

PR #9082, opened 2020-08-13
i2s audio support

gemu2015:
## Description:

plays mp3 files from file system 

and synthesizes text to speech (no need for file system e.g.:  i2ssay hello)

supports various i2s audio devices like:
https://www.amazon.de/Adafruit-I2S-Class-Amplifier-Breakout/dp/B01K5GCFA6
and TTGO ESP32 watches

not yet tested with esp8266 because of library issues in esp8266
at least for mp3 files even the ESP32 must work with 240 Mhz

text to speech: 
i2ssay  text
set gain:
i2sgain volume  (1-100)
play mp3
i2splay /sound.mp3

## Checklist:


arendst:
Is there any chance this will be proted to ESP8266 too? If not I want to rename the driver to ``xdrv_83_i2s_audio.ino`` to distinguish it as ESP32 only.

gemu2015:
the libraries seem to be in active development. I could not figure out yet why they don't compile in ESP8266.
there are several unresolved references to files.
if will check why this happens when i find the time. Any help appreciated.

gemu2015:
ok, got it running.
platformio needs a deeper library search
lib_ldf_mode = deep+

this resolves the references in audiolib and seems not to have any side effects.

text to Speech works now an esp8266 even on 80Mhz

however mp3 will not work in tasmota esp8266 because is needs at least 30k of RAM

will do a pr later

jesse-schein:
Nice work on this. Was able to get it to mostly work with M5stack core 2. I switched the partition to spiffs though so I could upload files via platformio. I didn't understand how to use ffat and upload files.

gemu2015:
the Tasmota file system is bound to scripting.
you need to enable scripting and if you have an SD card also SPI.
without an SD card you may use part of the flash for flash file system.
you must enable the web file interface by
#define SDCARD_DIR

the file interface is then accessible in the script editor menu as a submenu 
you may upload (and download) files there.

check the Tasmota scripting docs for more info

jesse-schein:
> the Tasmota file system is bound to scripting.
> you need to enable scripting and if you have an SD card also SPI.
> without an SD card you may use part of the flash for flash file system.
> you must enable the web file interface by
> #define SDCARD_DIR
> 
> the file interface is then accessible in the script editor menu as a submenu
> you may upload (and download) files there.
> 
> check the Tasmota scripting docs for more info

Nice I'll have to check that out. My board has an asd card slot but I think it's way overkill to use it. I have 16MB of storage total. 

I did manage to get scripting working with the proper defines I just didn't realize there was a special interface to upload files with another define. Will mess with it some more, I'd like to use ffat system since it's faster.

gemu2015:
ok then i would suggest to use the linker file
esp32_partition_app1984k_ffat12M.csv
and have 12 mb of FFAT storage.
and use 
#define USE_SCRIPT_FATFS -1
#define SDCARD_DIR

gemu2015:
nice board by the way, i only knew the previous version.
i ordered this new one too now and will implement all features during the next few weeks.

jesse-schein:
> nice board by the way, i only knew the previous version.
> i ordered this new one too now and will implement all features during the next few weeks.

Yea I think this board is amazing for $40. I've pretty much been able to get every sort of framework to run on it. I got LVGL running on it just fine. The fact that is has 16MB total flash is awesome, plus the psram and all the components.

By the way I did add those defines and was able to see my files on the website script config page, so thanks for help with that. I was able to get it to work in addition to all my SPIFFS changes but I will work to change it back to FFAT since all that code I modified can be undone and work as intended. The only benefit to my way is the ability to upload the file system from the computer, which can be useful maybe I don't know.

gemu2015:
@jesse-schein 
i meanwhile received my core2 and implemented power manager, sdcard, display and i2s.
too early for pr but you might want to test my fork or even contribute.

https://github.com/gemu2015/Sonoff-Tasmota

#define USE_SCRIPT
#define USE_SCRIPT_FATFS 4
#define FAT_SCRIPT_SIZE 8192
#define USE_SCRIPT_FATFS_EXT
#define SDCARD_DIR

#define USE_I2C
#define USE_M5STACK_CORE2
#define SAY_TIME
#define USE_WEBRADIO
#define USE_DISPLAY
#define USE_DISPLAY_ILI9342

in Order to get the display driver to work you must define SCL and SDA on Pins 32,33 because the SPI bus cannot be defined  (it uses Pin 38 which cannot be defined in Tasmota GUI) and Tasmota requires a BUS to enable the display driver. (SPI bus is therefore hardcoded in CORE2 driver)

according i2s ram usage: i already preallocated some memory in psram however for speed reasons a lot of program code has been put to ram also.

jesse-schein:
> @jesse-schein
> i meanwhile received my core2 and implemented power manager, sdcard, display and i2s.
> too early for pr but you might want to test my fork or even contribute.
> 
> https://github.com/gemu2015/Sonoff-Tasmota
> 
> #define USE_SCRIPT
> #define USE_SCRIPT_FATFS 4
> #define FAT_SCRIPT_SIZE 8192
> #define USE_SCRIPT_FATFS_EXT
> #define SDCARD_DIR
> 
> #define USE_I2C
> #define USE_SPI
> #define USE_M5STACK_CORE2
> #define SAY_TIME
> #define USE_WEBRADIO
> #define USE_DISPLAY
> #define USE_DISPLAY_ILI9342
> 
> in Order to get the display driver to work you must define SCL and SDA on Pins 32,33 because the SPI bus cannot be defined (it uses Pin 38 which cannot be defined in Tasmota GUI) and Tasmota requires a BUS to enable the display driver. (SPI bus is therefore hardcoded in CORE2 driver)
> 
> according i2s ram usage: i already preallocated some memory in psram however for speed reasons a lot of program code has been put to ram also.

Nice work. I was also able to get the display to work. I just imported the m5core2 library and refactored the tasmota display driver a bit, mostly name rewrites, the signatures are fully compatible with tft_espi, must be what tasmota based it on a while ago. Ill check your repo out. 

By the way, I found out if you use the RTC deep sleep methods on the m5core2 current draw is around 250uA, so quite low. This could be a really good tasmota remote logging board. Would last a good while on the internal battery.

gemu2015:
ok, i implemented the command core2shutdown x   x= seconds >=30
works nice, now we have to test what battery life we can achieve with different timings .

PR #9095, opened 2020-08-14
Teleinfo Added setOption108

hallard:
## Description:

- Added `SetOption108 0/1` to enable Teleinfo telemetry into Tasmota Energy MQTT (0) or Teleinfo only (1) in this case MQTT will send RAW Teleinfo telemetry on each frame received and not into Tasmota energy calculation telemetry.
- Fix bug un `lib_teleinfo` when checksum was space

## Checklist:


tasmota/xnrg_15_teleinfo.ino
```diff
@@ -324,6 +319,62 @@ void NewFrameCallback(struct _ValueList * me)
...
+
+        // Total size we can use, remove { } and \0
+        int remaining_size;
+        char sep = ' ';
+        char * p ;
+        int len;
+        boolean isNumber ;
+
+        strcpy_P(mqtt_data, PSTR("{"));
+        len = strlen(mqtt_data);
```
arendst, line 336: Dive into the use of Response function. This increases code size with redundant code.

hallard: Theo,

Not sure to understand, the deal is that code happens on frame reception so mainly called in 250ms function not in Web or JSON answer.
just followed the https://tasmota.github.io/docs/Sensor-API/ MQTT part, now if a better solution exists I'll take it but not sure what to do.

arendst: Any manipulation of mqtt_data should be done using the provided respond functions. Search code and it wil become clear. No need for strcat or printf

hallard: Ok, got it, just followed documentation should have dig into the code, I'll fix that tomorrow.

```
      Response_P(PSTR("{"));
      ResponseAppend_P( bla bla );
      ResponseJsonEnd();
```

hallard: Theo,
That's done

s-hadinger (commented):
Maybe you could consider keeping some configuration in Ram instead of SetOption. This could be set via a Rule at startup.

PR #9104, opened 2020-08-15
Add Virtual CT for 4 channels lights (Phase 1)

s-hadinger:
## Description:

Users reported an increasing number of LED bulbs advertised as RGBCW (RGB + Cold White + Warm White) but actually containing only 4 channels (RGB + warm or cold White). The missing white is emulated using RGB. Users have been asking for the same feature to emulate the missing 5th channel.

Be aware that results may be very disappointing:
- RGB are often much less bright than the White LED
- RGB produce blurry white (imperfect mixing of the colors)
- Best results are obtained if the White LED is either COLD or WARM, and the RGB White produces the opposite white. All other variations will not give good results. Ex: 4000K (natural white) LEDS will not allow to produce good Cold or Warm White.

**This implementation is experimental and open to feedback** There may be a Phase 2 to improve results.

`SetOption106 1` enables Virtual CT. Tasmota reboots and shows a cold/warm slider.
Note: works only for 4-channels lights, and only if `SetOption68 0` and `SetOption37` < 128.

`SetOption107 1` indicates that the hardware White LED is COLD (warm is emulated). `SetOption107 0` indicates that the hardware LED is WARM (Cold is emulated).

### Change in White Blend mode
To avoid interfering with Virtual CT, White Blend Mode now uses `SetOption105 1` instead of `RGBWWTable`.

At startup, if white blend mode was enabled via `RGBWWTable`, `SetOption105 1` is automatically enabled and the 5th channel of `RGBWWTable` is set to 255 (its default value).

## Checklist:

### DIFF
--- a/tasmota/CHANGELOG.md
+++ b/tasmota/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Add better config corruption recovery (#9046)
 - Remove support for 1-step upgrade from versions before 6.6.0.11 to versions after 8.4.0.1
+- Change White blend mode moved to using ``SetOption 105`` instead of ``RGBWWTable``
+- Add Virtual CT for 4 channels lights, emulating a 5th channel
 
 ### 8.4.0.1 20200730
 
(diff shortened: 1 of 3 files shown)


gjmwolmarans:
@s-hadinger @arendst I have set up virtual ct, and white blend with warm white RGBW 4 channel lights. But I've noticed that when the color temperature is at its' warmest, and then setting saturation to 0, the lights don't use the white channel, therefore no light. With color temperature at its' coldest, it uses white blending and all the channels are on.

s-hadinger:
@gjmwolmarans Thanks for reporting this. Could you please open an issue with all the required fields, especially your precise version and steps to reproduce.

PR #9105, opened 2020-08-15
Add support for #Axdiv10 to ADS1115 sensor module

mrbig:
## Description:

The default ADC sensor module has a nice feature to report analog value changes if the input changed more then 1%. This feature was missing from the ADS1115 module, wich also provides analog inputs.

To provide the same functionality as the ADC sensor module this PR adds support for the #Axdiv10 selector wich can be used in rules to trigger immediately on a relevant input change. Works exactly like in the [ADC documentation](https://tasmota.github.io/docs/ADC/#rule-triggers).

## Checklist:
  - [ X] The code change is tested and works on Tasmota core ESP8266 V.2.7.3.2
  - [X ] The code change is tested and works on core ESP32 V.1.12.2

### DIFF
--- a/tasmota/xsns_12_ads1115.ino
+++ b/tasmota/xsns_12_ads1115.ino
@@ -180,6 +180,61 @@ void Ads1115Detect(void)
   }
 }
 
+#ifdef USE_RULES
+// Check every 250ms if there are relevant changes in any of the analog inputs
+// and if so then trigger a message
+void AdsEvery250ms(void)
+{
+  int16_t value;
+
+  for (uint32_t t = 0; t < sizeof(Ads1115.addresses); t++) {
+    if (Ads1115.found[t]) {
+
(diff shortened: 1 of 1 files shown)


arendst:
Thx for your PR.

The way you implemented it will only provide a 1% change at teleperiod time and not, like the default ADC at 1% change time. It is therefor not completely the same but could work on slow changing ADC inputs.

mrbig:
Thanks @arendst I'm new to Tasmota development, and might misunderstood something.
I didn't want to change the module a lot, and to me it looks like that it only checks the analog inputs when a teleperiod or a rule asks.
I'm using this solution from a rule, and since it will poll the sensor in each loop, the rule will react almost immediately.
If I understand you correctly then sending an event right when the change happens would need a more extensive rewrite of the sensor module.

PR #9166, opened 2020-08-26
Shelly 2.5 separate power measuring

cskarai:
Hi Tasmota developers,

Tasmota doesn't support separate channel measuring for Shelly 2.5.
The problem is difficult, so this pull request is rather a patch for those who want to measure the channels separately.

Limitations:
- EnergyReset1-3 equally splits the value amongst the channels
- Only the webserver displays separate measurement data, MQTT, Domoticz are not covered
- It measures only Shelly 2.5 separately, the other 3-phase products are not supported (yet)
- Structural change in RtcSettings (backwards incompatibility)

I think, that in the long term separate power measuring for dual channel products is important. While it's not done, this temporary patch might handle the problem.

shelly2 5


sfromis:
Withut MQTT support, I'd say that the value of this is, well, limited.

cskarai:
> Nice. Without MQTT support, I'd say that the value of this is, well, limited.

I don't have millions of time to develop software that I will never use.
If the developers are open to merge this feature, then I can work a bit more on this pull request.

Otherwise this patch is enough for me.

arendst:
Currently uses too many resources for too little profit.

PR #9176, opened 2020-08-27
Add ``SetOption109 1`` to force gen1 Alexa mode, for Echo Dot 2nd gen device only

s-hadinger:
## Description:

Echo Dot 2nd gen devices do not support simple switches devices with Hue Emulation - they need to always have a (fake) dimmer. The issue happens if you have only this type of device. If you have more modern Echo devices on the same network, there is no issue.

Before, the detection was automatic via User-Agent, but it does not work anymore. Instead you must now use `SetOption109 1` to force legacy support of Echo Dot 2nd gen only.

Also added more logging in Hue and Wemo emulation to track the source IP of the requests.

## Checklist:

### DIFF
--- a/tasmota/CHANGELOG.md
+++ b/tasmota/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add command ``PowerDelta1`` to ``PowerDelta3`` to trigger on up to three phases (#9134)
 - Add Zigbee web ui widget for Lights
+- Add ``SetOption109 1`` to force gen1 Alexa mode, for Echo Dot 2nd gen devices only
 
 ### 8.4.0.2 20200813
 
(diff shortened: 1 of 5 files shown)


slim0287:
@s-hadinger 
I have a question.  In my current setup with ~10 Tasmota devices, 3 of which use Hue emulation I have working thanks to you and another person on a different thread.  I have an Echo Gen 1 and an Echo Dot Gen 2.  As such, I use `SetOption109 1` on the devices with Hue Emulation.  I am replacing the Echo Gen 1 tomorrow with an Echo Dot Gen 3.  Should I change the Hue emulation devices back to `SetOption109 0`?  And should I delete and re-discover them?
Thanks.

s-hadinger:
@slim0287 Yes. As soon as you set `SetOption109 0`, the device may not be recognized any more. Remove it and trigger a new discovery.

This applies only to switches and simple bulbs (on/off), not to dimmer, RGB or CT lights - for which SetOption109 has no effect.

Issue #9183, opened 2020-08-28
Add support for infrared thermal sensor array MLX90641 / MLX90640


balajisraghavan:
**Have you looked for this feature in other issues and in the docs?**  
Yes
**Is your feature request related to a problem? 
No
**Describe the solution you'd like**  
Add support for the I2C Infrared thermal sensor array [MLX90641](https://www.melexis.com/en/product/MLX90641/High-Operating-Temperature-FIR-Thermal-Sensor-Array) & [MLX90640](https://www.melexis.com/en/product/MLX90640/Far-Infrared-Thermal-Sensor-Array). It has a variety of uses - human presence detection, air leaks, a low cost thermal camera & many more.

**Additional context**  
I'd like Tasmota to publish the temperature at each of the sensors in the array over MQTT during every tele-period. Some kind of rule support such as min/max temperature detected exceeding threshold would be a great addition as well. A reliable open source [Arduino driver](https://github.com/sparkfun/SparkFun_MLX90640_Arduino_Example/tree/master/Firmware/Example1_BasicReadings) is available for this sensor.

Staars:
I have a MLX90640 and will try write a driver in the next few weeks, which I need too for a personal project.

balajisraghavan:
Thanks @Staars. I got the sensor working over the weekend and raised a PR. It's my first contribution to the Tasmota codebase, would be great to receive feedback on what can be done better

ascillato2:
Closing this issue as this feature has been added. Thanks.

Erickrk:
Hello everyone,

I hope you are doing well. I wanted to check if anyone is still actively testing this project. I tried running it myself, but it doesn't seem to be very stable. I am not sure if it's because I am running a script while working with the camera or if there are issues with the driver itself.

During my testing, I noticed that my heap usage goes up to 130 on my ESP32 while running this project with an MLX90640. Additionally, it seems like the WEB UI might be causing some instability. I have noticed that whenever I was able to break the device, it was due to issues with the WEB UI.

If anyone has any suggestions or insights on how to improve the stability of this project, it would be greatly appreciated. Thank you for your time and help.

Staars:
Sounds like your system is indeed under heavy load.

The MLX90640 was primarily designed to be used with the MQTT messages, the Web UI should only make it easy to position the POI's (and it looks quite fancy 😄 ).
[...]

PR #9195, opened 2020-08-30
Support for MLX90640 IR sensor array

balajisraghavan:
## Description:

Changes to support the MLX90640 sensor. See https://github.com/arendst/Tasmota/issues/9183 

## Checklist:
  - [ ] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.1

### DIFF
--- a/lib/Adafruit_MLX90640-1.0.2/Adafruit_MLX90640.cpp
+++ b/lib/Adafruit_MLX90640-1.0.2/Adafruit_MLX90640.cpp
@@ -1,0 +1,202 @@
+#include <Adafruit_MLX90640.h>
+
+/*!
+ *    @brief  Instantiates a new MLX90640 class
+ */
+Adafruit_MLX90640::Adafruit_MLX90640(void) {}
+
+/*!
+ *    @brief  Sets up the hardware and initializes I2C
+ *    @param  i2c_addr
+ *            The I2C address to be used.
+ *    @param  wire
+ *            The Wire object to be used for I2C connections.
+ *    @return True if initialization was successful, otherwise false.
+ */
+boolean Adafruit_MLX90640::begin(uint8_t i2c_addr, TwoWire *wire) {
+    i2c_dev = new Adafruit_I2CDevice(i2c_addr, wire);
+
+    if (!i2c_dev->begin()) {
+        return false;
+    }
+
+    wire->setClock(400000);  // Speed it up, lots to read :)
+    
+    // MLX90640_I2CRead(0, MLX90640_DEVICEID1, 3, serialNumber);
+    delay(0);
+    uint16_t eeMLX90640[832];
+    if (MLX90640_DumpEE(0, eeMLX90640) != 0) {
+        return false;
+    }
+
+    delay(0);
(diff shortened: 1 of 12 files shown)


balajisraghavan:
MLX90640 sensor works fine with the ESP32. On ESP8266, I'm getting a watchdog reset due to the amount of math needed, despite yielding to WDT frequently at 160 MHz. Max temperature is shown on the UI, max temperature and all 24x32 temperatures are published on MQTT.

Jason2866:
#define MQTT_MAX_PACKET_SIZE 5700 is a no go.
You break a lot of stuff in Tasmota with this setting
You have to find a way to reduce to use the setting of 1200 for

balajisraghavan:
Thanks for the feedback @Jason2866. Could you please share more info on which features would break if the `MQTT_MAX_PACKET_SIZE` is `5700` on an ESP32? I see 111KB free mem in the Tasmota information page after these changes. The sensor is limited to the ESP32 for now (notice the `#ifdef ESP32` directive).

There are many ways to reduce this dataset such sampling, publishing min/max for each row, Shox96, if we have to.

Jason2866:
Your code will never be merged if you change the mqtt packet size. There are a lot of dependencies 
If you are really interested look in the code. We can not an we will not explain here all details.
Github is not the place for.

balajisraghavan:
> Main code for ESP82XX and ESP32 has to be the same
I see. This sensor provides 768 temperature readings, we need 4610 bytes to send all of it. With 1200 bytes, we can send 200 data points, which is also quite useful.
@Staars you'd expressed interest in this sensor earlier. Could you please share your thoughts on if you see any issues with downsampling to 200 readings? It'd be great if you could also help with the below WDT reset issue I'm facing on the ESP8266
@arendst could you please share your thought on using this sensor as an ESP32 specific sensor? That way, we can offer a more robust dataset to users who want to use the sensor. Also, on the ESP8266, I'm seeing a WDT reset when we're doing the math after reading the data (despite yielding frequently).

Staars:
I have no personal use case for sending that much unfiltered data points and prefer to define some points of interest (maybe 2 - 10 POIs) which should be observed via MQTT. These POI's should be definable at runtime.
Bonus would be a WebGUI to define these POI's, which I already tested as proof of concept.

For me there are no obvious reasons, why this should be impossible on the ESP8266. Some of the original library functions from Melexis would have to be refactored, but that should be doable.
To prevent a WDT you "only" have to divide the big chunk of work in many small sequential tasks and use the callback function inside Tasmota.

Staars:
Meanwhile I started with my effort to write a driver and I can safely say, that it will be possible on the ESP8266.
I have started with the WebGUI, which already works. My idea is, that the points of interest should be defined here (or at least validated).
At the core the interesting POI's will then be published via MQTT, which I still have to implement.

mlx90640_1

Jason2866:
Thx for your work. This PR will not be merged. Tasmota uses PR #9294 to support this sensor.

ascillato2:
Thanks for sharing your work.

Closing as this feature has been added.

PR #9210, opened 2020-09-01
Support for VL53L1X

Jason2866:
Thx @device111 for refactoring

I added Domoticz again (done like in driver VL53L0x). Hope that is correct. I have no clue about Domoticz

## Checklist:


stefanbode:
I checked the changes and also agree that it should work. Thumbs up

joba-1:
I just tested this and have some questions:

1.
According to [datasheet (Page 4 Section 1.1)](https://www.st.com/resource/en/datasheet/vl53l1x.pdf) default address is 0x52.
This does not conflict with existing tasmota I2C devices, so why use 0x29 which has conflicts?
I tried to change the default address in [the lib](https://github.com/arendst/Tasmota/blob/86e44b0bbfa14959ef284dc5029c3f68618f249f/lib/vl53l1x-arduino-1.01/VL53L1X.h#L1311) and [the driver](https://github.com/arendst/Tasmota/blob/86e44b0bbfa14959ef284dc5029c3f68618f249f/tasmota/xsns_77_vl53l1x.ino#L36) but that did not help - I probably misunderstand how tasmota handles I2C adresses?
 
2.
Why use illuminance for domoticz values? It should be a distance.

Jason2866:
I copy the perfect answer from sfromis from Discord

> the I2C address is given including the lsb read/write bit, meaning that if the actual address 0x29 is shifted one bit to the left, it becomes 0x52.

Dont ask me about Domoticz, just copied from driver VL53L0X :-)

Theo uses Domoticz, he did not changed this part so it seems to be correct

ghost (approved):

arendst:
Regarding Domoticz, Tasmota dies not support Domoticz distance directly. The domoticz command for distance is the same as for illuminance so that's why you select illuminance index in Tasmota's Domoticz configuration screen.

It's all a matter of using less code for the same result.

joba-1:
ok, fine with me then :)  Rest was clarified on discord. Thanks

joba-1:
works mostly. But if I use domoticz device general, subtype distance, then domoticz reports 1500 mm as 1500 cm.
Do you use something different or should this be adjusted in the tasmota driver?

arendst:
It would need a devide by 1000 in the VL53L1X driver.

Will do it now.

arendst:
Give it a try

Jason2866:
Probably the same issue in the VL53L0X driver. I copied 1:1

joba-1:
wait until you change that too. 1000 seems a bit to much... Testing...

joba-1:
ok, as suspected. Now it says 1.5cm. Divide by 10, not 1000 should fix it...

joba-1:
dtostrfd((float)vl53l1x_sensors.distance / 10, 1, distance);

tested. works fine.

Issue #9098, opened 2020-08-15
Additional shutter mode - single pulse motor


Pewibe:
**Have you looked for this feature in other issues and in the docs?**  
yes
**Is your feature request related to a problem? Please describe.**  
Shutters like garage doors usually are driven by ONE momentary switch / are controlled by ONE pulse meaning OPEN-STOP-CLOSE-OPEN-STOP-etc. In other words the direction is switched internally by the door control and a 2nd relay is NOT needed. Why shutter mode for garage doors? For „side sectional doors“ a shutter mode is conveniant because moving the car in and out you need opening 100% but stepping in/out an opening of 20% might good enough. 

**Describe the solution you'd like**  
As there is shutter mode 1 and 2 for continues operated motoers, there should be similar 2 shutter modes for pulse motors (first relay: OFF/DOWN, second relay: OFF/UP or first relay: ON/OFF, second relay: UP/DOWN). It can be used for a pulse motor to start/stop and change it's direction or just by leaving the 2nd relay unconnected for the garage door case.  Because an additional criteria is needed to select shutter mode 2a or 2b I like to suggest a new and clear command "ShutterMode<x> <value> (0 = off, 1/2 = duration based motors, 2/4 = pulse motors, 5 = stepper, . . ) - SetOption80 can be reused.

**Describe alternatives you've considered**  
1. Instead of "ShutterMode<x> <value>" the SetOption80 can be extended (all ShutterRelay<x> then have the same behaviour). 
2. Allow a virtual relay instead of a physical one as 2nd relay.

**Additional context**  
The command "ShutterRelay<x>" is too unflexible because of allowing only odd numbers and force the following number as 2nd relay. Why not explicitly define both relay numbers like "ShutterRelay<x> <number of 1st relay>,<number of 2nd relay>"? 

**(Please, remember to close the issue when the problem has been addressed)**

stefanbode:
This should already work. Please check the option for the momentary relays. The function should be as following:
- Pulsetime1 and 2 on the both relays. In the Wiki is something about the minimum time.
- ensure that interlock is OFF, or at least not defined on the relays 

Please give feedback if this solves Your problem. If not we can work to make it happen.

stefanbode:
[...]

Issue #8901, opened 2020-07-14
MG90s Servo Control with Dimmer command


helmar74:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._
I'm using the precompiled ESP32WEBCAM Image. I connected two MG90s servos to GPIO14 and 15. I can control the servos with the commands channel1 xx and channel2 xx.
Example:
channel1 19 moves the servo to 0°, channel 42 to 180° (which is the maximum of this servo)
The problem is, when I enter a value lower than 19 or higher than 42 than the servo moves very fast from 0 to 180°.

I read in a few other issues, that they managed to move the servos with dimmer command. (and than I could use dimmerrange paramter)
This doesn't work for me at all. The servos doesn't move.
As I'm using two servos, I do need two "dimmer comands" (dimmer1 / dimmer2)
I tried setoption37 128, but this doesn't help either.

Here are my current settings:
setoption15 ON
setoption37 0
setoption68 ON
setoption69 ON
GPIO14 is set to PWM1
GPIO15 is set to PWM2

Has anyone an idea, what I could try further?

Regards

Helmar

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [X] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [X] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [X] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [X] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [X] Device used (e.g., Sonoff Basic): ESP32CAM
- [X] Tasmota binary firmware version number used: 8.4.0
  - [X] Pre-compiled
  - [ ] Self-compiled
    - [ ] IDE / Compiler used: _____
- [X] Flashing tools used: esptool
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```
18:54:02 CMD: Backlog Template; Module; GPIO 255
[...]

PR #9244, opened 2020-09-05
Enhance new shutter modes and code cleanup. 

stefanbode:
## Description:
#9098 fixed enhanced
#8901 fixed enhanced

Refactor the driver to support different kinds of requirements to operate relays/stepper/servos and open for future requests.
Pulsetime on relays now can be defined independently from the operation mode of the shutter. Only define pulsetime on the relay.
Added new "shuttermode" command to specify an operation mode. Default is 0 and will autodetect the mode as in the previous version as the best guess. The following modes are supported: SHT_UNDEF, SHT_TIME, SHT_TIME_UP_DOWN, SHT_TIME_GARAGE, SHT_COUNTER, SHT_PWM_VALUE, SHT_PWM_TIME,
New modes: 
- Garage mode (3). Operate Garage door only with one pulse on relay1 with automatic reverse
- Servo value (5): Operates a servo where PWM value controls the position. Also supports fancy shuttermotordelay to smooth ramp up down servo like the stepper motor is already using.
- Servo 360° will come in the future (6):

#9098 fixed enhanced
#8901 fixed enhanced

## Checklist:
  - [x ] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.1
  - [ x] The code change is tested and works on core ESP32 V.1.12.2

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -585,6 +585,7 @@
 #define D_CMND_SHUTTER_TOGGLEDIR "ToggleDir"
 #define D_CMND_SHUTTER_UP "Up"
 #define D_CMND_SHUTTER_DOWN "Down"
+#define D_CMND_SHUTTER_MODE "Mode"
 #define D_CMND_SHUTTER_STOPOPEN "StopOpen"
 #define D_CMND_SHUTTER_STOPCLOSE "StopClose"
 #define D_CMND_SHUTTER_STOPTOGGLE "StopToggle"
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -613,8 +613,9 @@ struct {
   uint8_t       free_f43[1];               // F43
 
   uint16_t      energy_power_delta[3];     // F44
+  uint8_t       shutter_mode;              // F45
 
-  uint8_t       free_f4e[106];             // F4A - Decrement if adding new Setting variables just above and below
+  uint8_t       free_f4e[105];             // F4A - Decrement if adding new Setting variables just above and below
 
   // Only 32 bit boundary variables below
   SysBitfield5  flag5;                     // FB4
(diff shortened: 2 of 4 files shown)


arendst:
Stefan, is this stable? I plan to release v8.5 tomorrow but am reluctant to include this major change. pls advice.

stefanbode:
Hi arendst. Please do NOT integrate for the next stable release. I already did a lot of testing but you never know I I do not want to break something in a stable release. You can integrate it AFTER the release of 8.5. So I also have time to update the documentation. And also check for the stepper motors.

stefanbode:
I found an Issue in the Stepper driver. Please do not integrate now. Will post an update Tuesday

stefanbode:
Now I'm pretty sure that not only the normal shutter still operates This was also OK before. Now also the smooth ramp on stepper and servos is in sync and tested hours and hours and hours... :-) Driver is now tested 100% and I would also be fine with 8.5. Depends on you. In many cases the stepper runs better than before.

stefanbode:
@arendst : Hi Theo, the servo operation PWM min and max are currently hardcoded. I see there is already a PWMrange defined in the standard. But this just seems to limit the max value. Is there already a way to limit the PWM value instead of 0..1023 to 90..500 without inventing additional variables?

arendst:
Current PWMRange command only supports 255, 511 or 1023 (being 8, 9 and 10-bit) resolutions. These values make calculation PWM values very easy.

The bottom range is always 0. 

Changing this to any value like higher low and different high values will at least for lights introduce issues. The chances of having an esp8266 control both lights and shutters are dim.

I think the best way forward and keep lights and shutter not intertwined you'll need to introduce a ShutterPwmRange command with both a low and high parameter like ``ShutterPwmRange 90,500``. This way future additions won't interfere with your shutter code. Just take two uint16_t values from the settings.h list.

pawol:
Please look at issue #9458
Tasmota 8.5.1 is affected by roller shutter timming bug

PR #9280, opened 2020-09-10
Adding Lat/Long IP-based detection

tichris0:
This patch adds the ability for Tasmota to detect the user's latitude & longitude based on their IP

This is triggered by setting the latitude/longitude to 0.0 (nowhere in the Ocean)

This is done through 2 web services (one is a backup) that are both free and don't require keys.

Tested:
* Each site individually
* Failure case

## Description:

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -245,8 +245,8 @@
 #define TIME_STD_OFFSET        +60               // Offset from UTC in minutes (-780 to +780)
 
 // -- Location ------------------------------------
-#define LATITUDE               48.858360         // [Latitude] Your location to be used with sunrise and sunset
-#define LONGITUDE              2.294442          // [Longitude] Your location to be used with sunrise and sunset
+#define LATITUDE               0.0               // [Latitude] Your location to be used with sunrise and sunset
+#define LONGITUDE              0.0               // [Longitude] Your location to be used with sunrise and sunset
 
 // -- Application ---------------------------------
 #define APP_TIMEZONE           1                 // [Timezone] +1 hour (Amsterdam) (-13 .. 14 = hours from UTC, 99 = use TIME_DST/TIME_STD)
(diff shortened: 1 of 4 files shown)


tichris0:
I don't have an ESP32 to test this against.

arendst:
Thx. Nice idea but as it is used only once and takes code space which could be used for more "urgent" features I can only merge if this feature will be disabled by default using a define.

stefanbode:
I agree to let it undefine in default and maybe default on ESP32 where space is less an issue. This capability is more for a manufacturer, who want to sell 1000s of devices with tasmota. great idea, but haven't heard so far.

tichris0:
Alright I'll send another revision tomorrow to hide it by default with the appropriate documentation.  It would be nice to have on-boot, but I agree that it's fighting against basic functionality.

Let me know if you'd like it on for the ESP32.

arendst:
Pls off on both

tichris0:
I've updated the change.  It will only be enable if **USE_AUTOLATLONG** is defined.

stefanbode:
@tichris0 : there is something wrong with your build. All esp32 do not work. If this is correct disabled the error should not happen.

tichris0:
I'm not sure why it failed on ESP32.  It compiled on my side.  Regardless, I've addressed your cnocerns @stefanbode and also fixed a bug I noticed along the way.  Everything has compiled this time around & tested everything again except ESP32 (which I don't own).

I'm curious however as to your request to keep the LAT/LONG to the weird number it was set to.  <0, 0> seems like a safer number given that it's in the middle of the ocean whereas the currently set number seems like it could be valid for someone.

tichris0:
!ping.  This commit is ready to submit.  I'm not sure if I'm expected to do anything else here?

arendst:
Thx. Just did a test on both sites and they were able to locate me over 300km away from where I am. It seems they use the providers presence location.

I don't think this is a viable option to be merged soon.

Jason2866:
Did a test too. 
IPWHOIS.IO -> 300 km off 
ip-api.com -> **got me! Just 1km off**

tichris0:
Yah sounds good.  I'll close it for now.  Thanks

PR #9285, opened 2020-09-11
Scripter update

gemu2015:
## Description:

some bug fixes,
some reformatting
google charts new options
text variable replacement new option calculate formulas %(formula)%
fix display touch init
save memory on esp32 mp3 player with psram

## Checklist:

### DIFF
--- a/tasmota/xdrv_10_scripter.ino
+++ b/tasmota/xdrv_10_scripter.ino
@@ -136,19 +136,19 @@ Ticker Script_ticker4;
 
 void Script_ticker1_end(void) {
   Script_ticker1.detach();
-  Run_Scripter(">ti1", 4,0);
+  Run_Scripter(">ti1", 4, 0);
 }
 void Script_ticker2_end(void) {
   Script_ticker2.detach();
-  Run_Scripter(">ti2", 4,0);
+  Run_Scripter(">ti2", 4, 0);
 }
 void Script_ticker3_end(void) {
   Script_ticker3.detach();
-  Run_Scripter(">ti3", 4,0);
+  Run_Scripter(">ti3", 4, 0);
 }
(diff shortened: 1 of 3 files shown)


MomentumV:
Gemu, In the script docs, there's a typo for the google chart spline option; it reads 2f2 instead of lf2.
`- l2=linechart with exactly 2 lines and 2 y scales (must be given at end)
- 2f2 like above but with splined lines`
from https://tasmota.github.io/docs/Scripting-Language/

gemu2015:
thanks, fixed

MomentumV:
I also cannot figure out what this line means:
`("cntN" starts numbering entries with the number N an optional /X generates only an devides by X range)`

MomentumV:
After an evening trying things, I think this could be accurate:
`Using the label string "cntN" will produce labels that cycle through the array indexes, starting with the number N.`
`For instance, "cnt6" with an array of length 8 will produce the labels 6|7|0|1|2|3|4|5|
`Using "cntN/X" will then divide the numeric value of the label, so "cnt6/2" with an array of length 8 will produce the labels 3|3|0|0|1|1|2|2|`

gemu2015:
Yes this is accurate. The reason why you may need it is the following example.
lets say you want to have a value every hour. You have an array of 24 elements.
"cnt0" gives an x axis of 0-23 hours.
now you want to have more time resolution and have values every 15 Minutes.
you need an array of 96 Elements.
"cnt0/4" now gives an accurate axis of hours.
i know the documentation is preliminary. you are welcome to help by submitting pr´s to the Tasmota doc page.
i myself have a lot of examples i use in my devices. these are very specific to my needs and in german.
they show however what is possible. if i find the time i will translate and post some of them

PR #9294, opened 2020-09-12
Mlx90640

Staars:
## Description:

This driver works on the ESP8266 with a modified library.

Intended use case is to measure multiple points of interest of an object without the need to directly put a sensor on it, for instance if you are not allowed to apply hardware modifications to a heating.

The update frequency of the measurements is approximately 1 per second. In the first driver version 6 POI‘s are freely selectable (POI1..6). POI-0 is reserved for the ambient temperature of the sensor (Ta).
Data is published at Teleperiod as an array [POI-0...6].

Command:
MLXPOIn xxyy, n ist poi 1...6, xx 0...31, yy 0...23
Example: Point 2 at x:12 and y:08 results in mlxpoi2 1208

The configuration can be saved via RULES.

In order to find the correct coordinates a thermal view is provided via webGUI, where points can be validated and/or updated. A stable Wifi-setup is mandatory for reliable operation. 

mlx90640

Ram: +4,9k
Flash:  +18,6k (Unishox appr. -1k)

## Checklist:


arendst:
Well done!

PR #9350, opened 2020-09-19
Added SetOption112 - Use friendly name in zigbee topic

nicandris:
Use friendly name in zigbee topic (use with SetOption89)

## Description:

## Checklist:
  - [ ] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.1

### DIFF
--- a/tasmota/xdrv_23_zigbee_2_devices.ino
+++ b/tasmota/xdrv_23_zigbee_2_devices.ino
@@ -896,9 +896,15 @@ void Z_Devices::jsonPublishFlush(uint16_t shortaddr) {
     attr_list.reset();    // clear the attributes
 
     if (Settings.flag4.zigbee_distinct_topics) {
-      char subtopic[16];
-      snprintf_P(subtopic, sizeof(subtopic), PSTR("%04X/" D_RSLT_SENSOR), shortaddr);
-      MqttPublishPrefixTopic_P(TELE, subtopic, Settings.flag.mqtt_sensor_retain);
+      if (Settings.flag4.zb_topic_fname && fname) {
+        char frtopic[13];
+        snprintf_P(frtopic, sizeof(frtopic) + strlen(fname), PSTR("tele/%s/" D_RSLT_SENSOR), fname);
+        MqttPublish(frtopic, Settings.flag.mqtt_sensor_retain);
+      } else {
+        char subtopic[16];
+        snprintf_P(subtopic, sizeof(subtopic), PSTR("%04X/" D_RSLT_SENSOR), shortaddr);
+        MqttPublishPrefixTopic_P(TELE, subtopic, Settings.flag.mqtt_sensor_retain);
+      }
     } else {
       MqttPublishPrefixTopic_P(TELE, PSTR(D_RSLT_SENSOR), Settings.flag.mqtt_sensor_retain);
     }
--- a/tasmota/xdrv_23_zigbee_2_devices.ino
+++ b/tasmota/xdrv_23_zigbee_2_devices.ino
@@ -902,7 +902,7 @@ void Z_Devices::jsonPublishFlush(uint16_t shortaddr) {
         strlcpy(stemp, (!strlen(fname)) ? MQTT_TOPIC : fname, sizeof(stemp));
         MakeValidMqtt(0, stemp);
         //Create topic with Prefix3 and cleaned up friendly name
-        char frtopic[13 + strlen(stemp)];
+        char frtopic[TOPSZ];
         snprintf_P(frtopic, sizeof(frtopic), PSTR("%s/%s/" D_RSLT_SENSOR), SettingsText(SET_MQTTPREFIX3), stemp);
         MqttPublish(frtopic, Settings.flag.mqtt_sensor_retain);
       } else {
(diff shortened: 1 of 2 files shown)


tasmota/xdrv_23_zigbee_2_devices.ino
```diff
@@ -896,9 +896,15 @@ void Z_Devices::jsonPublishFlush(uint16_t shortaddr) {
     attr_list.reset();    // clear the attributes
 
     if (Settings.flag4.zigbee_distinct_topics) {
-      char subtopic[16];
-      snprintf_P(subtopic, sizeof(subtopic), PSTR("%04X/" D_RSLT_SENSOR), shortaddr);
-      MqttPublishPrefixTopic_P(TELE, subtopic, Settings.flag.mqtt_sensor_retain);
+      if (Settings.flag4.zb_topic_fname && fname) {
+        char frtopic[13];
+        snprintf_P(frtopic, sizeof(frtopic) + strlen(fname), PSTR("tele/%s/" D_RSLT_SENSOR), fname);
```
arendst, line 901: Buffer overflow. Won't work.

nicandris: hmmm tried it and it works. what could be the issue here?

nicandris: is this better? ```

        char frtopic[13 + strlen(fname)];
        snprintf_P(frtopic, sizeof(frtopic), PSTR("tele/%s/" D_RSLT_SENSOR), fname);
```

Comment (arendst) [reply tasmota/xdrv_23_zigbee_2_devices.ino:901]: Sure it may work but as you defined a buffer way too small it overwrites memory at the end resulting in mayhem someday. Redefine your buffer lenght and never use a length larger than your buffer

Comment (arendst) [reply tasmota/xdrv_23_zigbee_2_devices.ino:901]: It solves the buffer overflow indeed.

Comment (arendst) [reply tasmota/xdrv_23_zigbee_2_devices.ino]: What I dont like here also is the use of text tele. tele is user defined and you need to use the user defined name here instead of a fixed tele. Search for the correct user defined settings text. Again, what works for you may well not work for me.

Comment (s-hadinger) [reply tasmota/xdrv_23_zigbee_2_devices.ino]: I'm a little worried that you blindly copy "friendlyname" into the topic. I assume there are only limited characters allowed in a topic, and they are not escaped nor replace with a harmless character like '_'.

This will most certainly break MQTT.

Comment (nicandris) [reply tasmota/xdrv_23_zigbee_2_devices.ino]: fixed the SettingsText(SET_MQTTPREFIX3) thing. what do you suggest for replacing chars? do we have a ready method? maybe EscapeJSONString?

Comment (s-hadinger) [reply tasmota/xdrv_23_zigbee_2_devices.ino]: Ignore my last comment. According to http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc384800482 Topic names can contain any valid UTF-8 characters.

Only limitations are:
- must not contain NULL character
- must not contain '#' or '+'
I think we should avoid having '/' in the topic name, as it will create a new level.

I would suggest to simply scan `frtopic` (you need to carefully compute start and end of friendly name, and replace all characters matching `#+/` with `_`

Comment (nicandris) [reply tasmota/xdrv_23_zigbee_2_devices.ino]: `//Clean special characters and check size of friendly name
        char stemp[TOPSZ];
        strlcpy(stemp, (!strlen(fname)) ? MQTT_TOPIC : fname, sizeof(stemp));
        MakeValidMqtt(0, stemp);
        //Create topic with Prefix3 and cleaned up friendly name
        char frtopic[13 + strlen(stemp)];
        snprintf_P(frtopic, sizeof(frtopic), PSTR("%s/%s/" D_RSLT_SENSOR), SettingsText(SET_MQTTPREFIX3), stemp);
        MqttPublish(frtopic, Settings.flag.mqtt_sensor_retain);`

Comment (nicandris) [reply tasmota/xdrv_23_zigbee_2_devices.ino]: im using the same method we are using to clean up the topic when we change it in the web interface. I noticed a small thing though, in MakeValidMqtt 

`//        if ((str[i] == '/') || (str[i] == '+') || (str[i] == '#') || (str[i] == ' ')) {
    if ((str[i] == '+') || (str[i] == '#') || (str[i] == ' ')) {`

why did we disable the /?

File: tasmota/xdrv_23_zigbee_2_devices.ino [hunk 2]
```diff
@@ -896,9 +896,20 @@ void Z_Devices::jsonPublishFlush(uint16_t shortaddr) {
...
-      char subtopic[16];
-      snprintf_P(subtopic, sizeof(subtopic), PSTR("%04X/" D_RSLT_SENSOR), shortaddr);
-      MqttPublishPrefixTopic_P(TELE, subtopic, Settings.flag.mqtt_sensor_retain);
+      if (Settings.flag4.zb_topic_fname && fname) {
+        //Clean special characters and check size of friendly name
+        char stemp[TOPSZ];
+        strlcpy(stemp, (!strlen(fname)) ? MQTT_TOPIC : fname, sizeof(stemp));
+        MakeValidMqtt(0, stemp);
+        //Create topic with Prefix3 and cleaned up friendly name
+        char frtopic[13 + strlen(stemp)];
```
arendst, line 905: Use TOPSZ here too. Less code and solves longer user defined tele too.

PR #9472, opened 2020-10-05
Add support for stateful ACs using IRremoteESP8266

ayavilevich:
## Description:

Tasmota uses IRremoteESP8266 for IR HVAC functionality, the relevant code is in `tasmota\xdrv_05_irremote_full.ino` . Specifically, `IRAcUtils::decodeToState` and `ac.sendAc` functions are used. Those functions expect the caller to pass a previous AC state to them. This is required to support Air Conditioning devices that use a differential IR protocol (they send changes rather than state). Without considering the previous state, for these devices, any IRHVAC instruction where power==on will be transmitted as "toggle power" by IRremoteESP8266. This completely breaks all IRhvac functionality for such vendors as Whirlpool, Airwell and others.

This PR implements a working POC/feature where a state is added to the logic.
The functionality can be implemented as an "option".
Also a POC for setting the initial state is implemented as a piggy-back on top of an existing "model" field. This is useful if one wants to sync the state with the actual state of the AC without generating an IR transmission.

The code is relatively simple. State transitions are implemented in IRremoteESP8266. Tasmota merely needs to hold the structure and pass it around.

If you wish to improve on this implementation I would suggest an additional "option" for enabling this state management (can be false by default). Also a special field can be added to the IRhvac command (https://tasmota.github.io/docs/Tasmota-IR/#sending-irhvac-commands) to signal that only a state update is required.

This was not tested functionally on an ESP32. There is nothing in the code that seems to be platform specific. I don't have an ESP32 based IR blaster, those are not common. The build passes though. If this is a big issue LMK.

## Checklist:
  - [y] The pull request is done against the latest dev branch
  - [y] Only relevant files were touched
  - [y] Only one feature/fix was added per PR.
  - [y] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.1
  - [n] The code change is tested and works on core ESP32 V.1.12.2
  - [y] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

@crankyoldgit fyi


tasmota/xdrv_05_irremote_full.ino
```diff
@@ -412,7 +417,17 @@ uint32_t IrRemoteCmndIrHvacJson(void)
     }
   }
 
-  if (val = root[PSTR(D_JSON_IRHVAC_MODEL)]) { state.model = IRac::strToModel(val.getStr()); }
+  bool stateOnly = false; // flag that means to update the state of IRremote without sending actual IR command
+  // parse model of the ac
+  if (val = root[PSTR(D_JSON_IRHVAC_MODEL)]) {
```
s-hadinger, line 422: I'm not a big fan of changing the semantic of an existing field. I would rather have a "StateOnly":true field instead.

.gitignore
```diff
@@ -7,6 +7,7 @@
 .platformio
 .pioenvs
 .piolibdeps
+.pio/libdeps
```
s-hadinger, line 10: Why do you need this one?

s-hadinger (commented):
Thanks for the PR. I made a suggestion to introduce a new field for StateOnly.

.gitignore
ayavilevich: hmmm. I think I needed this because at some point I used a dev branch of IRremote and PIO would create some temp files in .pio/libdeps .
It seems the `.piolibdeps` folder (one line above in current file) has been replaced with `.pio/libdeps` in newer versions of PIO.
https://community.platformio.org/t/piolibdeps-folder/8830
we can remove it from here and add in another PR or somebody on the team can add it. whatever you prefer.

tasmota/xdrv_05_irremote_full.ino
ayavilevich: I totally agree. Just not sure about the right processes here to add a new field. Any guidelines? Is this trivial? Should I just try to do it or do you prefer to write this part yourself? Will you accept PRs that add fields, options, etc?
Overall I understand the IR/AC part well but more cautious about the Tasmota part.

arendst: Just add a new field. As it's JSON it's position isn't a problem and ignoring the field shouldn't be a problem either.

Jason2866:
I did a extra PR for the corrected .gitignore.
Thx for pointing to this old entrys.
Please remove from this PR.

ayavilevich:
hey, I have made the requested changes. Please review and send feedback.

s-hadinger:
Thanks, looks good to me.

ayavilevich:
great. should we do something with the documentation? what is the process?

arendst:
Yes pls do. Follow this https://tasmota.github.io/docs/Contributing/#editing-articles

PR #9519, opened 2020-10-11
Tweak template handling

emontnemery:
## Description:
Attempt to fix issues with the new templates:
1. Don't copy old template if `USER_TEMPLATE` is defined
2. ~~Fix logic when detecting old template~~ This was a mistake, reverted

## Checklist:


arendst:
What's the goal of this pr? the additional gpio16 setting is already done earlier in the function so this is a duplicate which also fails on esp32 when user template is enabled....

????????

emontnemery:
@arendst I was on an older version without 490e7d838e1e185e464264ce4beefc89128e26d4
This should teach me to always test on latest version..

arendst:
At least your solution solved the issue too.

Great minds having same thoughts!

PR #9544, opened 2020-10-14
Add support for backlog with no delay

emontnemery:
## Description:
Add support for executing backlog with no delay without setting `SetOption34 = 0`

The delay defined by `SetOption34` is omitted for any command in a backlog sequence following immediately after the special command `NoDelay`.
This must be used with care, and only for simple commands.

The main driver is to allow HomeAssistant to more precisely control lights, in particular transitions (`Fade` and `Speed` in Tasmota).

## Example:

Disable fade, and turn off a light without any delays:
`Backlog NoDelay;Fade 0;NoDelay;Power4 OFF`
Enable fade, and turn on a light without any additional delays:
`Backlog NoDelay;Fade 1;NoDelay;Speed 4;NoDelay;Power4 ON`

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -212,6 +212,7 @@
 // Commands tasmota.ino
 #define D_CMND_BACKLOG "Backlog"
 #define D_CMND_DELAY "Delay"
+#define D_CMND_NODELAY "NoDelay"
 #define D_CMND_STATUS "Status"
   #define D_STATUS1_PARAMETER "PRM"
   #define D_STATUS2_FIRMWARE "FWR"
--- a/tasmota/support_command.ino
+++ b/tasmota/support_command.ino
@@ -311,6 +311,7 @@ void CmndBacklog(void)
 {
   if (XdrvMailbox.data_len) {
 
+    bool firstcommand = true;
 #ifdef SUPPORT_IF_STATEMENT
     char *blcommand = strtok(XdrvMailbox.data, ";");
     while ((blcommand != nullptr) && (backlog.size() < MAX_BACKLOG))
@@ -329,6 +330,15 @@ void CmndBacklog(void)
           break;
         }
       }
+      // If the first command in backlog sequence is NoDelay, set flag to skip delay before executing the next command
+      if (firstcommand) {
+        if (!strncasecmp_P(blcommand, PSTR(D_CMND_NODELAY), strlen(D_CMND_NODELAY))) {
+          blcommand += strlen(D_CMND_NODELAY);
+          backlog_no_delay = true;
+        }
+        firstcommand = false;
+      }
+
       if (*blcommand != '\0') {
 #ifdef SUPPORT_IF_STATEMENT
         if (backlog.size() < MAX_BACKLOG) {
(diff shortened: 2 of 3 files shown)


effelle (approved):

arendst:
I like the idea but would like to play with ``backlog_delay`` value in order to execute the next command removing the code execution redundancy.

Let me have a look before optionally merging this.

arendst:
Thx. Implemented my way.

emontnemery:
@arendst I first implemented your way, because as you mention code is much cleaner.

However, it doesn't work because with this way there is a delay between handling of the `Backlog` command itslef and until `BacklogLoop` starts processing the first command in the Backlog.

What could work, while still keeping the `NoDelay` handling only in `BacklogLoop`, is to allow `BacklogLoop` to check if the next command is `NoDelay`.

What I mean is: revert 29e73da and instead do something like this:
```C++
void BacklogLoop(void) {
  if (TimeReached(backlog_delay) || NextCommandNoDelay()) {
```

arendst:
@emontnemery How about leaving the commit and add the following to ``support_command.ino``:
```

    }
//    ResponseCmndChar(D_JSON_APPENDED);
    mqtt_data[0] = '\0';
    backlog_delay = 0;
  } else {
```
So add line 347 with ``backlog_delay = 0;``

emontnemery:
@arendst that is the perfect solution if the delay between the backlog command itself and the first command in the backlog is not needed!

Otherwise, something like this:
```C++
// Get current backlog command without removing it from the backlog
String BacklogGet(void) {
  String cmd;
#ifdef SUPPORT_IF_STATEMENT
  cmd = backlog.get(0);
#else
  cmd = backlog[backlog_pointer];
#endif
  return cmd;
}

// Shift backlog - get current backlog command and remove it from the backlog
String BacklogShift(void) {
  String cmd;
#ifdef SUPPORT_IF_STATEMENT
  cmd = backlog.shift();
#else
  cmd = backlog[backlog_pointer];
  backlog[backlog_pointer] = (const char*) nullptr;  // Force deallocation of the String internal memory
  backlog_pointer++;
  if (backlog_pointer >= MAX_BACKLOG) { backlog_pointer = 0; }
#endif
  return cmd;
}

void BacklogLoop(void) {
  if (!BACKLOG_EMPTY && !backlog_mutex) {
      backlog_mutex = true;
      bool nodelay_detected = false;
      do {
        String next_cmd = BacklogGet();
        bool nodelay_detected = !strncasecmp_P(next_cmd.c_str(), PSTR(D_CMND_NODELAY), strlen(D_CMND_NODELAY));
        if (nodelay_detected) {
          BacklogShift();
          backlog_delay = 0;
        }
      }
      while (nodelay_detected && !BACKLOG_EMPTY);
      backlog_mutex = false;
  }

  if (TimeReached(backlog_delay)) {
    if (!BACKLOG_EMPTY && !backlog_mutex) {
      backlog_mutex = true;
      String cmd =  BacklogShift();
      ExecuteCommand((char*)cmd.c_str(), SRC_BACKLOG);
      backlog_mutex = false;
    }
  }
}
```

PR #9552, opened 2020-10-15
Fixed issue of cannot read data when restart and using Sensor18

VergLsm:
## Description:

I am using two pms7003. in order to extend life, I set ```Sensor18 300```. When restarting for unknown reasons (another bug?), it cannot reads data of  pms7003 anymore. Later I found out that to make it resume reading data, just resend the command ```sensor18 300```.

## Checklist:

### DIFF
--- a/tasmota/xsns_18_pms5003.ino
+++ b/tasmota/xsns_18_pms5003.ino
@@ -154,7 +154,6 @@ bool PmsReadData(void)
 #else
   memcpy((void *)&pms_data, (void *)buffer_u16, 30);
 #endif  // PMS_MODEL_PMS3003
-  Pms.valid = 10;
 
   return true;
 }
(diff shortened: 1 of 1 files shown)


arendst:
That's what rules are for. Just add a rule to be executed at reboot time execting command ``sensor18 300``

Jason2866:
Mhh, i have a PMS7003 sensor to. I have none of your issues. It does not restart. Runs over weeks without. Only planned maintainence restarts. No need for doing a Sensor18 again.
Did this one time for setup. Working nonstop nearly a year.

VergLsm:
So strange. 
I have two PMS7003 sensor, they working with different circuit boards, and they always restart at intervals(about 12h), and can’t read the sensor's data anymore.
Honestly, I cannot be 100% sure that it happened after the restart, maybe it happened during running. So I don’t agree to solve it through rules.
During the issue, I found that the program is calling PmsReadData(void) every second, until I send the command ```Sensor18 300```. Normally, reading should be paused after successful.
Although my code is not good enough, but it works for me. 
Please check the original judgment mechanism for bugs.

Environmental information:
The sensor is directly connected to the chip TX(GPIO1) and RX(GPIO3)
FW version: 8.5.1
kernel and SDK version: 2_7_4_1/2.2.2-dev(38a443e)

VergLsm:
v9.1.0 has no such issue

VergLsm:
@Jason2866 
Are you sure your sensor working nonstop nearly a year with Sensor18?
I found out that the Sensor18 command was just added on April 9, 2020

Jason2866:
I have the sensor a long time. never had any issue. I update always to latest version.
Since Sensor18 was introduced i could simplyfy my setup (no more rule needed for switch off device)
image
image

VergLsm:
@Jason2866
The air in your home is so clean.
Is your sensor directly connected to esp8266?

Jason2866:
Yes, directly connected to a Wemos mini via short soldered wires.
The Wemos mini weak orig. 3.3V reg is replaced with a 3.3V 500mA type.

VergLsm:
@Jason2866 @gururise
I think I caught the bug.
The restart occurs after the sensor goes to sleep, the first thing after the device restarts is to read the sensor data but not wakeup.
Note that the flag ```Pms.wake_mode``` is initialized to ```1```, it means the sensor will always be asleep, the device will not be able to read data from it.

VergLsm:
I want to reopen it and add only the initialization code to ensure that the initial state is correct

PR #9563, opened 2020-10-17
Add Switchmode Ignore 15

boaschti:
## Description:

## Checklist:


arendst:
??????

Why not unconfiguring the switch?

boaschti:
Classic example is a pir sensor which is on switch1 and the coupled relais1 which controlls the lamp. If i want to open the connection at daytime and use the pir also for a trigger to get a picture from a camera then i need the new switchmode and a immideatelly report from the sensor change via mqtt.
Another example:
If you have switch1..4 and relais1 and relais2. Then switch 3 and 4 will be coupled with relais1. With the new switchmode you can open the coupling and have just the input reported via mqtt.

arendst:
> then i need the new switchmode and a immideatelly report from the sensor change via mqtt.

That's a major change as it is reporting for EVERY switchmode flooding MQTT with mostly useless data.

It's better to send ONLY for this new switchmode.

boaschti:
Hm. Does that matter? Another Advantage is to get the switch state if you have switchmode = toggle. Actually you never know the switch state. (immidiatelly)

For example you have two or more tasmotas and linked via homeautomation. And you want to trigger a timer if the switch of tasmota1 is toggled. The tasmota2 an others does not trigger the timer or perhaps another timer.

arendst:
Implemented my way.

PR #9567, opened 2020-10-18
Adding support for EZOpH & EZOORB sensors

tichris0:
This commits adds support for EZOpH and EZOORP to Tasmota

## Description:

Adding sensor support for EZOpH and EZOORP.
There's many more EZO devices that could easily be supported, but I don't own the hardware to test them.

Some notes:
- Common EZO code was pulled out to save space (~1k code)
- Each subsequent EZO driver adds about 0.6k
- Supports multiple of the same EZO devices
- EZO devices support asciiz commands which are funneled directly through the Sensor78/79 commands
- Added support to target a specific device in the list when they are of the same type by adding a dash (ex: Sensor78-1 to address the first pH sensor only)
- All sensors can be sent the same command by omitting the dash ^.  It seems like functionality like this would be best implemented at the Tasmota level for other sensors types that would want to support sending commands to a specific sensor when multiple sensors of the same type are supported.
- EZO devices can be given any I2C address, but the implementation restricts itself to the range of default addresses that the EZO devices use (0x61-0x70).
- EZO devices are detected by issuing commands to them and seeing what identifier returns.  This code is common to all devices and they are thus all gated by XI2C_55.

PS: I do not own a ESP32 to test this against.

## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.3


arendst:
What a nice implementation of C templates! Very instructive!

The current check fails because the other language files miss the additions you made in en_GB.h

Could you pls add the english defines in all other languages too. In time local language decoders will pick it up.

arendst:
Ah. I see you did the languages already.

tichris0:
> 
> 
> Ah. I see you did the languages already.

Yah I didn't think of trying it locally :(.  It should be happy now.

> 
> 
> What a nice implementation of C templates! Very instructive!

Thanks.  It should be super easy to add other EZO sensors, but since I don't own them, at best, I could implement a theoretical implementation & test against test strings.

Jason2866:
Do you have a (example) source where to buy?

tichris0:
> 
> 
> Do you have a (example) source where to buy?

Yes.  There are many places you can buy these.  At least in America, Amazon sells them.  You can also always buy them directly from the manufacturer's website: https://atlas-scientific.com/

tichris0:
I'm not sure what you guys wanted to do with the default & which build to include it in, so I've left it all "off" by default :)

PR #9599, opened 2020-10-22
Refactor EZO drivers to only use one sns ID

tichris0:
Adds support for EZO humidity&temp sensor.

## Description:

Adding support for EZOHUM.  Without configuration, it will pick up the humidity only.  The user can enable the temperature through the sensor81 commands

## Checklist:
  - [X ] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.5
  - [X ] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).


arendst:
Would it be possible to consolidate all EZO sensors in one sns driver? This way I'm running out of sensor driver ids in no time while these sensors will hardly be used considering the price.

tichris0:
> 
> 
> Would it be possible to consolidate all EZO sensors in one sns driver? This way I'm running out of sensor driver ids in no time while these sensors will hardly be used considering the price.

Yah it should be feasible.  I'll still keep the individual #define so that they can be enabled individually considering how little space Tasmota has on the esp8266.

arendst:
Great! thx.

PR #9604, opened 2020-10-22
Optimize compile time (part 1)

Jason2866:
## Description:

Since more and more libs are used the compile time increases since PlatformIO 
does compile everything when not excluded with lib_ignore.
Starting with Tasmota-minimal

## Checklist:


eku:
Good point, but isn't there a risk that the configuration may at some point become outdated?

Jason2866:
It is a WIP. Tasmota-minimal is a good start. In worst case stuff will be added which is compiled
uneccessary (again).

PR #9609, opened 2020-10-23
MI_ESP32: add beacon functions

Staars:
## Description:

Introduction of MAC-based beacons to the driver, allows to have 4 beacons.

New command: `MI32BEACON`

Usage:
`MI32BEACON0` starts an BLE scan and prints result to the console

 `MI32BEACONx AA:BB:CC:DD:EE:FF` starts the observation of a beacon with number x and MAC-address AA:BB:CC:DD:EE:FF (address format AABBCCDDEEFF is valid too)

 `MI32BEACONx 00:00:00:00:00:00` stops the observation of a beacon with number x

Reporting scheme:
- for every beacon a count-up-timer starts with every received advertisement, starting with 0
- RULE-example: `on Beacon2#time==30 do SOMETHING endon` - is triggered 30 seconds after last packet was received

TELE-output:
Beacon1":{"MAC":"11:22:33:44:55:66","CID":"0x0000","SVC":"0x0000","UUID":"0x0000","Time":4,"RSSI":0}}

(following AD type, read here: https://www.bluetooth.com/specifications/assigned-numbers/generic-access-profile/)
CID - company identifier
SVC - service data
UUID - service or class UUID

Web-interface shows only data, that was part of the packet (typically only 1-2 AD type is part of the packet).

minor changes:
- add RSSI to every sensor at TELE-period (was only shown in bridge mode before)
- refactor some misleading 'connection messages' in the console output

## Checklist:
  - [ ] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.5

### DIFF
--- a/tasmota/xsns_62_MI_ESP32.ino
+++ b/tasmota/xsns_62_MI_ESP32.ino
@@ -20,6 +20,8 @@
   --------------------------------------------------------------------------------------------
   Version yyyymmdd  Action    Description
   --------------------------------------------------------------------------------------------
+  0.9.1.6 20201022  changed - Beacon support, RSSI at TELEPERIOD, refactoring
+  -------
   0.9.1.5 20201021  changed - HASS related ('null', hold back discovery), number of found sensors for RULES
   -------
   0.9.1.4 20201020  changed - use BearSSL for decryption, revert to old TELEPERIOD-cycle as default
@@ -57,7 +59,6 @@
 void MI32scanEndedCB(NimBLEScanResults results);
 void MI32notifyCB(NimBLERemoteCharacteristic* pRemoteCharacteristic, uint8_t* pData, size_t length, bool isNotify);
 
-
 struct {
   uint16_t perPage = 4;
   uint32_t period;             // set manually in addition to TELE-period, is set to TELE-period after start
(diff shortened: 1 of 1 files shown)


Staars:
Additional comments:

The reported CID, SVC and UUID are published as strings in the the JSON message, although they are 16-bit-values.
This is done (for now) for easier human readability, because these values are commonly used only as hex and JSON standard does not allow hex numbers. (I know about JSON5)
This might be changed in the future.

rt400:
@Staars Great Job...
I waited a long time for this option in ESP32.
Will it be possible to reset IBEACONS data without restarting to ESP32 or running MIBEACON1 0 command?
The RSSI also does not change to 0 once it is no longer in range.

Staars:
>  without restarting to ESP32 or running MIBEACON1 0 command?
How would you reset/stop a beacon without a command (MIBEACONx 0)???

> The RSSI also does not change to 0 once it is no longer in range.
I scratched my head for a while about this topic and IMHO using a count-up-timer is the only solution, that is always correct.
There is no way to tell, if a BLE-device is out of range. You can only make an assumption, that after a certain period of time without a received advertisement from a certain device, this device might be out of reach.
You can never know, how the actual RSSI would be only one second after the last received packet. There could be an obstacle (a closed door or something) between sender and receiver immediately after last packet receive. You will only ever know the RSSI of the last received packet and that is what the driver reports.

So I am pretty sure, that every possible information is given in the current driver regarding the beacons. Additional information could only be the result of (subjective) interpretation of a the data by a  user (i.e. the time out, after you consider the beacon to be lost). Setting the RSSI to 0 is purely cosmetical.

That does not mean, that I will not try to improve the driver on user request.
BTW, I spotted a little bug in the neacons-rules-function and will push an PR soon.

PR #9621, opened 2020-10-25
Reduce compile time

Jason2866:
## Description:

by spliting librarys. Reason for: Since more and more features where added the compile time increased a lot.
This is generated by compiling all libs even the where not used in the builded firmware variant.

This PR defines for the Tasmota variants only the libs which are used. This is done in the ENV Platformio files.
The Tasmota version and all it language variants are untouched. All libs are enabled. User has to do the disabling.

So the build process for variant Tasmota is unchanged. The user has to do if he wants to optimize   

## Checklist:


Jason2866:
new PRs added which are conflicting

PR #9734, opened 2020-11-04
Swapped strstr for strchr where applicable

tichris0:
strstr calls with a single character were changed to strchr which should be more efficient.

## Description:

This is a very minor change.  While doing changes in the past, I noticed that in a few areas, calls were made to strstr where a single character was provided.  This isn't as efficient because the strings are forced to be 2 bytes (they're ASCIIZ) and it's extra work since the function needs to handle the arbitrary length of the second argument.

I wasn't able to test every change given that I don't have the appropriate sensors, however the change is pretty benign and should work without a hitch.

## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.5

### DIFF
--- a/tasmota/support_command.ino
+++ b/tasmota/support_command.ino
@@ -1242,7 +1242,7 @@ void CmndTemplate(void)
 
   bool error = false;
 
-  if (strstr(XdrvMailbox.data, "{") == nullptr) {  // If no JSON it must be parameter
+  if (strchr(XdrvMailbox.data, '{') == nullptr) {  // If no JSON it must be parameter
     if ((XdrvMailbox.payload > 0) && (XdrvMailbox.payload <= MAXMODULE)) {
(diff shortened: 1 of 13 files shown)


arendst:
We've seen this before. Does it save code size?

Jason2866:
Tasmota without this PR
````
RAM:   [======    ]  57.9% (used 47464 bytes from 81920 bytes)
Flash: [======    ]  59.6% (used 610452 bytes from 1023984 bytes)
```
with PR
```
RAM:   [======    ]  57.9% (used 47464 bytes from 81920 bytes)
Flash: [======    ]  59.6% (used 610452 bytes from 1023984 bytes)
```

tichris0:
No as Jason pointed out it doesn't since the parameter is an int anyways.  It does perform better however small it may be.

PR #9444, opened 2020-10-01
Change new IR Raw compact format

s-hadinger:
### New IR Raw compact encoding (tasmota-ir only)

We had numerous requests from users to expand the buffer sizes because many IR codes would exceed the MQTT/Web/Serial buffer size. Instead, we changed the IR Raw encoding to shrink the size necessary to encode almost any IR code.

**Before** (buffer overflow):
`{"IrReceived":{"Protocol":"PIONEER","Bits":64,"Data":"0xA55A50AFA55A50AF","DataLSB":"0xA55A0AF5A55A0AF5","Repeat":0,"RawData":[8574,4224,574,1558,572,502,570,1562,570,502,572,502,570,1562,570,502,570,1562,572,500,572,1560,572,500,572,1560,572,1560,570,504,568,1560,570,502,572,502,570,1562,570,502,570,1560,570,502,572,500,570,502,572,502,570,1560,570,504,572,1558,572,502,570,1564,568,1562,570,1560,572,1560,572,25258,8574,4222,572,1560,570,502,572,1558,572,502,570,502,572,1558,572,500,570,1560,570,502,570,1560,570,502,570,1560,570,1560,570,504,570,1560,572,502,570,502,570,1560,572,502,570,1560,570,502,570,502,570,502,570,502,570,1560,570,502,570,1560,572,502,570 ...`

**After** (no overflow):
`{"IrReceived":{"Protocol":"PIONEER","Bits":64,"Data":"0xA55A50AFA55A50AF","DataLSB":"0xA55A0AF5A55A0AF5","Repeat":0,"RawData":[+8570-4240+550-1580C-510+565-1565F-505Fh+570gFhIdChIgFeFgFgIhFgIhF-525C-1560IhIkI-520ChFhFhFgFhIkIhIgIgIkIkI-25270A-4225IkIhIgIhIhIkFhIkFjCgIhIkIkI-500IkIhIhIkFhIgIl+545hIhIoIgIhIkFhFgIkIgFgI],"RawDataInfo":[135,135,0]}}`

The new format still encodes timings for High/low pulses.
- First the timings are rounded to the closest 5 microsec value.
- Instead of using commas, values are prefixed with `+` if it's a HIGH signal, or `-` if it's a LOW signal.
- Each new timing value is assigned a letter starting with 'A'
- If a timing value matches a previously found value, it is replaced with the letter, in uppercase for a HIGH signal, or lowercase for a LOW signal.

Ex:

```
[...]

PR #9746, opened 2020-11-05
Fix IRsend raw

thefloweringash:
## Description:

The frequency parameter was inconsistently handled. It was removed from the data string by the outermost handler `IrRemoteCmndIrSendRaw`, which then called `IrRemoteSendRawStandard`, which parsed the entire data string as the frequency.

While i was here, i removed some dead code. This goes against the checklist but I thought it was appropriate since I'm changing the affected function. Let me know if I should drop that commit.

cc @s-hadinger as author of #9444

## Checklist:

### DIFF
--- a/tasmota/xdrv_05_irremote_full.ino
+++ b/tasmota/xdrv_05_irremote_full.ino
@@ -688,8 +688,7 @@ uint32_t IrRemoteParseRawCompact(char * str, uint16_t * arr, size_t arr_len) {
 //   count: number of commas in parameters, i.e. it contains count+1 values
 //   repeat: number of repeats (0 means no repeat)
 //
-uint32_t IrRemoteSendRawStandard(char ** pp, uint32_t count, uint32_t repeat) {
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks, but I'm not sure to understand why parsing `freq` inside or outside of `IrRemoteSendRawStandard()` makes a difference. I'm not against this change, but I don't see why it fixes anything.

You are right about the old dead code that I forgot to remove.

thefloweringash:
IIUC the data string passed to `IrRemoteSendRawStandard()` does not contain the frequency. So it's not available to parsed, and the first interval is being used as the frequency.

s-hadinger:
Sorry but I still don't get it. The string passed to `IrRemoteSendRawStandard()` is the same before and inside the call. The original code parses freq as a first statement of the function. Your version parses it before the call but it's still the same string.

s-hadinger:
Ok got it now, I was confusing `p` and `str`. You are completely right.

@arendst This should be merged :)

PR #9858, opened 2020-11-14
Fix switchmode 15 publish old switch state

barbudor:
##Description:

When using SwitchMode15, every change of the switch state sends an immediate forced SENSOR message
But the message is sent with the switch's last_state.
So the last_state must be updated before the SENSOR message is sent.

## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.7


barbudor:
The failuere doesn't seems to be my fault 😢

PR #9867, opened 2020-11-15
Sm2135: fixed power toggling

orsonmmz:
## Description:

Sm2135 driver did not correctly implement power toggling, switching the device on/off only affected CW channel. The problem was a condition in Sm2135SetChannels() that recognized turning off the device as a CW channel update. The new implementation makes a distinction between CW channel updates and power toggling.

## Checklist:

### DIFF
--- a/tasmota/xlgt_04_sm2135.ino
+++ b/tasmota/xlgt_04_sm2135.ino
@@ -1,7 +1,7 @@
 /*
   xlgt_04_sm2135.ino - sm2135 five channel led support for Tasmota
 
-  Copyright (C) 2020  Theo Arends and CrudelisPL
+  Copyright (C) 2020  Theo Arends, CrudelisPL and Maciej Suminski
 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
@@ -53,6 +53,9 @@
 #define SM2135_55MA         0x09
 #define SM2135_60MA         0x0A
 
+// Number of color channels (RGBCW)
+#define CHANNEL_COUNT       5
+
 enum Sm2135Color { SM2135_WCGRB, SM2135_WCBGR };
 
 //                              RGB current         CW current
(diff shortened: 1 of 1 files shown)


arendst:
Why the abundance of static?

orsonmmz:
Hi Theo,

I have added 'static' modifier to symbols that IMHO should not be visible outside the translation unit. If you prefer to leave them out then I can create a pull request without them.

arendst:
I prefer to leave them out as they are useless anyway in an ino environment; all files are merged first before compilation so they are still visible for all.

orsonmmz:
Alright, I have just removed 'static' keywords.

arendst:
Just did some tests and it appears my bulb stopped working as expected. A power On quickly blinks the led and it's off again.

You must know the original solution made sure that either RGB OR CW were on simply because the bulbs power supply cannot handle both colors fully on.

I'll revert this PR for now.

orsonmmz:
I am sorry about that, the change worked fine with my lightbulb. Since you have a lightbulb with SM2135 controller, are you able to turn it off completely? Do you see the RGB part always turned on, no matter what is the power state? If the answer is yes, then I can change the patch to drive either RGB or CW part.

arendst:
I can control both RGB and CW independently. By using the sliders in the GUI it turns on either RGB (if I touch RGB sliders) or CW (if I touch CW slider). It's governed by the bulbs logic only to allow onw of the two.

Doing a Power On/Off always turns on/off the bulb in all situations returning to it's color before the power off.

orsonmmz:
Thank you for the explanation, now I see what is the issue. Apparently my lightbulb does not have the logic that mutually excludes either RGB or CW channels, so I can use both at the same time, possibly out-of-spec.

Then there is another issue: electrical current setting. I imagine this is different for different lightbulbs, depending on the power. What is there right way to handle it in Tasmota? I have seen in SM2135 driver code that it selected RGB channels order depending on "switch 1" presence, but I wonder if there is a better alternative.

arendst:
By bulbs logic I meant the bulb driver being this driver. It was written in such a way that either RGB or CW will be allowed. This way overcurrent won't occur.

By using the "switch1" option it is possible to change the drivers features by using just another template instead of having to use "SetOption" commands. This makes user experience simple.

PR #9922, opened 2020-11-19
comparison should be >=

barbudor:
## Description:

I feel the comparison should be `>=` instead of `>` in 
```
uint16_t GpioConvert(uint8_t gpio) {
  if (gpio > ARRAY_SIZE(kGpioConvert)) {
    return AGPIO(GPIO_USER);
  }
  return pgm_read_word(kGpioConvert + gpio);
}
```
## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.7
  - [-] The code change is tested and works on Tasmota core ESP32 V.1.0.4.2

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -1168,7 +1168,7 @@ int ResponseJsonEndEnd(void)
 
 #ifdef ESP8266
 uint16_t GpioConvert(uint8_t gpio) {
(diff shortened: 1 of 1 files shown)


arendst:
You're right. Until then there was no gpio equal array size but indeed a user couls enter it in a template and fail.

Thx.

PR #9992, opened 2020-11-27
HM10: ports from MI32 to HM10

Staars:
## Description:

This should roughly bring feature parity compared to the ESP32-version (minus decryption).

- add commands HM10block and HM10option
- allows negative temperature values
- send generic BLE scan (with `Beacon0`) to MQTT
- minor bugfixes and code refactoring 

## Checklist:

### DIFF
--- a/tasmota/xsns_62_MI_HM10.ino
+++ b/tasmota/xsns_62_MI_HM10.ino
@@ -20,6 +20,8 @@
   --------------------------------------------------------------------------------------------
   Version yyyymmdd  Action    Description
   --------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Staars:
The default behavior is unchanged, but I encourage every user to at least test the options to seem what the sensor sees.
The virtual "HM10truth" command, goes like this (works after the initial connection round to all found sensors):
```
HM10option0 0
HM10option1 1
HM10option2 1
```

This reflects what really happens. The BLE-sensors send at their own frequency without any influence by this Tasmota driver and the HM-1x is able to catch a certain percentage of these packets (fur sure it will miss some).

PR #10027, opened 2020-11-30
Add option `#define UPGRADE_V8_MIN` to drop support for upgrading before v8.0 (saves 1.3kb)

s-hadinger:
## Description:

Dropping upgrade from pre-v8.0 Tasmota versions allows to save 1376 bytes of Flash.
This option is not enabled by default except for ZBBridge, whose version was always at least v8.0.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - Support for multiple WeMo devices by Magic73 (#9208)
 - Fallback NTP server from x.pool.ntp.org if no ntpservers are configured
 - TyuaMcu update 2/3 by Federico Leoni (#10004)
+- Add option `#define UPGRADE_V8_MIN` to drop support for upgrading before v8.0 (saves 1.3kb)
 
 ### Breaking Changed
 - KNX DPT9 (16-bit float) to DPT14 (32-bit float) by Adrian Scillato (#9811, #9888)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ All notable changes to this project will be documented in this file.
 - Support for multiple WeMo devices by Magic73 (#9208)
 - Fallback NTP server from x.pool.ntp.org if no ntpservers are configured
 - TyuaMcu update 2/3 by Federico Leoni (#10004)
-- Add option `#define UPGRADE_V8_MIN` to drop support for upgrading before v8.0 (saves 1.3kb)
 
 ### Breaking Changed
 - KNX DPT9 (16-bit float) to DPT14 (32-bit float) by Adrian Scillato (#9811, #9888)
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -45,7 +45,6 @@
 #define CFG_HOLDER             4617              // [Reset 1] Change this value (max 32000) to load SECTION1 configuration parameters to flash
                                                  // If following define is disabled it increases configuration corruption detection BUT
                                                  //  it only allows firmware upgrades starting from version 6.6.0.11
-// #define UPGRADE_V8_MIN                           // do not support upgrading from version below 8.0.0
 
 // -- Project -------------------------------------
 #define PROJECT                "tasmota"         // PROJECT is used as the default topic delimiter
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -45,6 +45,7 @@
 #define CFG_HOLDER             4617              // [Reset 1] Change this value (max 32000) to load SECTION1 configuration parameters to flash
                                                  // If following define is disabled it increases configuration corruption detection BUT
                                                  //  it only allows firmware upgrades starting from version 6.6.0.11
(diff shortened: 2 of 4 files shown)


meingraham:
> drop support for upgrading before v8.0

What does this mean? For example, I have several 6.x devices. The migration path currently allows upgrading from 6 to 7 to 8 and finally to 9 (or current). What does "drop" mean in this context?

s-hadinger:
It means that upgrading from v6 or v7 to v9 will fail to update settings. Actually I'm not sure how it fails and maybe I need to reset settings if this happens. 

This is only for self compile and zbbridge version. All other versions are unchanged

meingraham:
I was under the impression that one had to clear settings anyway when going through the major versions due to the settings structural changes. Otherwise you'd end up with corrupted values that could end up soft bricking the device. Thus, the process starts with record the current settings (e.g., decode-config to JSON), Reset 6, upgrade, Reset ... and then, once on the latest version, restore settings or manually reconfigure. In other words, you don't want settings to be retained when jumping major versions.

I think that automatically clearing/resetting settings might be too harsh. Perhaps the upgrade should fail if it detects custom settings are "present" and issue a warning as to why the upgrade failed. This would allow the user to take the necessary steps to record the current settings followed by the requisite `RESET 6` prior to performing an upgrade. Could it also make sure that the upgrade is to the next major version w/o skipping (e.g., don't upgrade to N+1 if not on N)?

s-hadinger:
I'm not very familiar with the correct update process. But the code in `SettingsDelta()` is run once the upgrade has completed and takes care of completing new settings that didn't exist in previous versions, as well as the structural change in v8.

If the current settings is below v8, then indeed we risk a soft bricking, this is why I was considering a full reset to default settings, which is always better than a soft bricking. But there is no way to abort the current upgrade, it's already too late for that.

s-hadinger:
I added a safeguard to force a clearing of setting if coming from a version before v8.0.

araczkowski:
Hi @s-hadinger 

> I added a safeguard to force a clearing of setting if coming from a version before v8.0.

if I understand correctly, with this safeguard we will be able to upgrade straight (without migration) from very old version to the newest one, without soft bricking the device? All we will lose are all old settings (wifi too). 
If this is the case, thent this fix is epic one :)

s-hadinger:
There is no change in any firmware. It's only an option for self compile.

araczkowski:
I did a test:
1. downgrade to Tasmota 5.14 via TTL
2. update via OTA to Tasmota 9.1.0 my own minimal compilation + your fix in settings.ino 
image

no soft brick :+1: 

But I'm not able to connect to WiFi and do the update via OTA to full version, course the minimal version is not able to store the wifi settings. I think I will be able to bypass this probelm via special ssid2.

I think this approach have a potential, at least for me.

arendst:
What's the reason NOT to follow the migration path and have all settings converted between version up to the latest without loosing wifi etc.?

The provided define is only useful for zigbee bridge as it is introduced in v8 and therefor has no history og settings to convert.

araczkowski:
Hi Theo!

I can explain :) we are using your great software in our project AI Speaker controlled locally. Not all ouer users are technicall advanced... they know how to add device to gate via our wizard and this is all they users need to know - all works. End users don't care about ttl, mqtt, wifi, ip, etc.. and some of the users have vision problems (but this is diferent story). The goal of the project AI-Speaker is to let all the users, to use devices with Tasmota software (and other open source great soft like zigbee2mqtt), connected to our local mqtt broker on AI-Speaker iot gate.

We are publishing new version of the Tasmota compilation and each time we need add some "Attention info - do not do this is you don't have to. There is a migration path..." and in case users will do something not according to path (with is possible) we can have a problem.

So we are looking for better way, the first idea was to allow only upgrade step by step (according to migration path).
Another one was to force upgrade each time the when the new version it is published (not a best idea too).
We thought that change in CFG_HOLDER will fix the problem... but not. 

Now I think that maybe using this fix, we can always allow to upgrade and if the upgrade is done not according to the migration path, we can just reset all the settings - the device should be as new - ready to be paired again with iot gate.
Of course the full configuration of the device is in our custom Tasmota build. 

Personally I can use ttl, but if my choice will as for end users, only to:
1. do OTA 4 times
or
2. jump directly to newest version and reconnect with wifi
I will vote for option 2.

Hope this is more clear now why I think this is needed. 

And the best way will be to allow the update without migration and keep only the wifi settings (like during reset 5). Do you think this is possible?

Thanks and have a nice evening :)

arendst:
Latest commit contains additional defines UPGRADE_V8_MIN_KEEP_WIFI and UPGRADE_V8_MIN_KEEP_MQTT to keep wifi and/or mqtt settings.

This feature is not guaranteed to work in future major version updates as it relies on certain settings which may not be there anymore.

YMMV.

araczkowski:
I see how this is done - thank you Theo, you save me a day of my work :) 
You know what, I'll ask now my secretary @asystentka-jolka to donate you because you are star! 
------------------------
The secretary is a joke, Jola is my colleague... we are not so commercial project yet ;) 
But the rest is 100% true.

PR #10037, opened 2020-12-02
More generic Bluetooth Low Energy

btsimonh:
## Description:
Logging needs to be reduced, and I believe there is room for web UI support, and scan results -> MQTT (although this could be a separate driver - that's kind of the point).

### The issue:

### xdrv_46 provides:
2/ generic BLE 'operations'.

xsns_52_ibeacon has been modified to make use of xdrv_46_BLE_ESP32, as an example of registering for advertisment callbacks.

I believe I have included all the features required for xsns_62_MI_ESP32 to be updated to use the generic BLE - happy to do this but want some feedback first.

### use:
xdrv_46 is enabled via 
USE_BLE_ESP32 and ESP32
It currently logs all adverts with extra data at LOG_LEVEL_DEBUG
To enable xsns_52_iBeacon as well, add USE_IBEACON_ESP32
Ensure USE_MI_ESP32 is not defined.

### documentation

### Personal motivation:

### Requests:
Scan results are currently not published - there is a lot of information in the debug loigs to work with, but I believe that scan results should be published - thoughts on this appreciated (e.g. should you have to ask for them, should they be a web UI feature, etc.).
Also thoughts on blocking devices in busy areas would be appreciated - this is a feature of xsns62 which is not in xdrv46.

## Checklist:


btsimonh:
should I convert this to a draft?

arendst:
Thx.

- As the driver id 46 is already used for ccloader you are requested to change it to 47.
- For testing I recommend you add the needed defines to your user_config_override.h file and chk it is compiled for both esp8266 and esp32. If OK then only provide the files needed but do not enable them in my_user_config.h

@staars can you have a look if this makes sense to replace/add your work so far?

btsimonh:
Hi Theo,

Renamed to xdrv_47 (sorry, was a few weeks back in my repo when I chose 47!).
Added a commented line in "my_user_config.h"

tested building with the defines
#define USE_IBEACON
#define USE_BLE_ESP32
//#define USE_MI_ESP32
fine for env:tasmota-sensors 
fine for env:tasmota32-sensors

My first visit back here in a couple of years, and things moved on so much :).  looking good as always.  Nice to find some ESP32 support, and BLE as a bonus!

This is heavily inspired by @Staars 's code.

Staars:
The fundamental idea to allow access to a shared resource is a good one.
I also can not spot any "completely wrong" design decisions (which of course is always a matter of personal taste).

Questions are: How stable does this (added overhead) work? How much memory do we burn? Who will maintain it?

If only one driver would use this, it would make zero sense to me. We would only add overhead and an additional chance for bugs.

I do not know, if the ESP32 has enough resources to run more than one BLE-driver (=micro application) inside Tasmota sufficiently. Only real world tests would show this. 
Bluetooth is still a moving target on the ESP32 and a few month ago my answer would have been: Forget it.
But we have seen some progress in the last weeks with the never core-versions and maybe it can really work this way.

My suggestion:
@btsimonh Please add additional version of xsns_62 and(!) xsns_52 which are user selectable at compile time.
These should be provided as a build option to see, if they work for the users (including the driver maintainers). Then a reasoned decision could be made, if all drivers should be converted by default.
I would not recommend to change to the new driver model before this (open) test.

arendst:
So pls go forward as @staars suggested allowing to test pros and cons

btsimonh:
sounds like a plan.

I will:
- duplicate the existing xsns_52 and add #ifndef USE_BLE_ESP32
- duplicate the existing xsns_62 and add #ifndef USE_BLE_ESP32
- strip the duplicates of native BLE code for clarity (leaving only support the xdrv_47 form).
- Hide them behind USE_BLE_ESP32 and USE_BLE_EXAMPLES
- Hide them behind USE_BLE_ESP32 and USE_BLE_EXAMPLES

once stable and happy, if @Staars could do an initial checkover, then we ask for a couple more interested parties to test?

I'll also put together some form of NodeRed flow for hammering the exposed interfaces over periods of time.

br,  Simon

btsimonh:
I'm going to close this PR and start another cleaner version.

btsimonh:
ok, new PR ready for analysis.
Lost a day to my real work, so be kind :).

PR #10041, opened 2020-12-02
Disable IR receive while sending

s-hadinger:
## Checklist:

### DIFF
--- a/tasmota/xdrv_05_irremote.ino
+++ b/tasmota/xdrv_05_irremote.ino
@@ -119,7 +119,6 @@ const char kIrRemoteProtocols[] PROGMEM = "UNKNOWN|RC5|RC6|NEC";
 #include <IRsend.h>
 
 IRsend *irsend = nullptr;
-bool irsend_active = false;
 
 void IrSendInit(void)
 {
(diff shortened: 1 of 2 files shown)


Jason2866:
Tested. Great! No more mirror ghost commands.

pilaGit:
If a feature (IrReceived#Data) is removed, it should be selectively removed. If nobody had this above problem for years, most likely, problem here is caused by funky IR blaster of this particular user.

It should be reversible for all other users (except for this one). This change killed my hybrid central heating / heat pump heating/cooling system.

So, it should at least be selectable, like Shutters or HVAC, by e.g.

#define ir_disable_recv

or even better with a corresponding

SetOptionXX

Pleas, not just killed after many years.

s-hadinger:
What you call a "feature" was considered as an anomaly for many users:
- it created unnecessary MQTT messages
- it was unreliable and device dependent (some devices tended to return garbage)
- it was most certainly disturbing the timing of IR signals

Unless you explain why this "feature" is useful for many users, it will stay like this.

I'm sorry that it killed your central heating / heat pump heating/cooling system; but be very careful to rely on unreliable features.
If you absolutely want to reliably receive the IR sent, I highly suggest to add a second Tasmota device just for listening messages.

s-hadinger:
@pilaGit I added the compile time `#define IR_RCV_WHILE_SENDING 1` so you can easily compile your own version with the feature you want.
The standard version will remain with receiving disabled.

Jason2866:
There where open issues. If you had done a search in issues you would have found.
This PR is clearly a bug fix. 
Tasmota has breaking changes now and then. Updating a productive working setup can have always unwanted effects.

PR #10046, opened 2020-12-03
Generic ESP32 BLE driver, plus sample modified drivers

btsimonh:
## Description:
Introducing xdrv_47_BLE_ESP32.
I believe there is room for web UI support, and scan results -> MQTT (although this could be a separate driver - that's kind of the point).

### The issue:

### xdrv_46 provides:
2/ generic BLE 'operations'.

### Example modified drivers:

These have been adjusted to instead of using the NimBLE stack directly, to subscribe to services provided by xdrv_47_BLE_ESP32.

### use:
xdrv_47 is enabled via 
USE_BLE_ESP32 and ESP32
To enable xsns_52_ibeacon_BLE_ESP32 as well, add USE_IBEACON_ESP32
To enable xsns_62_MI_ESP32_BLE_ESP32 as well, add USE_IBEACON_ESP32

### documentation

### Personal motivation:

### Requests:

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -821,6 +821,7 @@
 
 //#define USE_SPI                                  // Add support for hardware SPI
 #define USE_MI_ESP32                             // Add support for ESP32 as a BLE-bridge (+9k2 mem, +292k flash)
+//#define USE_BLE_ESP32                             // Add support for ESP32 as a BLE-bridge (+9k2? mem, +292k? flash)
 //#define USE_WEBCAM                               // Add support for webcam
 
 #endif  // ESP32
--- a/tasmota/xdrv_47_BLE_ESP32.h
+++ b/tasmota/xdrv_47_BLE_ESP32.h
@@ -106,6 +106,7 @@ namespace BLE_ESP32 {
 // states for runTaskDoneOperation
 #define GEN_STATE_IDLE 0
 #define GEN_STATE_START 1
+#define GEN_STATE_STARTED 2
 
 #define GEN_STATE_READDONE 7
 #define GEN_STATE_WRITEDONE 8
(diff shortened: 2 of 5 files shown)


btsimonh:
@Staars - ready for your consideration.

btsimonh:
last commit should prevent any Logging evil from the use of a task (and maybe extra tasks inside NimBLE?).
It's expensive on memory, and could be improved. BLE_ESP32::SafeAddLog_P( ) adds logs to a queue, which is serviced from the main thread.

btsimonh:
I'm happy with reliability and operation, but I'm struggling with keeping the memory use under control, which is so important in Tasmota.

I'm going to re-factor the advertisment callback (where I currently 'break out' all the data for easier usage) to eliminate memory copying, but putting the onus on other drivers to use the NimBLE APIs to get the data they need.

I'd like some form of getting every advert on MQTT - but may end up with a simple list of MAC/Name, and then the ability to request either a single advert complete from a device, or every advert from that device until told to stop.

i.e. I imagine functionality where a user's (external) application needs to identify the devices present, determine which it is interested in, and then perform operations on them.

btsimonh:
see Commands commented section - comments before each 'CmndXXX' document the options available for 
BLEAdv|BLEOp|BLEMode|BLEDetails.

btsimonh:
removed MOST of the static variables from the advert structure, and modified sensor drivers to suit.
Done for today.

btsimonh:
@Staars - nothing more to do.  I believe it is stable, relatively complete, and functional.
Comments please.

btsimonh:
Added some basic documentation on use here:
https://docs.google.com/document/d/1PMaGJwhxtKDgruHc79bVby8Wx6kbHyKp6WM6AriXNMo/edit?usp=sharing
I HAVE done more - slimming down on the structures used in operations.

btsimonh:
Pls await further updates.
The BLE stack is less than stable; but, I have it working (fairly well) now without the underlying issues affecting Tasmota operation in general.  However, I'm working on user experience - i.e. driving it from Node-Red, and there are some changes to the MQTT to make it easier to know what you have done.

btsimonh:
@arendst - what are the chances of going to esp idf >v4.0?
I've hit a blocker on BLE in general with a closed library bug which was fixed in march, and is probable in idf-v4.0.  
espressif/arduino-esp32 has a V4.0 branch.....
Either that or I fix it be shimming inside NimBLE-arduino to cover it up.
With the fix in place, it can be quite reliable.  Without, any connecting to any device risks BLE hangups (1/1000 in my tests).

btsimonh:
closing. will issue a new one on latest development branch

PR #10055, opened 2020-12-04
Improved Opentherm error handling.

definable:
## Description:
With Opentherm a frame might ocassionally get corrupted. According to the OT standard the conversation should be aborted in such a case.  With the original implementation a single corrupted frame resulted in change to disconnected, which after a "cooldown" time of 10 seconds attempted to start the conversation again.  This method has 2 drawbacks: 
- After a single frame error the communication is 'dead' for at least 10 seconds.
- According to the OT standard a boiler should fallback to OT-/PWM mode when the line level is low for > 5 ..15 seconds. Low level is the default idle state, depending on the boiler type it changes to PWM mode and does not respond easily  to new OT frames. Depending on boiler implementation it checks for a message at specific intervals.  

This fix reduces the cooldown period to < 5 seconds minimizing the risk of the boiler entering PWM  state. 
It adds an error counter, disconnected state is only entered after 3 errrors.  

## Checklist:

### DIFF
--- a/tasmota/xsns_69_opentherm.ino
+++ b/tasmota/xsns_69_opentherm.ino
@@ -33,7 +33,10 @@
 #define OT_BOILER_DEFAULT 85;
 
 // Seconds before OT will make an attempt to connect to the boiler after connection error
-#define SNS_OT_DISCONNECT_COOLDOWN_SECONDS 10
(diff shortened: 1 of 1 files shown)


yury-sannikov:
hi @definable, just curious, how stable your OT connection? How often do you get the disconnected state?

PR #10067, opened 2020-12-05
Add protection for AddLog_P for ESP32 tasks

### LINKED ISSUE #10049: logging corruption - observed on ESP32 with more that one 'task'
### PROBLEM DESCRIPTION
Whilst developing a new driver (xdrv_47_BLE_ESP32), serial logging corruption was observed.
the specific corruption was something like:
00:00:04 Startscan from on
00:00:04 Startscan from onesec

With ESP32, since multiple 'tasks' are possible (and essential for some operations like BLE if we are not to hold up the whole main process), it is possible to call, for example, Addlog_P from multiple threads.

We need to do some analysis of TasmotaGlobal and the data therein to see what data accesses must be protected by some form of mutex/semaphore when compiling for ESP32, else we will introduce instability and crash and burn badly and randomly.

Immediate danger would be:
Any logging from anything not on the main thread.
Any MQTT activity from anything not on the main thread.

Until we have some form of protection, all logging from secondary threads should be avoided, as should any activity against MQTT or other global data.
One issue is that it may not be obvious when using a 3rd party API like NimBLE if a callback is on the main thread or on a secondary task.

### REQUESTED INFORMATION

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): ESP32 dev board
  - [ ] Pre-compiled
  - [x] Self-compiled
```
  Configuration output here:


```
```
  Rules output here:


```
```
  STATUS 0 output here:


```
```
  Console output here:


```

### TO REPRODUCE
n/a

### EXPECTED BEHAVIOUR
n/a

### SCREENSHOTS
n/a

### ADDITIONAL CONTEXT
I would assume this only affects ESP32, as ESP8266 does not have multiple threads? (but i'm no expert).

Example of possible Semaphore use:
(what is port_MAX_DELAY anyway? - must check).
```
// this protects our global data, which can be accessed by multiple tasks 
SemaphoreHandle_t  GlobalsMutex;
...
  GlobalsMutex = xSemaphoreCreateMutex();



// example from support.ino
  xSemaphoreTake(GlobalsMutex, portMAX_DELAY);
    memmove(TasmotaGlobal.log_data + strlen(syslog_preamble), TasmotaGlobal.log_data, sizeof(TasmotaGlobal.log_data) - strlen(syslog_preamble));
    TasmotaGlobal.log_data[sizeof(TasmotaGlobal.log_data) -1] = '\0';
    memcpy(TasmotaGlobal.log_data, syslog_preamble, strlen(syslog_preamble));
    PortUdp_write(TasmotaGlobal.log_data, strlen(TasmotaGlobal.log_data));

  xSemaphoreGive(GlobalsMutex); // release mutex

```
The take/give could be a #define, making if compile to nothing when not ESP32?
There are a lot of references to .log_data, .mqtt_data, etc. so this is no small undertaking.

btsimonh:
## Description:
prevents AddLogP from calling other functions if the log call was from a task other than the main task.
Calls PerpLog_P with the format string from the AddLogP call with LOG_LEVEL_ERROR as a guide for diagnosing what called.

**Related issue (if applicable):** fixes #10049

## Checklist:

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -20,6 +20,11 @@
 IPAddress syslog_host_addr;      // Syslog host IP address
 uint32_t syslog_host_hash = 0;   // Syslog host name hash
 
+#ifdef ESP32
+  TaskHandle_t mainTaskHandle;
+  int logTaskMismatch = 0; 
+#endif
+
 extern "C" {
 extern struct rst_info resetInfo;
 }
@@ -75,6 +80,10 @@ void OsWatchInit(void)
   RtcSettings.oswatch_blocked_loop = 0;
   oswatch_last_loop_time = millis();
   tickerOSWatch.attach_ms(((OSWATCH_RESET_TIME / 3) * 1000), OsWatchTicker);
+
+#ifdef ESP32
+  mainTaskHandle = xTaskGetCurrentTaskHandle();
+#endif
 }
 
 void OsWatchLoop(void)
@@ -2003,8 +2012,25 @@ void PrepLog_P(uint32_t loglevel, PGM_P formatP, ...)
   TasmotaGlobal.prepped_loglevel = loglevel;
 }
 
+#ifdef ESP32
+int ThreadError(const char *reason){
+  SemaphoreHandle_t thisTaskHandle = xTaskGetCurrentTaskHandle();
+  if (thisTaskHandle != mainTaskHandle){
+    PrepLog_P(LOG_LEVEL_ERROR, PSTR("Thread error: context:%s"), reason);
+    return 1;
+  }
+  return 0;
+}
+#endif
+
 void AddLog_P(uint32_t loglevel, PGM_P formatP, ...)
 {
+#ifdef ESP32
+  if (ThreadError(formatP)){
+    return;
+  }
+#endif
+
   if (TasmotaGlobal.prepped_loglevel) {
     AddLog(TasmotaGlobal.prepped_loglevel);
   }
@@ -2019,6 +2045,11 @@ void AddLog_P(uint32_t loglevel, PGM_P formatP, ...)
 
 void AddLog_Debug(PGM_P formatP, ...)
 {
+#ifdef ESP32
+  if (ThreadError(formatP)){
+    return;
+  }
+#endif
   va_list arg;
   va_start(arg, formatP);
   vsnprintf_P(TasmotaGlobal.log_data, sizeof(TasmotaGlobal.log_data), formatP, arg);


arendst:
I redesigned logging hoping to solve the issue you provide this PR for.

PrepLog_P is gone.

Can you pls verify that the latest developments provide the same function as you hoped to get here?

What changed is that the (global) heap log buffer has been replaced by a (local) stack log buffer solving corruption of log buffer.

btsimonh:
Will take a look.  Just trying to finalise all the BLE stuff at the moment; making sure OTA works, etc.
So I'll drop it over latest dev branch and try the new logging.  a few days?

btsimonh:
ohh... so had a look.
We're going to need mutex protection on TasmotaGlobal.log_buffer and associated control variables.
The implementation using a global buffer with the various log dests pulling from there is nice (more memory efficient, and centralised, compared to my 'SafeAddlog_P' which had a fixed set of constant sized buffer), but if the buffer got screwed once, everything will fall apart really badly....

My next task is to drop my BLE stuff on top of development again, ready for a (hopefully final) PR.  Give me a few days, and I'll propose a global auto-mutex class which we can use in AddLogData and GetLog, which should be the protection we need for ESP32 - and it should be well tested by then too.  I'll drop it as a separate PR if it works....  
I'm not sure about interrupts - I don't even know if the interrupt context supports printf....  we probably should not be logging direct from interrupts except in fatal circumstances, and then only to serial; but I'll defer to the more knowledgeable on that.

btsimonh:
Theo, have a look at https://github.com/btsimonh/Tasmota/tree/loggingmutex two changed files.
The automutex class I added in support.ino can also be used elsewhere, and is only active in esp32.
If you think it's ok, feel free to pull it over.
It's been running here for an hour, no obvious defects on ESP32; but it's based on the code I made for the BLE stuff, so the idea is heavily tested over the last 6 weeks.

arendst:
OK will have a look once I have my SPI project finished which currently breaks stuff. Give me a few days.

btsimonh:
have thrown in a pr.

btsimonh:
@arendst - Theo, ref the logdata being in stack, this adds 700 bytes per thread/task.
1/ Can I move it back to be a static variable, so we only have one of them, protected by the mutex?
2/ You eliminated PrepLog_P - does this mean AddLog_P may be called from interrupt routines now? (i'm scared...)

arendst:
The point of moving logdata to the stack was fixing corruption and no need for a mutex. The advantage was also gaining 700 bytes RAM as most logging messages are way smaller than 700 bytes.

I'll plan and anticipate to make current 700 bytes stack use lower once I have a complete picture of current logging strings; the biggest once, which lead to the 700 size are already handled by their own stack reservations (mqtt logging) so I expect considering saving back to 300 bytes.

The function of PrepLogP was to save logging string before showing it as showing it was the showstopper for both udp and mqtt. The new logging environment does kind of the same so the need for PrepLogP is gone. As it is in use for over a week I haven't seen exceptions on emulation enabled devices caused by this (which was the main culprit).

So in short I do not want to go back to static logbuffer. Certainly not on esp8266.

btsimonh:
sounds good.  I have a small update on the mutex stuff, but the logging will remain unchanged.
Thanks, Simon.  
(p.s. I'm hoping to finish BLE before 'real work' on Monday! :) ).

btsimonh:
closing as really happy with where we got to....

PR #10071, opened 2020-12-06
Add new ShutterMode 7 (SHT_TIME_AWNING) to support some awnings

th-wilde:
## Description:

This will add support for some push-button controllable awnings.
The awning will extend/shrink when it detects a pulse. It will stop by
reaching its end-position or by another pulse while moving. The
direction changes with every move starting pulse.

This behavior is very similar to ShutterMode 3 (SHT_TIME_GARAGE)
use-case of garage-doors. But i encountered a problem with the pulse
generated by ShutterMode 3 when the end position is reached. It triggered
the awning again. So this is just a slightly modified ShutterMode 3 that
do not send pulses if an end-position (0 or 100%) of the shutter is
reached.

**Related issue (if applicable):** None. 

## Checklist:


ascillato2:
@stefanbode 

Please, can you take a look to this? What do you think? Thanks.

stefanbode:
I will check, if this implementation is a good default for the garage mode in any case. I agree that there is no stop pulse required if the garage door reaches final position.

stefanbode:
@th-wilde: Can we explore a bit more on this issue/enhancement. In general, it makes sense. I'm just thinking about some side effects and if we can solve the issue more elegant. I think we will still have issues if we send the pulse at 2% or 98% because it can never be that exact. Additionally, I assume every garage opener has the auto end stop. Therefore the configuration should be something like 0..10% and then it will e.g. not send a stop signal above and below this position. As a result, you cannot set the garage door to a position e.g. 2%. In this case, it will fully close and also recognize the garage position as 0% and not 2%. Better?

th-wilde:
@stefanbode: Since tasmota can't know the real state of the garage door control (or in my case the awning control), there will always be room for edge cases. The parameters for time based shutters (`ShutterOpenDuration`, `ShutterCloseDuration`, `ShutterEnableEndStopTime`) should already allow a lot of tweaking to keep those edge-case-rooms small. If i understand the `ShutterEnableEndStopTime`-Option correctly, than it's purpose is to  deal with the end-position-uncertainty (by adding an additional second when reaching an expected end-position; just to be sure that the real end-position will be reached). So enabling `ShutterEnableEndStopTime` and adjusting `ShutterOpenDuration`/`ShutterCloseDuration` in respect, should already have a similar effect like your suggestion. Or did I miss something?

peter--s:
Hello,
While I se no issues near fully open, I consider it desirable to be able to let a garage door move to a nearly closed position, keeping people and pets out while cables can be fed through without getting crushed upon closing (leaving a crack for ventilation can also be desirable).
Please consider these use-cases.
Regards, Peter

bradkeifer:
Any chance we can get this PR opened and progressed? It sounds as though there is a legitimate requirement that is currently not being met. From my reading of the request, this proposed ShutterMode will provide the equivalent of OSC (Open, Stop, Close) control used on many garage doors. The example I would use this for is also for control of an awning, as per @th-wilde requirement.

stefanbode:
Can you open a new ticket and describe in detail the use case and the expected behavior. The garage Mode is now already implemented. Please describe the delta. If possible I will make an enhancement. I assume you will support in testing because I have no hardware here to do so

But first take a closer look to the "garage" mode. 

Stefan

bradkeifer:
Thanks for looking at this @stefanbode. I was able to work out an alternate solution to my requirement using ShutterMode 1 and a second control option on the radio transmitter device I was interfacing to (Becker Centronic Variocontrol VC421A Transmitter). Sorry for any inconvenience caused. 
I did have a look though the shutter device code and there are two PWM modes in the code, but only one in the documentation - I'm not sure if you're aware of that?

stefanbode:
The PWM modes are for two different kind of devices. One is the stepper motor that controls speed through the frequency of the pwm and the position by the number of pulses and the other one is for SERVO motors. These use the PWM duty time to control position very similar to an LED. As far as I know stepper and servo are documented. But I will double check

PR #10082, opened 2020-12-07
Update Let's enctypt CA to use Intermediate R3

xhess:
## Description:
Recently Let's encrypt switched from using the "Let's Encrypt Authority X3" certificate for issuing new certificates to the "R3" Certificate. X3 is now retired. This uses their own certificate, not cross singed, since it valid till 2025. The cross singed would only be valid till Sept. 2021.

## Checklist:


s-hadinger:
Thanks for the update. Are the previously issued certificate compatible with the new R3 authority or do we need to keep both X3 and R3 for the next 3 months?

xhess:
with this change. No. (Didn't test but I am pretty sure)
After the request I saw that there is a variable called something like LetsEncryptX3. This should also be changed. But then other files need to be changed as well.

s-hadinger:
Thanks for your work. I have created a new PR to add R3 instead of replacing X3 #10086

Can you please confirm it works with certificates signed by R3 CA?

xhess:
I need to see when I have time. But I have to do it within the next two weeks because my old LE-Certificates will then expire. At the moment I am using the old certificate for mqtt.

WillemJansen:
I can confirm - however I can only confirm this for the cross-signed certificate. It seems that by default, LE signs with the cross-signed cert, not the X1 self-signed one.

best way to check your own cert: 
openssl s_client -showcerts  -verify 7 -connect <your.mqtt.server>:8883

arendst:
Duplicate from #10086

PR #10098, opened 2020-12-08
Moved from cross-signed X3 to cross-signed R3 Lets encrypt certificate

WillemJansen:
Committer: Willem Jansen <willem@jansen.xyz>

## Description:
Given the recent retirement of the Let's Encrypt X3 certificate (see https://letsencrypt.org/certificates/), no new LE-certificate can be recognized as valid by Tasmota anymore. Therefore, I upgraded the trust anchor to the cross-signed R3 certificate.

## Checklist:
  - [X The code change is tested and works on Tasmota core ESP8266 V.2.7.4.9


s-hadinger:
How do you know there aren't any valid X3 signed certificates out there? Shouldn't we wait for 3 months for all previously issues certificates to expire?

xhess:
Duplicate of #10082

WillemJansen:
This might be a valid concern, but could easily be healed by re-requesting a LE certificate (certbot  renew). The alternative would be to have both x3 and r3 in the Code 

In my opinion, the once-and-for-all-solution should involve the EC-keys, given that would speed up Tal ops, too.

s-hadinger:
Since PR #10086 was merged, Tasmota contains both X3 and R3 CA. We will remove X3 when all certs are renewed, which I hope will happen in 3 months max.

From my previous tests, ECDSA is much slower than ECDHE with RSA certificate. Do you have other numbers?

arendst:
Duplicate from #10086

PR #10115, opened 2020-12-12
Zigbee add visual map of network

s-hadinger:
### **Major feature / Experimental**

Added visual map of the Zigbee network, including LQI of each link (if known).

ZbMap_example

Pressing the `Zigbee Map` button triggers an automatic `ZbMap` command and waits until mapping is complete. It takes 2 seconds per device.

Red circle is the coordinator, orange boxes are routers, blue ellipses are end-devices.

Two buttons are also added on the main page, to allow joining of new devices (`ZbPermiJoin`) and Zigbee map (`ZbMap`), and a new visual indicator when joining of new devices is allowed.

ZbPermitJoin

Note: the visual map uses `vis.js` network library, which is not included as it is too big to fit in Tasmota memory. For this reason, viewing the map requires that your browser has internet access. But no data is tranferred to any site and the rendering is done locally in the browser.

## Checklist:


MattWestb:
Great work Stephan !!!!
And i can see that your Philips HUE bulb have the same bug as my = not deleting end devices that have timed out and keep reporting it in its neighbour table. Hard resetting the bulb and rejoining it and its OK (until next end device is leaving) and perhaps the same with the OSRAM plug (or the IKEA remote have changing parent for some seconds / minutes ago). 
Looks very nice and helping much then having problems with the network :-)))

s-hadinger:
Thanks. Indeed this will probably help many users understand better their zigbee network.

MattWestb:
Is it possible using the EEPROM on the Sonoff board as cache for the `vis.js` file then is being downloaded if the EEPROM is present and dont need waiting for it being downloaded every time loading the map ?
Only brainstorm as usual from my side :)

s-hadinger:
Unfortunately not. Vis.js is more than 600kb in size, way too big for eeprom and a waste for Flash

PR #10118, opened 2020-12-12
BLE on ESP32 Xdrv 47 ble esp32 - a generic driver

btsimonh:
## Description:
Introducing xdrv_47_BLE_ESP32.
I believe there is room for web UI support, and scan results -> MQTT (although this could be a separate driver - that's kind of the point).

### The issue:

### xdrv_46 provides:
2/ generic BLE 'operations'.

### Example modified drivers:

These have been adjusted to instead of using the NimBLE stack directly, to subscribe to services provided by xdrv_47_BLE_ESP32.

### use:
xdrv_47 is enabled via 
USE_BLE_ESP32 and ESP32
To enable xsns_52_ibeacon_BLE_ESP32 as well, add USE_IBEACON_ESP32
To enable xsns_62_MI_ESP32_BLE_ESP32 as well, add USE_IBEACON_ESP32

### documentation

### Personal motivation:

### Requests:

## Checklist:

### DIFF
--- a/lib/libesp32/NimBLE-Arduino/src/NimBLEClient.cpp
+++ b/lib/libesp32/NimBLE-Arduino/src/NimBLEClient.cpp
@@ -189,6 +189,8 @@ bool NimBLEClient::connect(const NimBLEAddress &address, bool deleteAttibutes) {
 
     int rc = 0;
 
+    m_waitingToConnect = true;
+
     /* Try to connect the the advertiser.  Allow 30 seconds (30000 ms) for
      *  timeout (default value of m_connectTimeout).
      *  Loop on BLE_HS_EBUSY if the scan hasn't stopped yet.
@@ -201,6 +203,7 @@ bool NimBLEClient::connect(const NimBLEAddress &address, bool deleteAttibutes) {
         }
     }while(rc == BLE_HS_EBUSY);
 
+    m_result = rc;
     if (rc != 0 && rc != BLE_HS_EDONE) {
         NIMBLE_LOGE(LOG_TAG, "Error: Failed to connect to device; "
                     "addr=%s, rc=%d; %s",
@@ -210,8 +213,6 @@ bool NimBLEClient::connect(const NimBLEAddress &address, bool deleteAttibutes) {
         m_waitingToConnect = false;
         return false;
     }
-
-    m_waitingToConnect = true;
 
     // Wait for the connection to complete.
     ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
@@ -229,6 +230,10 @@ bool NimBLEClient::connect(const NimBLEAddress &address, bool deleteAttibutes) {
     NIMBLE_LOGD(LOG_TAG, "<< connect()");
     return true;
 } // connect
+
+int NimBLEClient::getResult(){
+  return m_result;
+}
 
 
 /**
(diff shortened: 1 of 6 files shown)


Jason2866:
Dont get me wrong, this are just my personal thoughts.
I dont see a real benefit. If i understand it correct it adds code to read maybe?! somemore devices.
Why should this be done? A half supported device has no benefit. It will generate "only" entrys in issue section.
My suggestion only adding code for a real use case / device. All other code is just littering Tasmota.

btsimonh:
Hi Jason,
the benefits:
1/ a common way to create sensor drivers which can co-exist and use BLE.  (e.g. MI and iBeacon now co-exist).
2/ it copes with a number of BLE failure modes gracefully. (3 weeks of logging, 150Mbytes of log data).
3/ Allows developers to hear and access devices not supported directly.  If these become popular, it's then very easy to make them part of Tasmota.  i.e. it prevents bloat in Tasmota.
4/ BLE is evil.  This driver brings all that 'evil' into one maintainable place.  One 'task', one thread, less risk, more maintainable.  Hopefully some of the evil will go away as we ride the IDF version up.
5/ You are correct, it does not 'support' a specific device or use case - hence why it is a xdrv, not an xsns.  It's not really intended to be used directly by end users.  But all of it's features ARE used by the MI driver, and so all are required for 'general' BLE.
br, simon.

btsimonh:
would it help If I added an EQ3 radiator valve driver as a use case?
s

btsimonh:
@Jason2866 - pls see
https://github.com/btsimonh/Tasmota/tree/xdrv_48_EQ3_ESP32

This adds a functional EQ3 TRV valve driver on top of the PR; as a demonstration of how a driver can be built on top easily.
EQ3 control comes up multiple times in the forum, so I am not alone in this desire.
And of course, this works along side the MI driver and the iBeacon driver - so you can sense the room temp and control the valve at the same time from the same Tasmota device.

Of course, because there has been no feedback about other BLE devices, I can't confirm that the same will be the case for other desired drivers.  But it xdrv_47 does support everything required for all the MI reading that requires connection, plus all required for EQ3, so there will be many other devices it will support.

br,
Simon

blakadder:
Good approach to utilise BLE as a singular device a not treat is as different devices depending on the use as it is currently

Jason2866:
Mhh, since ESP32 flash and ram is not such a rare good as for the ESP82XX why not merging 
(after release 9.2)? Let us see how BLE stuff goes with this changes.

Staars:
Yes, I think this could be done after the next release.

btsimonh:
:).  If anyone fancies trying to drag us up to a later IDF, I'd be happy to run more long term tests to see if the BLE gets better.  (a few web related things complain?).
I am considering if NimBLE is the right thing for us, both in memory use and because we don't need anything other than ESP32.  But h2zero has been **very** supportive....

Jason2866:
NimBLE will be the way for Tasmota. Every other solution needs to much flash and memory.

btsimonh:
@Jason2866  tentatively agree... but thinking it may be good to disconnect the drivers that use the BLE from needing to know it's NimBLE behind it.  e.g. making functions on the BLE driver which get the bits currently got from NimBLEAdvertisedDevice in advertisment callbacks, and have the pointer supplied in the callback opaque.  Just to future-proof us.
Thoughts?

btsimonh:
need help with sensors..  have reports of a crash on sensors which need a connection.  Need someone who has the sensors to debug?
Latest push is more stable ref scanning....

btsimonh:
failing because of BG language!

btsimonh:
keep holding - improvements coming - have a 'partner in crime' and receiving more sensors in a few days, so don't add until we are done.
thanks,
simon

btsimonh:
issing a new one over latest development branch

PR #10120, opened 2020-12-13
scripter update

gemu2015:
## Description:

add up to 5 pwm channels  pwm1(a b)
add larger compressed flash storage on 1M devices up to 6.5k script size
google charts add time fraction to x axis
fix i2s pin overrides

## Checklist:


arendst:
As far as I know there is no flash space below arduino defined eeprom area as it is used by tasmota rotating settings. The area is explicitly used during ota to save current settings. All space below can be overwritten by otaed file. Once running again the space below eeprom area will be used as rotating settings area again.

So I wonder how you keep your extended script save area valid.

gemu2015:
this area is erased during OTA and serial flash. so you need to load the script again after that. however
there are a couple of users that wanted more space on an 1M device.
there would be another solution to have settings rotation use only 7 slots and leave one free for script usage depending on a setoption.

PR #10139, opened 2020-12-15
fix day_of_week in tuyamcu

device111:
Tasmota begin with 1 for sunday but Tuya MCU Protokoll begin with 1 at monday in "day_of_week"

## Description:

## Checklist:


arendst:
Tasmota numbers days from 1 (Sunday) to 7 (Saturday) indeed.

With your fix you are numbering Tuya days from 0 (Sunday) to 6 (Saturday). Is that what you want? I don't think so as it would make more sense to number days 1 (Monday) to 7 (Sunday). 

Pls re-check your solution or affirm it is correct.

device111:
Oh, your are right. Thats a bad solution.
I will correct.

PR #8558, opened 2020-05-27
Add support for SetOption88 to put each relay in a separate DGR

pcdiem:
## Description:

Add support for SetOption88 to put each relay in a separate device group. This PR adds 160 bytes of code. When SetOption88 is enabled:

- Updates to the state of relay 1 are sent to device group 1, updates to the state of relay 2 are sent to device group 2, etc.
- Power updates sent to each device group send a power mask shifted so affected bit is in position 0.
- Messages received from device group 1 update the state of relay 1, messages received from device group 2 update the state of relay 2, etc.

This enables devices with multiple relays (real or virtual) to control other single relay devices (eg. a device with three buttons and three relays can be set up with three device groups to toggle the power of three single relay devices).

SetOption88 was originally only used to enable Remote Device Mode of the PWM Dimmer module. This added functionality is essentially the same concept the same SetOption88 was used which causes no conflicts with it's use in the PWM Dimmer module.

**Related issue (if applicable):** fixes #<Tasmota issue number goes here>

## Checklist:
  - [x] The pull request is done against the latest dev branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR.
  - [x] The code change is tested and works on core ESP8266 V.2.7.1
  - [x] The code change is tested and works on core ESP32 V.1.12.2
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #10140, opened 2020-12-15
When SetOption88 is enabled, sync the power state of all relays that use the same DevGroupName

gbg108:
## Description:
The SetOption88 feature provides support for multiple relays using different device groups however there is an issue when multiple relays are assigned the same device group name. In this case only the first relay will be sync'ed and others will be ignored.

For example, if I have three devices, one device (D1) with four relays, a second (D2) with a single relay and a third (D3) with a single relay and I would like to keep two relays on D1 in sync with D2 and two relays on D1 in sync with D3. To do this I would set the following device group names:

**D1:**{"DevGroupName1":"flood_lights","DevGroupName2":"flood_lights","DevGroupName3":"christmas_lights","DevGroupName4":"christmas_lights"}

**D2:**
{"DevGroupName1":"flood_lights","DevGroupName2":"","DevGroupName3":"","DevGroupName4":""}

**D3:**
{"DevGroupName1":"christmas_lights","DevGroupName2":"","DevGroupName3":"","DevGroupName4":""}

With the current code when D2 is toggled, only relay1 on D1 will toggle. Similarly when D3 is toggled only relay3 on D1 will toggle.

The change in this PR will keep the power state of all local relays assigned the same device group name in-sync the same as is done with remote relays so any change to D2 would update relay1/2 on D1 and any change to D3 would update relay3/4 on D1. 

Also, assume now D2 and D3 do not exist. In that case I would still expect the power state of relay1/2 and relay3/4 on D1 to always stay in sync with each other since these relays share the same device group name and SetOption88 is enabled. This change will do that.

Note, I currently only have access to ESP8266 devices and have not tested on ESP32.

**Related issue (if applicable):** fixes #8558

## Checklist:

### DIFF
--- a/tasmota/support_tasmota.ino
+++ b/tasmota/support_tasmota.ino
@@ -586,8 +586,22 @@ void ExecuteCommandPower(uint32_t device, uint32_t state, uint32_t source)
       else
         SendLocalDeviceGroupMessage(DGR_MSGTYP_UPDATE, DGR_ITEM_POWER, TasmotaGlobal.power);
     }
+
+    if (device_groups && Settings.flag4.multiple_device_groups) {  // SetOption88 - Enable relays in separate device groups
+      for(int other_index=0; other_index < TasmotaGlobal.devices_present; other_index++) {
+        if (((device - 1) != other_index) &&
+            (!strcmp(device_groups[device - 1].group_name, device_groups[other_index].group_name))) {
+          uint32_t on = (TasmotaGlobal.power >> (device - 1)) & 1;
+          uint32_t other_on = (TasmotaGlobal.power >> other_index) & 1;
+          if (on != other_on) {
+            ExecuteCommandPower(other_index + 1, (on ? POWER_ON : POWER_OFF), source);
+          }
+        }
+      }
+    }
 #endif  // USE_DEVICE_GROUPS
     SetDevicePower(TasmotaGlobal.power, source);
+
 #ifdef USE_DOMOTICZ
     DomoticzUpdatePowerState(device);
 #endif  // USE_DOMOTICZ


arendst:
@pcdiem can you verify this PR pls.

pcdiem:
You can set multiple device group names to the same name. As you discovered, only the first one will stay in sync. I can add a note about this to the docs.

gbg108:
@pcdiem 

Thanks. Is not allowing multiple relays to use the same device group a design decision or would adding this functionality break existing code?

pcdiem:
Sorry, I missed the fact that this was a PR and not an issue. There are a couple problems with solving the issue with approach. 

One is that this will not keep the other device groups with the same name as a lower-numbered one in sync. When a device group message is received, only the first matching device group data is updated. If a device has relays 1 and 2 in device group Test, when the state of relay 2 changes, the existing code will send a device group message with a sequence number and member list that has not been kept up-to-date.

Second, this only addresses the relay states. The light module has been updated to send/receive device group updates to/from the device group associated with the device number of the light.

The goal here is to keep multiple relays on a device in sync with the relay(s) on another device. Keeping multiple relays in sync with each other on a single device would accomplish the same thing, wouldn't it? In your example, on device D1, if you set up a rule to keep relay 2 in sync with relay 1 and relay 4 in sync with relay 3, wouldn't that accomplish your goal?

Allowing the code to support multiple device groups with the same name should really be addressed in support_device_groups.ino so it can handle both incoming and outgoing messages and support all the items. Probably the best way to handle it would be to add a 16-bit bitmask to the settings together with a command to allow the user to map which device group (1-4) each relay is associated with. I suppose it would be nice that say that a relay should not be associated with any device group too but that would require 3 bits per relay.

gbg108:
@pcdiem Thanks for taking the time to review this.

> One is that this will not keep the other device groups with the same name as a lower-numbered one in sync

Perhaps I'm misunderstanding your point but this change works by checking all relays (0 through devices_present) to see if they share the same device group as the relay being changed. If they do share the same group and the power state is different then ExecuteCommandPower() will be called for the other relay. That call to ExecuteCommandPower() for the other relay should execute  SendDeviceGroupMessage() to update the state.

> Second, this only addresses the relay states. The light module has been updated to send/receive device group updates to/from the device group associated with the device number of the light.

This may very well be the case. I have no lights to test with, only switches.

> Keeping multiple relays in sync with each other on a single device would accomplish the same thing, wouldn't it?

I agree this could be accomplished with a rule, but that seems like a work around. As I understand it, SetOption88 ties device groups to individual relays, and the goal of device groups is to keep devices in sync. What would be the reason a user would assign the same device group to multiple relays but not expect they would be kept in sync? If they don't want them sync'ed they should assign a different group name or no group at all.

> Allowing the code to support multiple device groups with the same name should really be addressed in support_device_groups.ino

I originally did add the code in support_device_groups.ino and this worked fine to keep the relays in sync with remote devices. However it wasn't able to maintain sync with local relay changes. In other words, sticking with the original example, if I toggle the state of relay1 from the webUI relay2 wasn't being updated. As mentioned, I think it makes more sense to treat the concept of device groups consistently with SetOption88 enabled, even without a remote device and if a local relay is changed all other local relays in that group will follow.  To make this happen I moved the logic to ExecuteCommandPower() and not limited to SRC_REMOTE. If there's a way to achieve similar behavior in support_device_groups.ino then I agree that could be a better solution.

> Probably the best way to handle it would be to add a 16-bit bitmask to the settings together with a command to allow the user to map which device group (1-4) each relay is associated with

Decoupling the relay index from the DevGroupName index this way would be a good solution if there are plans to allow SetOption88 support for more than four relays. If not then I'm not sure what is gained. I can't think of a use case with multiple relays assigned the same DevGroupName and not wanting them to remain sync'ed.

Thanks again for the review and the great source!

pcdiem:
The code maintains details of each group member and the message sequence of the last message received and ack'ed from each member. When a device group message is received, it goes through the list of device groups and updates the member information for the first device group matching the device group name. If another device group has the same name, the member information will not be updated in that group.

The solution of adding a relay-to-device-group bit mask setting would allow mapping all eight relays to up to four device groups. The current code limits the number of relays to four if SO88 is enabled.

If we choose not to go the route of adding a bit mask setting, I think the cleanest and most efficient solution would be to dynamically create a bit mask at boot time to map the relays to device group indices. That way there would only be one set of information maintained per device group and the logic would be fully contained in device_group_support.ino. This removes the burden of handling this from any modules that make use of device groups (currently light and pwm_dimmer as well as support_tasmota).

I can code it up either way.

gbg108:
I'm understanding your concern a bit better now. Admittedly I haven't looked into the inner workings of device groups so I will defer to you on that.

I think either of your proposed solutions would work. Removing the four relay limitation in the process might be a good way to kill two birds with one stone. Also depending on how it's written, perhaps the need for SO88 could be eliminated by having the default mask apply to all relays/device groups. Changing the mask would effectively enable SO88 behavior without an explicit call. 

Whichever is chosen I don't believe there's any rush. My current code is working well enough for my purpose now. One the final solution is ready I will switch over.

Thanks

Jason2866:
@gbg108  If i understand the discussion correct this PR should be closed.

gbg108:
@Jason2866 Yes I will close it. Thanks

pcdiem:
Actually, I think this issue needs to remain open until Theo or whoever makes the call on which way he/they want to go with this. gbg108 is correct, this would eliminate the need for SO88 but that would require code to migrate the SO88 to the new bit mask during upgrades.

To summarize, we have two options: 1. Add a 16-bit mask to map relays to device groups. This will require a new command to set the mappings but adds the ability to map all eight relays. 2. Add logic to dynamically create a map at boot time to map device groups with the same name into the same device group index. This solution certainly has the smallest footprint but only supports four relays when SO88 is enabled.

pcdiem:
I'm sorry, I did it again and mistakenly took this as a feature request/issue. @gbg108 can you create the feature request? Then you can close this PR and and I'll take from there.

PR #10149, opened 2020-12-16
ESP32: correct div10 trigger to 1% ADC range

vic42:
## Description:

On ESP32 the DIV10 trigger did not function as documented over the full ADC range.

## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP32 V.1.0.4.2


arendst:
I've seen the discussion on discord.

How about looking at the number 10 as a deviation of 10 steps. In that case no change is needed. The goal of this number 10 was to get rid of small deviations from adc input readings (which go everywhere anyway, hence it is polled 32 times before returning a kind of stable output value).

So I'm not convinced this PR is needed. Still open to discussions though....

vic42:
1) current code chops all values over 1000 ADC counts so 3/4 ADC range may behave unexpectedly
2) The docs claim 1% range will trigger a DIV10 (docs can be changed of course)
3) ADS115 driver does claim to fire DIV10 at 1% change (although it may contain a bug that lets it fire at 0.5% change actually)

PR #10194, opened 2020-12-19
Fix WeMo Emu Issues

### LINKED ISSUE #10165: 9.1.0.2 Wemo and Alexa group problem
### PROBLEM DESCRIPTION
I have flashed some Devices with 9.1.0.2 and enabled WeMo Emulation.
Devices where found by Alexa with the right friendly Name (Stehlampe, Vitrine, Ventilator, Blume ...)
When i try to group the devices (Alexa App or at the PC) the wrong device is grouped and some devices are not possible to group.  

### REQUESTED INFORMATION

- [ ] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [ ] Searched the problem in the [forum](https://groups.google.com/d/forum/sonoffusers)
- [x] Device used (e.g., Sonoff Basic): Teckin sp22_
- [x] Tasmota binary firmware version number used: 9.1.0.2 and 9.2.0.1
  - [x] Pre-compiled
  - [ ] Self-compiled
- [x] Flashing tools used: OTA
```
  Configuration output here:


```
```
  Rules output here:


```
```
  STATUS 0 output here:


```
```
  Console output here:


```

### TO REPRODUCE
1. let discover alexa new devices with WeMo emulation
2. min. 2 devices (i have used 6 devices) (Stehlampe, Vitrine, Ventilator, Blume,...)
2. go to alexa app or alexa.amazon.de and create a new group
3. put the first device in the new group and save (e.g. Stehlampe)
4. save the group 
5. open the group and look which devices are in there. if you have luck, then Stehlampe is in your group. But if you have no luck, another device (e.g. Vitrine) is in your group. in my szenario with 6 devices i only can put 2 devices in the group. the other devices are displayed but never put in the group.

so you can not put the device vitrine in one group, because Stehlampe is always used when you choose Vitrine.
seems, that there is a unique id in the device, which is discovered by alexa and when you group the device, the first device with that id is used...?

### EXPECTED BEHAVIOUR
i would like to group devices in rooms, but some devices are not groupable

### SCREENSHOTS

### ADDITIONAL CONTEXT

peterchs:
## Description:
Wemo emulation issues were found after 9.2.0 upgrade:
1) On alexa discovery wemo devices were duplicated
2) Devices could not be added to Alexa routines

Testing showed these issues were due to changes to the UUID/Serial Number presented to Alexa discovery requests, where the device id was appended/included altering the format/content of these identifiers.  This fix ensures that for device/switch 1 the original single wemo discovery xml is returned.  For switch 2+ the current response will still be returned, so multiple device support is maintained.  

**Related issue (if applicable):** fixes #10165

## Checklist:
  - [X] The code change is tested and works on Tasmota core ESP32 V.1.0.5-rc4


arendst:
Doesn't your solution also mean that if you removed the wemo emulated device from your setup and then add it again with the new uuid it would solve this issue too?

peteakalad:
Yes correct that would fix the duplicate issue but does not fix the problems with in inability to add those devices to routines.  

There is something about the multi change to the serial number to add the device to the end of the serial number that Alexa doesn’t like.  This PR avoids doing that for device 1 effectively reverting the multi behaviour.

I would say that device 2+ will still have the same problem with this PR but exactly what serial number change would not break routines is not clear.

tasmota/xdrv_21_wemo_multi.ino
```diff
@@ -390,7 +394,14 @@ private:
...
 #endif
-    setup_xml.replace("{x1", SettingsText(SET_FRIENDLYNAME1 + (_deviceId - 1)));
+    if (_deviceId == 1)
+    {
+      setup_xml.replace("{x1", SettingsText(SET_FRIENDLYNAME1));
+    }
+    else
+    {
+      setup_xml.replace("{x1", SettingsText(SET_FRIENDLYNAME1 + (_deviceId - 1)));
+    }
```
arendst, line 404: This is not needed as in case of a single device ``_deviceId`` is 1 resulting in exactly the same Friendlyname as in your fix.

This adds to code size.

tasmota/xdrv_21_wemo_multi.ino
```diff
@@ -235,8 +235,12 @@ private:
...
   String WemoSerialnumber(void) {
     char serial[18];
-
-    snprintf_P(serial, sizeof(serial), PSTR("201612K%08X-%d"), ESP_getChipId(), _deviceId);
+    if (_deviceId == 1)
+    {
+      snprintf_P(serial, sizeof(serial), PSTR("201612K%08X"), ESP_getChipId());
+    } else {
+      snprintf_P(serial, sizeof(serial), PSTR("201612K%08X-%d"), ESP_getChipId(), _deviceId);
+    }
```
arendst, line 243: What hapens if you change this to 
```
snprintf_P(serial, sizeof(serial), PSTR("201612K%08X%02X"), ESP_getChipId(), _deviceId);
```
so removing the dash and make it hex.

OFreddy: I have tried it. This change works fine for me. The if/else is not necessary in both cases.

peteakalad: And you can add the device 1 to routines in Alexa app ok?

Still expect there to be issue with dupe devices after upgrade from pre-9.2.0 as the serial/uuid changed.

OFreddy: Yes, everithing works fine, now. I had problems adding the devices to a group or routine as not all of them where added to that group. I made several test and when adding a device to that group another device (I expect the one first discovered) was added. Now I can add every new device to every group as I want to. Before I can't add the devices "Terrasse", "Haustuer" and "Garten" to the same group. No it works.

Screenshot_20201220-131746_Amazon Alexa

arendst: Thx so the latest fix solved both issues.

Closing this PR.

peteakalad: I tested with these two changes to remove the if/else, and yes does fix the routine/group issue, but I do still end up with duplicate device upgrading when the device existed in Alexa before.  Both devices still work ok however.

PR #10213, opened 2020-12-21
Zigbee add unishox strings

s-hadinger:
## Description:

Moved most strings to language templates for localization.

Introduced a new way to compress a group of strings with Unishox, which achieves better compression than compressing each string separately.

For developers: see `UnishoxStrings.h` to know how to use.

Code decrease: 260 bytes (and will lead to more)

## Checklist:


arendst:
OMG! Do you really want to translate ``AddLog_P`` messages? 

I was silently assuming to get rid of translations for logging and use English instead....

arendst:
Oh. I just saw most messages are ResponseP so need translation. OK forget my concerns.

s-hadinger:
Those are only high level messages that can be translated. 

I plan to compress all the Zigbee attribute names. They occupy more than 6kb and I hope to save 3kb of Flash

PR #10229, opened 2020-12-23
Changed pzem DC device address

centja1:
## Description: Changed the default MODBUS address for PZEM003/017 DC Energy Meter from 0x01 to 0XF8.  While 0x01 is the default device address from the factory, it can subsequently be changed by sending the appropriate MODBUS command to the device.  0xF8 will always work in a single slave environment, avoiding a situation where the device address has been previously changed and is no longer known

[https://peacefair.en.made-in-china.com/product/FCsEdQqvwkVl/China-Pzem-017-DC-0-300V-Electric-Solar-Power-Meter-with-RS485-Modbus-with-300A-Shunt.html](url)

_"The address range of the slave is 0x01 ~ 0xF7. The address 0x00 is used as the broadcast address, the slave does not need to reply the master. The address 0xF8 is used as the general address, this address can be only used in single-slave environment and can be used for calibration etc.operation."_

## Checklist:

### DIFF
--- a/tasmota/xnrg_06_pzem_dc.ino
+++ b/tasmota/xnrg_06_pzem_dc.ino
@@ -31,7 +31,7 @@
 
 #define XNRG_06                    6
 
-const uint8_t PZEM_DC_DEVICE_ADDRESS = 0x01;  // PZEM default address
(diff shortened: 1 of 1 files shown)


arendst:
Thx for your effort and I understand yuour situation. But this driver supports up to three devices and tries to detect them based on their addresses.

So this won't be merged.

PR #10253, opened 2020-12-25
Improve WiFi scan result display

DL6ER:
## Checklist:

I cannot test ESP32 myself, however, compilation passes fine.

This improves WiFi scan results:

- Reveal if multiple AP have the same SSID but different signal strengths and show them group them together
- Color-code the signal quality to improve accessibility

Exemplary screenshots:

Screenshot from 2020-12-25 14-06-28

Further less important details like encryption type and RSSI are hidden in a tooltip, like
Screenshot from 2020-12-25 14-07-14


s-hadinger:
This is a nice feature but I'm not in favor of spending precious flash space for a cosmetic feature used only once in the lifetime of the device. 

Even if we were to add the feature, avoid `float` calculation as much as possible. It is very expensive in code size.

DL6ER:
Current `development`:
``` plain
RAM:   [======    ]  57.2% (used 46832 bytes from 81920 bytes)
Flash: [======    ]  58.4% (used 597500 bytes from 1023984 bytes)
```

My branch:
``` plain
RAM:   [======    ]  57.2% (used 46832 bytes from 81920 bytes)
Flash: [======    ]  58.4% (used 597692 bytes from 1023984 bytes)
```

If we leave off the description of the WiFi encryption type (because it shouldn't really matter), we get my new scanning display at a mere 96 bytes of extra space:
```
RAM:   [======    ]  57.2% (used 46832 bytes from 81920 bytes)
Flash: [======    ]  58.4% (used 597596 bytes from 1023984 bytes)
```

DL6ER:
The failed CI run send to be a mistake of the CI itself:

> Error: read ECONNRESET

tasmota/xdrv_01_webserver.ino
```diff
@@ -2114,36 +2111,46 @@ void HandleWifiConfiguration(void)
...
+            ssid.c_str(), WiFi.BSSIDstr(indices[i]).c_str(), WiFi.channel(indices[i]), rssi);
+
+          // Print SSID
+          WSContentSend_P(PSTR("<div><a href='#p' onclick='c(this)'>%s</a><br><ul>"),
+            HtmlEscape(ssid).c_str()
           );
-          delay(0);
+
+          int j = 0;
+          String nextSSID = "";
```
s-hadinger, line 2128: You can save code with
`String nextSSID;`

You don't need to give it a value, is it still initialized.

tasmota/xdrv_01_webserver.ino
```diff
@@ -2114,36 +2111,46 @@ void HandleWifiConfiguration(void)
...
+            int quality = WifiGetRssiAsQuality(rssi);
+            // Color-code quality
+            uint8_t colors[2] = { 0xFF, 0xFF };
+            if(quality > 50) {
+              // Scale red component to go from yellow to green (full green)
+              colors[0] = (0xFF * (100-quality))/50;
+            } else {
+              // Scale green component to go from red to yellow (full red)
+              colors[1] = (0xFF * quality)/50;
+            }
```
s-hadinger, line 2142: You can save even more here:

```
            if(quality > 50) {
              // Scale red component to go from yellow to green (full green)
              colors[0] = changeUIntScale(quality, 50, 100, 0xFF, 0);
            } else {
              // Scale green component to go from red to yellow (full red)
              colors[0] = changeUIntScale(quality, 0, 50, 0, 0xFF);
            }
```

I would also suggest to avoid pure Yellow #FFFF00 which is not readable with light background. Maybe going for a darker version would still do it:

```
            uint8_t colors[2] = { 0xAA, 0xAA };
            if(quality > 50) {
              // Scale red component to go from yellow to green (full green)
              colors[0] = changeUIntScale(quality, 50, 100, 0xAA, 0);
            } else {
              // Scale green component to go from red to yellow (full red)
              colors[0] = changeUIntScale(quality, 0, 50, 0, 0xAA);
            }
```

s-hadinger (commented):
With both suggested changes, you should be at only +64 bytes.

arendst:
As said in the original issue I won't do changes regarding preferred (b)ssid. This change is only half of what needs to be changed as the command line part has not been tackled.

Changing that too will increase code while current functionality is good enough for most users.

As it stands I won't merge but I will keep it in mind.

DL6ER:
This change is not meant at all into the direction of preferred BSSID, it just helps users to get a better picture of their WiFi neighborhood at the exact location of the Tasmota device. Hence, I also felt no need to change the command line part.

Concerning BSSID, I can see `SetOption56/57` is the preferred choice, even when I cannot use it right now as it leads to malfunction on my side (but that's a separate issue, https://github.com/arendst/Tasmota/issues/10254).

tasmota/xdrv_01_webserver.ino
DL6ER: I figured that most (all?) will run Tasmota in the standard configuration which seems to be a black background. This seems to be a wrong assumption and getting colors what are visible for all (e.g. green or orange background) is hard if not impossible. 

Removing the quality colors isn't much of a loss as the results are anyway sorted by RSSI (= quality) and I still think my improvement brings advantage to the display. Removing the color code, we're down to 597492 bytes so even ***less*** (8 bytes) than current `development`.

arendst:
No offence at all but did you notice you are using the same channel on equal ssids? The general approach is to NOT have the same channels on near APs. They should not overlap as if they do they interfere eachother.

If you manage to change them you'll see tasmota will much easier switch to the strongest signal AP.

DL6ER:
I have the very widely used Fritz!Box in its "Mesh" configuration. I noticed this defect as well, however, the manufacturer says that it is expected. I set both 2.4 GHz and 5GHz to "auto-channel". As result, the 2.4 GHz WiFi are all on the same channel. The interesting bit to note is that the 5GHz are all on different channels... And, of course, the channels can only be configured at the "master", not allowing any manual intervention as to what channels the individual APs use.

Feel free to close the PR if this is not where you think the project should be going. I'll rather concentrate on features rather than UI tweaks in the future (even when we have a variant right now that uses even less Flash than the current `development` build ;-)).

tasmota/xdrv_01_webserver.ino
DL6ER: Alternative - and color-independent - representation would be using HTML5 progress-bar
Screenshot from 2020-12-26 09-14-21

I haven't pushed this so far, but it'd come at `597536` bytes (`+36` compared to `development`). As said before, my change without color or progress-bar comes at `597542` bytes (`-8` compared to `development`).

arendst:
In the end it became this:

image

at the cost of over 500 bytes (!) but I would like to see the Signal Strength Indicator. It can be disabled by removing ``#define USE_ENHANCED_GUI_WIFI_SCAN`` (as I use it myself ;-)

Thx for the idea.

PR #10270, opened 2020-12-27
Fix problem of config loss if deepsleep loop is less than 10seconds

stefanbode:
## Description: 
Sometimes the deepsleep loop is too fast and causes a config reset. Fixed

## Checklist:
  - [x ] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.9
  - [ x] The code change is tested and works on Tasmota core ESP32 V.1.0.5-rc4

### DIFF
--- a/tasmota/xdrv_29_deepsleep.ino
+++ b/tasmota/xdrv_29_deepsleep.ino
@@ -142,6 +142,7 @@ void DeepSleepStart(void)
   WifiShutdown();
   RtcSettings.ultradeepsleep = RtcSettings.nextwakeup - UtcTime();
   RtcSettingsSave();
+  RtcRebootReset();
 #ifdef ESP8266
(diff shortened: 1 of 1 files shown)


arendst:
Was this stil true?

A few weeks ago I already did something like this:
```
  if (ResetReason() == REASON_DEEP_SLEEP_AWAKE) {
    RtcReboot.fast_reboot_count = 0;  // Disable fast reboot and quick power cycle detection
  } else {
    RtcReboot.fast_reboot_count++;
  }
```
to fix the issue.

stefanbode:
Yes, the device is running without any further issues with the current development branch. Anyhow from time to time it can happen that after a switch ON/OFF the device makes a reboot. But does not crashes anymore just over time and best it does not hang in this case. Currently all fine for me

stefanbode:
ok maybe I missed this one. The device with the error was not on the latest firmware and the error was obvious. Maybe we have now double fixed the issue.:-)

PR #10303, opened 2020-12-29
Unishox - enhance the supportability for the xdrv_01_webserver.ino scripts/headers etc.

btsimonh:
## Description:
Building on @Staars work on unishox for the webserver, this PR separates out the uncompressed html/js/style strings into obviously named separate header files in a folder.
A modified version of the python compress script:
`tools\unishox\compress-html-uncompressed.py`
is provided which compresses ALL the files found in the uncompressed folder, and writes the compressed version to a separate folder.  Both uncompressed and compressed versions are committed to the repository so that compression does NOT have to be done every build.

This PR does not change functionality or memory load of Tasmota - it only makes the maintenance of the html/JS/Style stuff easier, and will make the application of unishox to other strings easier in the future where relevant.

**Specifically, it does NOT change any of the html/style/js.**

The code has been built and tested on ESP32 with and without USE_UNISHOX_COMPRESSION.  
The code has been built for esp8266, but I don't have a board to hand.

Next steps would be to apply this to the other places where unishox compressed html/js is directly included within a module, 
e.g. 
```
tasmota\xdrv_09_timers.ino
tasmota\xdrv_21_wemo_multi.ino
tasmota\xdrv_21_wemo.ino
tasmota\xdrv_43_mlx90640.ino
```

## Checklist:


btsimonh:
p.s. if anyone develops tasmota on linux/mac, suitable testing of the python script and modifications to make it cross-platform very welcome.   Specifically about path separators?

arendst:
Nice solution. Thx.

s-hadinger:
That's a nice solution indeed. I have some additional plans.

Unishox works well when there are some repeating patterns, and is often not very interesting with short strings. I started to introduce grouped strings so they can be compressed in groups rather than one by one.

It is already present in `xdrv_23_zigbee_A_impl.ino`. Look for `ZB_WEB_U`.

The idea is to have a list of strings separated by `'\0`` are point to them via offsets (uint16_t). Compression and calculation of offsets is done online via `https://tasmota.hadinger.fr/util` by it should be scripted in the build process.

To indicate the offset names, I use comments in the form of `//=OFFSET_NAME`.

Example:
```
const char ZB_WEB_U[] PROGMEM =
    //=ZB_WEB_CSS

    "<style>"
   [...]
    "</style>"
    "\0"

    //=ZB_WEB_PERMITJOIN_ACTIVE
    "<p><b>[ <span style='color:#080;'>%s</span> ]</b></p>"

    "\0"
    ; // end of list
```

Then the compressed form has:
```
enum {
  ZB_WEB_CSS=0,
  ZB_WEB_PERMITJOIN_ACTIVE=507,
...
}

// Compressed from 1625 to 1111, -31.6%
const char ZB_WEB[] PROGMEM = "\x00\x66\x3D\x0E\xCA\xB1\xC1\x33\xF0\xF6\xD1\xEE\x3D\x3D\x46\x41\x33\xF0\xE8\x6D"
[...]
```

To use it, instanciate: (it can take either the compressed or not compressed version
```
UnishoxStrings msg(ZB_WEB);
```
While it is in scope, it decompresses only once the strings.

The use with:
```
WSContentSend_P(msg[ZB_WEB_CSS]);
```

`msg[offset]` returns a pointer to a string of type `const char *` and null terminated. It may be in PROGMEM if the template was not compressed, or in RAM if it was unishox compressed.

I hope it's clear...

s-hadinger:
Here if the python3 code to compress the strings and used in tasmota.hadinger.fr:

```

import unishox
import json
import ast
import re
from parse import parse

UNISHOX = unishox.Unishox()

# https://www.geeksforgeeks.org/break-list-chunks-size-n-python/
def chunked(my_list, n):
    return [my_list[i * n:(i + 1) * n] for i in range((len(my_list) + n - 1) // n )]


def unishox_strings(input):
    
    (raw, labels_out) = parse(input)
    
    len16 = (len(raw) + 15) // 16   # rounded up to 16 bytes multiple
    if len16 > 255:     return ("buffer too big", "", "")
    
    in_bytes = bytearray(raw, 'utf-8')
    in_len = len(in_bytes)
    out_bytes = bytearray(in_len * 2)
    
    out_len = UNISHOX.compress(in_bytes, len(in_bytes), out_bytes, len(out_bytes))
    ratio = "// Compressed from {i} to {o}, -{p:.1f}%".format(i=in_len, o=out_len, p=(100-out_len/in_len*100))
    out_bytes = out_bytes[:out_len]     # truncate to right size
    out_bytes = b'\x00' + bytes([len16]) + out_bytes
    
    # split in chunks of 20 characters
    chunks = chunked(out_bytes, 20)
    
    lines_raw = [ "\"\\x" + "\\x".join( [ '{:02X}'.format(b) for b in chunk ] ) + "\"" for chunk in chunks ]
    
    line_complete = "const char <var>[] PROGMEM = " + ("\n" + " "*29).join(lines_raw) + ";"
    
    return ("", labels_out + ratio + "\n" + line_complete, raw)
    
```

parse.py:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
"""

"""

mapping = {
	't': '\t',
	'n': '\n',
	'r': '\r',
	'f': '\f',
	'b': '\b',
	'0': '\x00',
}

def parse(input):
    labels = {}
    in_literal = False
    in_comment = False
    in_label   = False
    cur_label  = ""
    out = ""
    i = 0
    while i < len(input) - 1:
    	c = input[i]				# current char
    	cc = c + input[i+1]			# current char + next char
    
    	if (in_literal):			# inside a string
    		if c == '"':
    			in_literal = False
    		elif c == '\\':
    			out += mapping.get(cc[1], cc[1])
    			i += 1
    		else:
    			out += c
    	else:		# not in literal
    		if in_comment:
    			if in_label:
    				if c.isalnum() or c == '_':
    					cur_label += c
    				else:
    					# save label
    					labels[cur_label] = len(out)
    					in_label = False
    					cur_label = ""
    			if (c == '\n') or (c == '\r'):	in_comment = False
    		elif cc == '//':
    			in_comment = True
    			if (i < len(input) - 2) and (input[i+2] == '='):		# catch "//=LABEL"
    				in_label = True
    				i += 1
    			i += 1
    		elif c == '"':
    			in_literal = True
    
    	i += 1
    
    labels_out = "enum {\n" + "".join(["  {}={},\n".format(key, labels[key]) for key in labels.keys()]) + "};\n\n"
    return (out, labels_out)
```

Staars:
Cool, I thought about something similar (mapping in a combined F string) too.

s-hadinger:
@Staars Help yourself. Of course it's better to group strings that we be used in the same portion of code.

btsimonh:
@s-hadinger - I noticed your suggestions whilst searching issue & prs about unishox - a nice idea.
If we are heading for combined strings, then mods to the other modules should wait until after....

PR #10311, opened 2020-12-29
Commands `CTRange` and `VirtualCT`

s-hadinger:
### Virtual CT Phase 2.

This new version gives fine grain control to drive Colw/Warm white, for RGB, RGBW and RGBCW.

**New command**: `CTRange`
`CTRange <ct_min>,<ct_max>`

Speficies the CT range of the bulb. Default is `CTRange 153,500`. When you change this settings, it will only indicate the capabilities of the bulb. The slider will still allow to set CT from 153 to 500, but the rendering will be done within this new range.

Note: the settings is not persisted in Flash. You need to re-do the command at each restart. For example:

```
Rule1 on Power1#Boot do ctrange 200,350 endon
Rule1 1
```

**New command**: `VirtualCT`

This command allows to precisely specify the rendering of the bulb for Color Temperature.

Syntax: `VirtualCT {"<minct>":"<color1>","midct":"<color2>","maxct":"<color3"}`

The first and last ct values indicate the min and max CT and are equivalent to `CTRange`

Example: `VirtualCT {"200":"FFFFFF0000","400":"000000FF00"}`

Note: the settings is not persisted in Flash. You need to re-do the command at each restart. For example:

```
Rule1 on Power1#Boot do VirtualCT {"200":"FFFFFF0000","300":"000000FF00","400":"FF0000FF00"} endon
Rule1 1
```

`VirtualCT` needs `SetOption106 1` and works for 3,4,5 channels bulbs. It is not enables when `SetOption68 1` or when `SetOption37 128`. `VirtualCT` is compatible with white-blend mode and uses the current CT value.

**How to configure:**
1. Disable VirtualCT: `SetOption106 0`
2. Set your bulb to the coolest possible white with `Color xxxxxxxxxx` command. `xxxxxxxxxx` will be the value to be used for each point in VirtualCT
3. Do the same for each CT value you want
4. Enable `SetOption106 1` and after the reboot, use the values above for `VirtualCT <json>`

VirtualCT is included in Tasmota standard but is disabled in tasmota-lite. This feature depends on `#define USE_LIGHT_VIRTUAL_CT`.

VirtualCT supports up to 3 color points by default. This number can be increasd by changing `#define LIGHT_VIRTUAL_CT_POINTS 3`

**Important**: Use either the simplified `CTRange` or the more sophisticated `VirtualCT`. Both commands will override each others.

Adds 1.1kb to Tasmota standard.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 ## [9.2.0.2]
 ### Added
 - Basic support for ESP32 Odroid Go 16MB binary tasmota32-odroidgo.bin (#8630)
+- Command ``CTRange`` to specify the visible CT range the bulb is capable of
+- Command ``VirtualCT`` to simulate or fine tune CT bulbs with 3,4,5 channels
 
 ### Breaking Changed
 - Replaced MFRC522 13.56MHz rfid card reader GPIO selection from ``GPIO_SPI_CS`` by ``GPIO_RC522_CS``
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -425,6 +425,8 @@
 #define D_CMND_DIMMER_RANGE "DimmerRange"
 #define D_CMND_DIMMER_STEP "DimmerStep"
 #define D_CMND_HSBCOLOR "HSBColor"
+#define D_CMND_VIRTUALCT "VirtualCT"
+#define D_CMND_CTRANGE "CTRange"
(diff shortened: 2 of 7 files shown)


psisquared2:
Dear @s-hadinger, is there a chance you're interested in making VirtualCT work with `SetOption92` (#15337)? If not, I'll add a PR updating the documentation. I might try fixing this myself in a couple months, right now I don't have the time and hardware to do so (but I could of course test if a solution works).

PR #10318, opened 2020-12-30
Modify web refresh scripts (Main and Console)

btsimonh:
Allow up to 20s for an XHR request to succeed - only cancel & retry after that.  
Move refresh timer inside successful response.

Applied to Main menu and Console menu.
Fixes times when commands are issued from Web Console, but get the XHR gets cancelled after the default 2.345s webrefresh time.

## Description:
On the main menu and on the console menu, if an XHR request does not complete before the webRefresh time, it is currently cancelled.
This PR sets a timeout of 20s for completion of the XHR, and upon completion of the XHR, sets the normal refresh timer of 'webRefresh'.  If the timeout is reached, the XHR request is cancelled, and the next request made.

For MOST people, the XHR requests will complete in ~80-200ms, and so the PR will have no direct consequences.
For people in bad network conditions, the PR avoids commands issued from the console from being cancelled before they have been sent.
Not that 'bad network conditions' includes turning on ESP32 internal Bluetooth, where responses can take anything between 80ms and 15s.

With compression, we use no extra bytes in the main menu, but ~+30 bytes in the console script.
Very impressed with unishox - I tried to optimise the JS size myself, and found that unishox did a better job that I could.

## Checklist:


s-hadinger:
Nice job!  Indeed unishox gives us more latitude more js code with limited impact on flash size.

barbudor:
@arendst This PR seems to creates interop issues with Safari on iPad as reported on Discord by user Andy7
Apparently related to https://stackoverflow.com/questions/49247362/safari-bug-with-block-scoped-variables

Possibly offending code such as `let ft` in  https://github.com/arendst/Tasmota/blob/9c2b8040f185164d9e2e1cad320fa01275caa996/tasmota/html_uncompressed/HTTP_SCRIPT_ROOT_NO_WEB_DISPLAY.h#L2

btsimonh:
@barbudor - good spot.  There is no need for scoping the variables, just seemed safer with scripts being combined.
Risk will be easier to understand when we have multi-string compression, as all the code for a page is likely to be in one uncompressed file....

arendst:
Thx.

I just reverted this commit. It may also be the cause of this #10329.

Let them try if this solve their issue.

barbudor:
Adding @assachs (Andy7 on Discord)

btsimonh:
@arendst - issue could be that they are running safari?

assachs:
The newest build is working on ipad.

btsimonh:
theo, should I re-submit with this fixed?

arendst:
Yes pls and prefferably if tested on Apple stuff.

assachs:
I can test on apple if the new version is here: http://ota.tasmota.com/tasmota/. Or if someone could provide me the build

btsimonh:
new PR issued.
https://github.com/arendst/Tasmota/pull/10332
tested on OSX, needs a spin on iOS safari.....

arendst:
Just merged.

Jason2866:
@assachs  development builds are here https://github.com/arendst/Tasmota/tree/firmware right after a CI tests are successfully.
There is just no OTA possible since it is a https source...

assachs:
The new javascript code is working on iPad. (testcases: status on main screen and console)

btsimonh:
:)

PR #10334, opened 2020-12-31
universal file system

gemu2015:
## Description:

initial commit

this driver adds universal file system support for
ESP8266 (sd card or littlfs on  > 1 M devices with special linker file e.g. eagle.flash.4m2m.ld)
(makes no sense on 1M devices without sd card)
and
ESP32 (sd card or fatfile system)
the sd card chip select is the standard SPI_CS or when not found #define SDCARD_CS_PIN
initializes the FS System Pointer ufsp which can be used by all standard file system calls
the only specific call is ufs_fsinfo() which gets the total size (0) and free size (1)
a button is created in the setup section to show up the file directory to download and upload files
subdirectories are supported

console calls :

ufs       fs info
ufstype   get filesytem type 0=none 1=SD  2=Flashfile
ufssize   total size in kB
ufsfree   free size in kB

driver enabled by

#define USE_UFILESYS

this code was part of scripter and if merged will remove that part from scripter.
there is no dependency left to scripter code

## Checklist:


arendst:
Thx.

Two things:
- I'm in the process of enforcing fixed SPI_CS for differen devices. This will allow multiple SPI devices on the same bus. Currently rewriting SPI drivers.
- I also want to have a generic filesystem driver. The basics have been released as ``support_filesystem.ino`` mainly for ESP32. My goals was to use SPIFFS on standard tasmota flash images (needed to save zigbee data). On larger images it should be FFAT. Still need to make this transparant based on used partition table. In addition I see a Card filesystem next to the internal 12M FFAT filesystem. Open and close files would be different in that case.

So for now this is on hold until I have a working environment for all. If I release this I see future incompatibilites.

gemu2015:
there are NO api differences on sd card and fat or littlefs. i use these all for a couple of month.
on both esp8266 and esp32 systems. they only differ on initialization and storage size calls.

gemu2015:
i would suggest to remove spiffs completely.  i always use the 983ffat linker file on 4m esp32 devices

arendst:
But the smallest ffat sector size is 4k (as far as I know) so a 64K ffat could only hold 16 files...

gemu2015:
yes that would not make sense FFAT and 64k . Thats why i use the 983 linker file so far.
AND littlefs is already merged into esp32 and the API is completely compatible.
so we could switch to littflefs easily

Jason2866:
littlefs is still not merged in ArduinoESP32 (IDF3.3) which is the base for the released ESP32 Arduino core (s). It is merged in ArduinoESP32 (IDF 4.2) branch which is not compatible
without (many?) changes in Tasmota. Anyway doing a littlefs release for ArduinoESP32 (IDF3.3)
shouldnt be complicated. The upcoming ArduinoESP32 release 1.0.5. will probably not include it.

My2cents using littefs would be a good choice. It is better than SPIFFS. Fully supported from ESP8266 and the port for the ESP32 is already done. It just needs to be implemented in the ESP32 core. Doing this we have ONE file system that can handle all needs.

gemu2015:
ok little file system works with esp32 64k spiffs linker file!

gemu2015:
changed all linker files to spiffs which is used by little fs.
now esp32 and esp8266 both use little fs as default

arendst:
@gemu2015 just merged. Before you want to make changes let me do some changes to integrate in Tasmota.

I'll let you know when I'm finished and you can do your (optional) changes.

gemu2015:
ok will wait for your changes

gemu2015:
when you are implementing SD card we need also a global sync of Free RTOS RTC system
because sd card time stamps come from there.

something like this when time is set from NTP

void Sync_RTOS_TIME(void) {
  struct timeval tv;
  tv.tv_sec = Rtc.local_time;
  tv.tv_usec = 0;
  settimeofday(&tv, NULL);
}

arendst:
Done.

Now needs PCI ``GPIO_SDCARD_CS`` or use ``#define SDCARD_CS_PIN``

arendst:
Missed the RTOS_TIME thing.

Will need to chk as I removed the sync on pupose to get NTP functional again. Oh sh*t.

arendst:
Ah I see. We simply sync RTOS time with Tasmota time. No problem.

I'll see if I can integrate it with the current RTC code...

gemu2015:
next step would be to replace all esp32 linker file references with the spiffs variant. (which means little FS)
we are still compatible with FFAT partitions because the driver falls back when no spiffs is found.

arendst:
That takes some time as I just started to use SPIFFS for storing zigbee data using my ``support_filesystem.ino`` which would become obsolete.

Give me a day.

gemu2015:
NO all linker files need to be spiffs because little fs assumes spiffs partitions. 
FFAT assumes FFAT partitions and fails on SPIFFS. 
 so "old" partitions fall back to FFAT and continue to work while new linker files use little FS

gemu2015:
for info:

the rtos RTC sync fails, since at time of calling Rtc.local_time is not yet valid

PR #10352, opened 2021-01-01
RuleTimer0 applies to all RuleTimers

barbudor:
## Description:

Following request https://github.com/arendst/Tasmota/discussions/10338
This implements `RuleTimer0` which applies to all rules timer simultaneously.
The major interest is `RuleTimer0 0` to clear all pending rule timers.

Independently of this change (already present in older version of Tasmota), I noticed during the test that if no rule is enabled and a RuleTimer is started, RuleTimers do not "stop" at 0 because `RulesEverySecond()` do not process rules nor rules timers. This means if a rule is enabled later on, the RuleTimer may trigger unexpectedly. (testing `if (Settings.rule_enabled && !Rules.busy)` 
This doesn't seems to be a major issue. However a small rewrite, moving the test inside the Timer loop would do the trick.

Happy to add it to this PR if you want.

## Checklist:

### DIFF
--- a/tasmota/xdrv_10_rules.ino
+++ b/tasmota/xdrv_10_rules.ino
@@ -2123,21 +2123,29 @@ void CmndRule(void)
 
 void CmndRuleTimer(void)
 {
-  if ((XdrvMailbox.index > 0) && (XdrvMailbox.index <= MAX_RULE_TIMERS)) {
-    if (XdrvMailbox.data_len > 0) {
-#ifdef USE_EXPRESSION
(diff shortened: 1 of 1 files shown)


arendst:
> Happy to add it to this PR if you want.

Yes pls do.

barbudor:
Hi, Theo. Done. 
Now ruletimers started while no rule is enabled will properly stops at 0.
I also checked with a few test that it doesn't break `on time#minute`

PR #10364, opened 2021-01-02
Ble esp32

btsimonh:
## Description:
Introducing xdrv_48_BLE_ESP32.
This PR is Ready.

### The issue:

### xdrv_48 provides:
2/ generic BLE 'operations'.

### Example modified drivers:

These have been adjusted to instead of using the NimBLE stack directly, to subscribe to services provided by xdrv_48_BLE_ESP32.

### use:
xdrv_48 is enabled via 
USE_BLE_ESP32 and ESP32
BLE may be enabled through a new Configuration button on the web UI, or 'setoption115 1'.
To enable xsns_52_ibeacon_BLE_ESP32 as well, enable USE_IBEACON_ESP32
To enable xsns_62_MI_ESP32_BLE_ESP32 as well, enable USE_MI_ESP32

### documentation
A pr is in the wings for updaiing the documentation on the docs site.

### Personal motivation:

### Requests:
I have prepared the xdrv_48 driver based on requirements gathered from xsns62, xsns52, and from my own EQ3 BLE requirements.  

## Checklist:

### DIFF
--- a/lib/libesp32/NimBLE-Arduino/src/NimBLEClient.cpp
+++ b/lib/libesp32/NimBLE-Arduino/src/NimBLEClient.cpp
@@ -201,6 +201,7 @@ bool NimBLEClient::connect(const NimBLEAddress &address, bool deleteAttibutes) {
         }
     }while(rc == BLE_HS_EBUSY);
 
+    m_result = rc;
     if (rc != 0 && rc != BLE_HS_EDONE) {
         NIMBLE_LOGE(LOG_TAG, "Error: Failed to connect to device; "
                     "addr=%s, rc=%d; %s",
@@ -230,6 +231,9 @@ bool NimBLEClient::connect(const NimBLEAddress &address, bool deleteAttibutes) {
     return true;
 } // connect
 
+int NimBLEClient::getResult(){
+  return m_result;
+}
 
 /**
  * @brief Initiate a secure connection (pair/bond) with the server.\n
--- a/lib/libesp32/NimBLE-Arduino/src/NimBLEClient.h
+++ b/lib/libesp32/NimBLE-Arduino/src/NimBLEClient.h
@@ -59,6 +59,7 @@ public:
     void                                        setClientCallbacks(NimBLEClientCallbacks *pClientCallbacks,
                                                                    bool deleteCallbacks = true);
     std::string                                 toString();
+    int                                         getResult();
     uint16_t                                    getConnId();
     uint16_t                                    getMTU();
     bool                                        secureConnection();
@@ -92,6 +93,7 @@ private:
     int32_t                 m_connectTimeout;
     NimBLEClientCallbacks*  m_pClientCallbacks;
     ble_task_data_t         *m_pTaskData;
+    int                     m_result;
 
     std::vector<NimBLERemoteService*> m_servicesVector;
 
--- a/platformio_tasmota32.ini
+++ b/platformio_tasmota32.ini
@@ -85,6 +85,7 @@ build_flags                 = ${esp_defaults.build_flags}
                               -Dsint16_t=int16_t
                               -Dmemcpy_P=memcpy
                               -Dmemcmp_P=memcmp
+                              -Wl,-Map,firmware.map
 
 
 [core32]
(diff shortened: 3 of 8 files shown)


btsimonh:
@Jason2866 - I note your change to Arduino core.  It helps with some BLE/Wifi interaction, but this set works with the release Arduino core too.
As soon as this goes in, I have an EQ3 driver which depends upon it ready as a PR & example of BLE control.

fightforlife:
I would love to see this! I  also have these EQ3 valves and want to conrol them via MQTT. Currently I am using my own simple ESP32 program. But a tasmota integration would be way better!

gemu2015:
nice project!
Although i am not a fan of BLE in home automation (except ibeacon presence monitoring)
i think this is a good approach.

since i also use ESP32 Tasks i want to draw your attention to the fact that the task stack of ESP32 BLE drivers in Tasmota and probably in your project is much too low for stable operation.

in original RTOS stack size is in words, so 2048 would be 8192 and far enough, but on ESP32 RTOS stack size is given in bytes (see ESPRESSIF docs). And 2048 is definitely too low for stable operation.

btsimonh:
@fightforlife 
> I would love to see this! I also have these EQ3 valves and want to conrol them via MQTT. Currently I am using my own simple ESP32 program. But a tasmota integration would be way better!
You can try it here:
https://github.com/btsimonh/Tasmota/tree/EQ3_BLE_ESP32
We will propose a PR for EQ3 once the BLE PR is accepted.
p.s. expect this PR to close, and a new one to open int he next few days....

btsimonh:
@gemu2015 - noted on stack size.  We're already above 2048, should really measure the max used - i understand there are APIs to do this (@Jason2866 - we could do something general here on ESP32 and tasks?).

Jason2866:
@gemu2015 @btsimonh  if the change is possible via sdkconfig for the idf 3.3 build we can do a customized build variant (same process as for the ESP32solo1 build).

gemu2015:
you may determine the needed stack size  by calling uxTaskGetStackHighWaterMark(NULL) in the task itself. this call returns the max needed stack size. you then can set the stack size to a slightly higher value.

i found that at least 4k stack was needed, depending on what is done in the task. when doing HTTPS i did even need about 8k stack.

if stack is too low you may get weird results like random crashes.

Jason2866:
@gemu2015  Thx for the info!

btsimonh:
I believe you can also get a list of tasks, get the name, current stack size, and high water for each - it's probably a core thing - do we have a 'support esp32' type 'driver' to which we could add a command (maybe it's already there somewhere)?

gemu2015:
Tasmota (Arduino) runs with a stack size:
#define CONFIG_ARDUINO_LOOP_STACK_SIZE 8192

this is on the save side. i would not choose a much lower size until you are sure it is enough.

if have tasks that worked with 4096, but i would not go lower.

i only wanted to warn that the existing Tasmota BLE drivers have much too low stack size (2048) to run stable

this documentation has some apis for listing tasks but these are claimed to be only for debug purposes because they burden the cpu and use memory. 

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/freertos.html

gemu2015:
i will try to implement a task list cmd. think would be useful.

gemu2015:
@Jason2866 
you may probably help here. i implemented the list tasks command but can't get it linked in.

vTaskList needs 2 special defines, but these don't help, still unresolved references.

i went through all net references but didn't find a solution yet.

build_flags                 = ${core32.build_flags} -DCONFIG_FREERTOS_USE_TRACE_FACILITY -DCONFIG_FREERTOS_USE_STATS_FORMATTING_FUNCTIONS

[support_command.ino.zip](https://github.com/arendst/Tasmota/files/5808096/support_command.ino.zip)

Jason2866:
It is not in the Arduino ESP32 build. https://github.com/espressif/esp32-arduino-lib-builder/blob/master/sdkconfig#L622

gemu2015:
yes i also saw this, so for the time being start with fairly hight stack size and then determine the stack watermark in the task and adjust the stack size thereafter.

btsimonh:
see revised PR https://github.com/arendst/Tasmota/pull/10600

PR #10328, opened 2020-12-30
Feature: Adafruit SSD1331 Display support

jeroenvermeulen:
## Description:
Support for the [SSD1331](https://www.aliexpress.com/item/32790785282.html) SPI 94\*64 Full Color OLED Display.
I have used **DisplayModel 14**.
Based on [release 1.2.0 of the Adafruit library](https://github.com/adafruit/Adafruit-SSD1331-OLED-Driver-Library-for-Arduino/releases/tag/1.2.0).
Inspired by @gemu2015's xdsp_09_SSD1351 driver.
Both software and hardware SPI work. For hardware SPI the OLED Reset is optional.

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.9
  - [x] The code change is tested and works on Tasmota core ESP32 V.1.0.5-rc4 => works but was only able to use SSPI on ESP32.
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #10376, opened 2021-01-02
fixes compile error of SSD1331 with tasmota-display

jeroenvermeulen:
## Description:
fixes compile error of SSD1331 with tasmota-display + improves layout of DisplayMode 1.

**Related issue (if applicable):** fixes [compile error](https://github.com/arendst/Tasmota/pull/10328#issuecomment-753327374)

## Checklist:

### DIFF
--- a/lib/lib_display/Adafruit_SSD1331-1.2.0/Adafruit_SPITFT_Renderer.cpp
+++ b/lib/lib_display/Adafruit_SSD1331-1.2.0/Adafruit_SPITFT_Renderer.cpp
@@ -1,5 +1,10 @@
+/*
+ * This class is basically the same as Adafruit_SPITFT.
+ * The only difference is: it extends Renderer which extends Adafruit_GFX.
+ * The original Adafruit_SPITFT class directly extends Adafruit_GFX.
+ */
 /*!
- * @file Adafruit_SPITFT.cpp
+ * @file Adafruit_SPITFT_Renderer.cpp
  *
  * @mainpage Adafruit SPI TFT Displays (and some others)
  *
@@ -92,7 +97,7 @@
 
 #endif // end USE_SPI_DMA
 
-// Possible values for Adafruit_SPITFT.connection:
+// Possible values for Adafruit_SPITFT_Renderer.connection:
 #define TFT_HARD_SPI 0  ///< Display interface = hardware SPI
 #define TFT_SOFT_SPI 1  ///< Display interface = software SPI
 #define TFT_PARALLEL 2  ///< Display interface = 8- or 16-bit parallel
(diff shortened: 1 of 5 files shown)


jeroenvermeulen:
I will explain why I have created the class `Adafruit_SPITFT_Renderer`, maybe there is a better way, I am not a C++ expert.

- Normally the Adafruit_SSD1331 class [extends Adafruit_SPITFT](https://github.com/adafruit/Adafruit-SSD1331-OLED-Driver-Library-for-Arduino/blob/master/Adafruit_SSD1331.h#L57)
- Adafruit_SPITFT [extends Adafruit_GFX](https://github.com/adafruit/Adafruit-GFX-Library/blob/master/Adafruit_SPITFT.h#L119)
- I want Adafruit_SSD1331 to extend Renderer, which in turn [extends Adafruit_GFX](https://github.com/arendst/Tasmota/blob/development/lib/lib_display/esp-epaper-29-ws-20171230-gemu-1.1/src/renderer.h#L18), the Adafruit_SSD1306 in Tasmota [does this also](https://github.com/arendst/Tasmota/blob/development/lib/lib_display/Adafruit_SSD1306-1.3.0-gemu-1.1/Adafruit_SSD1306.h#L119)
- However in Adafruit_SSD1331 I can't do it directly like the SSD1306 driver does because SSD1331 is an SDI display and it needs Adafruit_SPITFT in between instead of directly extending Adafruit_GFX
- That's why I created a custom version of Adafruit_SPITFT called Adafruit_SPITFT_Renderer which [extends Renderer](https://github.com/jeroenvermeulen/Tasmota/blob/fix_ssd1331_with_tasmota-display/lib/lib_display/Adafruit_SSD1331-1.2.0/Adafruit_SPITFT_Renderer.h#L125)
- The [original version of Adafruit_SPITFT](https://github.com/jeroenvermeulen/Tasmota/blob/fix_ssd1331_with_tasmota-display/lib/lib_display/Adafruit-GFX-Library-1.5.6-gemu-1.0/Adafruit_SPITFT.h#L119) is also included in Tasmota, it is used by [ILI9341](https://github.com/arendst/Tasmota/blob/development/lib/lib_display/Adafruit_ILI9341-1.2.0-Tasmota-1.0/Adafruit_ILI9341.h#L126) which does not use Renderer.h

I somebody knows how I can make Adafruit_SSD1331 extend both Adafruit_SPITFT and Renderer (multiple inheritance?) I would like to learn how.

arendst:
Thx.

As the current ILI9341 driver fails to run with your tweaked SSD1331 library I just removed it.

I suggest you revisit the whole SSD1331 implementation and test it with at least the original ILI9341 driver.

jeroenvermeulen:
What exactly goes wrong?
Did you have the problem with the version from this PR? Because my whole purpose was to fix the conflict with other display drivers.
With my latest version I was able to build and use ‘tasmota-display’

arendst:
OK. In that case I'll try again tomorrow (it's almost 24:00 here) as I'm glad I managed to get it going again.

jeroenvermeulen:
Here too, I am Dutch and live in Belgium just across the border :-)

arendst:
Thx. Tested and works.

PR #10422, opened 2021-01-06
Ufsys

gemu2015:
## Description:

language files change sd card to UFS
directory listig -> scroll view

## Checklist:

### DIFF
--- a/tasmota/language/af_AF.h
+++ b/tasmota/language/af_AF.h
@@ -864,7 +864,7 @@
 #define D_CONFIGURE_SCRIPT     "Wysig skrif"
 #define D_SCRIPT               "wysig skrif"
 #define D_SDCARD_UPLOAD        "lêer oplaai"
-#define D_SDCARD_DIR           "sd card directory"
+#define D_UFSDIR               "ufs directory"
 #define D_UPL_DONE             "Klaar"
 #define D_SCRIPT_CHARS_LEFT    "karakters oor"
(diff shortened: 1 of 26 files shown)


arendst:
I think we have to rethink this ufs stuff Unless it already works this way).

What I want is this on ESP32:
- As every ESP32 has a filesystem partition, either LittleFS OR FFat it needs to be the primary filesystem to store future settings.
- Any connected SD card must be an ADDITIONAL filesystem and should not take over the embedded partiton filesystem. If it does take over it ruins any settings stored on the embedded partition.

On ESP8266 I can live with your current SD-Card usage.

Can you verify this is the case already and if not we have to make it this way.

gemu2015:
since only very few people will use sd cards i see no conflict for the time being.  simply do not define USE_SDCARD
i can think about how to have an extra filesystem in this case for the future.

gemu2015:
ok checked it.  its easy to accomplish.  i need a day or 2 to implement this.

gemu2015:
ok is already implemented in commit.  
a new FS pointer *ffsp holds a pointer to Flashfile in any case. (nullptr if not found) even when sd card found
ufsp points to the global file system including sd.
the only thing is that currently ufsp points to global directory. if an SD card found the directory shows the SD card not the flash. the flash is hidden in this case.
i may easily switch between the 2 if wanted.

arendst:
Just keep it this way.

I will move the zigbee settings to the new ffsp pointer.

Thx.

PR #10518, opened 2021-01-11
Pid branch revived

marcvs:
## Description:

The additional commmits fix the following:
- Code now compiles with `tasmota-minimal` because I've surrounded it by `#ifndef FIRMWARE_MINIMAL`
- Now including publishing the `%topic%/PID {"power": <PidPower>}` by default (can be turned off), since it was requested by @colinl  . My issue with this was that it does access `TasmotaGlobal.mqtt_data`, which was discouraged earlier. I tried to find alternatives for publishing, but failed. Pointers be helpful.
- Re-inserted the code in `xdrv_48_timeprop.ino` for advanced handling of incoming commands (in bool Timeprop_Command), so that the `timeprop_setvalue_x` can be received to set the timeprop value for relay `x`. Also it is being used from `xdrv_49_pid.ino`.
I failed moving this to DecodeCommand, because of the `_x` feature. 
Is it really necessary to move to DecodeCommand? If so: How?

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -811,10 +811,10 @@
 // -- Prometheus exporter ---------------------------
 //#define USE_PROMETHEUS                           // Add support for https://prometheus.io/ metrics exporting over HTTP /metrics endpoint
 
-// -- PID and Timeprop ------------------------------
-// #define use TIMEPROP                            // Add support for the timeprop feature (+0k8 code)
+// -- PID and Timeprop ------------------------------ // Both together will add +12k1 code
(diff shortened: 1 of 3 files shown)


ascillato:
Hi,

The message from github that is above, is telling you that this PR can not be merged because has conflicting files.

image

Please, update your branch before applying your changes.

marcvs:
Yep; I see this, yet I feel stupid in trying to solve it.  At the moment I can't even reproduce the conflicts when merging this branch into my local (updated) development branch.

PR #10666, opened 2021-01-22
Zigbee don't do auto-bind if device is already known

s-hadinger:
## Description:

Some devices (ex: IKEA PIR) tend to re-join and send `DeviceAnnounce` although they were already paired, which triggers a new auto-binding.

Now auto-binding happens only if PermitJoin is active, or if the IEEEAddress was not already known. Simple rejoins will not trigger auto-binding anymore.

Additional changes:
- fixed PermitJoin indicator on WebUI for CC2530
- changed BatteryVoltage reporting threshold to 5% instead of 1%, and 0.2V instead of 0.1V
- Fixed devices that were mistakenly shown as router in Zigbee map

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -760,8 +760,8 @@
 
   // Auto-binding constants, see `Z_autoAttributeReporting`
   // Below are the threshold for attribute reporting
-  #define USE_ZIGBEE_AUTOBIND_BATTVOLTAGE   0.1     // V
-  #define USE_ZIGBEE_AUTOBIND_BATTPERCENT   1       // %
+  #define USE_ZIGBEE_AUTOBIND_BATTVOLTAGE   0.2     // V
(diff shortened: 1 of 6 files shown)


MattWestb:
By zigbee 3 standard shall one device that is rejoining the network sending one `DeviceAnnounce` so the trust center can knowing where it is. so 110% OK.

Example is one light is being repowered or have doing on software reboot or one end device have changing its parent.

Zigbee 3 devices is also sending 1 `DeviceAnnounce ` the joining no zigbee 3 network and 3 if its joining on zigbee network (its getting one new TC-link key).

Also one battery saver for end devices that like jumping around and changing parent all the time  like my Philips HUE motion sensor dont need being repaired every time :-)))

s-hadinger:
Thanks for the additional information. Now Tasmota auto binding behaves in a better way

PR #10680, opened 2021-01-24
Added ext_snprintf to support extended types

s-hadinger:
## Description:

This is a proposal for a way to extend `snprintf()` and all related commands `AddLog()` etc. with higher level arguments, simplifying code and avoiding superfluous code:
- support for `float`
- support for `uint64_t`
- support for `IPAddress`

Example: 
```
uint64_t u64 = 0x1122334455667788LL;
ext_snprintf(PSTR("Int64 = 0x%_X"), &u64);

uint32_t ip = 0x10203040;
ext_snprintf(PSTR("IP = 0x%_I"), ip);

float fl = 3.14;
ext_snprintf(PSTR("f1=%*_f f2=%*_f"), 4, &fl, -4, &fl);
// outputs:
// "f1=3.1400 f2=3.14"
```

Full doc is here: https://tasmota.github.io/docs/Sensor-API/#extension-to-vsnprintf

Code size increases by 412 bytes for `tasmota` but tends to be a lower impact thanks to the code saved with the primitives, in `tasmota-ir` and `tasmota-zbbridge`.

Note: the implementation is CPU architecture specific, and works for all Xtensa CPUs. It will require small adaptation for RISC-V or ARM.

## Checklist:

### DIFF
--- a/lib/default/Ext-printf/library.properties
+++ b/lib/default/Ext-printf/library.properties
@@ -1,0 +1,7 @@
+name=Ext-printf
+version=1.0
+author=Stephan Hadinger
+maintainer=Stephan <stephan.hadinger@gmail.com>
+sentence=Extension of snprintf() and vsnprintf()
+paragraph=This library provides extended types support for snprintf (float, uint64_t)
+architectures=esp8266, esp32
--- a/lib/default/Ext-printf/src/ext_printf.cpp
+++ b/lib/default/Ext-printf/src/ext_printf.cpp
@@ -1,0 +1,298 @@
+/*
+  ext_printf.ino - Extended printf for Arduino  objects
+
+  Copyright (C) 2021  Stephan Hadinger
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "ext_printf.h"
+#include <Arduino.h>
+#include <IPAddress.h>
+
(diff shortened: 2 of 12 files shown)


sfromis:
Looks like this change broke payloads published to MQTT. Before and after:
`{"IrReceived":{"Protocol":"RC5","Bits":12,"Data":"0x9F9","DataLSB":"0x909F","Repeat":0}}`
`"Protocol":"RC5","Bits":12,"Data":"0x00000000000001F9","DataLSB":"0x000000000000809F","Repeat":0}}`

The issue of lacking the leading `{"IrReceived":{` part may be due to sendIRJsonState using Response_P instead of ResponseAppend_P when building the rest of the JSON response string.

And all those extra zeroes in Data and DataLSB is quite a breaking change for users of the inputs. Is it intentional to now format the 64-bit value including leading zeroes? Previously the code called Uint64toHex for conversion to string, but now relies on ResponseAppend_P doing sprint-like formatting, with different output.

sfromis:
Ok, that fixed the issue of invalid JSON, now I'm getting
`{"IrReceived":{"Protocol":"RC5","Bits":12,"Data":"0x00000000000001F9","DataLSB":"0x000000000000809F","Repeat":0}}`
instead of the previous format "only" transmitting the actual number of bits instead of always 64 bits, mostly leading zeroes.
`{"IrReceived":{"Protocol":"RC5","Bits":12,"Data":"0x1F9","DataLSB":"0x809F","Repeat":0}}`
As such a breaking change is likely to result in much moaning, I'd suggest fixing.

Also, I've noticed that the spurious messages with "UNKNOWN" protocols now seems to include a format specifier instead of actual hash value. Before and after:
`{"IrReceived":{"Protocol":"UNKNOWN","Bits":52,"Hash":"0x2AFCC184","Repeat":0}}`
`{"IrReceived":{"Protocol":"UNKNOWN","Bits":52,"Hash":"0x08X","Repeat":0}}`
I've not tested this for "actual unknown" protocols where I suppose that the hash could be useful.....

arendst:
Thx.

The ``"Data":"0x00000000000001F9"`` issue is due to the ``ext_snprintf_P`` parsing of ``0x%_X`` which in turn calls ``U64toHex()`` which seems to always print 16 digits.

Let me chk if I can solve this by implemnting some kind of "decimal" value....

arendst:
QandD fix.

sfromis:
Now, my old automations work again :100:

PR #10700, opened 2021-01-26
Update NimBLE to release verison 1.1.0 + Adds MQTT discovery for MI sensors using new BLE

btsimonh:
## Description:

1/ Updates NimBLE to the release version 1.1.0

2/ Adds MQTT discovery for Homeassistant users under 'Setoption19 1'
- yes, I know this is the 'old' way, but the 'new' way is unsuited to tasmota discoverable sensors
- this also ADDS publish of each sensor on a separate topic, not related to the publishing tasmota (only if setoption19).

3/ Removes BLE stats from /sensors mqtt (never should have been there) - publishes instead to /BLE

4/ Sets the default address filter in BLE to 0 - public static addresses only
 - this hides random addresses created by mobiles.  If you want these (e.g. you want to hear covid announcments), do `BLEAddrFilter 1`.  Most users will not want to see them.

5/ add MI32Option5
- if `MI32Option5 1` then only MI devices which have BLEAlias defined for them will be heard by MI.  All devices are removed at the point of setting 1.

## Checklist:

### DIFF
--- a/lib/libesp32/NimBLE-Arduino/CHANGELOG.md
+++ b/lib/libesp32/NimBLE-Arduino/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.  
 
-## [Unreleased]
+## [1.1.0] - 2021-01-20
 
 ### Added
 - `NimBLEDevice::setOwnAddrType` added to enable the use of random and random-resolvable addresses, by asukiaaa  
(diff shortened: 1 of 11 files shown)


btsimonh:
Note for EQ3 self build, branch with the above included is here:
https://github.com/btsimonh/Tasmota/tree/HA_plus_EQ3
PR for EQ3 still delayed pending thoughts on discovery/control from HA.

btsimonh:
@theo -haha - did not think of changing it to a const char rather than a define to solve those warnings!  thankyou.  was scratching my head about how to resolve.

PR #10711, opened 2021-01-27
Rename and make USE_BLE_ESP32 the default for sensors.

btsimonh:
## Description:

title says it all :)
Rename files as per instruction.
Added additional comments hopefully making it obvious when a file will be used.

The two iBeacon drivers could still do with some cleanup.  e.g. the esp32 version has both serial and BLE_ESP32 - really, we should completely split, and allow the HM10 version to be on ESP8266 or ESP32 in ADDITION to the ability to enable the BLE_ESP32 version - but exactly what call there would be for HM10 +  native at the same time, I do not know.  Or even if anyone has ever attached a HM10 to an ESP32.....  so no priority for the moment.

## Checklist:

### DIFF
--- a/tasmota/tasmota_configurations.h
+++ b/tasmota/tasmota_configurations.h
@@ -170,6 +170,8 @@
 #define USE_IBEACON                              // Add support for bluetooth LE passive scan of ibeacon devices (uses HM17 module)
 //#define USE_GPS                                  // Add support for GPS and NTP Server for becoming Stratus 1 Time Source (+ 3.1kb flash, +132 bytes RAM)
 #define USE_HM10                                 // (ESP8266 only) Add support for HM-10 as a BLE-bridge for the LYWSD03 (+5k1 code)
+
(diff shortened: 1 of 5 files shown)


btsimonh:
pls check :)
builds & runs on esp32.
builds for 8266.

btsimonh:
@arendst - Q: the EQ3 driver depends on the BLE_ESP32 driver, therefor must be a higher number?  is this correct?  
Is the new numbering meant to be > 79 for esp32?
(I'm trying to keep the EQ3 'on top' of development, so will rename now.  PR will only come when we've got more experience with actually controlling them, etc.  so I'll just keep rolling the number up if others are added in the meantime....).

That is, unless you Want me to add the PR now, and follow up with further PRs as we modify the way they are discovered/controlled?  There are quite a few people who are using it now already, home-building from my fork....
br,
S

arendst:
> s the new numbering meant to be > 79 for esp32?
Yes.

If the EQ3 has to follow the BLE driver then I propose to rename BLE to 78 and the EQ3 driver to be 79.

PR #10844, opened 2021-02-05
Fix teleinfo standard

Joda89:
## Description:

## Checklist:


barbudor:
Hi @meingraham and team
Indeed this PR (still draft) needs some comments ;)

**Context:**
Teleinfo is the protocol of the French National Grid power meter installed in every french house. The protocol is serial (thus the relation with TasmotaSerial). It exists in 2 flavours depending on your power meter model and configuration:
- Legacy mode (used by old meters) : 1200 baud with data lines up to 64 bytes and no timestamping
- Standard mode (used by all recent meters) : 9600 baud with data lines up to 128 bytes and timestamping on some lines
The protocol is intense : at 9600 baud, the effective data rate is more than 90% of the capacity.

@Joda89 and I tried to make the Teleinfo feature to work on both ESP8266 (myself+Joda) and ESP32 (Joda) on a recent meter because it was not working at all. This plus other build problems related to USE_TELEINFO_STANDARD let us suspect it has only been tested on Tasmota with legacy mode.
The debug lead us to fix issues in the LibTeleinfo library from @hallard and to enhance TasmotaSerial in order to work reliably with the high throughput of the 9600 baud serial.

**Scope of the changes:**
_TasmotaSerial_:
1) Addition of the `TasmotaSerial::read(buffer,size)` method to read multiple bytes from the serial buffer at once.
The method is based based on the underlying hardware serial `::read(buffer,size)` method, and coded manually for software serial.
2) Add support for the `buffer_size` argument on ESP8266/HardwareSerial as it was skipped up to now. We need that to increase the serial rx buffer above the default 256 bytes.

Both changes were required to improve processing of the 9600 baud intense throughput.

_xnrg_15_teleinfo_:
As the processing function of the driver `TInfoEvery250ms()` is called every at 250ms, up to 250 characters could be received between 2 calls. The default size of the serial rx buffer being 256 bytes, this is not enough to cope with Tasmota relaxed real-time as we have witnessed sometime more than 400 ms between 2 calls, especially when playing and refreshing the web GUI.
In order to cope with this, the change in the drivers are:
1) Use the `buffer_size` argument of `TasmotaSerial` to set a 512 bytes rx buffer. This is a safe value to handle variations in the periodicity of `FUNC_EVERY_250_MSECOND`.
2) Use a local buffer of the same size and read at once all the content of the serial buffer, then process by calling LibTeleinfo.
3) The process function has been renamed from `TInfoEvery250ms` to `TInfoProcess`

At some point I considered to move the function from `FUNC_EVERY_250_MSECOND` to `FUNC_LOOP`. However, feedback from Joda was that Tasmota was less reactive. So keeping it at 250 ms needed the above buffer size increase.
Performance measurements show that processing up to 400 bytes don't last more than 15ms. So this low enough for the 250ms call period.

_LibTeleinfo_:
A few problems were present, 2 related to the protocol implementation itself (checksum), and 1 to the Tasmota build environment:
_@hallard you are welcome to comment on the below. Thanks_
1) Removing `USE_TELEINFO_STANDARD`
LibTeleinfo.h had some code dependencies on `USE_TELEINFO_STANDARD`: definition of line buffer (128 vs 64 bytes) and addition of a timestamp field into an internal structure to support the Standard mode timestamp.
Unfortunately, when `USE_TELEINFO_STANDARD` is #defined into `user_config_override.h` (which is the proper Tasmota way), Tasmota code is correctly compiled with it and `LibTeleinfo.h` was going through the #if branches of USE_TELEINFO_STANDARD.
But when `LibTeleinfo.cpp` was compiled on its own when building `lib_div`, the macro is not defined and the library is compiled ignoring the USE_TELEINFO_STANDARD.
This could have been workarounded by adding `-DUSE_TELEINFO_STANDARD` on the build flags of a custom environment but this is not the "Tasmota way".
Considering that the impact of `USE_TELEINFO_STANDARD` is only to increase the line buffer size from 64 to 128 and adding a timestamp field, ignoring the flag and enforcing those in the code is a very little price to pay to simplify builds.

2) The main checksum calculation in `TInfo::calcChecksum` was wrong and was missing 1 separator character. Each and every frame was rejected because of that. I fixed that and added the frame format as a comment to justify the change.

3) The checksum is also calculated again in `TInfo::addValue()` but the way the function was called from `TInfo::checkLine()` didn't included the timestamp (`pts`). This lead to erroneous checksum validation in `TInfo::addValue()` for every frame that are including a timestamp.

I also added quite a few calls to `AddLog()` that I left in the code with comments.
I only lefts 3 calls with loglevel 2 to trigger a log in case of checksum error or line buffer overflow.
Those have proven to be very useful.

**Next steps**
We will appreciate any comments from the team (@arendst , @Jason2866 , ...) before providing a final PR.
If/When the content pass the review, we would squash the commits to make it clean.
I believe as well that we should split the PR into 2 parts:
a) TasmotaSerial enhancements
b) Teleinfo driver and library

Thanks in advance

arendst:
Thx for the elaborate information!

I will investigate the serial part and probably merge this after doing some tests. So you can focus on the teleinfo part.

arendst:
Serial part has been merged incrementing serial version to v3.2.0

meingraham:
@barbudor 

Thanks for the clarification. The use of the term `TeleInfo` is what confused me.

arendst:
I cannot test but it looks fine to me. If a French user can test it and he/she agrees I'll merge.

barbudor:
Thanks @arendst
@Joda89 is not available for now, let's wait for some final tests

@Jason2866 Any comments regarding USE_TELEINFO_STANDARD ?
Should we totally clean any reference to it ?

barbudor:
I just found a additional difference between Legacy mode and Standard mode which explain the checksum problems.
Previous version was (apparently) compatible only with Legacy mode, and those change makes it compatible only with Standard mode.
More to do....

hallard:
Dear guys,
Thanks for investigating on new téléinfo. Correct checksum has changed. I have no new meter to test but if you have any question on existing code do not hesitate to ask.

Jason2866:
Took a look to save this 64 bytes. Didnt find a nice solution (with PlatformIO).
So i agree with you, lets sacrifice the 64 bytes

barbudor:
@hallard Could you please provide a raw capture of Legacy mode for me to complete the testing ?
Thanks

hallard:
@barbudor 

here the full frame, without STX and ETX (not sure it will pass over this preview, also not sure about CR/LF interpreted by git

````
ADCO 030522024356 7
OPTARIF HC.. <
ISOUSC 45 ?
HCHC 084879914 8
HCHP 107060110 #
PTEC HP..  
IINST 007 ^
IMAX 039 K
TENSION 233 8
PAPP 01530 *
HHPHC H 3
MOTDETAT 000000 B
````

barbudor:
Bonsoir Charles @hallard 
I think everything works fine, both in Historique and Standard modes
No need for a define for Standard mode, all is set by SetOption102
However I tested Historique mode only with your above data, so it would be good someone do a more complete test
Do you have some time to test it from Joda's fork ? https://github.com/Joda89/Tasmota/tree/fix_teleinfo_standard

@Jason2866 the only way to save 64 bytes between Standard and Legacy mode would be to change the Teleinfo library to do a malloc() instead of a static allocation. Do you think it's worth ? Happy do to it you think so.
I also enforce the large 512 serial buffer only in Standard mode (9600 baud) and leave default 256 in Legacy mode (1200 baud).


I will also ask on Charles' forum for people to test

Thanks

Jason2866:
@barbudor  I was interested if there is a way doing it via Platformio. For me it is fine as it is.

barbudor:
Hi
The patch has been tested with different types of meters by user on @hallard's [forum](https://community.ch2i.eu/topic/676/tasmota) an it's working well.
So it's a good to go.
Do you want a squash ?

arendst:
Yes pls do

barbudor:
Moved to squashed PR : https://github.com/arendst/Tasmota/pull/10911
this PR can be closed

PR #10920, opened 2021-02-11
Tasmota32 bluetooth variant

Jason2866:
## Description:
since Bluetooth feature is not included in other build variants.

## Checklist:


arendst:
Tricky changes.

The order of entries in ``kModuleNames`` and ``kModules`` needs to be the same as the order of entries in kModuleNiceList. 

As the BLE version doesn't need extra hardware I suggest to put it right after the first entry (WEMOS) so in kModuleNiceList it should read like this:

```
const uint8_t kModuleNiceList[] PROGMEM = {
  WEMOS,
#ifdef USE_BLUETOOTH
  BLUETOOTH,
#endif  // USE_BLUETOOTH
#ifdef USE_WEBCAM
  ESP32_CAM_AITHINKER,
#endif  // USE_WEBCAM
...
```

resulting in 

```
const char kModuleNames[] PROGMEM =
  "ESP32-DevKit|"
#ifdef USE_BLUETOOTH
  "ESP32-BLE|"
#endif  // USE_BLUETOOTH
#ifdef USE_WEBCAM
  "ESP32-Cam|"
#endif  // USE_WEBCAM
...
```

and the location of the AGPIOs also need to be corrected.

PR #10936, opened 2021-02-12
display update

gemu2015:
## Description:

add ili9341-2  CS and DC
make ILI9341-2 compatible with ESP8266
faster ST7789 drawing
support text scaling for all EPD fonts

## Checklist:


arendst:
Would it be possible to replace my "old" ILI9341 driver by yours? That way we keep one driver as now it is not clear when to use which driver.

In that case no template and language changes are needed.

BTW your template changes corrupt current systems as you inserted the GPIO's and Text in the fixed lists. It should have been inserted at the end of the lists.

gemu2015:
ok that is probably the best solution. i will work on it.

gemu2015:
i will close this and make a new pr when ready

PR #10982, opened 2021-02-15
Independent dimmer range

twischer:
## Description:

**Related issue (if applicable):** non

With this PR each PWM dimmer channel can be configured with a
different DimmerRange. Therefore the minimal and maximal brightness of
each channel can be configured independently.
With `dimmerrange ...` still the range for all PWM channels will be set.
But for example with `dimmerrange2 ...` only the range of PWM channel 2 will be modified.

The migration of `Settings` is not yet implemented. First I would like to get a feedback if this PR gets a chance to be merged.

## Checklist:

### DIFF
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -561,16 +561,18 @@ struct {
   uint8_t       ot_boiler_setpoint;        // E8D
   uint8_t       ot_flags;                  // E8E
   uint8_t       ledpwm_mask;               // E8F
-  uint16_t      dimmer_hw_min;             // E90
-  uint16_t      dimmer_hw_max;             // E92
+  uint16_t      ex_dimmer_hw_min;          // E90
+  uint16_t      ex_dimmer_hw_max;          // E92
   uint32_t      deepsleep;                 // E94
   uint16_t      hass_new_discovery;        // E98  ex2_energy_power_delta on 8.4.0.3, replaced on 8.5.0.1
   uint8_t       shutter_motordelay[MAX_SHUTTERS];    // E9A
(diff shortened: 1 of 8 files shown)


twischer:
@s-hadinger 

> It breaks previous settings

When you are intrested in this modification I will extend `SettingsDelta()` in https://github.com/arendst/Tasmota/blob/development/tasmota/settings.ino#L1116 to migrate the settings.
When using `dimmerrange` nothing changes with this PR but `dimmerrange1` to `dimmerrange5` will be supported.

> Why do you need this?

I have different LED types connected to the same ESP8266. One PWM channel is connected to a LED light string with 10 LEDs in series. Therefore in my case I need to generate 32V for the LED light string from the 5V supply connected via USB. I wanted to keep the circuit small and cheap. Therefore I am using the sigma delta hardware of the ESP to generate a 312kHz PWM to drive a transistor part of a step up converter.
The required software modifications for the hardware PWM and an example circuit can be found in the following commit
https://github.com/twischer/Tasmota/commit/0a8735cfeb6f9d60db590956990d1b9ee8813415
I need to limit the maximum duty cycle of this hardware PWM to limit the maximal voltage to 32V to not break the LED light string.

s-hadinger:
Thanks for sharing the details. This sounds like a very specific use case. I'm reluctant to spend previous Flash Settings for a feature that is really specific to your usage.

In the end it's Theo's decision @arendst

barbudor:
As this is pointed to be very specific, why not keeping only the default DimmerRange in the flash settings.
Use-case specific could be setup at start with a system#boot rule or autoexec, thus not using any additional generic flash-settings.

PR #10986, opened 2021-02-16
update sendmail

gemu2015:
## Description:

new email library for esp8266 and esp32
this new enhanced library supports nearly all email providers

however the esp8266 verision needs some more heap.

therefore i kept both versions for esp8266 .

with ESP32 
#define USE_ESP32MAIL 

with ESP8266
#define USE_SENDMAIL
but selecting
#define USE_ESP32MAIL  on esp8266 also works with the benefits of the new lib but more heap needed
(about 14k)

## Checklist:


Jason2866:
Is it possible to remove the SPIFFS part completely?

gemu2015:
yes i can strip that out of the email driver since we do not use it at all.

PR #10889, opened 2021-02-09
Added support for TM1637 Seven-Segment Display

ajithvasudevan:
## Description:

Added support for TM1637 Seven-Segment Display.

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.9
  - [ ] The code change is tested and works on Tasmota core ESP32 V.1.0.5-rc6
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

Issue #3487, opened 2018-08-14
TM1638, all buttons same topic


wunderbaum:
I installed the newest version to try out the tm1638 function. Unfortunately, I always get the same topic on console and mqtt whatever button I press.
I can set different switchmodes to each button, but the topic is always stat/sonoff/POWER
How can I change this? I know, tm1638 support is really new, so I am very curious about it.

**Describe the bug**

_A clear and concise description of what the bug is._

Also, make sure these boxes are checked [x] before submitting your issue - Thank you!

- [ X] Searched the problem in issues and in the wiki
- [X ] Hardware used : Wemos D1 
- [ X] Provide the output of command ``status 0`` :
```
STATUS 0 OUTPUT HERE
20:26:14 MQT: stat/sonoff/STATUS = {"Status":{"Module":18,"FriendlyName":["Sonoff"],"Topic":"sonoff","ButtonTopic":"0","Power":0,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"ButtonRetain":0,"PowerRetain":0}}


**To Reproduce**
Install 6.1.1.6, connect a TM1638-module and see what happens.

**Expected behavior**
Get information which button was pressed.

**(Please, remember to close the issue when the problem has been addressed)**

**Comment by Jason2866** [comment]
See #3214

**Comment by wunderbaum** [comment]
Hi,
what I tried on the console to give each of the 8 buttons of the tm1638 a different topic. What I did:
```
switchtopic1 2 bla
switchtopic2 2 blubb
switchtopic3 2 lala
```
Console responses:
```
18:42:18 MQT: stat/tm1638/RESULT = {"SwitchTopic":"2_bla"}
18:43:31 MQT: stat/tm1638/RESULT = {"SwitchTopic":"2_blubb"}
18:44:11 MQT: stat/tm1638/RESULT = {"SwitchTopic":"2_lala"}
```
and the behaviours of the buttons
```
switchmode1 3
switchmode2 1
switchmode3 4
...
```

So, I expected to get now different topics published. But its always the last given topic for every button.
`18:45:06 MQT: cmnd/2_lala/POWER = TOGGLE`
I think I am doing something completely wrong, what do I need to do to see WHICH button on the tm1638 is pressed?

Jason2866:
Take a look:
`18:51:08 MQT: cmnd/sonoff-1A2AFC/POWER1 = ON
[...]

PR #11031, opened 2021-02-18
Added support for TM1638 Seven-Segment Display

ajithvasudevan:
## Description:
This is an update of the _TM1637_ driver, which adds support for the **TM1638** seven-segment display ~~(+ onboard buttons +LEDs)~~ module.
The update for the TM1638 includes all the "Display" [commands](https://github.com/arendst/Tasmota/pull/10889#issuecomment-779803867) of the _TM1637_~~, and additionally, has the ability to set the onboard LEDs and react to button presses.~~

~~This PR also adds a new `Settings.display_type` parameter, used here to switch between different display variants supported by the same (_TM1637_) driver. It now uses the new `Settings.display_options` to switch between display variants.~~

~~The **DisplayType** command now manipulates the `Settings.display_options.data` value.~~

~~Thus the _TM1637 4-digit_ is **DisplayType 0**, _TM1637 6-digit_ is **DisplayType 1** and _TM1638_ is **DisplayType 2**. A future addition could be _DisplayType 3_ for the _MAX7219_.~~

https://github.com/arendst/Tasmota/pull/10889

## Checklist:


ajithvasudevan:
Presently, the proper initialization and working of the TM1637/8 driver depends on the existence of `USE_SPI`. 

However, this is not really required for this driver to work.

Unnecessarily enabling **USE_SPI** also causes the SPI pins to be assigned to 3 GPIOs in Tasmota Configuration page, and it cannot be unset.

To this end, I have made the following changes so that the dependency of TM1637 on SPI is removed.

arendst:
As there is already a TM1638 button driver I would have expected this driver should use the same GPIOs. I see no need for adding TM1638 GPIOS as there are already TM16xx GPIOs only to be used by the TM1638 driver.

The new defined GPIOs in the template file also break any user GPIO configuration as they are inserted in the list instead of adding them to the list.

I made changes to TM1637 to allow GPIO selection without SPI enable. This should also be the case for this TM1638 display driver.

There should be no changes neede to the ``xdsp_interface.ino`` file. If so something else is wrong.

ajithvasudevan:
> As there is already a TM1638 button driver I would have expected this driver should use the same GPIOs. I see no need for adding TM1638 GPIOS as there are already TM16xx GPIOs only to be used by the TM1638 driver.
> 
> The new defined GPIOs in the template file also break any user GPIO configuration as they are inserted in the list instead of adding them to the list.
>

Noted. I have made changes to use the existing TM16XX GPIOs
 
> I made changes to TM1637 to allow GPIO selection without SPI enable. This should also be the case for this TM1638 display driver.
> 

Please note that this is not a **new** driver. It is an update to the **existing** TM1637 Driver which adds support for TM1638.

I am unable to get the `xdrv_13` driver loaded if I revert my changes.
Specifically, in the callback  `bool Xdrv13(uint8_t function)` , the condition

`if ((TasmotaGlobal.i2c_enabled || TasmotaGlobal.spi_enabled || TasmotaGlobal.soft_spi_enabled) && XdspPresent()) {`

is never true. 

I must be missing something. Can you help me figure it out?



> There should be no changes neede to the `xdsp_interface.ino` file. If so something else is wrong.

I have reverted the changes to `xdsp_interface.ino` in the latest commit.

ajithvasudevan:
@arendst 

> I made changes to TM1637 to allow GPIO selection without SPI enable. This should also be the case for this TM1638 display driver.
> 

It seems allowing TM1637 GPIO selection isn't the only thing necessary for the TM1637 driver to work. 
Currently, the TM1637 support itself is broken in Release 9.3.1.  Please see new issue https://github.com/arendst/Tasmota/issues/11358

ajithvasudevan:
@arendst 
After merging the fix https://github.com/arendst/Tasmota/issues/11358 everything works fine now. You can merge this PR with development if and when you see fit :)

arendst:
I see a new command called ``DisplayType`` which does nothing more than fixing the column count. So the current command ``DisplayCols`` could also be used.

I hate all these new and redundant Display commands. They add redundant code and need studying the docs to find what you need.

Could you remove ``DisplayType`` please and replace by using the output of command ``DisplayCols``.

tasmota/xdsp_15_tm1637.ino
```diff
@@ -132,52 +154,133 @@
...
+           <state> = state of the corresponding LED. It can be "ON" or "OFF" 
+
+  For example, if Button3 is pressed, LED3 would light up and a STAT message is sent as follows:
+  
+  stat/TM1638/RESULT = {"TM1638_BUTTON3":"ON"}
+
+  Pressing Button3 again would turn OFF LED3 and a STAT message is sent as follows:
+
+  stat/TM1638/RESULT = {"TM1638_BUTTON3":"OFF"}
+
```
arendst, line 208: As far as I know TM1638 button support has always been present as it emulates additional standard buttons. What I see here is adding a newkind of button just to support the ones from TM1638 without integration in the Tasmota button pool. It is nice playing but not really functional and again adds redundant code.

ajithvasudevan:
> I see a new command called `DisplayType` which does nothing more than fixing the column count. So the current command `DisplayCols` could also be used.
> 
> I hate all these new and redundant Display commands. They add redundant code and need studying the docs to find what you need.
> 
> Could you remove `DisplayType` please and replace by using the output of command `DisplayCols`.

I added ```DisplayType``` to essentially switch between 4 variants of 7-Segment Displays, while using the same driver (TM1637) - 

* TM1637, 4-digit
* TM1637, 6-Digit
* TM1638
* MAX7219

Incidentally, it also affects the column-count.

I thought it better to keep this separate for 7-segment displays supported by the TM1637 driver and any future 7-segment display variants.

I think it is not appropriate to check the value of ```DisplayCols``` to do this switching between 7-segment variants.

Let me know if you still think I should replace ```DisplayType``` with ```DisplayCols```

arendst:
> I think it is not appropriate to check the value of DisplayCols to do this switching between 7-segment variants.

Switching between hardware needs to be done based on GPIO type. So the TM1637 code is used only when TM1637 GPIO's are selected; TM1638 code is used when TM1638 GPIO's are selected; MAX7219 code will be used when appropriate GPIO's are selected. No need for additional commands an user error due to wrong driver trying to access wrong hardware.

ajithvasudevan:
> 
> Switching between hardware needs to be done based on GPIO type. So the TM1637 code is used only when TM1637 GPIO's are selected; TM1638 code is used when TM1638 GPIO's are selected; MAX7219 code will be used when appropriate GPIO's are selected. No need for additional commands an user error due to wrong driver trying to access wrong hardware.

Alright. I will fix this so that ```DisplayCols``` is used for the switch between 4-digit to 6-digit of the TM1637, and the GPIO selection will switch drivers.

tasmota/xdsp_15_tm1637.ino
ajithvasudevan: > As far as I know TM1638 button support has always been present as it emulates additional standard buttons. What I see here is adding a newkind of button just to support the ones from TM1638 without integration in the Tasmota button pool. It is nice playing but not really functional and again adds redundant code.

I can try merging the functionality of the TM1638 Sensor (buttons) driver into this TM1638 Display Driver. But that would mean truly redundant code.
Or, would you rather have me remove the buttons functionality altogether from this TM1638 Display Driver?

ajithvasudevan:
> As far as I know TM1638 button support has always been present as it emulates additional standard buttons. What I see here is adding a newkind of button just to support the ones from TM1638 without integration in the Tasmota button pool. It is nice playing but not really functional and again adds redundant code.

I can try merging the functionality of the TM1638 Sensor (buttons) driver into this TM1638 Display Driver. But that would mean truly redundant code.
Or, would you rather have me remove the buttons functionality altogether from this TM1638 Display Driver?

arendst:
Pls remove the buttons functionality from the TM1638 display driver. It's a display driver after all.

I haven't looked at the leds code but I guess the same is true for the leds as they should be handled as relays.

ajithvasudevan:
> Pls remove the buttons functionality from the TM1638 display driver. It's a display driver after all.
> 
> I haven't looked at the leds code but I guess the same is true for the leds as they should be handled as relays.

Will remove buttons, and LEDs support from TM1638 Display Driver, for now.

arendst:
BTW I see your PR'ed updated TM driver still contains bad init code.

Just before 9.3.1 release I had to make changes in order to make the TM1637 driver work within the display eco-system without claiming the display_model.

I suggest you base your TM1638 changes on the current released TM1637 driver as that works fine within the display ec-system.

ajithvasudevan:
> BTW I see your PR'ed updated TM driver still contains bad init code.
> 
> Just before 9.3.1 release I had to make changes in order to make the TM1637 driver work within the display eco-system without claiming the display_model.
> 
> I suggest you base your TM1638 changes on the current released TM1637 driver as that works fine within the display ec-system.

Will take care of that.

ajithvasudevan:
Its done. Please see latest commit.

arendst:
Thx

swampdogmash:
Hi, Thanks for adding the support for TM1638 as a display - I wonder what options you consider there are for adding back the support for the button and the LEDs? It seems like a pretty useful little board for doing control with. It's sort of mostly a redundant piece of hardware if it's connected to something running Tasmota without the buttons and LEDs. I'd love to offer my help, but I can't code to save my life... hence using Tasmota!

I saw that some time back there was some support for the buttons (I think) [https://github.com/arendst/Tasmota/issues/3487](url) - is there a way to implement it going forward?

mkohns:
Hi! I am also wondering how the buttons and leds could be included. Are there any concerns around this feature.

ajithvasudevan:
I think it was an issue of separation of concerns in the driver
implementation. A driver could either be an output driver (display, in this
case), or an input driver (one that reads from buttons or sensors). Not
both together.

barbudor:
While it is best to have indépendant drivers for indépendant component that can be re-usedin a different context, it happens that sometime a driver has to be done for a specific board and not for si gle components when they are tight together

How are the buttons connected?
To the esp? To an MCU?

mkohns:
The buttons are part of the TM1638
image
I am using the library TM1638plus in my projects.
The buttons are read by the functions readButtons()
So my answer to your question would be: The buttons are not connected directly to the GPIOs of the ESP.
Rather the chipset TM1638 is connected via SPI and retrieves the buttons over this interface.

From my point of view this is an edge case. 
I understand the SoC principal - but here the controller has input and output.
Separating them would not make sense as they always come together.
On top - separation would make the code complicated - as both use the same SPI interface.

mkohns:
Another question: I am not yet familiar with the new esp32 berry script. Just reading the docs - I have the feeling that reading the buttons could be possible with berry. What is your feeling about this?

barbudor:
The good thing is that Tasmota driver is also based on TM1638plus
I'm looking at it

barbudor:
There is already support for the buttons as Switches in the driver
Aren't this doing anything ?
I'm trying to understand how it's working...

barbudor:
@mkohns 
I discussed with Theo and what is done currently in the driver needs to be changed.
I'm working on it but I need to order a TM1638 to test properly

arendst:
Try latest dev release. It introduces the keys as switches and the leds as relays. All switchmodes do apply.

Optionally compile options can change switch support to button support allowing use of button features.

#define USE_TM1638                               // Add support for TM1638 switches copying Switch1 .. Switch8 (+1k code)
//  #define TM1638_USE_AS_BUTTON                   // Add support for buttons
  #define TM1638_USE_AS_SWITCH                   // Add support for switches (default)
  #define TM1638_MAX_DISPLAYS    8               // Add support for power control 8 displays
  #define TM1638_MAX_KEYS        8               // Add support for 8 keys
  #define TM1638_MAX_LEDS        8               // Add support for 8 leds

mkohns:
HI @arendst,

i compiled the current dev branch and tested the TM1638.
This all looks very good. I got a bar of 8 Web Buttons + States on top.
Pressing the hardware buttons -> turns the hardware led (relay) on + correct log in console.
Pressing the Web buttons -> turns the hardware led (relay) on + correct log in console.
Look great.

I am using a board tasmota32s2.
I now tried to get DISPLAY + your code running. To see if both is working.
I have no glue how to do this.
This only enables DISPLAY (which works)

```
[env:tasmota32s2-display]
extends                 = env:tasmota32_base
board                   = esp32s2
build_flags             = ${env:tasmota32_base.build_flags} -DFIRMWARE_DISPLAYS
lib_extra_dirs          = lib/libesp32, lib/lib_basic, lib/lib_display, lib/lib_ssl
lib_ignore              =
                          TTGO TWatch Library
                          NimBLE-Arduino
                          Micro-RTSP
                          epdiy
```

But your code seems to be in -DFIRMWARE_TASMOTA32
Apparently I can not mix them.

Can you give me a hint?

mkohns:
I reverted this new tasmota32s2-display from above and
just added #define FIRMWARE_DISPLAYS in tasmota/my_user_config.h

By this both is now working. Looks like this

image

Where "9" is now the display.

barbudor:
That's a hack, not officially supported :roll_eyes: 

The proper way is to compile tasmota32 with all the needed defines in user_config_override.h
```
#define USE_DISPLAY
#define USE_DISPLAY_TM1637

#define USE_TM1638 // Add support for TM1638 switches copying Switch1 .. Switch8 (+1k code)
// #define TM1638_USE_AS_BUTTON // Add support for buttons
#define TM1638_USE_AS_SWITCH // Add support for switches (default)
#define TM1638_MAX_DISPLAYS 8 // Add support for power control 8 displays
#define TM1638_MAX_KEYS 8 // Add support for 8 keys
#define TM1638_MAX_LEDS 8 // Add support for 8 leds
```

barbudor:
Just received today my 2 TM1638
Hooked one, compiled tasmota (ESP8266) with the above defines
I also tested the BUTTONS vs SWITCHES

Works like a charm !

barbudor:
A little ruleset to enter a 8 digit code (with only digits 1 to 8)
```
DisplayModel 15
DisplayWidth 8

rule1 
  on system#init do event reset endon 
  on event#reset do backlog ruletimer1 0; displaymode 1; var1 1; var2 " endon

rule2 
  on button1#state do event key=1 endon
  on button2#state do event key=2 endon 
  on button3#state do event key=3 endon 
  on button4#state do event key=4 endon 
  on button5#state do event key=5 endon 
  on button6#state do event key=6 endon 
  on button7#state do event key=7 endon 
  on button8#state do event key=8 endon 
  on event#key do backlog var2 %var2%%value%; displaymode 0; displaytext %var2%%value%; ruletimer1 10; add1 1 endon 
  on rules#timer=1 do event reset endon 
  on var1#state==9 do backlog ruletimer1 10; var3 The Code is %var2% endon

rule1 1
rule2 1
```

Will do a little update to the docs

barbudor:
docs updated
https://tasmota.github.io/docs/TM163x/#tm1638-leds-and-buttons

FraatTailscale:
Hallo, can you please add also support for the tm1638 4x4 16 button board??

[https://www.amazon.de/ASHATA-Digital-LED-R%C3%B6hrenanzeigemodul-MCU-Steuermodul-8-Bit-LED-Anzeigen-default/dp/B084LFQMV4/ref=sr_1_7?__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&crid=1VPM2GPR6EJBK&keywords=tm1638&qid=1680809319&refinements=p_85%3A20943776031&rnid=20943775031&rps=1&sprefix=tm1638%2Caps%2C96&sr=8-7](url)

barbudor:
I have one in the AliEx mail but expected in may :(
Is there some sample code somewhere so I can have a look at the difference?

barbudor:
@FraatTailscale I haven't received my unit yet but looking at the code, I think everything is there
When self compiling you can specifiy the number of Buttons, LEDs and 7-segment 
Just add in your `user_config_override.h` with the proper values:
```
#define TM1638_MAX_DISPLAYS   8
#define TM1638_MAX_KEYS       16
#define TM1638_MAX_LEDS       0
```

barbudor:
I have received my TM1638 4x4 buttons boards
Unfortunately it doesn't work as expected: neither LCDs display correct digits nor buttons works as expected

There must be some more differences between the 2 boards in trms of wiring/muxing

Does anyone knows where I can find reference/sample code that works for both model of board so I can compare ?

@Jason2866 Didn't you also ordered a 4x4 board ?

Regards

Jean-Michel

barbudor:
Ok, I found the library and there will be some changes to do
Working on it now

Jason2866:
@barbudor did not order the 4x4 version.

FraatGIT:
Hallo, are there any news regarding the TM1638 4x4 board?

barbudor:
I apparently burned down my 1st 4*4 board.
So I ordered and received another one but haven't much time to play with it
Before destroying the 1st one I had very partial result even by upgrading the library.
I need to find some time to work on it again

FraatGIT:
Hallo, are there any news regarding the TM1638 4x4 board?

PR #11087, opened 2021-02-22
Added option to configure ledlink to show a heartbeat

jeroenst:
## Description:
The link led is a status led which is often off or on when connections (wifi and mqtt) are established. This option enables the user to use the command "ledlinkheartbeat" with payload 0 or 1 to enable or to disable this feature which is disabled by default to prevent changing default behaviour of the led.

When enabled the linkled will be (depending on the inverted flag) on for 3750ms and off for 250ms continuously when connections are established.

## Checklist:

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -82,6 +82,7 @@ The attached binaries can also be downloaded from http://ota.tasmota.com/tasmota
 ### Added
 - Animate PWM dimmer brightness LEDs during transitions and with variable brightness [#11076](https://github.com/arendst/Tasmota/issues/11076)
 - Commands ``StateRetain`` and ``InfoRetain`` [#11084](https://github.com/arendst/Tasmota/issues/11084)
+- Command ``LedLinkHeartbeat`` https://github.com/arendst/Tasmota/pull/11087
 
 ### Changed
 - Remove the need to start filenames with a slash (/) in Ufs commands
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -82,7 +82,7 @@ The attached binaries can also be downloaded from http://ota.tasmota.com/tasmota
 ### Added
 - Animate PWM dimmer brightness LEDs during transitions and with variable brightness [#11076](https://github.com/arendst/Tasmota/issues/11076)
 - Commands ``StateRetain`` and ``InfoRetain`` [#11084](https://github.com/arendst/Tasmota/issues/11084)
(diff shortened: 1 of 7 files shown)


s-hadinger:
Thanks for the PR. However I'm not convinced by the usefulness of this feature. Who wants a device that blinks endlessly? Blinking should denote a temporary action or an anomaly

jeroenst:
> Thanks for the PR. However I'm not convinced by the usefulness of this feature. Who wants a device that blinks endlessly? Blinking should denote a temporary action or an anomaly

Blinking also gives the impression the device is on and working (not stalled) which in some cases is very useful imo.

s-hadinger:
Did you try rules to achieve the same result?

jeroenst:
I didn't try to make a rule for this.

Jason2866:
I like that the user can decide if she/he wants blinking as a sign of "life"
This is not uncommon for electronic devices. All my APs have this. AND all have the feature to switch this off. I use this option
because i dont like "Disco" in my living and sleeping rooms (especially at night).

jeroenst:
An additional benefit of this change is that it also works with sonoff zbbridge, where the configuration of ledlink in the template configuration doesn't work at this moment. (i have to create a bugreport for that).

s-hadinger:
What do you mean it doesn't work with zbbridge?
Be aware that this led is also used as a pulsating indicator for permit join. What happens if both modes are on simultaneously?

jeroenst:
> What do you mean it doesn't work with zbbridge?
> Be aware that this led is also used as a pulsating indicator for permit join. What happens if both modes are on simultaneously?

The pulse when permit join differs from the heartbeat pattern so it's easy to see that it's in join state.

ascillato:
Hi, thanks a lot for your idea. It is well programmed but this change is not going to be added, sorry. It uses the blink as ok while in Tasmota a blink is something that needs attention.

jeroenst:
> I like that the user can decide if she/he wants blinking as a sign of "life"
> This is not uncommon for electronic devices. All my APs have this. AND all have the feature to switch this off. I use this option
> because i dont like "Disco" in my living and sleeping rooms (especially at night).

True, in my bedroom the

> Hi, thanks a lot for your idea. It is well programmed but this change is not going to be added, sorry. It uses the blink as ok while in Tasmota a blink is something that needs attention.

Well, it's optional, not default..

arendst:
Sorry, another waste of code space currently on hold.

ascillato2:
Thanks for sharing your work but it is not going to be merged. Sorry.

PR #11095, opened 2021-02-22
Power margins for all channels/phases

vic42:
## Description:
This patch allows PowerLow, PowerHigh, CurrentLow, CurrentHigh to be triggered based on the sum of all phases/channels. It came to my attention when a Shelly 2.5 would only evaluate channel 1 of 2.
I think that it makes more sense to act on the total device power/current instead of power/current of just the first phase or channel.

## Checklist:

### DIFF
--- a/tasmota/xdrv_03_energy.ino
+++ b/tasmota/xdrv_03_energy.ino
@@ -377,11 +377,17 @@ void EnergyMarginCheck(void)
     ResponseAppend_P(PSTR("\"" D_CMND_POWERDELTA "\":%s"), EnergyFormat(value_chr, power_diff_chr[0], 1));
   }
 
-  uint16_t energy_power_u = (uint16_t)(Energy.active_power[0]);
+  uint16_t energy_power_u = 0;
+  for (uint32_t phase = 0; phase < Energy.phase_count; phase++) {     // Sum up power of all channels/phases
+    energy_power_u += (uint16_t)(Energy.active_power[phase]);
+  }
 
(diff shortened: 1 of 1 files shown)


Jason2866:
I think it depends o the use case. Imho one solution is not "better" than the other.

barbudor:
Came across this one
I feel that we have 2 different use cases:
- Multi-phases, generally managed with 2 or 3x PZEM like described by TrDA
- Multi-channel, like Shelly 2.5
The current implementation that only looks at phase/channel 1 doesn't seems satisfying for any of those.
- In multi-phases case, I would expect the power/current to be monitor for a single device (or installation/house) where user mainly concerned by the total power. In that case looking only at power/current of phase 1 (out of 2..3) doesn't make any sense.
- In multi-channels, user would probably prefer to look separately at each channel as they are likely to be different pieces of equipment.
Of course, I'm sure there could be use case totally opposite.

Unless we go for a larger PR that provide the choice (SetOption) between sum of phases/channels or each of the 2..3 phases/channels separately, dealing only with the sum seems still better than dealing with only phase/channel 1.
Could this be a 1st step ?

vic42:
I totally agree that this could be done even more elegant with a `SetOption` and having margins for each channel, however, likely at the cost of additional complexity.

The reasons why I chose exactly this design:
* It is minimally invasive
* It doesn't need extra options
* It is very unlikely to break something
  * Case 1 (single channel): no change in behavior.
  * Case 2 (multiple channels): The feature is currently useless for multiple channels so most likely very rarely used. In the rare case that someone was monitoring channel 1 only with margins it could impact the outcome.
* It will most likely meet the expectations of an unbiased user, at least for `PowerHigh`, `PowerLow`, `MaxPower`.
* Per channel values would somewhat duplicate the `PowerDelta` feature

vic42:
No love for this one ;-)

PR #11188, opened 2021-03-02
[irremote_full] Save the current status even when sending IR

nao-pon:
## Description:

If the IR transmission is successful, it should save the current status. If it does not notify the current state next time, the toggle function data will be inconsistent.

see https://github.com/crankyoldgit/IRremoteESP8266/issues/1424

## Checklist:


s-hadinger:
I don't agree with your change that creates a regression with the previous version. Saving the state should depend on `stateMode`.

We can discuss though of changing the default of `stateMode` from `StateModes::SEND_ONLY` to `StateModes:: SEND_STORE `

nao-pon:
Yes, I understand. What is the effect of defaulting `stateMode` to` StateModes::SEND_STORE`?

nao-pon:
Looking at the source code, I noticed that SendMode can be modified by the JSON payload. I would like to withdraw this PR as my wishes can be achieved by it.

Thank you for your support.

PR #11241, opened 2021-03-06
Fix WS2812 ESP32

s-hadinger:
## Description:

The latest version of GCC prevent the assignment of attributes to methods and requires to specify for each specialization.
See https://stackoverflow.com/questions/36279162/section-attribute-of-a-function-template-is-silently-ignored-in-gcc

This creates a fundamental problem with NeoPixelBus BitBang implementation, and prevents the BitBang code from being into IRAM, which triggers display errors because of timing inaccuracies.

I've tried many options, but the only work-around for now it to explicitly create specialization of the `send_pixels` method for each option. This creates duplication of source code, but has no impact on compiled code.

Currently this work-around is only applied to `WS2812` inverted and non-inverted.

There will be more improvements to make `I2S` and `RMT` schemes available too, which are required for long led stripes.

## Checklist:
  - [x] The code change is tested and works with core ESP32 V.1.0.5

### DIFF
--- a/tasmota/xlgt_01_ws2812.ino
+++ b/tasmota/xlgt_01_ws2812.ino
@@ -123,6 +123,113 @@ void (* const Ws2812Command[])(void) PROGMEM = {
 
 #endif  // No USE_WS2812_DMA
 
+#ifdef ESP32
+//
+// Below is a quick work-around to ESP32 gcc that prevents templated methods to be in IRAM unless explicitly marked per specialization
+// It will work only for the default WS2812 (800KHz) and inverted and non-inverted
+//
+template <>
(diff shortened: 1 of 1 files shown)


hallard:
Oh yeah @s-hadinger 
For some days my test board was unable to drive my WS2812 correctly, which was driving me confused because it worked before but as I was on other issues, I did not took time to investigate to this issue (but it was on my todo)
So applied, compiled, flashed, and guess what, solved my issue
Thanks because it avoided my long time to understand why and when it was occurring. 
Nice shot.

s-hadinger:
Yes. I was also confused when trying to use Denky and M5stack Atom Lite

hallard:
Nice to see you're using your Denky :-)

PR #11354, opened 2021-03-15
Add support for MPU6686 on primary or secondary I2C bus

s-hadinger:
## Description:

Multiple improvements:
- removed MPU6886 from M5Stack Core 2 code, removed all hard-coded values and implemented as a proper driver
- selects automatically primary or secondary I2C bus on ESP32.
- simplified calculation to avoid unnecessary `float` calculation.
- displays acceleration (in 1/1000th of G) and gyroscope (dps - degree per second)
- removed complex AHRS code that was not used

## Checklist:
  - [x] The code change is tested and works with core ESP32 V.1.0.5


arendst:
Why is the MFRC522 code included? Currently the MFRC522 is only supported on SPI....

Jason2866:
More a optic issue. Should be moved https://github.com/arendst/Tasmota/blob/development/tasmota/tasmota_configurations_ESP32.h#L84
to i2c section

s-hadinger:
Oops, my bad. It was not supposed to be in this PR but in a future one.

The reason is that MFRC522 is provided as an external I2C device by M5Stack. Let me remove it for now.

PR #11361, opened 2021-03-16
loghost resolve failure leads to broadcast logs

barbudor:
### Problem
Discord user "eddieb" reported strange behavior of syslog in his use case.
https://discord.com/channels/479389167382691863/490244847568158751/821275083342741535

When using a name for loghost instead of an IP, right after a restart, Tasmota start to broadcast syslog to 255.255.255.255 instead of the correct server's IP. Moreover his logs never starts on the expect 1st line of log  `CFG: Loaded from flash at...` but on `ESP-WIF: Connected`.

### Analysis
As I was unable to reproduce on my mini-D1, eddieb seems to be in a rare race-condition case where `SyslogAsync()` is called before Wifi is fully initialized which makes the call to Wifi.getHostByName() to fail and use 0xFFFFFFFF as a broadcast address to the log server. 1st lines of logs are lost for the same reason as there are send out while Wifi is not yet ready.
Similar problem occur if one define a loghost to a name that cannot be resolved to an IP (like DNS is down or typo in the name). As the resolution would be done only once unless the loghost hash change, Tasmota will continuously broadcast the logs.

### Proposed fix
- Move the test on loghost hash outside of the loop as there is no reason that the logserver is changed during this loop
- Check result of Wifi.getHostByName() and immediately returns if it fails which prevent sending the 1st logs to the void
- Save the addr and the hash only if the name resolution is ok

This prevent Tasmota to broadcast the logs whether loghost cannot be resolved because it is wrong or specific eddieb's race-condition.
Broadcasting the syslog is still a "feature" as one can enter `loghost 255.255.255.255`

### Drawback 
If loghost cannot be resolved because of a typo in the name (or DNS server is down), syslog index will not move and SyslogAsync() will continuously try to resolve the name at the max call frequency of SyslogAsync() (250ms ?).
An intermediate solution would be to leave the code in the loop, still checking the return code of Wifi.getHostByName() and returning if resolution fails without saving bad address and hash.
This would lose 1st lines of syslog in eddie's case but would limit the CPU load for anyone entering a wrong syslog name as resolution will try to occur only when new lines are added to the log buffer.

Happy to change for that option if you prefer.

## Checklist:

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -2128,6 +2128,15 @@ void SyslogAsync(bool refresh) {
   if (!TasmotaGlobal.syslog_level) { return; }
   if (refresh && !NeedLogRefresh(TasmotaGlobal.syslog_level, index)) { return; }
 
+  uint32_t current_hash = GetHash(SettingsText(SET_SYSLOG_HOST), strlen(SettingsText(SET_SYSLOG_HOST)));
+  if (syslog_host_hash != current_hash) {
+    IPAddress temp_syslog_host_addr;
+    int ret = WiFi.hostByName(SettingsText(SET_SYSLOG_HOST), temp_syslog_host_addr);  // If sleep enabled this might result in exception so try to do it once using hash
+    if (!ret) return; // name resolution dosn't work, too early ? or wrong name
+    syslog_host_hash = current_hash;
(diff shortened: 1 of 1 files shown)


barbudor:
Hold !
More test to do

Issue #8965, opened 2020-07-23
MCP23017 outputs and DS18B20 double telemetry 


hebdam:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._

Tasmota 8.2.0.4 fimware

I have both MCP23017 I/O expander and DS18B20 connected and reading it in OPENHAB. That works well  except, that telemetry publishes twice state of SENSOR. Once for DS18B20 and MCP23017 and immediatly after MCP23017 ouput state.
Both of them are published as tele/*****/SENSOR, so OPENHAB reads the DS18B20 correctly but since the next publish is as SENSOR and only for MCP23017 then my OPENHAB log files reports it as a transformation path incorrect for DS18B20.

Tasmota publishes
  10:17:27 MQT: tele/garage_control/SENSOR = {"Time":"2020-07-23T10:17:27","DS18B20":{"Id":"020A92464368","Temperature":22.31},"MCP230XX":{"D0":0,"D1":0,"D2":1,"D3":1,"D4":0,"D5":0,"D6":0,"D7":0,"D8":1,"D9":1,"D10":1,"D11":1,"D12":0,"D13":0,"D14":0,"D15":0},"TempUnit":"C"}
10:17:27 MQT: tele/garage_control/SENSOR = {"Time":"2020-07-23T10:17:27","MCP230_OUT":{"OUT_D0":"OFF","OUT_D2":"OFF","OUT_D3":"OFF","OUT_D4":"OFF","OUT_D5":"OFF","OUT_D6":"OFF","OUT_D7":"OFF","END":1}}

and Openhab log
2020-07-23 11:17:27.345 [WARN ] [t.generic.ChannelStateTransformation] - Executing the JSONPATH-transformation failed: Invalid path '$.DS18B20.Temperature' in '{"Time":"2020-07-23T10:17:27","MCP230_OUT":{"OUT_D0":"OFF","OUT_D2":"OFF","OUT_D3":"OFF","OUT_D4":"OFF","OUT_D5":"OFF","OUT_D6":"OFF","OUT_D7":"OFF","END":1}}

Not a big deal but it fills the log file with every 5 minutes messages obscuring any other issues. Ys I can probably disable those warnings in OPENHAB, but perhaps soeone can offer a solution

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

[...]

PR #11394, opened 2021-03-20
Mcp230xx output mode 2

theOzzieRat:
## Description:
This adds an additional mode when setoption0 is 0.
```sensor29 pin,5|6,2```
on a restart the current value is read from the MCP230xx ic and synced internally. 
The display of the sensor pin,? command when the pin is configured as an output has also been updated to display properly. Before it displayed values as if the pin was an input which made no sense.
Updated the code sizes in the comments of the config file.

## Checklist:

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,9 +12,9 @@
 [build_envs]
 default_envs =
 ; *** Uncomment by deleting ";" in the line(s) below to select version(s)
-                tasmota
+;                tasmota
 ;                tasmota-ircustom
-                tasmota-minimal
+;                tasmota-minimal
 ;                tasmota-lite
 ;                tasmota-knx
 ;                tasmota-sensors
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,9 +12,9 @@
 [build_envs]
 default_envs =
 ; *** Uncomment by deleting ";" in the line(s) below to select version(s)
-;                tasmota
+                tasmota
 ;                tasmota-ircustom
-;                tasmota-minimal
+                tasmota-minimal
 ;                tasmota-lite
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: platformio.ini [hunk 1]
```diff
@@ -12,9 +12,9 @@
 [build_envs]
 default_envs =
 ; *** Uncomment by deleting ";" in the line(s) below to select version(s)
-;                tasmota
+                tasmota
 ;                tasmota-ircustom
-;                tasmota-minimal
+                tasmota-minimal
```
arendst, line 17: Pls remove this change

theOzzieRat: Hopefully that is fixed.

barbudor:
@theOzzieRat 
This section seems strange with a little bit too much of #ifdef USE_MCP230xx_OUTPUT
https://github.com/arendst/Tasmota/blob/ba8ef7feb5d0e5cdf1211cfedcbfe2c4828d0059/tasmota/xsns_29_mcp230xx.ino#L127-L140

theOzzieRat:
@barbudor 
I agree. What was I thinking. I also left some commented out code in there that I should have removed (at line774). I will clean this up.

There is also some previously commented out code at line 818 that possibly should be removed? Should I do that as well?

barbudor:
Indeed I agree that 770-783 would probably benefit to be cleared up. 
For 818 and the subsequent call in FUNC_EVERY_SECOND, I see that Theo commented that out after issue #8965.
He may have had good reasons to comment it only. In case someone complains and what to implement a SO to control it ? 
If you do a cleanup, may be just add a note to reference the issue number at the beginning of the commented code ?
thanks

PR #11410, opened 2021-03-21
only perform interlock delay once

t-8ch:
## Description:

This make sure that the delay after turning off interlocked devices is only executed once and not for each device turned off.

NOTE: This issue was not reproduced with a real device or tested with one. I stumbled upon it while looking at the code with no device (yet) at hand.
If somebody could validate it that would be much appreciated.

## Checklist:

Also mentioned in https://github.com/arendst/Tasmota/discussions/11407

### DIFF
--- a/tasmota/support_tasmota.ino
+++ b/tasmota/support_tasmota.ino
@@ -556,17 +556,21 @@ void ExecuteCommandPower(uint32_t device, uint32_t state, uint32_t source)
         ((POWER_ON == state) || ((POWER_TOGGLE == state) && !(TasmotaGlobal.power & mask)))
        ) {
       interlock_mutex = true;                           // Clear all but masked relay in interlock group if new set requested
+      bool perform_interlock_delay = false;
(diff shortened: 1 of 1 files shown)


tasmota/support_tasmota.ino
```diff
@@ -544,6 +544,7 @@ void ExecuteCommandPower(uint32_t device, uint32_t state, uint32_t source)
   SetPulseTimer((device -1) % MAX_PULSETIMERS, 0);
 
   static bool interlock_mutex = false;    // Interlock power command pending
+  bool perform_interlock_delay = false;
```
arendst, line 547: Wrong location. As ExecuteCommand is re-entrant it will clear this flag every time disabling the needed delay. This should be moved between the interlock_mutex flags.

t-8ch: As the variable is not `static` that shouldn't be an issue, or?

From a stylistic point it would make sense to move it into the scope where it is useful which would be the same place.

Issue #11358, opened 2021-03-16
tasmota-display build isn't enabling TM1637 driver without enabling SPI. 


ajithvasudevan:
### PROBLEM DESCRIPTION
Tasmota's _Display_ build (_tasmota-display.bin_) is supposed to support **TM1637** in release **9.3.1**. However, this is not happening if we configure just the **TM1637** specific GPIO pins (TM1637 DIO and TM1637 CLK). 

After configuring the _TM1637 DIO_ and _TM1637 CLK_ pins, none of the ```DISPLAY-``` commands are available, indicating that ```xdrv_13_display``` is not executed.

I have traced the issue (in the latest development branch) to the following in ```xdrv_13_display.ino``` :

```
bool Xdrv13(uint8_t function)
{
  bool result = false;

  if ((TasmotaGlobal.i2c_enabled || TasmotaGlobal.spi_enabled || TasmotaGlobal.soft_spi_enabled) && XdspPresent()) {
```
The above condition is ```false``` even if **TM1637** pins are configured 

This could be because of ```TasmotaGlobal.soft_spi_enabled``` not being ```true``` from the following code in ```support_tasmota.ino```

```
    bool valid_cs = (ValidSpiPinUsed(GPIO_SPI_CS) ||
                     ValidSpiPinUsed(GPIO_RC522_CS) ||
                     (ValidSpiPinUsed(GPIO_NRF24_CS) && ValidSpiPinUsed(GPIO_NRF24_DC)) ||
                     ValidSpiPinUsed(GPIO_ILI9341_CS) ||
                     ValidSpiPinUsed(GPIO_ILI9341_DC) || // there are also boards without cs
                     ValidSpiPinUsed(GPIO_EPAPER29_CS) ||
                     ValidSpiPinUsed(GPIO_EPAPER42_CS) ||
                     ValidSpiPinUsed(GPIO_ILI9488_CS) ||
                     ValidSpiPinUsed(GPIO_SSD1351_CS) ||
                     ValidSpiPinUsed(GPIO_RA8876_CS) ||
                     ValidSpiPinUsed(GPIO_ST7789_DC) ||  // ST7789 CS may be omitted so chk DC too
                     ValidSpiPinUsed(GPIO_ST7789_CS) ||
                     (ValidSpiPinUsed(GPIO_SSD1331_CS) && ValidSpiPinUsed(GPIO_SSD1331_DC)) ||
                     ValidSpiPinUsed(GPIO_SDCARD_CS)
[...]

PR #11422, opened 2021-03-22
Added support for another variant of the 6-digit TM1637 display module

ajithvasudevan:
## Description:

This adds support for another variant of the 6-digit TM1637 display module, where the digit order is different in hardware.
Also, this fixes the issue reported here https://github.com/arendst/Tasmota/issues/11358#issuecomment-803649954

## Checklist:

### DIFF
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -332,7 +332,7 @@ typedef union {
   struct {
   uint8_t ilimode : 3;
   uint8_t Invert : 1;
-  uint8_t spare2 : 1;
+  uint8_t tm1637_variant : 1;
   uint8_t spare3 : 1;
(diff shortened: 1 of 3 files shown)


arendst:
Thx.

I see what you need but I don't like it.

The issue is that both the ILI driver and your driver seem to need additional submodel parameterization. That's fine but adding another command to do this doesn't make sense regarding keeping the code small.

I will rewrite your PR by using a new command for both ILI and your driver. 

Stay tuned...

arendst:
Merged using command ``DisplayType 1`` for the second variant.

ajithvasudevan:
> Merged using command `DisplayType 1` for the second variant.

Thanks.

PR #11472, opened 2021-03-26
Ds18x20 arithm mean

pharaocurt:
## Description:

if DSx20 sensors are powered by VCC, the temperature is read every 2 seconds and updated mostly in the WEB_SERVER. MQTT gets only a new value every telemetry period. An arithmetic mean can be used reduce noise variations and increase the resolution.

## Checklist:

Please ignore my changes to platformio.ini

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,7 +12,7 @@
 [build_envs]
 default_envs =
 ; *** Uncomment by deleting ";" in the line(s) below to select version(s)
-                tasmota
+;                tasmota
 ;                tasmota-ircustom
 ;                tasmota-minimal
 ;                tasmota-lite
@@ -81,11 +81,11 @@ build_flags                 = ${core.build_flags}
 board_build.f_cpu           = 80000000L
 board_build.f_flash         = 40000000L
 monitor_speed               = 115200
-monitor_port                = COM3
+monitor_port                = COM5
 upload_speed                = 115200
 ; *** Upload Serial reset method for Wemos and NodeMCU
 upload_resetmethod          = nodemcu
-upload_port                 = COM3
+upload_port                 = COM5
 extra_scripts               = ${scripts_defaults.extra_scripts}
 lib_ldf_mode                = chain
 lib_compat_mode             = strict
--- a/platformio.ini
+++ b/platformio.ini
@@ -12,7 +12,7 @@
 [build_envs]
 default_envs =
 ; *** Uncomment by deleting ";" in the line(s) below to select version(s)
-;                tasmota
+                tasmota
 ;                tasmota-ircustom
 ;                tasmota-minimal
 ;                tasmota-lite
@@ -81,11 +81,11 @@ build_flags                 = ${core.build_flags}
 board_build.f_cpu           = 80000000L
 board_build.f_flash         = 40000000L
 monitor_speed               = 115200
-monitor_port                = COM5
+monitor_port                = COM3
 upload_speed                = 115200
 ; *** Upload Serial reset method for Wemos and NodeMCU
 upload_resetmethod          = nodemcu
-upload_port                 = COM5
+upload_port                 = COM3
 extra_scripts               = ${scripts_defaults.extra_scripts}
 lib_ldf_mode                = chain
 lib_compat_mode             = strict
--- a/tasmota/xsns_05_ds18x20.ino
+++ b/tasmota/xsns_05_ds18x20.ino
@@ -26,12 +26,11 @@
 #define XSNS_05              5
 
 //#define USE_DS18x20_RECONFIGURE    // When sensor is lost keep retrying or re-configure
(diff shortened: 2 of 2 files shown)


ascillato:
Hi,

Platformio.ini file shouldn´t be modified

platformio.ini
```diff
@@ -12,7 +12,7 @@
 [build_envs]
 default_envs =
 ; *** Uncomment by deleting ";" in the line(s) below to select version(s)
-;                tasmota
+                tasmota
```
ascillato, line 15: Platformio.ini file shouldn´t be modified

ascillato (changes requested):
Platformio.ini file shouldn´t be modified

platformio.ini
pharaocurt: Yes, this is right. Sorry, this is my 1st pull-request and i did not succeed in removing my platform.ini changes from pull-request. So i mentioned in the description to remove the platformio.ini changes.

ascillato: > I did not succeed in removing my platform.ini changes from pull-request.

you can:

1- go to https://github.com/arendst/Tasmota/blob/development/platformio.ini
2- click on the pen on the upper right
3- copy everything
4- go to your platformio.ini of your branch https://github.com/pharaocurt/Tasmota/blob/ds18x20-arithm-mean/platformio.ini
5- click on the pen on the upper right
6- paste everything
7- click on the green button to commit and you are done.

pharaocurt:
The arithmetic mean is calculated for mqtt json message and the resulting curve of temperture progression is more accurate and has lesser steps and reduced noise, especially with tempres 3.

pharaocurt:
The behaviour in Ds18x20EverySecond(void) in non parasitic powered sensor environment is a temperature reading every 2nd second, which is even, the odd second is used for converting the temperature in the ADC of DS18x20. So with minimal 10 sec telemetrie period, there will be 5 temperature readings in every telemetry period, for default 300 sec, there are 150 readings.

pharaocurt:
I assume, You are fighting for every byte in tasmota code. The new feature is only enabled by #define DS18x20_ARITH_MEAN.
I compiled 4 variants of main (plain vanilla) tasmota:
1. tasmota develop, when i forked the repository:
RAM:   [======    ]  55.0% (used 45076 bytes from 81920 bytes)
Flash: [======    ]  59.1% (used 605308 bytes from 1023984 bytes)
2. tasmota-arithm-mean, disabled new feature //#define DS18x20_ARITH_MEAN
RAM:   [======    ]  55.0% (used 45076 bytes from 81920 bytes)
Flash: [======    ]  59.1% (used 605276 bytes from 1023984 bytes)
3. tasmota-arithm-mean, enabled new feature #define DS18x20_ARITH_MEAN
RAM:   [======    ]  55.1% (used 45108 bytes from 81920 bytes)
Flash: [======    ]  59.1% (used 605420 bytes from 1023984 bytes)
4. tasmota-arithm-mean, enabled new feature #define DS18x20_ARITH_MEAN, uint16 numread
RAM:   [======    ]  55.1% (used 45108 bytes from 81920 bytes)
Flash: [======    ]  59.1% (used 605420 bytes from 1023984 bytes)
From previous message, i learned that uint8 for variable numread is not sufficient, if a user extents tele_period over 512 seconds.
The reduduction in Flash for 2nd compilation results from streamline the code, using break; in switch statement and avoiding copying same code for several DS18xtypes.
4.

arendst:
Implemented using command ``SetOption126 1``

Thx.

sfromis:
I suppose that this could make as much sense for many other sensors too ...

arendst:
Yes but it takes a lot of extra memory (and rewriting).

arendst:
And I'm still not convinced this is a correct way. You won't see quick temperature rises in JSON.

At least I won't use it.

sfromis:
I see your point. What I might have preferred instead would be a much shorter window than TelePeriod, which indeed will tend to be a very hard flattening of the curve. Having, say, a moving average over 4 samples could smooth a lot of jitter while still allowing quick changes. Averaging jitter could improve operation of rules-based thresholds to take actions based on the value.

arendst:
Thx.

Do you see so much jitter on temperature sensors? I never see it. They are just slow responding sensors with incrementing or decrementing values. I never see unexplainable jumps in temperature...

sfromis:
Not so much on ambient temperature sensors, where I may see 0.1 jitter, which is not a real issue. The type of sensors where I experience notable jitter are light sensors, sometimes jumping in up to a 10% band, even without perceivable illuminance differences. Of course real fluctuations from things moving around can also cause jitter here. CO2 sensors can also be jumping a bit, with nearby breath a possible factor.

Of course, temperature sensors which are not for ambient temperature, but to measure a "more local" process (could be around a computer or 3D printer) is a case where some curve-smoothening may make numbers more practical.

Still my phrase "as much sense" was meant to imply that I did not see temperature sensors as where smoothening is especially called for. Not trying to make a push for a lot of change :-)

PR #11557, opened 2021-04-02
Report serial buffer overruns with LOG_LEVEL_ERROR

eriktews:
## Description:

Should a serial buffer overrun be detected, it will be logged with LOG_LEVEL_ERROR and a hint how to fix this using the SerialBuffer command is shown as well.

## Checklist:


arendst:
Thx.

This won't be merged as the cost of code size increase doesn't weight agains functionality. If we had an unlimited amount of code we could tell the user all that's in the documentation ;-)

PR #11629, opened 2021-04-08
Update to iBeacon

btsimonh:
to use only MAC to recognise different beacons - this brings it into line with common use.

A further update would be nice to display major & minor in the web ui, not included here.

## Description:

## Checklist:

### DIFF
--- a/tasmota/xsns_52_esp32_ibeacon_ble.ino
+++ b/tasmota/xsns_52_esp32_ibeacon_ble.ino
@@ -337,7 +337,7 @@ uint32_t ibeacon_add(struct IBEACON *ib) {
   if (!strncmp(ib->MAC,"FFFF",4) || strncmp(ib->FACID,"00000000",8)) {
     for (uint32_t cnt=0;cnt<MAX_IBEACONS;cnt++) {
       if (ibeacons[cnt].FLAGS) {
-        if (!strncmp_P(ib->UID,PSTR("00000000000000000000000000000000"),32)) {
+//        if (!strncmp_P(ib->UID,PSTR("00000000000000000000000000000000"),32)) {
           if (!strncmp(ibeacons[cnt].MAC,ib->MAC,12)) {
             // exists
             strncpy(ibeacons[cnt].NAME,ib->NAME,sizeof(ibeacons[cnt].NAME));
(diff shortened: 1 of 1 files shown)


joba-1:
you mean uuid, not mac? mac could change (e.g. iphone does this) but beacon uuid stays the same

btsimonh:
@joba-1 - does the below make sense?
The issue it addresses is the intended use case for iBeacons 
(read the apple docs:
https://developer.apple.com/ibeacon/Getting-Started-with-iBeacon.pdf
page 2) 
where you may have 50 iBeacons all with the same UUID, but different major & minor.  the UUID's intent is to indicate that those beacons belong to a group. (e.g. 'warehouse 1').  Major & minor are intended to distinguish which beacon it is.  The issue was that the driver thought all beacons with the same UUID were ONE beacon, and so only reported one, not 50...
When a phone is used to send iBeacon broadcast, the 'new' issue is that the MAC changes every 5 minutes. 

This has two implications:
1/ you have to use 'BLAAddrFilter 1', else you won't hear the broadcasts (as they are on a random mac, and we by default don't listen to them because they change, and that blasts the BLE device list out of the park with Covid messages).
2/ when the mac changes, I believe currently you will get the old UUID/Major/Minor with RSSI:0 ONCE at the beacon timeout (probably after a few messages with the new MAC).

TODO: 
Modify the iBeacon driver to compare incoming UUID+Major+Minor, and if the same as one it has, replace the MAC silently.

btsimonh:
@joba-1 - are you in a position to test
https://github.com/btsimonh/Tasmota/tree/ibeaconUUIDMAJORMINOR
this should be the 'TODO' done.
Also adds major & minor in the web UI.

joba-1:
Will test and monitor requests/infos here and in the discord ble channel over the weekend. Will also try to make an esp32 an apple compliant iBeacon so I have full control over what gets sent when.

joba-1:
I tested current tasmota dev instead (I saw your PR #11647).
Works like I would expect. Great work :)
Android and iPhone BLE scanners show major/minor as I set them, without endian swap and tasmota shows the same.

One question left: 
I adapted the NimBLE iBeacon example for ESP32 to my testing needs and produced a LOT of iBeacons with it in tasmota (major/minor are used to show boot count and ESP32 sleeps for only 500ms...)
In case you want to try yourself and have a spare ESP32, you can use https://github.com/joba-1/NimbleBeacon.git
How is this list cleaned up?

PR #11634, opened 2021-04-08
Add command EnergyConfig

barbudor:
## Description:

Proposal to add command `EnergyConfig` which allows xdrv_03 to pass the command to the lower level xnrg_xx driver
This is a proposal regarding the discussion I have with Charles Hallard regarding xdrv_15_teleinfo.
I tried to add the index in a similar way to `SensorXX` but didn't find a simple way.

Tested with xnrg_20_dummy

## Checklist:

### DIFF
--- a/tasmota/xdrv_03_energy.ino
+++ b/tasmota/xdrv_03_energy.ino
@@ -43,7 +43,7 @@
 
 enum EnergyCommands {
   CMND_POWERCAL, CMND_VOLTAGECAL, CMND_CURRENTCAL, CMND_FREQUENCYCAL,
-  CMND_POWERSET, CMND_VOLTAGESET, CMND_CURRENTSET, CMND_FREQUENCYSET, CMND_MODULEADDRESS };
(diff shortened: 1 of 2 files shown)


vic42:
Definitely adds versatility to the energy monitoring system being able to pass device specific configuration options down to the xnrg driver.

arendst:
Yeah. Fine. I'll tweak it a bit but it's functionality will stay as you designed.

barbudor:
Many Thanks @arendst

Still I think it would be great to be able to specify the driver index like with `SensorXX` to insure the command is handled by the right driver
Would you have a suggestion on how to achieve that ?
(or may be you already did it part of your tweak ? haven't looks yet)

arendst:
No need to.

At restart the energy drivers are polled from lowest to highest index usinf function FUNC_PRE_INIT. The first to satisfy the requirements (correct GPIO) returns it's index to the master energy handler.

From then on ONLY the satisfied energy driver is accessed so no need for specifying the driver id.

Another fine feature of Tasmota dedicated drivers ;-)

hallard:
Awesome, thanks guys, I'll adapt my code and use this new feature.

PR #11723, opened 2021-04-14
Update Italian language

bovirus:
## Description:

Update Italian language.

## Checklist:


ascillato:
Hi @bovirus 

Thanks a lot for the translations.

As you have been very precise in the italian translation, I have a question to you about the reasoning behind the key:

```cpp
#define D_SHOW_MORE_OPTIONS "Altre opzioni"
```

I would like to understand why is this more suitable for the everyday use.

That key is used in the Initial WiFi Configuration Page.

image

and when clicking on it, you see

image

So, as this button is just showing more options that are hidden behind it, and it is not taking the user to another page, in English is more suitable the **_MORE OPTIONS_** label for this. Also, in Tasmota main configuration menu, there is the **_Other Options_** label that actually takes the user to another page.

I'm totally fine with the translation of this key, I just want to undestand this detail for further translations.

Thanks.

ascillato:
Checking the previous image I realize that the following keys should be different. The restore configuration is equal to reset configuration and in the previous menu it shows both buttons with the same label but they do different things.

```cpp
#define D_RESET_CONFIGURATION          "Ripristino impostazioni"

#define D_RESTORE_CONFIGURATION        "Ripristino impostazioni"
```

bovirus:
"Restore" is "Ripristino"
"Reset" should be "Ripristino impostazioni predefinite" but it's too long.

If you should have two different string "you can use "Impostazioni predefinite"

ascillato:
Great!!! Thanks for the fast answer 👍

ascillato:
I have just changed `D_RESET_CONFIGURATION` to `Impostazioni predefinite`

Thanks.

bovirus:
OK. Thanks.

bovirus:
I have an idea for next release

#define D_RESTORE_CONFIGURATION        "Carica impostazioni"

Restore Settings = Load setting from external file.
Then ifor me if it's true

Backup settings -> Save settings
Restore settings -> Load settings
 
For me is more clear.

ascillato:
Yes, I agree, it is more clear that way.

bovirus:
@ascillato 
I will change in Italian on next update.
I propose you to change them in English file on next update.

ascillato:
I have just changed those keys now in the italian translation file.

Backup settings -> "Salva impostazioni"
Restore settings -> "Carica impostazioni"

Thanks a lot for the help.

In English seems that Backup and Restore should be better understandable for new users or for non tech users, it feels like more accurate. What Tasmota is really doing is just a Save and Load, but seems better that in the English translation is better as it is now. In the other hand, in Italian, seems to be much much better to be "Salva impostazioni" and "Carica impostazioni" as you have proposed. These feels like having a more accurate meaning to what Tasmota is doing.

arendst:
No we won't!

There is a major difference between load/save and backup/restore. The first allows for any layout, the seconds just is supposed to restore a made backup. This is common data administration.

Discussion is futile.

Issue #5957, opened 2019-06-14
Device failed to switch


kenmercer:
My fishtank light is controlled by domoticz using a “every day at 6.30am” timer and the sonoff (Tasmota 6.4.1.17) failed to switch on this morning. It’s only happened once before but I do want a system configured so it is  100% reliable as I have other, more important, switches.

The domoticz log shows it switched on at 6:30:01
The sonoff consol log shows:
06:29:35 MQT: Attempting connection...
06:29:36 MQT: Connected
06:29:36 MQT: tele/MTSC/LWT = Online (retained)
06:29:36 MQT: cmnd/MTSC/POWER = 
06:29:58 MQT: domoticz/in = {"idx":45,"nvalue":0,"svalue":"","Battery":97,"RSSI":6}
06:30:04 MQT: Attempting connection...
06:30:06 MQT: Connected
06:30:06 MQT: tele/MTSC/LWT = Online (retained)
06:30:06 MQT: cmnd/MTSC/POWER = 
06:31:07 MQT: tele/MTSC/STATE = {"Time":"2019-06-15T06:31:07","Uptime":"3T12:32:57","Vcc":3.591,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":19,"POWER":"OFF","Wifi":{"AP":1,"SSId":"mySSId","BSSId":"9C:C1:72:0D:C6:69","Channel":1,"RSSI":66}}
06:31:07 MQT: domoticz/in = {"idx":46,"nvalue":0,"svalue":"24.6","Battery":97,"RSSI":6}

No one was up at 06.30 so what caused the normally reliable connection to break? If a connection does break is there anything I can do to ensure a switch command isn’t lost? Should I run a script to switch all my devices on/off every minute?
Thanks in advance.

ascillato2:
Please, address this to the [Tasmota Support Chat](https://discord.gg/Ks2Kzd4). The chat is a better and more dynamic channel for helping you. Github issues are meant for Tasmota Software Bug Reporting.

Please check the [Contributing Guideline and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md) and the [Support Guide](https://github.com/arendst/Sonoff-Tasmota/blob/development/SUPPORT.md).

Thanks.

---------------------------

### Support Information 
See [Wiki](https://github.com/arendst/Sonoff-Tasmota/wiki) for more information.
[...]

PR #11786, opened 2021-04-18
Demo of PWM 100% problem on ESP32 (for discussion)

barbudor:
## Description:

This is to demo and track a problem with ESP32's PWM not able to reach 100% duty cycle
The most annoying problem is not able to reach 0% on PWM_i inverted outputs which prevent powering off the output.

How to see the problem : 
- connect a LED to a GPIO with a 150..220 ohms resistor toward 3V3
- set the GPIO as PWM_i
- Try to power off the led
- Without this patch, the led will remain slighly on

Looking with a scope the GPIO signal shows that while we would expect the signal to be a continuous HI to reach 0% duty cycle for a PWM_i, a 1us pulse to LOW remains every cycle (1/1023th) but that is enough to slighly lite the LED.

The sample patch I provide force the pwm value to 1024 if the calculated value is 1023 which is enough to do the trick.

While this seems to fix the problem for ligthts (using color or channel), the same problem can be seen everywhere analogWrite() is used : PWM command, LedPWM, ....

I'm not sure to understand all the intrinsic around PWMrange and what impact modifying the whole PWMrange logic would have on Tuya dimmers for example

I can't tell this is a bug of the ESP32 HAL has I haven't found a definite description of what the ESP32 LedControl module should do. But this is clearly for now a difference with the ESP8266 that do not show this behavior.

## Checklist:
  - [X] The code change is tested and works with core ESP32 V.1.0.6

### DIFF
--- a/lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
+++ b/lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
@@ -89,6 +89,7 @@ inline void analogAttach(uint32_t pin, uint32_t channel) {
 inline void analogWrite(uint8_t pin, int val)
 {
   uint32_t channel = _analog_pin2chan(pin);
+  if ( val >> (_pwm_bit_num-1) ) ++val;
   ledcWrite(channel + PWM_CHANNEL_OFFSET, val);
 //  Serial.printf("write %d - %d\n",channel,val);
 }
--- a/tasmota/xdrv_04_light.ino
+++ b/tasmota/xdrv_04_light.ino
@@ -1967,12 +1967,7 @@ void LightSetOutputs(const uint16_t *cur_col_10) {
           cur_col = cur_col > 0 ? changeUIntScale(cur_col, 0, Settings.pwm_range, Light.pwm_min, Light.pwm_max) : 0;   // shrink to the range of pwm_min..pwm_max
         }
         if (!Settings.flag4.zerocross_dimmer) {
-          uint16_t pwm = bitRead(TasmotaGlobal.pwm_inverted, i) ? Settings.pwm_range - cur_col : cur_col;
-          #ifdef ESP32
-          if (pwm = Settings.pwm_range) pwm = Settings.pwm_range+1; // ESP32 full PWM is 1024 (1023 on ESP8266)
-          #endif
-          //AddLog(LOG_LEVEL_DEBUG_MORE, PSTR("LIT: PWM%d = %d"), i, pwm);
-          analogWrite(Pin(GPIO_PWM1, i), pwm);
+          analogWrite(Pin(GPIO_PWM1, i), bitRead(TasmotaGlobal.pwm_inverted, i) ? Settings.pwm_range - cur_col : cur_col);
         }
 #ifdef USE_PWM_DIMMER
         // Animate brightness LEDs to follow PWM dimmer brightness
--- a/tasmota/xdrv_04_light.ino
+++ b/tasmota/xdrv_04_light.ino
@@ -122,7 +122,7 @@
 \*********************************************************************************************/
 
 #define XDRV_04              4
-// #define DEBUG_LIGHT
+ #define DEBUG_LIGHT
 
 enum LightSchemes { LS_POWER, LS_WAKEUP, LS_CYCLEUP, LS_CYCLEDN, LS_RANDOM, LS_MAX };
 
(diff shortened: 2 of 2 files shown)


s-hadinger:
@arendst please put on hold. We are discussing this issue on Discord and I would favor a skipping 1 at mid range (512) rather than at end of range

Dlloydev:
I've recently added an AnalogWrite library that will resolve the missing full-on PWM issue. For any bit width, the missing step is included. For example, 8-bit resolution has 257 steps (zero-based) 0 is fully off, 1-255 is the PWM duty cycle range and 256 is fully on (high).

 [arduino-library-badge](https://www.ardu-badge.com/ESP32%20ESP32S2%20AnalogWrite)
GitHub link ...  [https://github.com/Dlloydev/ESP32-ESP32S2-AnalogWrite](https://github.com/Dlloydev/ESP32-ESP32S2-AnalogWrite)

barbudor:
@Dlloydev This is already how the ESP32 analogWrite(), based on lecdWrite(), is behaving whereas the ESP8266 has 100% duty cycle at 255.
I feel our position is that 0..100% should be matched with 0..255 (or 0..1023) (like the ESP8266) and not 0..256.

Dlloydev:
Actually, I agree that 255 should be 100% full on. I'm new with the ESP32, but it seems that the hardware will support this as I briefly looked at their technical reference manual.

hallard:
@barbudor discovered this issue on new Denky D4 board which as a Common Cathode RGB Led  (so that is why it's PWM inverted). The fact is with common Anode RGB led, setting PWM to 255 (or 1023) even if it's not really full illuminated it could be 99.6% or 99.9% so I can imagine no human eye can see the difference (so the issue)
I already had this bug on ESP8266 long time ago where I needed an `analogWrite(256)` to light off the led and looks like it has been fixed since but also explained on [#5957](https://github.com/esp8266/Arduino/issues/5957)

The deal is that `analogWrite` is not native to ESP32 and has been added in compatibility module in tasmota so opening this issue to ESP32 team won't be supported. The new api is `ledcWrite` ( atfer setup of `ledcAttachPin()` and `ledcSetup()`) and even the official demo PWM sample of ESP32 reference this issue, see comments on sample [code](https://github.com/espressif/arduino-esp32/blob/404a31f4455d2459dad0b6533b7bfa70347e0044/libraries/ESP32/examples/AnalogOut/ledcWrite_RGB/ledcWrite_RGB.ino#L22) so my guess we need to live with that one

As @barbudor state the easier fix is that if we are in PWM inverted mode and PWM value is the latest of the range, add one or set it to 1024 on ESP32. But I'll try to open a issue on ESP32 core team, worth trying to fix at the source.

barbudor:
As you pointed, tasmota is not using Arduino for analogWrite on ESP32 but a compatibility layer.
I've already adapted the code but was unable to publish yesterday due to ISP problems.
I'll push it today after a last check with the scope.

barbudor:
## TL;DR;
- Arduino doc say between 0 (always off) and 255 (always on)
- ESP8266 : 0..100% is 0..2^N-1 (compliant)
- ESP32 : 0..100% is 0..2^N (not compliant)

Best place to adapt is probably in libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
Draft PR updated following @s-hadinger for performing the fix at mid-range
Will do a final check tonight with a scope to be 100% sure

## Details

### ESP8266
The esp8266/Arduino [code](https://github.com/esp8266/Arduino/blob/cfbbd6f22db6b9d76808424ea1a813b02ba9c5ab/cores/esp8266/core_esp8266_wiring_pwm.cpp#L62-L64) is referring to the Arduino official doc for [`analogWrite()`](https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/):
> value: the duty cycle: between 0 (always off) and 255 (always on)

The code defines the `high` and `low` duration of the PWM signal:
```C++
  uint32_t high = (analogPeriod * val) / analogScale;
  uint32_t low = analogPeriod - high;
```
Where `analogScale` is the value  provided with `analogWriteRange()` which makes:
- val = 0            => high = 0 and low = analogPeriod => signal is low 100%
- val = analogScale  => high = analogPeriod and low = 0 => signal is high 100%

### ESP32
On the ESP32 the hardware PWM is based on a 
> counter which will count up to the value specified in LEDC_HSTIMERx_DUTY_RES. An overflow interrupt will be generated once the counting value reaches 2^LEDC_HSTIMERx_DUTY_RES − 1, at which point the counter restarts counting from 0
> A channel takes the 20-bit value from the counter of the selected high-speed timer and compares it to a set of two values in order to set the channel output. The first value it is compared to is the content of LEDC_HPOINT_HSCHn;
if these two match, the output will be latched high. The second value is the sum of LEDC_HPOINT_HSCHn and LEDC_DUTY_HSCHn[24..4]. When this value is reached, the output is latched low

_(ESP32 Technical Reference Manual section 14.2.2 and 14.2.3)_

The ESP32 immplementation is in https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-ledc.c
#### Initialisation
- LEDC_HSTIMERx_DUTY_RES is initialized with the `analogWiteRange()`
- LEDC_HPOINT_HSCHn is initialized at 0
- define the idle_level to LOW (pin value when PWM is stopped)
#### analogWrite() => ledcWrite()
- if duty == 0 : the PWM is stopped and pin is set LOW (idle_level)
- if duty != 0 : duty is written to LEDC_DUTY_HSCHn
So with duty != 0, the signal is HIGH from 0 to duty-1 and low from duty to DUTY_RES (range, typically 1023)
When DUTY_RES=1023 and duty=1023, the signal is HIGH from 0 to 1022 and LOW for 1023. Thus the ratio is 1023/1024.
When DUTY_RES=1023 and duty=1024, the signal is always HIGH because the counter never reach 1024.

s-hadinger:
Thanks @barbudor
Your last fix looks good to me, simple and slick. All that I love.

@arendst I suggest you merge.

PR #11814, opened 2021-04-20
add variant ESP32-solo1

Jason2866:
## Description:

as official build variant.

## Checklist:


arendst:
I guess we can get rid of the ``tasmota32solo1`` environment? See above.

Jason2866:
The extra env for solo1 is needed since the IDF 3.3 libs are compiled for unicore use.
This is done in sdkconfig from IDF
So the code base is the same BUT generated libs used in Arduino are different for ESP32 and solo1

PR #11865, opened 2021-04-24
Lvgl update

gemu2015:
## Description:

remove all global vars from renderer to callbacks
LVGL glue now takes renderer class. (no separate LVGL udisp needed)
this allows  concurrent action of LVGL and display text. (all renderer display drivers, not only udisp)

verified with Berry examples. however @s-hadinger  please verify commit

(touch support in preparation)

## Checklist:

### DIFF
--- a/lib/lib_display/Display_Renderer-gemu-1.0/src/renderer.cpp
+++ b/lib/lib_display/Display_Renderer-gemu-1.0/src/renderer.cpp
@@ -153,6 +153,13 @@ void Renderer::DrawStringAt(int16_t x, int16_t y, const char* text, uint16_t col
     int refcolumn = x;
     sFONT *xfont = selected_font;
 
+/*
+    if (font == 5 && !drawmode) {
+      // clear bckground
+      int16_t x1,y1;
+      uint16_t w,h;
+      Adafruit_GFX::getTextBounds(text, 0, 0, &x1, &y1, &w, &h);
+    }*/
 #ifndef USE_EPD_FONTS
     font=0;
 #endif
@@ -297,13 +304,15 @@ void Renderer::setTextFont(uint8_t f) {
 
 
 void Renderer::SetRamfont(uint8_t *font) {
+
   ramfont = (GFXfont*)font;
-  uint32_t bitmap_offset = (uint32_t)ramfont->bitmap;
-  uint32_t glyph_offset = (uint32_t)ramfont->glyph;
+  if (font) {
+    uint32_t bitmap_offset = (uint32_t)ramfont->bitmap;
+    uint32_t glyph_offset = (uint32_t)ramfont->glyph;
 
-  ramfont->bitmap = (uint8_t*)((uint32_t)font + bitmap_offset);
-  ramfont->glyph = (GFXglyph*)((uint32_t)font + glyph_offset);
-
+    ramfont->bitmap = (uint8_t*)((uint32_t)font + bitmap_offset);
+    ramfont->glyph = (GFXglyph*)((uint32_t)font + glyph_offset);
+  }
   setFont(ramfont);
 }
 
@@ -586,6 +595,33 @@ void Renderer::setScrollMargins(uint16_t top, uint16_t bottom) {
 }
 void Renderer::scrollTo(uint16_t y) {
 
+}
+
+void Renderer::SetPwrCB(pwr_cb cb) {
+
+}
+void Renderer::SetDimCB(dim_cb cb) {
+
+}
+
+uint16_t Renderer::fgcol(void) {
+  return 0;
+}
+uint16_t Renderer::bgcol(void) {
+  return 0;
+}
+int8_t Renderer::color_type(void) {
+ return 0;
+}
+
+void Renderer::Splash(void) {
+
+}
+
+const char dname[1] = {0};
+
+char *Renderer::devname(void) {
+  return (char*)dname;
 }
 
 void VButton::xdrawButton(bool inverted) {
--- a/lib/lib_display/Display_Renderer-gemu-1.0/src/renderer.h
+++ b/lib/lib_display/Display_Renderer-gemu-1.0/src/renderer.h
@@ -16,6 +16,8 @@
 // a. in class GFX setCursor,setTextSize => virtual
 // b. textcolor,textbgcolor => public;
 
+typedef void (*pwr_cb)(uint8_t);
+typedef void (*dim_cb)(uint8_t);
 
 class Renderer : public Adafruit_GFX {
   //Paint(unsigned char* image, int width, int height);
@@ -42,12 +44,21 @@ public:
   virtual void setScrollMargins(uint16_t top, uint16_t bottom);
   virtual void scrollTo(uint16_t y);
   virtual void TS_RotConvert(int16_t *x, int16_t *y);
+  virtual void SetPwrCB(pwr_cb cb);
+  virtual void SetDimCB(dim_cb cb);
+  virtual uint16_t fgcol(void);
+  virtual uint16_t bgcol(void);
+  virtual int8_t color_type(void);
+  virtual void Splash(void);
+  virtual char *devname(void);
 
   void setDrawMode(uint8_t mode);
   uint8_t drawmode;
   virtual void FastString(uint16_t x,uint16_t y,uint16_t tcolor, const char* str);
   void setTextSize(uint8_t s);
   virtual uint8_t *allocate_framebuffer(uint32_t size);
+  pwr_cb pwr_cbp = 0;
+  dim_cb dim_cbp = 0;
 private:
   void DrawCharAt(int16_t x, int16_t y, char ascii_char,int16_t colored);
   inline void drawFastVLineInternal(int16_t x, int16_t y, int16_t h, uint16_t color) __attribute__((always_inline));
--- a/lib/lib_display/UDisplay/uDisplay.cpp
+++ b/lib/lib_display/UDisplay/uDisplay.cpp
@@ -31,6 +31,17 @@ uint16_t uDisplay::GetColorFromIndex(uint8_t index) {
   return udisp_colors[index];
 }
 
+uint16_t uDisplay::fgcol(void) {
+  return fg_col;
+}
+uint16_t uDisplay::bgcol(void) {
+  return bg_col;
+}
+
+int8_t uDisplay::color_type(void) {
+  return col_type;
+}
+
 
 uDisplay::~uDisplay(void) {
   if (framebuffer) {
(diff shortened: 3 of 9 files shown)


s-hadinger:
Thanks this is awesome. I didn't expect so much.

Currently it requires:
```
#define USE_LVGL
#define USE_DISPLAY
#define USE_XPT2046
```
to compile successfully. Is it the only touchscreen supported? Or would it make sense to have a non-touchscreen version?

gemu2015:
it supports several I2C (luckily the I2C touch chips have a common command set) and the SPI touch controllers.  
if the descriptor does not contain a touch definition touch is not initialized.
you may enable both I2C and SPI touch in compiler. the descriptor selects which to use.

however i expect problems with SPI BUS.
since there are probably up to 3 devices on the bus (display, touch, SD card)
and we have no mutex in place. this would require some work to protect all these calls (including file system)

s-hadinger:
Excellent! I submitted a PR to simplify configuration from Berry.

My compilation options are finally:
```
#define USE_LVGL
#define USE_DISPLAY
#define USE_XPT2046
#define USE_UNIVERSAL_DISPLAY
#undef USE_DISPLAY_MODES1TO5
```

s-hadinger:
One last thing, I get a red screen displayed just before LVGL displays. How to remove it?

gemu2015:
red screen ?

the initial display (index) fill color is given in the descriptor and normally black.

s-hadinger:
Ah yes, I set it to red in the template to make some tests...

gemu2015:
according to your pr autoconfig.
this is already in place in display driver 17.   * char is replaced by the tasmota config pin.
i generally avoid calls from driver to non os parts.

s-hadinger:
Hmmm. This is what I thought initially but it didn't seem to work when I first tried. Ok let me undo the PR and keep only the space delimiter.

s-hadinger:
I'm just missing a last piece. Which SPI_CS is being used? I configured `SPI_CS` but maybe you are looking for another GPIO?

Here is my ouput from putting `*` in all SPI gpios:

```
xs : 320
ys : 240
bpp: 16
Nr. : 1
CS  : -1
CLK : 18
MOSI: 23
DC  : 27
BPAN: -1
RES : 33
MISO: 19
SPED: 40000000
Pixels: 16
SaMode: 255
opts: 2c,ff,ff
SetAddr : 2a,2b,2c
Rot 0: 36,8 - 0 - 0
cmd, args ef, 3 03 80 02
cmd, args cf, 3 00 c1 30
cmd, args ed, 4 64 03 12 81
cmd, args e8, 3 85 00 78
cmd, args cb, 5 39 2c 00 34 02
cmd, args f7, 1 20
cmd, args ea, 2 00 00
cmd, args c0, 1 23
cmd, args c1, 1 10
cmd, args c5, 2 3e 28
cmd, args c7, 1 86
cmd, args 36, 1 48
cmd, args 37, 1 00
cmd, args 3a, 1 55
cmd, args b1, 2 00 18
cmd, args b6, 3 08 82 27
cmd, args f2, 1 00
cmd, args 26, 1 01
cmd, args e0, 15 0f 31 2b 0c 0e 08 4e f1 37 07 10 03 0e 09 00
cmd, args e1, 15 00 0e 14 03 11 07 31 c1 48 08 0f 0c 31 36 0f
cmd, args 11, 0
cmd, args 29, 0
Dsp Init complete
```

gemu2015:
this was a recent change,  cs as input parameter and i coded case 1 instead of case '1'
in driver 17

should look like this
if (cs < 0) {
        switch (*cp) {
          case '1':
            cs = Pin(GPIO_SPI_CS);
            break;
          case '2':
            cs = Pin(GPIO_SPI_CS, 1);
            break;
          default:
            cs = Pin(GPIO_SSPI_CS);
            break;
        }
      } 

i am busy currently with private things, can't do a pr currently

gemu2015:
should be fixed by pr.

a remark to e-paper displays:
i just finished epd47 16 gray level display. it has 960x540 pixel which would cost too much ram.
i introduced a variable :B,XX in descriptor for flush buffer lines. it defaults to zero and then uses 60 lines from define.
for epd47 the non udisp driver sets this to 10 being 20k of RAM, slow but doesn't help
if ram is an issue we could also use PSRAM as long we do not use DMA.

e-papers need a frequent refresh, either partial or full. this can not be done with display flushing. it would happen to often.
so for berry you would need to have a call for that.
in display text it  is [I] for full and [i] for partial refresh.
renderer->DisplayInit(mode, 0, 0, 0);
mode = 1 => partial
mode = 2 => full

s-hadinger:
Thanks, I confirm it's now working. I will just push a short PR to allow space in addition to `\n` to change section.

I don't have any experience in e-paper, and I don't understand why they need frequent refresh. Aren't they supposed to keep the image on the screen almost forever?

s-hadinger:
Thanks again for your awesome work!!!

Do you have plans to introduce DMA for SPI transfer? I suspect this could almost double the pixel rate to ILI9342.

gemu2015:
e-papers don't need frequent refresh, only when new content shall be shown. however LVGL does update in slices, so i can not use this update for paper screen refresh. i did not find a callback that signals something like "screen done"
and even partial refresh e-papers from time to time need a full refresh to clean the surface from remaining washed out pixels

DMA would  speed things up of course. first i will optimize the existing code for speed.
SPI DMA has some side effects. Other devices on the bus have to recognize this. SD card and Touch drivers have to be adapted.
will check the net if someone has already thought about that. (similar problem with missing mutexes) not sure if its worth the effort.

s-hadinger:
Ah ok, I get it. Indeed slices updates are not suited for e-paper.

With DMA, indeed you can't mix DMA and non-DMA so it needs to be very careful.
I found 2 libs that seem to succeed in doing SPI DMA on ESP32:
- TFT\_eSPI: https://github.com/Bodmer/TFT_eSPI/blob/master/Processors/TFT_eSPI_ESP32.c
- lvgl\_port\_esp32 (although I saw weird behavior with ILI9341: lavish colors) https://github.com/lvgl/lvgl_esp32_drivers/blob/master/lvgl_tft/disp_spi.c

gemu2015:
in the meantime i optimized significantly (see latest pr, still no DMA)
as a compromise could use DMA and wait for completion (no conflict with other drivers) and still faster than N single SPI cmds.

s-hadinger:
Thanks. I added speed measure:

Before:
```
20:38:56.822 LVG: Refreshed 19200 pixels in 61 ms (314 pix/ms)
20:38:56.882 LVG: Refreshed 19200 pixels in 56 ms (342 pix/ms)
20:38:56.945 LVG: Refreshed 19200 pixels in 58 ms (331 pix/ms)
20:38:57.018 LVG: Refreshed 19200 pixels in 56 ms (342 pix/ms)
```

Now
```
20:43:24.152 LVG: Refreshed 19200 pixels in 18 ms (1066 pix/ms)
20:43:24.184 LVG: Refreshed 19200 pixels in 21 ms (914 pix/ms)
20:43:24.209 LVG: Refreshed 19200 pixels in 21 ms (914 pix/ms)
20:43:24.250 LVG: Refreshed 19200 pixels in 21 ms (914 pix/ms)
```

Considering the 40MHz SPI bus, the theoretical max should be around 2000 pix/ms (taking ~20 bits per pixel, although I don't know the SPI overhead). Currenlty a full refresh takes 80ms + ~40ms of compute. Using DMA we should be able to go down to ~50ms (with double buffer, one buffer is copied to SPI while the next slice is calculated). That would mean 50Hz :)

gemu2015:
ok, initial DMA works. (see my universal fork)
very fast, however strange effects , distorted Text, noisy pixels. reducing frequency does not help
needs further research

enable with
:B,60,1
60 = number of display lines (in DMA use 2 * 30)
1 = dma enable

gemu2015:
tried all kinds of mods of the spi transaction modes, with no success.
made a pr with current status.
will check with a completely different display (ST7789) this afternoon.

s-hadinger:
What do you mean with no success? Distorted display?

gemu2015:
yes distorted display. as if there were some pixels slightly wrong (bit errors)
no sharp lines, characters 
no success in getting a sharp image.

s-hadinger:
Confirmed, some bits are off but it's hard to tell why. Some colors are good, some are not
IMG_1149

I also put an explicit measure, to see the actual pix/ms:
```
11:30:17.114 LVG: Refreshed 19200 pixels in 9 ms (2133 pix/ms)
11:30:17.125 LVG: Refreshed 19200 pixels in 8 ms (2400 pix/ms)
11:30:17.135 LVG: Refreshed 19200 pixels in 8 ms (2400 pix/ms)
11:30:17.160 LVG: Refreshed 19200 pixels in 9 ms (2133 pix/ms)
```

s-hadinger:
Oh wait, it's simply inverted bytes...

gemu2015:
Yes, conformed. now it works

s-hadinger:
I believe you needed the commented line:
```
    for (uint32_t i = 0; i < len; i++) (image[i] = image[i] << 8 | image[i] >> 8);
```

s-hadinger:
The downside is that adding the line above kills the performance back to 1000 pix/ms

s-hadinger:
I'm trying with `#define LV_COLOR_16_SWAP   1` in `tasmota_lv_conf.h`

s-hadinger:
More tests (I'm counting the overall time spent, not only the DMA transfer)
- with software byte inversion: 1600-1900 pix/ms
- with LV_COLOR_16_SWAP 1: 2100-2400 pix/ms

So we definitely need to enable `#define LV_COLOR_16_SWAP 1`

gemu2015:
this is perfect!

do you have issues with your SPI touch controller now ?

i only have I2C touch controllers.

s-hadinger:
I don't have any SPI touch controller, the only one I have is M5Stack Core2 (I2C)

s-hadinger:
I detected a small issue, the first `fillScreen(bg_col);` run at the beginning (in red in my case) is very slow, like if it was software SPI. That's not blocking for now.

s-hadinger:
Now that DMA is straight, non-DMA is reversed and I don't see any option in `spiWritePixelsNL()` to reverse byte order... Rats.

I guess we need `for (uint32_t i = 0; i < len; i++) (image[i] = image[i] << 8 | image[i] >> 8);` for non-DMA version now. This is where it will have the least impact.

s-hadinger:
Would it be possible to not clear the screen if section `:S` is absent?

gemu2015:
yes i also see this problem.
but we have to know who the caller is. display text sends normal 16 bit colors.
i suggest to use the no longer need flag first in push colors to identify the caller 
displaytext uses true, so LvGL may use false.
and we may then invert on LVGL non DMA calls

gemu2015:
why not set the background index color to -1, this could be easily implemented.
 normally index colors are positive

s-hadinger:
Ok, I need to go now. I submitted https://github.com/arendst/Tasmota/pull/11897
So you can add the test on top of it.

PR #11876, opened 2021-04-25
lvgl monochrome support

gemu2015:
## Description:

LVGL support for monochrome displays (also I2C, and currently only udisp drivers)

due to the fact that LVGL does not support dynamic pixel depth (only compile switches)
we use color display frames and convert to BW during flush

## Checklist:


s-hadinger:
Thanks. That's a smart trick. I will try it. 

Alternatively we can explore compile time option to of LVGL in monochrome mode. That would save ram and CPU. But this will be more useful to have a universal version that yiu can tune at runtime

Great work!

PR #11951, opened 2021-04-29
Adding support for SAS Token Generation

ksaye:
Azure IoT Hub uses a Sha256 has as opposed to a password.  This change enables the admin to add the 'Preshared Key' and lets Tasmota calculate the SAS token, vs the original changes where the end user had to calculate the SAS Token.

## Description:

## Checklist:

### DIFF
--- a/tasmota/xdrv_02_mqtt.ino
+++ b/tasmota/xdrv_02_mqtt.ino
@@ -24,7 +24,7 @@
 #endif
 
 #ifdef USE_MQTT_AZURE_IOT
-#include <bearssl\bearssl.h>
+#include <t_bearssl.h>
 #include <base64.hpp>
 #include <JsonParser.h>
 #undef  MQTT_PORT
--- a/tasmota/xdrv_02_mqtt.ino
+++ b/tasmota/xdrv_02_mqtt.ino
@@ -24,6 +24,8 @@
 #endif
 
 #ifdef USE_MQTT_AZURE_IOT
+#include <bearssl\bearssl.h>
+#include <base64.hpp>
 #include <JsonParser.h>
 #undef  MQTT_PORT
 #define MQTT_PORT         8883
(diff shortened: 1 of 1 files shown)


ascillato:
Hi,

Why you add that last change in 5b90754 ? Seems that was ok as it was before that wasn't including libraries that were not used.

tasmota/xdrv_02_mqtt.ino
```diff
@@ -23,9 +23,9 @@
 #define MQTT_WIFI_CLIENT_TIMEOUT   200    // Wifi TCP connection timeout (default is 5000 mSec)
 #endif
 
-#ifdef USE_MQTT_AZURE_IOT
 #include <bearssl\bearssl.h>
 #include <base64.hpp>
+#ifdef USE_MQTT_AZURE_IOT
```
ascillato, line 28: Why is this last change used for?

ascillato (changes requested):
Why is this last change used for?

Seems that was ok as it was before that wasn't including libraries that were not used.

tasmota/xdrv_02_mqtt.ino
ksaye: I received the failed to build message when they were not global, as the method azurePreSharedKeytoSASToken was global and not under the #ifdef

tasmota/xdrv_02_mqtt.ino
```diff
@@ -24,6 +24,8 @@
 #endif
 
 #ifdef USE_MQTT_AZURE_IOT
+#include <bearssl\bearssl.h>
```
ascillato, line 27: What do you think of using instead the library that Tasmota already has?

```cpp
#include <t_bearssl.h>
```

ascillato:
I have tested your PR and works fine. Great Job!

What do you think of using the bearssl library that already Tasmota has?

So, instead of using:
```cpp
#include <bearssl\bearssl.h>
```
Change to use:
```cpp
#include <t_bearssl.h>
```

ascillato2 (approved):

ascillato2:
I end up changing that to t_bearssl.h

PR #11968, opened 2021-04-30
Add firmware for Tube's Zigbee EFR32 EZSP 6.7.9

s-hadinger:
## Checklist:


MattWestb:
Great that you have getting both the EFR32 and the TI Ethernet coordinators !!

Im interested is you if getting better stability with the tube0013 EFR32MG21 then the Sonoff one.
The latest is pointing that the very fast battery draining and other problems is coming from bad hardware of the SM-011 module and also the new USB-Zigbee 3 stick is having the same and more worse problems.

I have running one LIDL (tuya) LED strip controller (its one EFR32MG21) with EZSP 6.9.2.0 on Wemos D1 mini (the same as was doing the first test with tasmota ZBB) and tasmota connected to ZHA and is very stable so the problems is not coming from tasmota part !!

s-hadinger:
It's been running for 24 hours yet, but it looks very stable. I ran the CC2652P2 for a week and didn't see the battery drain problem. But I don't have it anymore with the ZBBridge either.

MattWestb:
Great !!
The EZSP 6.7.8.0 is fixing one verified bug in the NCP that was hitting Sonoff ZBB so very likely its fixing it in your hardware.

Some (ZHA) user still getting very fast battery draining (48 hours) but no user / devs have capture it in on sniffing.
I think the last apart is hardware related (miss tuned RF) that is partly fixed (work around by tuning the RF in the firmware and not in the chip that is the standard) in the factory EZSP 6.7.9.0 but is not working well on all devices then its not so god then have SED connected direct to the coordinator but working well then have on router as there parent.

I have recommended the EZSP 6.7.9.0 for ZHA user for getting the ZBB working better and most is working OK.

I think tube0013 with on Silabs module that is calibrated from the factory is having much better stability in the real world.

Im being exited what is coming out from you new hardware investments :-))

s-hadinger:
Before I forgte, I didn't change the settings for Tube's EFR32 coordinator, so you don't benefit yet from increased capacity. For example max devices is still set to 32, whereas the EFR32 from this device can handle 64.

MattWestb:
I think its not helping putting up some of the parameters then was reading in Silabs forum that commercial devices is overriding the settings and the network is using the "normal" ones and ignoring the "hot tuned" parameters. If using them you need having all devices in your network with the same parameters that is not possible with commercial devices.

Also its better having some routers in the network so it can benefit for the mesh network functionality and SED is not leaving if the coordinator is off line.

Did you finding where putting the HFXO if you is doing on Sonoff FW ?
In HW configurator and CMU you is having it.

PR #11975, opened 2021-05-01
feature: apply/execute rules after firmware flash

orcema:
## Description:

After flashing a firmware i alwyas had to reapply the rules and activate them manually. When you have a lot of devices this can take some time and you have to remember the rules to be applied.
Whit this pull request you can apply the rules in an automatic manner after flashing the firmware.

Rules are implemented within the file user_config_override.h as last statement. Up to 9 rules can be configured using the statement  **_#define defaultSetting_01_** up to **_#define defaultSetting_09_**.

Here below a sample:

>   #ifdef APPLY_DEVICE_GARAGE
>     #define PROJECT "DRAGON"
>     #define FIRMWARE_BASIC
>     #define MODULE USER_MODULE
>     #define USER_TEMPLATE "{\"NAME\":\"CustomDragon\",\"GPIO\":[0,0,0,0,22,23,0,0,0,21,9,0,0],\"FLAG\":0,\"BASE\":18}"
>     #define defaultSetting_01 "Rule1 \
>     ON Power2#state=1 DO RuleTimer2 2 ENDON ON Rules#Timer=2 DO Power2 off ENDON  \
>     ON Power3#state=1 DO RuleTimer3 2 ENDON ON Rules#Timer=3 DO Power3 off ENDON"
>     #define defaultSetting_02 "Rule1 ON"
>  #endif

Then in you platform.ini file you define the usage like this:

> [env:DRAGON_PorteGarage]
> board = esp01_1m
> build_flags = ${common.build_flags} 
>   -DAPPLY_DEVICE_GARAGE

## Checklist:
  - [ x] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.9

### DIFF
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -738,8 +738,10 @@ struct {
   uint8_t       shd_leading_edge;          // F5B
   uint16_t      shd_warmup_brightness;     // F5C
   uint8_t       shd_warmup_time;           // F5E
+  uint8_t       rulesAfterFlash;           // F5F - flag for running rules only once after firmare flash
 
-  uint8_t       free_f5f[65];              // F5F - Decrement if adding new Setting variables just above and below
+  uint8_t       free_f5e[64];              // F60 - Decrement if adding new Setting variables just above and below
+
 
   // Only 32 bit boundary variables below
 
@@ -762,6 +764,8 @@ struct {
   uint32_t      i2c_drivers[3];            // FEC  I2cDriver
   uint32_t      cfg_timestamp;             // FF8
   uint32_t      cfg_crc32;                 // FFC
+
+
 } Settings;
 
 typedef struct {
--- a/tasmota/tasmota.ino
+++ b/tasmota/tasmota.ino
@@ -391,6 +391,43 @@ void setup(void) {
 #endif
 
   TasmotaGlobal.rules_flag.system_init = 1;
+  Serial.println("rulesAfterFlash status:"+Settings.rulesAfterFlash);
+  if (!Settings.rulesAfterFlash){
+      // commands to execute at startup
+      #ifdef defaultSetting_01 
+        ExecuteCommand(defaultSetting_01, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_02 
+        ExecuteCommand(defaultSetting_02, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_03 
+        ExecuteCommand(defaultSetting_03, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_04 
+        ExecuteCommand(defaultSetting_04, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_05 
+        ExecuteCommand(defaultSetting_05, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_06 
+        ExecuteCommand(defaultSetting_06, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_07 
+        ExecuteCommand(defaultSetting_07, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_08 
+        ExecuteCommand(defaultSetting_08, SRC_IGNORE);
+      #endif
+      #ifdef defaultSetting_09 
+        ExecuteCommand(defaultSetting_09, SRC_IGNORE);
+      #endif
+      Settings.rulesAfterFlash=true;
+      SettingsSave(0);
+      AddLog(LOG_LEVEL_INFO, PSTR(D_LOG_RULES " " D_RULE_SETUP_AFTER_FLASHING));
+  }else{
+      Serial.println("\nskipping - Rules setup after flashing firmware\n");
+  }
+
 }
 
 void BacklogLoop(void) {
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -738,6 +738,8 @@
 #define D_LOG_TCP "TCP: "          // TCP bridge
 #define D_LOG_BERRY "BRY: "        // Berry scripting language
 #define D_LOG_LVGL "LVG: "         // LVGL graphics engine
+#define D_LOG_RULES "RUL:"
+
 
 /********************************************************************************************/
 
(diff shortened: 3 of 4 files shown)


sfromis:
Looks like you'll have a boot loop if any command in those rules causes a restart or some failure.

barbudor:
@sfromis, there's a flag to prevent re-executing those commands on next loop.
This is somehow duplicating the new CMND feature field in the template.
As the template can be burned in the binary this PR would provide similar usage.

sfromis:
@barbudor, if you mean `Settings.rulesAfterFlash` that flag is only set after all commands successfully executed. Will not hinder boot loops if not. While the title here implies "rule setup only", the code suggests any command, meaning that restarts/failures are plausible.

barbudor:
@sfromis AFAIK The reboot would not happen within `ExecuteCommand` but only when Tasmota enters  its loop.
Just like with a `backlog`, the reboot occurs only at the end.
Still executing multiple `ExecuteCommand()` like this in a raw is probably not the right way of doing it (Backlog isn't doing like that).

sfromis:
To me it looks like ExecuteCommand dives straight into CommandHandler to execute. Anyway, scheduled restarts is not the "most risky" issue, but failures. Tasmota has to be able to survive commands crashing, testable by running the command `Restart -1` to force an exception.

tasmota/tasmota.ino
```diff
@@ -391,6 +391,43 @@ void setup(void) {
...
+      #ifdef defaultSetting_09 
+        ExecuteCommand(defaultSetting_09, SRC_IGNORE);
+      #endif
+      Settings.rulesAfterFlash=true;
+      SettingsSave(0);
+      AddLog(LOG_LEVEL_INFO, PSTR(D_LOG_RULES " " D_RULE_SETUP_AFTER_FLASHING));
+  }else{
+      Serial.println("\nskipping - Rules setup after flashing firmware\n");
+  }
+
```
arendst, line 430: The above implemntation is a no-go. Performing ExecuteCommand in sequence without any loop access is bound to exception or watchdog. That's where backlog kicks in.

tasmota/settings.h
```diff
@@ -738,8 +738,10 @@ struct {
   uint8_t       shd_leading_edge;          // F5B
   uint16_t      shd_warmup_brightness;     // F5C
   uint8_t       shd_warmup_time;           // F5E
+  uint8_t       rulesAfterFlash;           // F5F - flag for running rules only once after firmare flash
+
```
arendst, line 742: If you want only execution once then there is no need for this. If a user does initial installation of formware the settings are written completely. The next restarts will not execute settings init again so the commands will not execute either.

arendst:
Just scanned over the code and found major issues easily solvable by a total rethink of this implemntation.

As the OP just wants to execute rules once at firmware install time overwriting the settings it should be implemented like the USER_TEMPLATE define and therefor at the end of the initial settings configuration.

I'll have a look at this option.

The current PR won't be merged anyway.

tasmota/settings.h
orcema: without the **rulesAfterFlash** at each reboot the setup function is processed and this will involve execution of 
**SettingsSave(0)**, so  it will write the flash at each reboot, or not ?

The execute command for applying rules is a fast process why should it trigger the watchdog ?

What is exactly the new CMND feature field in the template ?

tasmota/tasmota.ino
orcema: without the rulesAfterFlash at each reboot the setup function is processed and this will involve execution of
SettingsSave(0), so it will write the flash at each reboot, or not ?

The execute command for applying rules is a fast process why should it trigger the watchdog ?

What is exactly the new CMND feature field in the template ?

sfromis: CMND in the template is a backlog-style command sequence executed when the template is applied. #11788

orcema: If my understanding is right, the CMND feature is intended for actions execution or activate/deactivate options but how would I setup a timer based rule triggered by GPIO states ?

arendst:
Implemented my way.

You now have three different defines for each rule and one define for a sequence of backlog commands.

Use ``define USER_RULE1``, ``define USER_RULE2`` and/or ``define USER_RULE3`` in addition to ``define USER_BACKLOG``.

In your case it would be something like
```
#define USER_RULE1 "ON Power2#state=1 DO RuleTimer2 2 ENDON ON Rules#Timer=2 DO Power2 off ENDON ON Power3#state=1 DO RuleTimer3 2 ENDON ON Rules#Timer=3 DO Power3 off ENDON"
#define USER_BACKLOG "Rule1 1"
```

sfromis:
Looks like resetting to firmware defaults does not run `USER_BACKLOG` again.

Did a test with `USER_BACKLOG "SetOption64 1"` to use a really simple case, along with some other settings including wifi details. Erase/flashed the build result, and SetOption64 indeed came up as "ON". Unsurprisingly, I could also confirm my `#define LATITUDE` value.

Great. Next step in the test was `SetOption64 0`, and `Latitude 0`. After letting these be updated to flash, `Reset 1`. My expectation was that this would, as docs suggest "reset device settings to firmware defaults", including running the baked-in SetOption64 command. While the normal precompiled settings (including Latitude) were indeed my baked-in defaults, I saw that `SetOption64` now was OFF, reflecting the firmware default without having run the USER_BACKLOG command.

Shouldn't the baked-in commands be run again after resetting to firmware defaults?

sfromis:
Tested again adding `SetOption90` which is not available to `user_config_override.h`. This confirmed my expectation of this not making a difference.

arendst:
When doing a ``reset 1`` they should be run again yes.

Let me chk...

arendst:
Ah. Nice issue.

What happens is that after the ``reset 1`` execution the defaults are set and the backlog is filled with the requested commands. It then immediatly saves the settings changes the last time before the restart AND before the backlog commands have been executed.

I will play with the ``restart_flag`` to find an optimum value.

As an intermediate workaround just add the command ``SaveData`` at the end of the backlog command string.

I'll fix ASAP.

arendst:
Pls retest.

sfromis:
Works :-) 
I'll tell that to the user who was not happy with my suggestion about using `Reset 1` to get the commands executed after OTA upgrade.

FerFemNemBem:
In case I was that user - I would be absolutely happy if ```Reset 1``` had helped - but this was not the case. Thanks @arendst for fixing! Now everything works great and like expected!
Cheers Jan

PR #11995, opened 2021-05-03
feat: added units of mi flora sensors like shown in the ui 

conrad784:
## Description:
The units of the mi flora devices are shown in the Tasmota WebUI, but are not sent in the mqtt autodiscovery config messages.

## Checklist:

### DIFF
--- a/tasmota/xsns_62_esp32_mi_ble.ino
+++ b/tasmota/xsns_62_esp32_mi_ble.ino
@@ -2822,17 +2822,17 @@ const char *classes[] = {
   // 7
   "", //- empty device class
   "Moisture",
-  "",
+  "%",
 
   // 8
   "", //- empty device class
   "Illuminance",
-  "",
+  "lx",
 
   // 9
   "", //- empty device class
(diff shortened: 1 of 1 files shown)


Jason2866:
It is usual for Tasmota NOT to send units.

conrad784:
> It is usual for Tasmota NOT to send units.

I had a look at any other Tasmota device I have in my MQTT Broker and all auto-discovery configs have the "unit_of_meas" filled out with a unit.

Jason2866:
Ahh, for autodiscovery, yes.

conrad784:
From what I read (and tested in my own environment), the values I changed are the values used in the auto-discovery only. 
To be honest, I did not perfectly get why they are not there even if the Tasmota-WebUI already shows those units.

One could argue that those units could be different for different connected devices, but I would argue that at least humidity will always be in percent and the "Mi-Devices system" will always have those values for the Flora devices.

Maybe somebody familiar with #11161 could enlighten me, which arguments there are not to use those default units.

PR #12014, opened 2021-05-04
Introduced tasmota_options.h for libs

s-hadinger:
## Description:

Introduced `include/tasmota_options.h` to be included in libraries that need conditional compiles depending on Tasmota configuration. This was not easily possible previously.

Any module that needs to depend on global features needs a file `library.json` with the following lines:
```
  "build": {
    "flags": [ "-I$PROJECT_DIR/include" ]
  }
```

and then needs to include:
```
#include "tasmota_options.h"
```

 The inclusion could be forced to the entire library with `-includetasmota_options.h` option, but this breaks the auto-detection of dependencies; meaning that any change in tasmota configuration would not cause a recompilation of modules.

Changes:
- added `tasmota_options.h` in `include` folder
- moved cpp files needed by tls to a new lib `tls-mini`
- conditional compile added to Display_Renderer and uDisplay. @gemu2015 can you please check it's ok
- `LVGL_RENDERER` renamed `USE_DISPLAY_LVGL_ONLY` which is now a sub-option of `USE_DISPLAY` and reduces code size by removing code not needed by LVGL.

## Checklist:


Jason2866:
Nice. Like as long not used with `-includetasmota_options.h` since this is a trap.

gemu2015:
@s-hadinger 
found no issues so far. since display drv13 stays active we can control power, rotation and dimm but only save 16 not 30k

PR #12018, opened 2021-05-05
TasmotaSerial: Add support for all serial transmission modes

ul-gh:
This commit adds support for all common serial transmission modes
with different bit legth (5, 6, 7 and 8), one or two stop bits,
even, odd or no parity.

This is done by turning TasmotaSerial back into a thin wrapper
for pin-based automatic switching between hardware UART / Serial class
and upstream Arduino SoftwareSerial class.

This is based on the observation that Arduino SoftwareSerial has
been the subject of lots of development and fixes since TasmoataSerial
was first created.

The changes made to TasmotaSerial are (supposed to be) completely
transparent / backwards-compatible.

Use case for this is the various kinds of modern/smart energy meters for
which Tasmota is a very much suitable remote readout gateway solution:
Quite some of these meters use obscure serial transmission modes
(e.g. seven bit character size as in 7E1 mode)

As current Tasmota version of esp8266/Arduino is outdated and the Stream
class inherited by TasmotaSerial does not implement two methods which
are marked as overridden in recent esp8266/Arduino versions of the
SoftwareSerial class, in order to be able to compile without updating
all of Arduino, this patch pulls in a recent fork of
upstream version of espsoftwareserial as a submodule in folder:
lib/default/espsoftwareserial.

("git submodule update --recursive --remote" or "git pull --recurse-submodules")

## Description:

## Checklist:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/default/espsoftwareserial"]
+	path = lib/default/espsoftwareserial
+	url = https://github.com/ul-gh/espsoftwareserial.git
--- a/lib/default/TasmotaSerial-3.3.0/src/TasmotaSerial.cpp
+++ b/lib/default/TasmotaSerial-3.3.0/src/TasmotaSerial.cpp
@@ -18,26 +18,9 @@
 */
 
 #include <Arduino.h>
-
-// The Arduino standard GPIO routines are not enough,
-// must use some from the Espressif SDK as well
-extern "C" {
-#include "gpio.h"
-}
-
 #include <TasmotaSerial.h>
 
-#ifdef ESP8266
-
-void IRAM_ATTR callRxRead(void *self) { ((TasmotaSerial*)self)->rxRead(); };
-
-// As the Arduino attachInterrupt has no parameter, lists of objects
-// and callbacks corresponding to each possible GPIO pins have to be defined
-TasmotaSerial *tms_obj_list[16];
-
-#endif  // ESP8266
 #ifdef ESP32
-
 #if CONFIG_IDF_TARGET_ESP32           // ESP32/PICO-D4
 static int tasmota_serial_index = 2;  // Allow UART2 and UART1 only
 #elif CONFIG_IDF_TARGET_ESP32S2       // ESP32-S2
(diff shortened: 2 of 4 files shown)


Jason2866:
Thx for your work. Tasmota serial is highly optimized (functions and size) and works very reliable.
Outdated is the wrong wording. Tasmota serial has all functions which are needed for Tasmota.
You did not mentioned an specific use case why a change would make sense and should be considered. The change would increase code size without an benefit.

ul-gh:
Hi, 

the submodule is only for illustration and to avoid any changes to the packaged version of the upstream Arduino which is currently shipped with Tasmota.

* The submodule can be removed once the packaged version is updated or the following two lines hot fix are apllied:

```
diff --git a/src/SoftwareSerial.h b/src/SoftwareSerial.h
index 1a2490d..e452d9e 100644
--- a/src/SoftwareSerial.h
+++ b/src/SoftwareSerial.h
@@ -129,7 +129,9 @@ public:
 
     int available() override;
 #if defined(ESP8266)
-    int availableForWrite() override {
+    // TEMPORARY fix for compatibility with current Tasmota toolchain version of Arduino-espressif8266
+    //int availableForWrite() override {
+    int availableForWrite() {
 #else
     int availableForWrite() {
 #endif
@@ -158,7 +160,8 @@ public:
     /// The read(buffer, size) functions are non-blocking, the same as readBytes but without timeout
     int read(uint8_t* buffer, size_t size)
 #if defined(ESP8266)
-        override
+        // TEMPORARY fix for compatibility with current Tasmota toolchain version of Arduino-espressif8266
+        //override
 #endif
```

As for the size difference: Yes, code size is show as follows:

```
TasmotaSerial_all_modes:

RAM:   [======    ]  57.9% (used 47420 bytes from 81920 bytes)
Flash: [======    ]  62.6% (used 640720 bytes from 1023984 bytes)


development:

RAM:   [======    ]  58.0% (used 47484 bytes from 81920 bytes)
Flash: [======    ]  62.3% (used 637760 bytes from 1023984 bytes)
```

I guess you can weed out a lot of the functionality of the upstream toolchain package version of ESPSoftwareSerial, but this is not only a trade-off between code size and features, but you can also avoid all maintenance of duplicated or diverting codes when using the Arduino component.


* Anyways, if the size difference is significant, having a choice between two alternatives would be best..

* What amount of work would be needed to make TasmotaSerial support the different character size and parity options that ESPSoftwareSerial offers?

ul-gh:
As for the use case (already replied via e-mail, on the risk of having a duplicate comment):

From the commit message:

> Use case for this is the various kinds of modern/smart energy meters for
> which Tasmota is a very much suitable remote readout gateway solution:
> 
> Quite some of these meters use obscure serial transmission modes (e.g.
> seven bit character size as in 7E1 mode)

Out of the three devices I tested, one uses 4800-7E1 and one uses
9600-7E1 mode, both for which TasmotaSerial does not work without any
changes.

Jason2866:
Theo will have a look to implement the additional transmission modes to software serial.

s-hadinger:
Thanks for the PR, but I'm afraid we can't just blindly switch to this other lib. We have spent a huge amount of work to stabilize 115200 reception (for ex ZBBridge) and taking into account timing noise that causes Wifi interrupts.

Support for 5/6/7/8 can easily be done in software before calling Tasmota-Serial. 

espsoftwareserial uses interrupts at each change in the serial signal, which is not reliable enough at 115200 (CPU at 80MHz) in my experience. I would love to be proved wrong.

115200 means a transition every 8.7 microsecond, that's around 700 cycles of CPU at 80MHz, knowing that ISR takes ~500 cycles, I don't see how this could work.

s-hadinger:
One more note: I described in details the issues that cause wifi interrupts to receiving and the counter measures taken in Zigbee module: https://tasmota.github.io/docs/Zigbee-Internals/

ul-gh:
Thanks anyways for your attention!

I admit, my view is limited and if the Arduino lib causes too many issues, this is not an option.

In case you are planning on working on TasmotaSerial implementation n the future, if adding /all/ of the obscure serial transmission modes is too much work, then I guess that the five or six-bits modes ("TELEX?") are really a bit far-fetched.

But the 7 and 8 bits with odd and even parity I encountered more often than I expected..

s-hadinger:
I've never seen 8 bits with parity, and this would not work with the current implementation.

7 bits with parity is trivial to do before calling TasmotaSerial (sending) and after receiving bytes.

Just grab the code from espsoftwareserial for parity calculation and set bit 7 accordingly
```

            case SWSERIAL_PARITY_EVEN:
                // from inverted, so use odd parity
                parityBit = byte;
                parityBit ^= parityBit >> 4;
                parityBit &= 0xf;
                parityBit = (0x9669 >> parityBit) & 1;
                break;
            case SWSERIAL_PARITY_ODD:
                // from inverted, so use even parity
                parityBit = byte;
                parityBit ^= parityBit >> 4;
                parityBit &= 0xf;
                parityBit = (0x6996 >> parityBit) & 1;
                break;
```

ul-gh:
Thanks!

I do not know why, but the 8E1 mode is quite common for Modbus RTU devices, e.g. ([PDF manual OR-WE-517](https://files.orno.pl/manuals/old/OR-WE-513,%20516,%20517%20manual%20ENG.pdf), on last page, paragraph 8.1)

Jason2866:
There is still hardware serial. Always 1st choice for using with a esp82xx

TheChatty:
@ul-gh I can't get your branch [ul-gh:TasmotaSerial_all_modes](https://github.com/ul-gh/Tasmota/tree/TasmotaSerial_all_modes) to compile in gitpod. Could you please reapply your patch to current Tasmota? I'm working on [this](https://github.com/gemu2015/Sonoff-Tasmota/issues/40) which needs bit-inverted serial. Swapping HIGH/LOW in TasmotaSerial.cpp worked for sending but not receiving. I'd like to try your (Ardunios) library for the same approach.

ul-gh:
Hi,


first of all:

==> Yes, the ESP8266 hardware is clearly capable of running the serial
communication lines in inverse polarity, as this is documented in the SDK:

>
https://docs.espressif.com/projects/esp8266-rtos-sdk/en/latest/api-reference/peripherals/uart.html#_CPPv421uart_set_line_inverse11uart_port_t8uint32_t

Back in 2021, I put some effort not only into making the Arduino
software serial module work with tasmota but also to allow configuring
the less common UART modes. This was in order to make some model of
smart meter (easymeter) work which required a seven-bit serial
communication mode.

When these patches were rejected however, I did not put any more effort
into this. If I remember correctly, the software serial module did have
some hiccups, reading was not very reliable.

When using the hardware serial (UART) however, all of these UART modes
should be working fine without these issues.

I have not had a look into the tasmota sources since a year ago, I don't
know if this can be configured in tasmota as of today.

It might be worth asking the tasmota maintainers on how to configure the
advanced UART features when using the hardware serial.


Regards,
Ulrich

TheChatty:
As you can see in my thread it's about using a factory-assembled ESP having two IR diodes on its board which are connected to GPIO4/14 which enforces software serial. The original purpose is to use it for remote controlling devices I guess - you can easily decode codes from real remotes and later transmit programmatically. However using it for IR serial communication is another matter. Inverting polarity for transmitting diode already works but I was not successful yet with receiving data. So at the moment I'm waiting for delivery of a plain IR receiver to replace the existing (factory-assembled) IR receiver IC.

PR #12075, opened 2021-05-10
Adding support for ILI9488/XPT2046 display

mikep1998:
## Description:

## Checklist:


mikep1998:
Yes, you are right the mapping can be done before the rotation. Or after not much of a difference. The Negative values are seen because the the values of rotmap_max and rotmap_min are imprecise and the map function can return values outside of rotmap_max or rotmap_min. 

Example if the xmax is 100 and the read from the touch is 90 then map will return negative value.

gemu2015:
then you should rethink the mapping function. it should limit the values inside the unrotated screen bounds

gemu2015:
these variables hold the unrotated size

   uint16_t gxs;
   uint16_t gys;

mikep1998:
I don't understand the issue. 
if not map then what function to use? Or write your own.

mikep1998:
Ok let me rework it.

thanks for the suggestion.

gemu2015:
map and then

constrain(x, a, b)

PR #12090, opened 2021-05-12
Expand decode-status to support all feature bits

sfromis:
Output of decode-status.py lacks information about the latest two new sets of feature bits, due to the constant for array size not being updated when the array was expanded. Code changed to avoid this magical number, but instead check size of array with texts, and array with features from the JSON format in Status 4, using the lowest number to provide as many bits as both Tasmota version and Python code supports. A warning is printed if the number of feature bits from Tasmota is higher than supported in the current iteration of Python code.

Based on history, I'm assuming that updates are first done in arendst/Tasmota development, later replicated to tasmota/Tasmota-decode-status. If wrong, I can move the change there.

## Description:

## Checklist:

### DIFF
--- a/tools/decode-status.py
+++ b/tools/decode-status.py
@@ -337,7 +337,11 @@ def StartDecode():
     if "StatusMEM" in obj:
         if "Features" in obj["StatusMEM"]:
             features = []
-            for f in range(7):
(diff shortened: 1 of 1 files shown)


arendst:
Thx. This is the correct location for changes.

PR #12292, opened 2021-06-04
Add Wi-Fi Modes Control Command

ascillato:
## Description:

Tasmota in a BLANK device, supports by default all the Wi-Fi modes available in the ESP8266 that are IEEE 802.11b/g/n.
But, if the device is not blank or comes from another firmware, sometimes the supported modes are restricted like having only 11b or 11b/g. Tasmota don't force the full support. That is why, this PR adds new options for the command `WIFI` to FORCE which modes are going to be used.

The `WIFI` command now have the following options with this PR:

0 - Turn Off Wi-Fi
1 - Turn On Wi-Fi
**2 - Force the device to ONLY connects as a 11b device
3 - Force the device to ONLY connects as a 11b/g device
4 - Force the device to connects as a 11b/g/n device**

When changing modes, the device disconnects and reconnects to the router without restarting. Because of that, sometimes you need to refresh your browser.

image

**_(REMEMBER that if your router don't support 11b, or is configured to not support 11b and you force 11b, your device won't connect. In the serial console you see AP TIMEOUT no matter if the SSId and Password are Ok)_**

Besides that, with this PR, it has been added to show the actual mode in `STATUS 11`, in the response of the `WIFI` command and in the INFORMATION menu of the WebUI, for user troubleshooting. Also this PR have some code cleaning.

image

Forcing modes is useful to troubleshoot and solve issues like:
+ A device that had another firmware before,
+ A device that is far from a router (sometimes forcing 11b gives more range <- this depends on the router due to new routers have wide and better range on 11n),
+ If in the users' network, there is a very old device that only supports 11b, sometimes old or cheap routers can't manage correctly different devices in different modes and put the entire network on 11b. With this, you can force to use 11n in Tasmota and find which device is lowering the performance in the network,
+ If a new device can't connect and from the serial interface it is being showed AP TIMEOUT, it can be checked now which modes are supported in the Tasmotized device to match the Router's.

In the AP mode, the ESP8266 supports only b/g modes.

These code changes were tested too on ESP32 and works fine. The ESP32 have all mode enabled by default, so this PR don't provide the new options of the `WIFI` command for ESP32. It just provide showing the actual mode.

The ESP32 supports b/g/n/LR on both AP and STA modes. The LR mode (Long Range-Low Rate of 1km) is not supported in Tasmota ATM and can not be used.

**Proposed changes to the docs:** https://github.com/tasmota/docs/pull/751

## Checklist:

### DIFF
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -200,6 +200,7 @@
 #define D_JSON_VOLUME "Volume"
 #define D_JSON_WEIGHT "Weight"
 #define D_JSON_WIFI "Wifi"
+#define D_JSON_WIFI_MODE "Mode"
 #define D_JSON_WRONG "Wrong"
 #define D_JSON_WRONG_PARAMETERS "Wrong parameters"
 #define D_JSON_YESTERDAY "Yesterday"
--- a/tasmota/support_command.ino
+++ b/tasmota/support_command.ino
@@ -2142,8 +2142,12 @@ void CmndWifi(void)
   if ((XdrvMailbox.payload >= 0) && (XdrvMailbox.payload <= 1)) {
     Settings.flag4.network_wifi = XdrvMailbox.payload;
     if (Settings.flag4.network_wifi) { WifiEnable(); }
+#ifdef ESP8266
+  } else if ((XdrvMailbox.payload >= 2) && (XdrvMailbox.payload <= 4)) {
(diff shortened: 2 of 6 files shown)


Jason2866:
@ascillato  i think this discussion is from interest for you, and i think you can give feedback https://github.com/esp8266/Arduino/issues/7965

PR #12335, opened 2021-06-10
Move "includes"" from c3 env to esp32 default settings

Jason2866:
`sdkconfig.h` is included for all ESP32 builds. Avoiding strange errors when compiling. `fix_esp32c3.h` is only activated when riscv flag is set, so it is okay here too.

Doing this cleans the env and reduces the chance to overlook needed compile flags for user generated env

## Checklist:

### DIFF
--- a/platformio_tasmota32.ini
+++ b/platformio_tasmota32.ini
@@ -53,8 +53,8 @@ build_flags                 = ${esp_defaults.build_flags}
                               ;for TLS we can afford compiling for 4K RSA keys
                               -DUSE_4K_RSA
                               -I$PROJECT_DIR/include
-                              -include "fix_esp32c3.h"
                               -include "sdkconfig.h"
+                              -include "fix_esp32c3.h"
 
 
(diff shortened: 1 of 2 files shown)


platformio_tasmota32.ini
```diff
@@ -52,6 +52,9 @@ build_flags                 = ${esp_defaults.build_flags}
                               -Dmemcmp_P=memcmp
                               ;for TLS we can afford compiling for 4K RSA keys
                               -DUSE_4K_RSA
+                              -I$PROJECT_DIR/include
+                              -include "fix_esp32c3.h"
```
s-hadinger, line 56: Including "fix_esp32c3.h" is not needed for non-C3. But it does not harm either.

Jason2866: I wrote the same. Placing here avoids accidently forgetting

s-hadinger:
@arendst Good to merge

PR #12377, opened 2021-06-16
Allow Captive Portal to be optional

ascillato:
## Description:

This PR just adds a new precompiler `#define` so as to allow the captive portal feature to be included or not in the build. By default, the captive portal is enabled and Tasmota's behaviour is the same as always.

If an user adds the following line in _user_config_override.h_ file, Tasmota won't pop up the captive portal for the Initial Wi-Fi configuration. If the user opens a web browser to go to 192.168.4.1, the Initial Wi-Fi configuration page will be shown.
```cpp
#define NO_CAPTIVE_PORTAL 
```

This optional is useful if the Wi-Fi credentials are going to be entered by a third party software, like a phone app. In the case of a phone app, it just needs to connect to the Tasmota AP, and then, by HTTP API, it can send the Wi-Fi Credentials. If the captive portal feature was active, when the phone app connects to Tasmota AP, the Tasmota WebUI page will pop up difficulting the user to return to the phone app. 

This PR don't increase the FLASH, nor RAM usage.

## Checklist:

### DIFF
--- a/tasmota/xdrv_01_webserver.ino
+++ b/tasmota/xdrv_01_webserver.ino
@@ -1017,7 +1017,9 @@ void WebSliderColdWarm(void)
 
 void HandleRoot(void)
 {
+#ifndef NO_CAPTIVE_PORTAL
   if (CaptivePortal()) { return; }  // If captive portal redirect instead of displaying the page.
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks. Makes sense.

PR #9418, opened 2020-09-28
Feature: IRMP Support for Tasmota

eku:
## Description:

Adds support for [IRMP](https://github.com/ukw100/IRMP) to Tasmota.

## Checklist:
  - [x] The pull request is done against the latest dev branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR.
  - [x] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.1
  - [x] The code change is tested and works on core ESP32 V.1.12.2
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

IRMP/IRSND is a library for decoding and encoding infrared signals. It supports 50+ protocols and proves to be more robust in decoding than IRremoteESP8266, which was the motivation to provide IRMP for Tasmota.

For details see
https://www.mikrocontroller.net/articles/IRMP_-_english
https://github.com/ukw100/IRMP

The functionality is completely deactivated by default (see my_user_config.h). As decoding is done in the ISR, the IRAM of the ESP8266 is not sufficient to activate all 50 protocols, so only the required protocols should be activated. The same applies to IRSND, the encoder/transmitter.

I am running the module from an ESP32 and have activated all supported protocols.

IRMP also supports a feedback LED, but I didn't know how to include it using the device profiles.

PR #12382, opened 2021-06-16
Feature: IRMP Support for Tasmota

eku:
## Description:

Adds support for [IRMP](https://github.com/ukw100/IRMP) to Tasmota.

## Checklist:

IRMP/IRSND is a library for decoding and encoding infrared signals. It supports 50+ protocols and proves to be more robust in decoding than IRremoteESP8266, which was the motivation to provide IRMP for Tasmota.

For details see
https://www.mikrocontroller.net/articles/IRMP_-_english
https://github.com/ukw100/IRMP

The functionality is completely deactivated by default (see my_user_config.h). As decoding is done in the ISR, the IRAM of the ESP8266 is not sufficient to activate all 50 protocols, so only the required protocols should be activated. The same applies to IRSND, the encoder/transmitter.

I am running the module from an ESP32 and have activated all supported protocols.

IRMP also supports a feedback LED, but I didn't know how to include it using the device profiles.


frank-f:
I'd really love to see this included as it supports some more rare protocols.

Jason2866:
The reasons for not merging are still the same as for the orig. PR #9418 
Not all protocols can be included (ESP8266) and IRMP is interrupt driven which is a nogo for Tasmota.

eku:
> The reasons for not merging are still the same as for the orig. PR #9418
> Not all protocols can be included (ESP8266) and IRMP is interrupt driven which is a nogo for Tasmota.

Yes, I'm aware of that and I don't expect the feature to be enabled and included in a separate build.

Since there is obviously a need for IRMP from the crowd of users, just including it would probably be enough. Users can then build the firmware themselves.

Jason2866:
The goal of Tasmota is to deliver a stable firmware  version with any combination of drivers (limiting factor is only RAM and flash usage). So every driver included in Tasmota has to work without any known problems / issues with. IRMP PR does not statisfy this requirements

PR #12651, opened 2021-07-14
Feature/can bus

mbezuidenhout:
## Description:

Added support for MCP2515 CAN controller.
Reading and reporting of Freedom Won Lite battery (RJ45 pins 7 & 8) State of Charge on CAN bus.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.1.0.7.1


tasmota/tasmota_template.h
```diff
@@ -158,6 +158,7 @@ enum UserSelectablePins {
   GPIO_MAX7219CLK, GPIO_MAX7219DIN, GPIO_MAX7219CS, // MAX7219 interface
   GPIO_TFMINIPLUS_TX, GPIO_TFMINIPLUS_RX,  // TFmini Plus ToF sensor
   GPIO_ZEROCROSS,
+  GPIO_MCP2515_CS,                     // MCP2515 Chip Select
```
arendst, line 161: For compatibility this needs to be moved to the end of the list

tasmota/tasmota_template.h
```diff
@@ -347,6 +348,7 @@ const char kSensorNames[] PROGMEM =
   D_SENSOR_MAX7219_CLK "|" D_SENSOR_MAX7219_DIN "|" D_SENSOR_MAX7219_CS "|"
   D_SENSOR_TFMINIPLUS_TX "|" D_SENSOR_TFMINIPLUS_RX "|"
   D_SENSOR_ZEROCROSS "|"
+  D_SENSOR_MCP2515_CS "|"
```
arendst, line 351: Ditto

tasmota/tasmota_template.h
mbezuidenhout: Moved to above GPIO_INTERRUPT

tasmota/tasmota_template.h
mbezuidenhout: Moved to below D_SENSOR_INTERRUPT at end of list

tasmota/xsns_90_mcp2515.ino
```diff
@@ -0,0 +1,275 @@
...
+#endif
+
+#ifndef MCP2515_BMS_CLIENT
+  #define MCP2515_BMS_CLIENT
+  // Look for Freedom Won BMS data in CAN message
+  #ifndef MCP2515_BMS_FREEDWON
+    #define MCP2515_BMS_FREEDWON
+  #endif  // MCP2515_BMS_FREEDWON
+#endif // MCP2515_BMS_CLIENT
+
```
arendst, line 69: Nice set of defines.

If the user does not define the BMS client it is defined anyway. On the other hand I understand that without any user the current implementation is of no use either.

pkkrusty:
@mbezuidenhout Seems like this driver is pretty specific to the Freedom Won BMS? What would it take to expand it to a general purpose MCP2515 Canbus bridge/driver? Just diving into Canbus, specifically the VW e-Golf battery implementation, where @Tom-EVNut has done some [good work](https://github.com/Tom-evnut/VW-bms) already.

PR #12806, opened 2021-08-01
Add IP filter to TCPStart command

jamesorlakin:
## Description:

This is an optional second parameter to enable some *very basic* security around the `TCPStart` command. I run Home Assistant with the ZHA integration using the TCP Serial feature. The HTTP interface can be protected with a password and MQTT uses authentication, but there's no security measures on the TCP Serial bits. Of course there's no protocol to add authentication as such, but old NFS relied upon host-based security and it seems harmless to implement.

I'll update the docs as well, but I thought it best to see your thoughts as well beforehand.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.1.0.7.3 *(I don't have ESP32 hardware)*


s-hadinger:
Thanks for your PR. Indeed there was no security on the TCP connection. 

Looks good to me.  @arendst

PR #12827, opened 2021-08-05
Set CORS to be an optional feature

ascillato:
## Description:

As reported in #6767, CORS is not safely implemented in Tasmota, so until a better solution is submitted, with this PR we disable by default all code related to CORS. If any user want to enable it, just add `#define USE_CORS` in _user_config_override.h_ file and compile. Just remember to set a `webpassword`!

**Note:** In a following PR (separated from this just for better organization) will be added a message telling the user when they don't have a password set.

**Related issue (if applicable):** https://github.com/arendst/Tasmota/issues/6767

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [x] The code change is tested and works with Tasmota core ESP32 V.1.0.7.3
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

Issue #6767, opened 2019-10-27
Local access security issue due to HTTP CORS header


pseudosavant:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._

Tasmota has CORS HTTP headers enabled by default. This is a major security issue that could easily be exploited by any website with some simple Javascript, especially since Tasmota does not require a web password by default.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): _Shelly 1 and Sonoff Basic, but this issue isn't device specific_
- [x] Tasmota binary firmware version number used: _6.7_
  - [x] Pre-compiled
  - [ ] Self-compiled
- [x] Flashing tools used: _FlashESP8266.exe_
- [x] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:
17:36:26 RSL: stat/sonoff/RESULT = {"NAME":"Shelly 1","GPIO":[0,0,0,0,21,82,0,0,0,0,0,0,0],"FLAG":0,"BASE":46}
17:36:26 RSL: stat/sonoff/RESULT = {"Module":{"46":"Shelly 1"}}
17:36:27 RSL: stat/sonoff/RESULT = {"GPIO":"Not supported"}
```
- [x] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```lua
  Rules output here:
NA
```
- [x] Provide the output of this command: `Status 0`:
```lua
  STATUS 0 output here:
[...]

PR #12828, opened 2021-08-05
Add a Warning in UI if the web admin password is not set

ascillato:
## Description:

Add a Warning in Tasmota User Interface if the **_web admin password_** is not set.

To set the **_web admin password_**, you can go to CONFIGURATIONS -> CONFIG OTHER -> Web Admin Password.
Or
In the console by `webpassword` command.

Then, after restart, your browser will ask for Username and Password. By default the username is `admin`. The username can be changed by using the key `WEB_USERNAME` in the _user_config_override.h_ file.

As the password is for securing the HTTP API and the web interface, this warning is not published by MQTT nor Serial. Only in the web UI.

This continues the PR: https://github.com/arendst/Tasmota/pull/12827

image

**Related issue (if applicable):** https://github.com/arendst/Tasmota/issues/6767

## Checklist:

### DIFF
--- a/tasmota/language/af_AF.h
+++ b/tasmota/language/af_AF.h
@@ -240,6 +240,7 @@
 #define D_ERASED_SECTOR "Uitgewis sektor"
 
 // xdrv_02_webserver.ino
+#define D_NO_PASSWORD_SET "Device Insecure<br>Please set a Web Admin Password"
 #define D_NOSCRIPT "Skakel JavaScript aan om Tasmota te gebruik"
 #define D_MINIMAL_FIRMWARE_PLEASE_UPGRADE "MINIMAL firmware <br> gradeer asseblief op"
 #define D_WEBSERVER_ACTIVE_ON "Webbediener aktief op"
--- a/tasmota/language/bg_BG.h
+++ b/tasmota/language/bg_BG.h
@@ -240,6 +240,7 @@
 #define D_ERASED_SECTOR "Изтрит сектор"
 
 // xdrv_02_webserver.ino
+#define D_NO_PASSWORD_SET "Device Insecure<br>Please set a Web Admin Password"
 #define D_NOSCRIPT "Разрешете JavaScript, за да използвате Tasmota"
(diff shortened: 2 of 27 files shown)


arendst:
I know there are people jumping on the security bandwagon and want to add passwords to every item on the internet. I hate security. It has always been a showstopper for me. I liked tasmota for being as open as possible and swa over the years with some dismay hidden passwords being introduced.

I don't like this PR as it now even forces me to use a webpassword on my own local intranet.

As it stands I won't merge this. There must be a better way for resellers to force their users to use a password without bothering me.

ascillato:
Yes, I agree. Let's find a better way for this. I did this PR just because I was following the advises in #6767.

bananenfisch:
> I don't like this PR as it now even forces me to use a web password on my own local intranet.

Is it local with this API implementation? :-D
If you think, its local (and safe), just visit my website: https://www.bananenfisch.net/fun.php

> I don't use MQTT passwords and don't want to use webpage passwords either.

I dont use MQTT passwords too, because i have only trusted devices in my LAN. But: sometimes is visit untrusted websites from a browser inside my LAN - thats the bridge, which could open my garage door or modify all firmware on every tasmota device in my LAN.

Well, i've setup my passwords for the Web-API, but most users out there never heard about this issue. And the issue is public. Please print the warning (maybe with a "dismiss" button, whatever). It's really an ugly bug... if my next mate, never heard about cross-site attacks, using Tasmota without web-password, i could simple send him a link and take control over his whole house!

>     * or have this enable only on new installations
>     * and having the option of not providing a password.

Yes, please! ;-)
Tasmota get more and more users - this issue is going to be a huge problem if some attackers are smelling blood!

arendst:
Retry with latest change. It won't tackle every attack but I think the basic functionality as demonstrated by your ``fun.php`` is disabled by default.

In fact all HTTP requests not coming from the device are disabled this way.

If users want to use HTTP web commands they have to enable ``SetOption128 1`` and additionally set an admin password.

Let me know what you think.

arendst:
Another try. See above.

PR #12890, opened 2021-08-15
Slower than 1Hz PWM support

lmagyar:
## Description:

See detailed description below (from driver file).

I think **v0 than v2 in Further development chapter below** should be done before it can be part of an official release without #define USE_SLOW_PWM In that case I can modify the tasmota/docs documentation also.

**Add #define USE_SLOW_PWM to user_config_override.h to use this feature!** Itt adds 608 bytes to firmware.bin.

> PWM with slower than 1Hz periods
>
> Slow PWM is good for eg. heating elements, thermoelectric valve actuators that are connected
> to mains AC voltage, where you don't need to use phase angle dimmers, where you can use
> simple SSRs (Solid State Relays), or even regular relays.
>
> Slow PWM works as plain PWM and also as brightness only Light (SetOption15 0 and
> SetOption15 1), as Light it has built in HA support by default.
>
> Current implementation:
> - can be turned on only by build flag in user_config_override.h: #define USE_SLOW_PWM
> - adds new command SlowPwmPeriod, and adds new setting slow_pwm_period
> - replaces Pwm\<x> command functionality, extends PwmRange command functionality
> - replaces Pwm\<x> pin behaviour (normal Pwm and Light, SetOption15 0 or 1)
> - does not modify other PWM related behavior (eg. status led, buzzer, shutter, pwm dimmer,
>   backlight, etc.) - only pwm status led is tested on ESP8266
>
> Tested with:
> - ESP8266 - NodeMcu v2: D1, D2, D5, D6, D7 pins (GPIO 5, 4, 14, 12, 13)
> - ESP32 - Wemos D1 Mini ESP32: D1, D2, D5, D6, D7 pins (GPIO 22, 21, 18, 19, 23)
>
> How to use it:
> - rebuild with #define USE_SLOW_PWM in user_config_override.h
> - upload new firmware
> - Important: in case of ESP32, if PWM was set up on any pin before new firmware is uploaded,
>   power cycle the board, because ESP32 has hardware PWM, and power cycling the device will
>   detach pins from hardware PWM (that were previously attached to hardware PWM by previous
>   firmware), and Slow PWM can use them as digital IO pins
> - use Generic(18) or ESP32-DevKit(1) module type
> - set GPIO pins as PWM\<x> or PWM_i\<x>
> - use SlowPwmPeriod \<x> command to set slow PWM cycle time
>   Note: Slow PWM's time resolution is 50ms, setting period times lower than 5s will decrase
>   even Dimmer 1..100 resolution. Use low values only for test purposes.
> - from now on, there are 3 scenarios:
>   - as plain PWM output:
>     - execute: SetOption15 0
>     - use PWM\<x> commands to set duty cycle
>     - use PwmRange \<x>, SlowPwmPeriod \<x> commands to change parameters
>   - as Light, only with 1 PWM pin (this has built in HA support also, as normal Light entity)
>     see: https://tasmota.github.io/docs/Lights/#1-channel-dimmable-light
>     - execute: LedTable 0  - or recompile with Settings->light_correction = 0;
>     - use Power\<x>, Dimmer\<x> commands to set duty cycle
>     - use PwmRange \<x>, SlowPwmPeriod \<x> commands to change parameters
>       Note: restart device when PwmRange changed, Light updates range only when reinitialized
>   - as Light, but 1..5 PWM pins (this has built in HA support also, as normal Light entity)
>     see: https://tasmota.github.io/docs/Lights/#independent-pwm-channels
>     - execute: LedTable 0  - or recompile with Settings->light_correction = 0;
>     - execute: MultiPwm 1 (previously SetOption68 1)  - or recompile with #define LIGHT_CHANNEL_MODE true
>     - use Power\<x>, Channel\<x> commands to set duty cycle
>     - use PwmRange \<x>, SlowPwmPeriod \<x> commands to change parameters
>       Note: restart device when PwmRange changed, Light updates range only when reinitialized
>
> I'm aware of https://tasmota.github.io/docs/Time-Proportioned-Output-support/
> - Timeprop has more features but no user interface. Slow PWM has less functions but has user
>   interface and built in HA support as Light entity.
> - main Timeprop features that are missing in Slow PWM:
>   - safety mechanism to revert to a specified fallback value when MQTT communication is lost:
>     - can be implemented with Tasmota rules
>   - actuator dead time:
>     - with other control specific features I think it has better place in the main control
>       logic in HA or other controller software
>   - different cycle times per pin:
>     - that could be useful
>
> Further development:
> - v0: as is now, it remains a build option, only hard core users will use it
>   (though we should rename it as xdrv_57 instead of xdrv_128)
> - v1: enable it with a new SetOption, still all PWM pin behavior will be changed when turned on
> - v2: create new pin config, like "SPWM\<x>", where only pins defined as Slow PWM will be
>   affected with this behavior
> - v3: extend scripting (Berry?) with the implemented v1/v2 solution

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 - Support for (Yeelight) Mi Desk Pro using binary tasmota32solo1.bin
 - Initial support for influxdb using ``#define USE_INFLUXDB`` and several ``Ifx`` commands
 - Command ``SetOption128 1`` disabling web referer check default blocking HTTP web commands (#12828)
-- Initial support for PWM with slower than 1Hz periods using ``#define USE_SLOW_PWM`` and ``SlowPwmPeriod`` command
+- Initial support for PWM with slower than 1Hz periods using ``#define USE_SLOW_PWM`` and ``SlowPwmPeriod`` command (#12890)
 
 ### Changed
 - NeoPixelBus library from v2.6.3 to v2.6.7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 - Support for (Yeelight) Mi Desk Pro using binary tasmota32solo1.bin
 - Initial support for influxdb using ``#define USE_INFLUXDB`` and several ``Ifx`` commands
 - Command ``SetOption128 1`` disabling web referer check default blocking HTTP web commands (#12828)
+- Initial support for PWM with slower than 1Hz periods using ``#define USE_SLOW_PWM`` and ``SlowPwmPeriod`` command
 
 ### Changed
 - NeoPixelBus library from v2.6.3 to v2.6.7
--- a/tasmota/i18n.h
+++ b/tasmota/i18n.h
@@ -281,6 +281,7 @@
 #define D_CMND_GPIOS "GPIOs"
 #define D_CMND_PWM "PWM"
 #define D_CMND_PWMFREQUENCY "PWMFrequency"
+#define D_CMND_SLOWPWMPERIOD "SlowPWMPeriod"
 #define D_CMND_PWMRANGE "PWMRange"
 #define D_CMND_BUTTONDEBOUNCE "ButtonDebounce"
 #define D_CMND_SWITCHDEBOUNCE "SwitchDebounce"
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -587,8 +587,9 @@ typedef struct {
 
   uint16_t      influxdb_port;             // 4CE
   power_t       interlock[MAX_INTERLOCKS_SET];  // 4D0 MAX_INTERLOCKS = MAX_RELAYS / 2
+  uint16_t      slow_pwm_period;           // 508
 
-  uint8_t       free_508[36];              // 508
+  uint8_t       free_50A[34];              // 50A
 
   uint16_t      mqtt_keepalive;            // 52C
   uint16_t      mqtt_socket_timeout;       // 52E
--- a/tasmota/settings.ino
+++ b/tasmota/settings.ino
@@ -1044,6 +1044,7 @@ void SettingsDefaultSet2(void) {
   flag4.virtual_ct_cw |= LIGHT_VIRTUAL_CT_CW;
 
   Settings->pwm_frequency = PWM_FREQ;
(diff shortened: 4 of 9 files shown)


stefanbode:
Him when I looked to PWM and the capabilities defined by the waveform function this was limited to 100Hz. As I read your reading the official library is now capable down to 1Hz. Is this correct? In this case, I have to do some changes to my shutter code. Anyhow the below 1Hz is an interesting approach. There are use cases also with motors that would benefit from more flexibility here because PWM is the only stable process besides RTC calls but can work at higher frequencies

s-hadinger:
From what I read in the code, the ultra-low frequency is not managed by the lib but by ad-hoc code based on Tasmota ticks. I'm not sure how reliable this is (what is the resolution).

lmagyar:
Yes, ultra low frequency is managed by my code, and it has 50ms resolution by Tasmota ticks. The goal is to control heating elements, or anything that can be regularly turned on and off (regularly means tens of seconds or even minutes, anything faster than 5s is only good for testing, so I think it is not good for motors). Yes, the 50ms tick has irregularity, but for a heating element it is not an issue.

You can use normal PWM and a proper phase angle dimmer to control mains AC stuff, but with this ultra low frequency PWM you can use cheap solid state relays instead of dimmers.

I'm using this eg. to control a small mains voltage heating element to add some small temperature offset to an old gas heater's mechanical thermostat, making this old gas stuff "smart" without modifying anything in it.

Only after I implemented my version I saw the xdrv_48_timeprop.ino, that has very similar functionality, with more parameters, but without UI or HA integration. I thought it can be useful for somebody else, that's why I made this PR, but I'm absolutely new to Tasmota and the design considerations around it!

stefanbode:
I checked the  xdrv_48_timeprop.ino and agree that we see already here similar functions. Therefore I find it easier if we just add there you changes to make Homeassistant work and add some nice UI. (UI is always costly for bytes and often will only be used once...)

arendst:
Agree too. The current implementation also heavily intrudes existing code making maintenance more dificult.

For now you'll have to use the less intruding timeprop driver.

lmagyar:
Agree. Should I have create a new xlgt driver as UI/API for timeprop? I'm new to Tasmota, isn't that also intrusive?

ascillato2:
> Should I have create a new xlgt driver as UI/API for timeprop?

No, you can add all that in the same timeprop driver. That would be great if you can do it.

Issue #12837, opened 2021-08-08
Error building tasmota32idf4 with latest development


sillyfrog:
### PROBLEM DESCRIPTION

With the latest development release (https://github.com/arendst/Tasmota/commit/0b28019544b922a2cd15d882f818f60a1a51366c), using the `platformio_tasmota_cenv_sample.ini`, you can no longer build `tasmota32idf4`.

From my testing, this PR is the cause of the issue https://github.com/arendst/Tasmota/pull/12830/ (ping @Jason2866 ).

Commenting out this line allows me to build `tasmota32idf4` again:
```diff
--- a/pio-tools/gzip-firmware.py
+++ b/pio-tools/gzip-firmware.py
@@ -27,7 +27,7 @@ def map_gzip(source, target, env):
     # remove map file
     if os.path.isfile(bin_file): os.remove(bin_file)
 
-env.AddPostAction("$BUILD_DIR/${PROGNAME}.bin", [map_gzip])
+# env.AddPostAction("$BUILD_DIR/${PROGNAME}.bin", [map_gzip])
 
 # gzip only for ESP8266
 if env["PIOPLATFORM"] != "espressif32":
```

I have also tried using the latest release at https://github.com/Jason2866/esp32-arduino-lib-builder/releases (just in case - I'm not claiming to understand why it's an issue just for `tasmota32idf4`)

Please let me know if I can do anything to help.

PS: This is the output from the end of the build:
```
Merged 25 ELF sections
bin_map_copy([".pio/build/tasmota32idf4/firmware.bin"], [".pio/build/tasmota32idf4/firmware.elf"])
map_gzip([".pio/build/tasmota32idf4/firmware.bin"], [".pio/build/tasmota32idf4/firmware.elf"])
*** [.pio/build/tasmota32idf4/firmware.bin] build_output/map/tasmota32idf4.map: No such file or directory
```

I also tried on a clean Ubuntu (docker) install, same result.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [X] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
[...]

PR #12893, opened 2021-08-16
Python pio updates

sillyfrog:
## Description:

Further to the discussion at the end of https://github.com/arendst/Tasmota/issues/12837 this PR refactors some of the PIO build tools to make them more "pythonic", and to have a common library so both the naming and gziping processes use the same functions for deriving file names. This also addresses an issue where the firmware files were incorrectly named when using non-default paths.

I have tested with the default setup, `platformio_overides.ini` changes and using environment variables. For the following:

- workspace_dir
- build_cache_dir
- cache_dir
- build_dir

**Please note:** I have _not_ tested on Windows (I don't have a Windows machine), however on Linux and Mac it has been OK. The process used should be cross platform. If you find any issues, please let me know and I'll fix ASAP.

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.1.0.7.3

### DIFF
--- a/pio-tools/gzip-firmware.py
+++ b/pio-tools/gzip-firmware.py
@@ -1,33 +1,32 @@
-Import('env')
+Import("env")
 import os
 import shutil
 import gzip
+import pathlib
 
-platform = env.PioPlatform()
-board = env.BoardConfig()
-mcu = board.get("build.mcu", "esp32")
+import tasmotapiolib
 
-OUTPUT_DIR = "build_output{}".format(os.path.sep)
 
 def map_gzip(source, target, env):
-    variant = str(target[0]).split(os.path.sep)[2]
-
     # create string with location and file names based on variant
-    bin_file = "{}map{}{}.map".format(OUTPUT_DIR, os.path.sep, variant)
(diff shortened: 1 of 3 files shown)


arendst:
Thx.

Tested on windows and works fine.

I'm glad I still can easily disable map gzip as I don't need map gzip locally to be able to read it in VSC.

sillyfrog:
@arendst Do you disable by deleting the file `gzip-firmware.py` file? I could make it so if you have an environment variable set it won't do the gzip file stuff. (Assuming you delete the file): that way your source could stay the same, just the `.gz` files are not created (I was thinking about doing something with environment variables to also have a way to put all of the source files somewhere else outside of the source tree).

arendst:
No I comment out line 31 like
```
#env.AddPostAction("$BUILD_DIR/${PROGNAME}.bin", [map_gzip])
```
 I might still need the binary zipped.

sillyfrog:
I have just pushed https://github.com/arendst/Tasmota/pull/12896 that provides options to disable the gzip creation, and override the locations of the `.bin` and `.map` files.

PR #12896, opened 2021-08-16
Python PIO tool controls using environment variables

sillyfrog:
## Description:

A minor enhancement to the PIO tools to allow you to disable the compression of `.map` and `.bin` files using environment variables. It also allows overriding the final location of the `.map` and `.bin` files.

For example, to disable the compression of the `.map` files, put the following in your environment:
```
TASMOTA_DISABLE_MAP_GZ=1
```

All of the options are at the top of `tasmotapiolib.py`.

(The top of this page outlines ways to set environment variables: https://docs.platformio.org/en/latest/envvars.html)

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.1.0.7.3


arendst:
Thx again! Testing...

arendst:
I don't like the OS environment useage. On windows many shells are running and setting the correct environment value is not that easy.

I prefer to have a value set in the ``platformio_override.ini file`` to be handled by the python gzip script.

Jason2866:
@sillyfrog  Take a look here https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html#custom-firmware-program-name
You can add for example to `build_flags` and parse for. Now you are in the Platformio rabbit hole. Since it is Python, a easy task for a Pythonic ;-)

sillyfrog:
@arendst While writing the description for the PR I had a similar thought... However after looking further at the docs, I some how missed yesterday that there is a way to add custom options to the `.ini`. I'll do that now and add another PR.

PR #12900, opened 2021-08-16
WebUI: Add option for enabling HTTP API 

ascillato:
## Description:

This PR just adds in the WebUI an option for enabling or disabling the SetOption128.

In the commits https://github.com/arendst/Tasmota/commit/15312f88290df43888dbf6ac25023c6f238e1b6b and https://github.com/arendst/Tasmota/commit/4ed63354288cb7a689fd793a2b26a523253ef97e (2021-Aug-15) it has been added a new command `SetOption128` into Tasmota in order to enable or disable the web referer check for blocking HTTP web commands. In other words, the use of HTTP API commands. **_By default the HTTP API is disabled._**

As that was a necessary **_breaking change_**, with this PR it is added an option to enable or disable the HTTP API through the WebUI too. For new users this makes an easier transition through this new security feature. The option is added in OTHER CONFIGURATIONS Menu:
image

Besides that, it is also added in the INFORMATION Menu:
image

And in the STATUS 5:
```lua
23:34:29.624 RSL: STATUS5 = {"StatusNET":{"Hostname":"tasmota-C658CC-6348","IPAddress":"192.168.1.112","Gateway":"192.168.1.1","Subnetmask":"255.255.255.0","DNSServer1":"179.233.128.14","DNSServer2":"179.233.128.19","Mac":"A4:CF:12:C6:58:CC","Webserver":2,"WifiConfig":4,"WifiPower":17.0,"HTTP API":"Disabled"}}
````
## NOTE:
Remember that when enabling HTTP API, it is highly recommended to set the web admin password for security (https://github.com/arendst/Tasmota/pull/12828#issuecomment-898067728, https://github.com/arendst/Tasmota/issues/6767). To do that, you can go to CONFIGURATIONS -> CONFIG OTHER -> Web Admin Password. Or In the console by `webpassword` command.
Then, after restart, your browser will ask for **Username** and **Password**. By default the username is `admin`. The username can be changed by using the key `WEB_USERNAME` in the _user_config_override.h_ file. This password will secure the web interface too.

**Related issue (if applicable):** fixes #12828, #6767

## Checklist:


arendst:
Thx for your PR.

I'll suggest to leave this open for some days to have more user experience with my referer solution. I'm not quite sure it solves the issue of easy remote access or even if it doesn't hamper local access.

Give it until coming friday.

joeH68:
Here’s my two cents. 

I stumble into this PR while migrating from Shelly to Tasmota. I hope this is the right place to submit this feedback and that it turns out to be helpful. My background: IT, midsize residential admin/user (smarthome), 40+ productive Tasmota units (pre-compiled versions), 10+ in experimental mode, openhab as backend.

Very positive: One of my reasons for migrating to Tasmota was security, so I’m very glad to see this area being further developed

Less positive : To configure my switches I use a Python script over HTTP. I had accidentally changed OtaUrl to the latest development and got 5.0.5.5/6 installed (my mistake). The script stopped working and it took my a while to discover setoption128

For me, installing a Tasmota switch involves 3 stages: 
- Installation and configuration
- Operational
- Maintenance

Here my comments:

1. Installation and configuration:
- to reduce/avoid manual clicks, pre-compiled Tasmota versions could/should come with HTTP and MQTT API’s enabled, webserver=2 and no password (*)
- I would always set an admin password to get a minimum protection even during configuration/experimental stage
- before going operational I would: set webserver=1 (or even =0) and setoption128=0
- Tasmota Reset commands preserving WiFi settings could/should be implemented to force setoption128=1
- Setoption128=0 should be rejected if MQTT is also disabled (and viceversa)

2. Operational
- I would run Tasmota with MQTT enabled, HTTP disabled (setoption128=0), password protection and webserver=1 (or even 0)

3. Maintenance
- Using MQTT I would enable Admin Gui (webserver=2) and HTTP API (setoption128=1) 

(*) even at its first boot in AP mode, Tasmota could be receptive to HTTP commands and, before rebooting, allow configuring MQTT, Template, password and of course Wifi. Not sure if this is already possible. GUI could be extended to allow/require this input (optional)

Own compilations may of course choose different defaults and pre-settings.

arendst:
I'll merge this one.

Once merged I enable ``SO128 1`` so not to break backward compatibility.

People into security then can disable the HTTP API if they like.

bananenfisch:
Many thanks to @arendst and @ascillato for the fix! The idea, to solve this security issue via referrer-check is IMO a really good and simple solution, because the referrer cannot be manipulated via javascript for other hosts. And this solution does not break all other things! VG job, thx! :-)
If option 128 should be enabled/disabled by default is another topic (depends on focus for security or backwards compatibility). IMO maybe a short warning in the UI, when enabling 128 should be enough.

Thanks again!

PR #12966, opened 2021-08-24
Removed hard dependency on USE_LIGHT from USE_HOME_ASSISTANT

Avamander:
## Description:

This PR rearranges a few preprocessor directives to make `USE_HOME_ASSISTANT` not depend on `USE_LIGHT`.

Tuya is remains incompatible with undefining `USE_LIGHT` however.

## Checklist:

### DIFF
--- a/tasmota/xdrv_12_home_assistant.ino
+++ b/tasmota/xdrv_12_home_assistant.ino
@@ -414,6 +414,7 @@ void TryResponseAppend_P(const char *format, ...) {
 #endif
 }
 
+#ifdef USE_LIGHT
 void HAssAnnounceRelayLight(void)
 {
   char stopic[TOPSZ];
@@ -535,7 +536,6 @@ void HAssAnnounceRelayLight(void)
           }
           TryResponseAppend_P(HASS_DISCOVER_DEVICE_INFO_SHORT, unique_id, ESP_getChipId());
 
-  #ifdef USE_LIGHT
         if (i >= Light.device) {
           if (!RelayX || PwmMod || (TuyaDim > 0 && TuyaMod)) {
             char *brightness_command_topic = stemp1;
@@ -586,7 +586,6 @@ void HAssAnnounceRelayLight(void)
           ind_light = false;
           max_lights--;
         }
-  #endif  // USE_LIGHT
         TryResponseAppend_P(PSTR("}"));
       }
     }
@@ -594,6 +593,7 @@ void HAssAnnounceRelayLight(void)
     MqttPublish(stopic, true);
   }
 }
+#endif  // USE_LIGHT
 
 void HAssAnnouncerTriggers(uint8_t device, uint8_t present, uint8_t key, uint8_t toggle, uint8_t hold, uint8_t single, uint8_t trg_start, uint8_t trg_end)
 {
@@ -1107,8 +1107,10 @@ void HAssDiscovery(void)
     // Send info about shutters
     HAssAnnounceShutters();
 
+#ifdef USE_LIGHT
     // Send info about relays and lights
     HAssAnnounceRelayLight();
+#endif // USE_LIGHT
 
     // Send info about status sensor
     HAssAnnounceDeviceInfoAndStatusSensor();


ascillato:
Hi,

Thanks for sharing your work, but this PR brakes down the discovery due to it is being sent less parameters than expected.

The parameter you just delete by not defining USE_LIGHT won't replace all vars in the HASS_DISCOVER_DEVICE and scramble the shutters data

```cpp
const char HASS_DISCOVER_DEVICE[] PROGMEM =                         // Basic parameters for Discovery
  "{\"ip\":\"%_I\","                                                // IP Address
  "\"dn\":\"%s\","                                                  // Device Name
  "\"fn\":[%s],"                                                    // Friendly Names
  "\"hn\":\"%s\","                                                  // Host Name
  "\"mac\":\"%s\","                                                 // Full MAC as Device id
  "\"md\":\"%s\","                                                  // Module or Template Name
  "\"ty\":%d,\"if\":%d,"                                            // Flag for TuyaMCU and Ifan devices
  "\"ofln\":\"" MQTT_LWT_OFFLINE "\","                              // Payload Offline
  "\"onln\":\"" MQTT_LWT_ONLINE "\","                               // Payload Online
  "\"state\":[\"%s\",\"%s\",\"%s\",\"%s\"],"                        // State text for "OFF","ON","TOGGLE","HOLD"
  "\"sw\":\"%s\","                                                  // Software Version
  "\"t\":\"%s\","                                                   // Topic
  "\"ft\":\"%s\","                                                  // Full Topic
  "\"tp\":[\"%s\",\"%s\",\"%s\"],"                                  // Topics for command, stat and tele
  "\"rl\":[%s],\"swc\":[%s],\"swn\":[%s],\"btn\":[%s],"             // Inputs / Outputs
  "\"so\":{\"4\":%d,\"11\":%d,\"13\":%d,\"17\":%d,\"20\":%d,"       // SetOptions
  "\"30\":%d,\"68\":%d,\"73\":%d,\"82\":%d,\"114\":%d,\"117\":%d},"
  "\"lk\":%d,\"lt_st\":%d,\"sho\":[%s],\"ver\":1}";                 // Light SubType, Shutter Options and Discovery version
```

Sorry, as it is can not be merged due to feature breaking.

tasmota/xdrv_12_home_assistant.ino
```diff
@@ -346,7 +348,12 @@ void NewHAssDiscovery(void)
               TasmotaGlobal.version, TasmotaGlobal.mqtt_topic, SettingsText(SET_MQTT_FULLTOPIC), PSTR(SUB_PREFIX), PSTR(PUB_PREFIX), PSTR(PUB_PREFIX2), Hass.RelLst, switch_mode, switch_name,
               stemp5, Settings->flag.mqtt_response, Settings->flag.button_swap, Settings->flag.button_single, Settings->flag.decimal_text, Settings->flag.not_power_linked,
               Settings->flag.hass_light, Settings->flag3.pwm_multi_channels, Settings->flag3.mqtt_buttons, Settings->flag4.alexa_ct_range, Settings->flag5.mqtt_switches,
-              Settings->flag5.fade_fixed_duration, light_controller.isCTRGBLinked(), Light.subtype, stemp6);
+              Settings->flag5.fade_fixed_duration, 
+#ifdef USE_LIGHT
+              light_controller.isCTRGBLinked(), Light.subtype, 
+#endif // USE_LIGHT
```
ascillato, line 354: This breakes the discovey. Parameters ofHASS_DISCOVER_DEVICE got scrambled.

ascillato (changes requested):
This breaks the discovery. Parameters of `HASS_DISCOVER_DEVICE` got scrambled.

ascillato:
Anyway, why do you want to do this?

Avamander:
> Anyway, why do you want USE_HOME_ASSISTANT to not depend on USE_LIGHT?

Both because this attempt at decoupling already exists and because I have no need for `USE_LIGHT`.

tasmota/xdrv_12_home_assistant.ino
Avamander: Oh whoops. Didn't notice it, what would be reasonable default values for this?

ascillato: Well, it is required from HATasmota module of Home Assistant, so data must be there. May be forcing to be relay?

arendst:
Pls revisit your PR as I changed the HASS discovery conforming Tasmota Discovery today. The latter already fixed the case where USE_LIGHTS was disabled.

Avamander:
@arendst 

> Pls revisit your PR as I changed the HASS discovery conforming Tasmota Discovery today. The latter already fixed the case where USE_LIGHTS was disabled.

Should be better now.

tasmota/xdrv_12_home_assistant.ino
```diff
@@ -232,10 +232,10 @@ void HassDiscoverMessage(void) {
                    SettingsText(SET_MQTTPREFIX2),
                    SettingsText(SET_MQTTPREFIX3));
 
+#ifdef USE_LIGHT
   uint8_t light_idx = MAX_RELAYS + 1;                          // Will store the starting position of the lights
```
arendst, line 236: Don't change this. As said before these values are needed for discovery.

Avamander: Should be resolved.

ascillato (changes requested):
This change don't allow HA to see the relays. So with this, for example a sonoff basic (that is not a light, just a relay) won't have anything to control.

Sorry, this PR still don't work and if USE_LIGHT is not defined, basic functionalities won't work in HA either.

Sorry, I still can't see the benefit of making HA Discovery to not depend on USE_LIGHT. The flash space that can be saved is minimal and the overcomplexity that is needed to be added to the code don't pay off the benefits. Can you explain a bit more why is this needed? Thanks.

ascillato:
> Well I did encounter this hidden dependency when trying to create a minimal build for something that is not a light. Would be nice to improve.

I agree.

I found that if you undef `USE_LIGHT` in standard Tasmota version, it can't compile even with your changes. It would be required to add this compiler preprocesor in **_xdrv_35_pwm_dimmer.ino_** and in **_xsns_01_counter.ino_**.

So, I think that it is better to improve the `HAssAnnounceRelayLight` routine (there are just 3 lines that uses light variables) instead of just not including it (due to it is a necessary part of the driver), and also to add the compiler preprocesor in the other files.

Do you want to do that? If not, I think I should be able to do it in a couple of weeks.

Avamander:
@ascillato 
> So, I think that it is better to improve the HAssAnnounceRelayLight routine (there are just 3 lines that uses light variables) instead of just not including it (due to it is a necessary part of the driver), and also to add the compiler preprocesor in the other files.
> What do you think about this?

The other files seem like content for an another PR, but in general I would say yeah that refactoring would be the nicest way. Though I'm not familiar enough with Tasmota's and HA's codebase to do it with a reasonable amount of time.

ascillato2:
> The other files seem like content for an another PR, but in general I would say yeah that refactoring would be the nicest way. Though I'm not familiar enough with Tasmota's and HA's codebase to do it with a reasonable amount of time.

Ok, no problem. Anyway, thanks a lot for pointing this out.

Closing this PR in favour of https://github.com/arendst/Tasmota/pull/13004

PR #12976, opened 2021-08-25
Add support for Ikea vindriktning

linuxrrze:
## Description:

This patch adds ESP8266 support for the IKEA Vindriktning air quality sensor
(ESP8266 not included from IKEA package, needs to be soldered in by yourself)

## Checklist:
  - [-] The code change is tested and works with Tasmota core ESP32 V.1.0.7.3


frank-f:
Am I understanding this correctly? The PR has been merged and should be included in the tasmota-sensors builds 9.5.0.7+. Do I need to set any runtime options to be able to configure a pin as Vindriktning input?

I flashed "9.5.0.8(sensors)", selected "Generic (0)" and went through every available value of (for example) "D1 GPIO5", but I don't see Vindriktning. What am I doing wrong?

barbudor:
@frank-f It has been added to the code based but not included in the default build.
You will need to compile your own binary with `#define USE_VINDRIKTNING` in your `user_config_override.h` 

Normally I should point you to [BUILDS.md](https://github.com/arendst/Tasmota/blob/development/BUILDS.md) in order to check if a specific feature or sensor is included in a given binary.
But, this time, it is not even listed there ;) 
We need to update the file

frank-f:
Thank you! That worked, I can now assign Vindriktning to a GPIO. 😊

frank-f:
> Merged in some way. Pls test.

I just finished putting mine back together and tested it right away. It works flawlessly. Would be nice to have this included in the official builds.

Bildschirmfoto 2021-09-10 um 01 25 22

(The BME sensor was added separately, of course)

frank-f:
Maybe it's a new feature request, but I want to ask here first before opening a new issue. I noticed that the sensor needs a minute or two of "warmup" time before it reports stable values. Is it possible to set an initial delay before Tasmota starts sending it's sensor data via MQTT?

probonopd:
https://github.com/timbocgn/way2dustlogger#access-the-sensor-data mentions

> pm1 is the number of 1um particles per m^3
> pm2 is the number of 2.5um particles per m^3
> pm10 is the number of 10um particles per m^3

Can TASMOTA be changed to show all of these?

Currently it only shows:
VINDRIKTNING PM 2.5 µm

frank-f:
There‘s compiletime options for those already, but the numbers are bogus from what I understand so far, since this specific sensor can only detect 2.5 um particles, but it’s protocol is shared between different models.

barbudor:
@probonopd Please refer to Discussion https://github.com/arendst/Tasmota/discussions/12810#discussioncomment-1259303
The datasheet confirm that only PM2.5 is valid on that model
The protocol is generic for multiple model but this one is a stripped down version for cheap devices
This will not be come back.

If you want better particule sensor, go for PMS5003.
HM330X is also on its way

probonopd:
Ah, I see - thanks for the clarification. Hard to complain for a 10 EUR device.

PR #12973, opened 2021-08-25
SwitchMode11 & 12: New SetOption129 for delaying single press event

Seestern93:
## Description:
Adds a new SetOption129 which can be used with SwitchMod11 & 12 to delay the reaction to single press events so that double press events are not triggering single press events too.

**Related issue (if applicable):** fixes #12712  (and discussion #12713)

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [x] The code change is tested and works with Tasmota core ESP32 V.1.0.7.3
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #13007, opened 2021-08-29
SwitchMode11 & 12 (Dimmer) new state#10 as delayed single-press

Seestern93:
## Description:
New proposal for solving #12973.

With this, at SwitchMode11 (and SM12 inverted) at first button press the Switch#State=2 event is triggered ("POWER_TOGGLE") and after release and without second button press, NEW Switch#State=10 event is triggered ("POWER_DELAYED").
No SetOption is required, since both events are triggered -> needed behavior can be decided over rules which fits to the other behavior of SwitchMode11 (&12).

Only thing which I could not achieve is that State#10 is also sent over MQTT with SetOption114 active (as it does for "POWER_CLEAR", "POWER_INCREMENT", "POWER_RELEASE" etc. even though I edited the tasmota.h file which holds the enum variable as well as kSwitchPressStates variable inside the support_switch.ino file. (But this is not mandatory for the needed functionality).

Changes are tested and are working as expected.

NOTE: Previous changes which implemented SwitchMode17 & 18 were reverted in the support_switch.ino, but changelog is not touched.

**Related issue (if applicable):** fixes  #12973

## Checklist:


arendst:
Thx. Looks much better than mine.

Seestern93:
> 
> 
> Thx. Looks much better than mine.

And also better than my original proposal (as you can use both events simultaneously and individually for each switch) 😃 
I have already created a PR [#796](https://github.com/tasmota/docs/pull/796) for updating the docu with explaining that new state.
Thank you for merging! 😁

PR #13034, opened 2021-09-01
add command WebGetConfig

barbudor:
## Description:

This is a proposal for an optional command WebGetConfig that can be enabled with `#define USE_WEBGETCONFIG`.
As the command name suggest, this command can pull a configuration file (DMP) from an HTTP URL.
What can this be used for ?
This can be useful for some form of auto-provisioning with very limited settings in `user_config_override.h`, a new device, or a known device that has been reset, will reconfigured itself from the last known good configuration.

Simple `user_config_override.h`:
```
#define STA_SSID1         "MYSSID"             // [Ssid1] Wifi SSID
#define STA_PASS1         "MYPASSWD"     // [Password1] Wifi password
#define USER_BACKLOG      "WebGetConfig http://myserver:8000/tasmota/conf/%id%.dmp"
#define USE_WEBGETCONFIG
```
%id% will be replaced by the MAC address which on a brand new unit is the only element to identify the device.

Possible use-case:
- If the device is known and a DMP file has been save named after the MAC address, if configuration got corrupted or too much twisted, a simple `reset 1` and the device will reconfigure itself from the last good configuration.
- If a known device as died and replaced by another of the same model, the DMP file can be renamed to the MAC address of the new device, and right after flashing, the new device will configure itself automatically to replaced.
- Slightly more advanced: using decode-config and a reference DMP or JSON, it could be possible to automate the provisioning of new devices by editing the JSON and rebuild a new DMP named after the MAC address of the new device.
The URL could point to a web script (`WebGetConfig http://myserver:8000/tasmota/provision.Php?id=%id%`) that would automate the use of decode-config to build the DMP from a database.

Location of this command in `xdrv_01_webserver.ino` makes it dependent to `USE_WEBSERVER` although it is not related to WebServer at all. Someone disabling WebServer could not use it.
It could be moved to another module.

What do you think ?

Beside, I also took the opportunity to replace numerical status in `WebSend` (and `WebGetConfig`) with an `enum WebCmndStatus`.

## Checklist:


arendst:
I'll merge this but do not announce yet. Let users play with this and see what happens.

arendst:
Released! I like it.

barbudor:
Thanks
I think at some point it will be interesting to have a doc page on automated provisioning to list different ways of doing it.
I will also look at similar command through Mqtt to leverage your Mqq-file-transfer

crispy78:
I like this feature and it could be great. Maybe I forgot some options but whenever I save my config from the GUI it's beens saved as "Config_[devicename]_[softwareversion].dmp". Could this be changed so it's more in line with this change?

barbudor:
For most users using the web GUI to save/restore  configuration files, it is much more convenient to have a human readable file name.
Nothing prevent you to use the command with such a name like for example a 'webgetconfig' triggered through MQTT from a provisioning backend.
Alternatively you can also use symbolic link to make a link based on the Mac address of the device to use an hard-coded rule with "%id"

Lot of options. 😁

eku:
You could also use existing mechanisms of the DHCP protocol and send the name/path of the config file in the DHCP reply. Just my opinion.

barbudor:
Most of users doesn't have a DHCP server that allows to configure such feature. My ISP's router doesn't.
There was a time where Tasmota was also able to retrieve MQTT server and/or NTP server from DHCP but this was removed. I think there was some reliability issues on this with Arduino's DHCP.

paolomeraviglia:
Hey there,
Running Tasmota 13.4.0(tasmota-4M) here, but seems something is not working

17:27:29.233 CMD: WebGetConfig
17:27:29.241 MQT: stat/sonoff/RESULT = {"Command":"Unknown"}
17:27:51.339 CMD: WebGetConfig http://test.io/dmp
17:27:51.347 MQT: stat/sonoff/RESULT = {"Command":"Unknown"}

Is this thing merged and automatically enabled on official builds?

sfromis:
It is explicitly not enabled in standard precompiled builds, as the first line of the post says:
> This is a proposal for an optional command WebGetConfig that can be enabled with
> #define USE_WEBGETCONFIG

As the post suggests, the core use case is for use as a command you can include when doing a custom build. "Normal" restore of config is in the web UI.

Issue #10010, opened 2020-11-29
Enable data with more than 32 bits in xdrv_17_rcswitch.ino


joetriskaide:
**Have you looked for this feature in other issues and in the docs?**  Yes
The rc-switch library in tasmoa is capable of sending such payload , the CmndRfSend command in xdrv_17_rcswitch.ino not.

**Is your feature request related to a problem? Please describe.**  
I won a Dooya 1603 remote which uses payload data of 40 bits

**Describe the solution you'd like**  
xdrv_17_rcswitch  xdrv_17_rcswitch.ino
line 101, change to  `unsigned long long data = 0;`
line 112 change to ' `data = root.getULong(PSTR(D_JSON_RF_DATA), data);`'
**Describe alternatives you've considered**  
None

**Additional context**  
I tried the changes on gitpod and they work

**(Please, remember to close the issue when the problem has been addressed)**

Jason2866:
If the code size increase is not too big a provided PR will be merged

nosaj66au:
Hi Guys, Been working and testing in this area it appear the fix above only addresses the transition section of the code, enabling 40 bits of data to be transmitted. Can you please also add it to the receive section.
Jason

PR #13076, opened 2021-09-06
implement rcswith rfreceive more than 32 bits

barbudor:
## Description:

Following https://github.com/arendst/Tasmota/issues/10010
and comment https://github.com/arendst/Tasmota/issues/10010#issuecomment-913284145

I extended the data to support more than 32 bits RF codes.

Limited testing with 2 ESP connected back to back. Not very reliable but the fix looks ok
Note that when SO28 is on (Display as decimal) and the code length is >32 bits, it will still be displayed as HEX because we are lacking a function to convert 64 bit integers to ASCII.

Need more testing, may be by @nosaj66au

## Checklist:


tasmota/xdrv_17_rcswitch.ino
```diff
@@ -51,22 +51,21 @@ uint32_t rf_lasttime = 0;
 void RfReceiveCheck(void) {
   if (mySwitch.available()) {
 
-    unsigned long data = mySwitch.getReceivedValue();
+    unsigned long long data = mySwitch.getReceivedValue();
```
s-hadinger, line 54: Please prefer `uint64_t` as it makes its size explicit

tasmota/xdrv_17_rcswitch.ino
```diff
@@ -51,22 +51,21 @@ uint32_t rf_lasttime = 0;
...
     unsigned int bits = mySwitch.getReceivedBitlength();
     int protocol = mySwitch.getReceivedProtocol();
     int delay = mySwitch.getReceivedDelay();
 
-    AddLog(LOG_LEVEL_DEBUG, PSTR("RFR: Data 0x%lX (%u), Bits %d, Protocol %d, Delay %d"), data, data, bits, protocol, delay);
+    char stemp[32];
+    ext_snprintf_P(stemp, sizeof(stemp), PSTR("\"0x%_X\""), &data);
+    AddLog(LOG_LEVEL_DEBUG, PSTR("RFR: Data 0x%s, Bits %d, Protocol %d, Delay %d"), stemp, bits, protocol, delay);
```
s-hadinger, line 61: You can use `0x%_X` directly in AddLog, it calls ext_snprintf_P behind the scenes

tasmota/xdrv_17_rcswitch.ino
barbudor: done

tasmota/xdrv_17_rcswitch.ino
barbudor: Yes, but we may need also the Hex 64 Bits later for the JSON.
So I think it is more efficient to do it once.
Let me know @s-hadinger

s-hadinger (approved):

nosaj66au:
Hi Guys,
   Thanks for you effort on this. I've compiled the changes and it works for existing remotes, I've tested a 12bit (protocol 6) and a 24bit (protocol 8). However it has changed the output format adding leading zero's e.g. -12bit - 'Data":"0x7D" is now "Data":"0x000000000000007D". (not a problem for me but just letting you know).
 My initial problem was with the Dooya 1603 protocol (35). MQTT command .../RFsend  payload {"Data":"0x3A354E5155","Bits":40,"Protocol":35}. This a still not returning any message to the console?? 

Thanks again for you help. 
I'm happy to compile and help where possible. :)   

image

barbudor:
| My initial problem was with the Dooya 1603 protocol (35). MQTT command .../RFsend payload {"Data":"0x3A354E5155","Bits":40,"Protocol":35}. This a still not returning any message to the console??

@nosaj66au in https://github.com/arendst/Tasmota/issues/10010#issuecomment-913284145 you explicitly requested implementation of >32 bits on Receive.
Same on https://github.com/arendst/Tasmota/discussions/13055 where you only refer on Receive

I can't find any report from you regarding RfSend. Did you reported it ?
If you haven't reported it yet, please open a new issue and tag me

nosaj66au:
Hi barbudor,
      Sorry should have explained myself better, my problem is with the RFReceive. I noticed you said you were testing using 2 devices back to back, for this reason I include my RFSend was to give you something "really" to test the RFreceive function with.
This RFsend message I provided works correctly and operates the blinds using the current build. and I have confirmed this with and adruino sketch (uno) that I used initial to sniff the RF from remote to get the hex message. 
Also I too have back to backed 2 generic devices with an RF transmitter and an RF receiver and also tried directly wiring data TX to data RX (no RF) neither provide the RF receive data in the console for the Dooya remote (40 bit), however but the 12bit and 24bit remotes still work.

barbudor:
Ok, but I'm still confused as I am unsure if there are still problems
Can you know receive 40 bits from the Dooya remote with Tasmota ? (or does only your Arduino sketch works ?)

Regarding 2 devices connected back to back, I found that unreliable. Even with 24 bits, I didn't always got the right data. Although I was using short (10cm) twisted wires between the NodeMCU and the D1. I didn't bothered testing with shielded cable.

nosaj66au:
Sorry if I confusion you. Please accept my apology.  I'm still have a problem not getting the Dooya RFRecieve Messaging in the Console. The Blind and Arduino sketch both work/receive the message fine. I'm sending the message using RFsend from a second device. If use the original remote same result only the blind and arduino work.
Hope this help. again thankyou very much
Jason

Jason2866:
The protocol is not recognized / decoded from the current implementation in Tasmota.
Since you have a working Arduino sketch to decode, provide the sketch and the detailed data (logs, protocol...). 
Since you are the only one having the device, no one else can do!

nosaj66au:
Hi Jason2866,
   Not sure you understanding . This is not a problem with the protocol or tasmota’s ability to recognise the protocol. Dooya 1602 and 1603 are the same chipset and RF bit pattern.  I believe that for some reason  rfrecieve will not return data from rfswitch when the protocol has  a bit length > 32 bit. At the moment the only device and protocol I have to test this is protocol 35 (dooya) Now regardless of the protocol being supported or not, the timing  etc parameters in the rfswitch.cpp protocol definitions should be bye directional . In other words if I use a protocol definition to RFsend a message it should align when being received. Thus returning data in the console and mqtt json output. I’m happy to provide the arduino sketch but I believe it will be little use if my assumption are correct.
Please correct me if you feel my assumption are incorrect.  
mall the best.
jason

barbudor:
Have you tried enabling WebLog 3 ?
The code of xdrv_17_rcswitch.ino is very simple. Function `RfReceivedCheck()` calls the library and if anything is returned , it should at least display a LOG_DEBUG.
If you still get nothing it is likely that the library is not returning anything to Tasmota.

In that case, as you have a working Arduino sketch, can you try pickup the library used by Tasmota (from Tasmota/lib/lib_rf/rc-switch/) and use this one with your sketch.
This will help to understand if there is a limitation between the "official" library in the Arduino IDE and the one used by Tasmota.

nosaj66au:
I’ll try the web log. However my arduino sketch is one I wrote and does not use a library. It is merely bit banging or to be more accurate simple time measurement.
Jason

barbudor:
Ok, so that means we have no clue if rcswitch can support this remote (except that one model is listed).
In that case I would suggest that you write a test sketch based on the official rcswitch library.
If this prove non-successfull it could be escalated as an Issue with rcswitch.
Once fixed in rcswitch, we would backport the update.

If it works with the official library, back to my previous message, testing with the tasmota version of the lib.

Jason2866:
Just did a commit to improve documentation of protocol format in lib rc-switch

nosaj66au:
Hi Guys, Did some quick testing before work, I'll continue tonight. I used RFsend to send my example 40bit message over several protocols. On reviewing the RFswitch code length is not a part of the definition. interesting results,
protocol 1,2,6,8,10,17 returned a value. 3,4,5,7,8, 11 to 16, 19 didn't. However 9 returned as 8, 16 returned as 8, and 22 return as 2, this is most likely  because of similar timing very understandable. For the ones that returned correctly they also worked on 12 and 24 bit. The ones that didn't work just don't work on any size.  I'll continue investigating.
Jason

nosaj66au:
Hi Guys, I have done a bit more investigation, In theory If we send the data from one device to another in a back to back config., it should work every time because the same timing is used to produce and detect/receive right??? There is nothing in RFswitch.CPP send or receive code (tasmota or RCswitch repo)  to account for drift cause by the code path/code execution. The protocols that work seem have longer start bits and therefor are more tolerant when detecting, and less susceptible to code execution time. this also may account for the 60% tolerance factor in the code. To make the whole thing harder the receive code uses an interrupt, but the send is bit banging, its ok in theory but compounds the problem and should really be done with a timed interrupt. unfortunately, Anyway, I don't have time to boil the ocean, so I may just try to get it to work for my protocol.      
Another thing, the Tasmota RFswitch.cpp is vastly different to the current RFswitch version, not sure why? Is it just an old version or was changed for some reason.  In short this may be a bigger problem than just getting one protocol to work. Is there any more details on protocol definition in the rfswitch.cpp. Is does not explain the protocol params very well.
Jason

Jason2866:
As explained earlier, we do not have the device to test. Without it makes no sense to work on this.
The implementation of rc-switch in Tasmota is a advanced version, mainly based on this fork https://github.com/1technophile/rc-switch. All infos we have are in the source file.
Sorry, we can help you here any further.

PR #12249, opened 2021-05-31
WebSend http PUT with JSON payload

cjstephan:
## Description:

WebSend may now use http PUT for starting my vacuum cleaner.

## Checklist:
  - [X] The pull request is done against the latest development branch
  - [X] Only relevant files were touched
  - [X] Only one feature/fix was added per PR and the code change compiles without warnings
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.9
  - [ ] The code change is tested and works with core ESP32 V.1.0.6
  - [X] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

Sorry, I did not found out how to run CI tests. Is Github doing this automatically on push event?
This is my first try of a pull request so please be lenient.

PR #13209, opened 2021-09-24
Implement WebQuery Command

JVital2013:
## Description:

Implements new WebQuery command. The command allows for GET, POST, PUT, and PATCH HTTP Queries, complete with Request Headers and request body (when applicable).

Syntax examples:
`WebQuery http://192.168.1.1/path GET`: Sends HTTP GET http://192.168.1.1/path
`WebQuery http://192.168.1.1/path POST {"some":"message"}`: Sends HTTP POST to http://192.168.1.1/path with body {"some":"message"}
`WebQuery http://192.168.1.1/path PUT [Autorization: Bearer abcdxyz|Second:Header] potato`: Sends HTTP PUT to http://192.168.1.1/path with authorization header, a second header, and body "potato"
`WebQuery http://192.168.1.1/path PATCH patchInfo`: Sends HTTP PATCH to http://192.168.1.1/path with body "patchInfo"

An unlimited number of headers can be sent per request, and a body can be sent for all command types. The body will be ignored if sending a GET command

This command enables Tasmota to send advanced web queries in addition to GET. It is my vision that this is an open-ended command that can control other SmartHome Devices, such as Philips Hue Hubs, the vacuum cleaner mentioned in #12249, and anything else that can be controlled over HTTP. It is designed to be more flexible than other Tasmota HTTP PRs I've seen

Let me know if you think anything needs changed!

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.1.0.7.4

### DIFF
--- a/tasmota/xdrv_01_webserver.ino
+++ b/tasmota/xdrv_01_webserver.ino
@@ -3098,14 +3098,86 @@ int WebSend(char *buffer)
       }
       url += F("cmnd=");                      // url = |http://192.168.178.86/cm?cmnd=| or |http://192.168.178.86/cm?user=admin&password=joker&cmnd=|
     }
-    url += command;                           // url = |http://192.168.178.86/cm?cmnd=POWER1 ON|
+    url += command;
+    url += F(" GET");                         // url = |http://192.168.178.86/cm?cmnd=POWER1 ON|
 
     DEBUG_CORE_LOG(PSTR("WEB: Uri |%s|"), url.c_str());
+    status = WebQuery(const_cast<char*>(url.c_str()));
+  }
+  return status;
+}
 
-    WiFiClient http_client;
-    HTTPClient http;
-    if (http.begin(http_client, UrlEncode(url))) {  // UrlEncode(url) = |http://192.168.178.86/cm?cmnd=POWER1%20ON|
-      int http_code = http.GET();             // Start connection and send HTTP header
+int WebQuery(char *buffer)
+{
+  // http://192.168.1.1/path GET                                         -> Sends HTTP GET http://192.168.1.1/path
+  // http://192.168.1.1/path POST {"some":"message"}                     -> Sends HTTP POST to http://192.168.1.1/path with body {"some":"message"}
+  // http://192.168.1.1/path PUT [Autorization: Bearer abcdxyz] potato   -> Sends HTTP PUT to http://192.168.1.1/path with authorization header and body "potato"
+  // http://192.168.1.1/path PATCH patchInfo                             -> Sends HTTP PATCH to http://192.168.1.1/path with body "potato"
+
+  // Valid HTTP Commands: GET, POST, PUT, and PATCH
+  // An unlimited number of headers can be sent per request, and a body can be sent for all command types
+  // The body will be ignored if sending a GET command
+
+  int status = WEBCMND_WRONG_PARAMETERS;
+  int http_code;
+  WiFiClient http_client;
+  HTTPClient http;
+  String methodStr;
+  String bodyStr;
+  char *url;
+  char *header;
+  char *headerBody;
+  char *body;
+  char *method;
+  char *temp;
+  bool headerFound;
+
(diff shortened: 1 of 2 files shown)


tasmota/xdrv_01_webserver.ino
```diff
@@ -3098,14 +3098,86 @@ int WebSend(char *buffer)
...
+          }
+        }
+        else { //No header information was sent, but there was a body
+          body = temp;
+        }
+      }
+      
+      //No body sent; define empty string for the body
+      else {
+        body = (char*)malloc(1);
```
s-hadinger, line 3164: I'm afraid that's a memory leak, I don't think this is ever freed.

s-hadinger:
Thanks that's a smart approach. Did you measure the difference in code size?

arendst:
Code size increase is 1k

arendst:
Implemented today.

tasmota/xdrv_01_webserver.ino
JVital2013: Thanks for the feedback! I missed that. It looks like Theo fixed this in his implementation

ygelt:
Nice function, I use it extensively :-)
In case the wifi OR the web service is unavaiable, is there a way to detect it ?

barbudor:
You can use rules to detect wifi connectivity:
on wifi#disconnected do ...
on wifi#connected do ...

By webservice I assume you mean the server you are trying to connect to ?
That should raise an error such as "Connect Failed" instead of "Done"

ygelt:
Many thanks for your fast answer, I will start to try your solution. I'm using WebQuery in scripting but still newbie ! ;-)

sfromis:
Did a small test with WebQuery, first a working URL (got data), and switching the other device off, I got the expected `{"WebQuery":"Connect failed"}`. Also if failing DNS lookup. Any caching done would be outside of Tasmota, like in your web service framework.

ygelt:
Thanks for your check.

On my side I checked what happens when calling the web service which is removed, it returns a 404 error and a system fault string.
I did not find how to get the http error code. 
In my case the system fault string which is returned is 408 characters long and it is unfortunately too large ? (I can't change it :.( )

sfromis:
Then the connection was ok, hence no reason for getting a "Connect failed". And it does indeed look like code 404 does not give error symptoms, and still no data with a very short 404 response text.

ygelt:
Understood !
So it would be useful to have a system variable with the last HTTP status code to improve web services integration.

My script needs to trigger an audible alarm when the tasmota device does not get a correct web service answers after N bad retries.
In my case (but not only!) there is a "long" body returned on error 404 but unfortunately not standardized, not even always JSON-compliant :-(

sfromis:
One workaround could be to yell if you did **not** get a valid response within a reasonable time frame.

ygelt:
It would be a good solution if I could distinguish a good answer from a bad answer.
I use a JSON {"alv":1} to say "I'm alive" but unfortunately when the web service "goes out", WebQuery/WebSend always returns the last answer, probably because either the 404 error has no "alv" or is not JSON. So I indefinitely get 1 in WebQuery#alv :.(

sfromis:
WebQuery is **not** returning the last answer in case of errors. Any caching you may experience is outside of what Tasmota controls or can act on.

Inside of what Tasmota gets, it can work great if you do not worry about trying to detect badness, but only detect good answers. Then you can yell if last good answer was too long ago.

sfromis:
Not like that. What you *can* do is to read the precise messages you get in JSON format, and create rules to match that. `{"WebQuery":"Done"}` or `{"WebQuery":"Connect failed"}` if not able to connect at all, but 404 still gets "Done".

No "hidden" features like status code. If you use a build with `USE_WEBSEND_RESPONSE`, you can get the response text in the JSON instead of just `"Done"`. Still no status code.

If you run on ESP32, instead of `WebQuery` you can use the Berry language with the `webclient` module giving you regular features including status code, headers etc and all that.

PR #13250, opened 2021-09-30
Add support for HM330X for SeedStudio Grove Particule sensor

barbudor:
## Description:

Discussion https://github.com/arendst/Tasmota/discussions/13212

### Driver for Particule sensor SeedStudio Grove HM3301
This driver use the I2C mode. Only address 0x40 is supported by the sensor so it is hardcoded  the driver without any possibility to change it (without editing and recompiling)

By default the sensor remains active all the time but no data are provided for the first `HM330X_WARMUP_DELAY` seconds (to allow the sensor to warm up). During warmup, no data are returned in the GUI or through SENSOR message in order to provide invalid data. In case of any doubt, logs with level 3 will confirm right at boot if the sensor has been detected.

Unless decided otherwise after this PR, this sensor is not included by default in any standard build.
It must be enabled in a custom build with `#define USE_HM330X`.

### Low power / life extension
It is possible to connect a GPIO to pin SET of the module and assign it to `HM330X SET` for automatic sleep in order to extend the life time of the sensor. When this is used, activity of the sensor in synced with Teleperiod as follow
1) At start, sensor is immediately enabled by setting `HM330X SET` HI and start warmup. If any Telemetry occurs before the end of the warmup, no data are provided and warmup goes on.
2) After HM330X_WARMUP_DELAY, data are available and reported in web and SENSOR
3) Immediately after the next teleperiod SENSOR message, if Teleperiod is greater than (HM330X_WARMUP_DELAY + 2) the sensor is put to sleep by taking the pin `HM330X SET` LOW and a new cycle starts.
4) _New cycle_: (HM330X_WARMUP_DELAY + 2) seconds before the next teleperiod, the sensor is awaken by taking the pin `HM330X SET` HI for warmup. No data are available yet.
5) 2 seconds before Telemetry, warmup is completed and data are available to Web until Telemetry messages are sent
6) Back to 2)

### Overridables (in user_config_override.h)
- `HM330X_DEFAULT_ADDRESS       0x40` : in case SeedStudio makes this changeable in the future
- `HM330X_WARMUP_DELAY          30`    : time for warmup, data are not pulled before that delay
- `HM330X_HIDE_OUT_OF_DATE      false`  : if changed to true web data and sensor will not be published when outdated

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.1.0.7.4


barbudor:
Tagging @Teleportist for merge in Tasmota notification

PR #13253, opened 2021-09-30
Berry add `tasmota.global`

s-hadinger:
## Description:

Experimental feature: now `tasmota.global` allows read/write access to `TasmotaGlobal`. It is still early access and for now only 2 attributes are available:
- `tasmota.global.energy_driver`
- `tasmota.global.uptime` (in seconds)

Also fix in `energy` attributes mapping.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.1.0.7.4


arendst:
For my understanding why do you want towrite to ``uptime`` and/or ``energy_driver`` from Berry?

Both action will possibly fail "normal" tasmota functionality;

- uptime is used as a "sequence motor" where at specific times specific actions take place in order of uptime.
- energy_driver is zero if the default energy driver has not found a sub-driver and stops amy action. If it becomes non-zero afterwards the energy_driver might start to function and calls the other function calls within an energy driver.

What's wrong with adding a dedicated Berry energy driver in line with the other energy drivers?

s-hadinger:
It was more a proof-of-concept to see how easily I could map some constants to surface them in Berry. `uptime` is a random choice to check that it works. It's not specially useful.

For energy driver, we're still exploring. I hoped that we could save adding a Berry energy driver and do everything from Berry. Let's dive into it.

PR #13268, opened 2021-10-03
PMS3003 doesn't need a specific build option

briaeros:
## Description:
With theses modification, Tasmota could read data from sensor pms3003, 5003 and 7003 without different version.
The code does it by managing the different framelength beetween 3003 and the other sensor.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.1.0.7.4

I could test only with a pms3003, since I doesn't have a pms5003/7003

### DIFF
--- a/tasmota/xsns_18_pms5003.ino
+++ b/tasmota/xsns_18_pms5003.ino
@@ -44,14 +44,6 @@
 
 TasmotaSerial *PmsSerial;
 
-struct PMS5003 {
-  uint16_t time = 0;
-  uint8_t type = 1;
-  uint8_t valid = 0;
-  uint8_t wake_mode = 1;
-  uint8_t ready = 1;
-  bool discovery_triggered = false;
-} Pms;
 
 enum PmsCommands
 {
@@ -70,6 +62,7 @@ const uint8_t kPmsCommands[][7] PROGMEM = {
     {0x42, 0x4D, 0xE1, 0x00, 0x00, 0x01, 0x70},  // pms_set_passive_mode
     {0x42, 0x4D, 0xE2, 0x00, 0x00, 0x01, 0x71}}; // pms_passive_mode_read
 
+typedef struct pmsX003data pmsX003data;
 struct pmsX003data {
   uint16_t framelen;
   uint16_t pm10_standard, pm25_standard, pm100_standard;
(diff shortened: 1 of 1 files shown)


barbudor:
Hi @briaeros 
Could you take the opportunity of this refactoring to consolidate `pms_data` as a member of `Pms` and make this a dynamically allocated structure that would be allocated only if a PMS is used (after having tested for PinUse()) ?
That would save memory when a driver is included into a build but not used.
Thanks

tasmota/xsns_18_pms5003.ino
```diff
@@ -269,45 +290,43 @@ void PmsInit(void)
...
   "{s}PMS5003 " D_PARTICALS_BEYOND " 2.5 " D_UNIT_MICROMETER "{m}%d " D_UNIT_PARTS_PER_DECILITER "{e}"
   "{s}PMS5003 " D_PARTICALS_BEYOND " 5 " D_UNIT_MICROMETER "{m}%d " D_UNIT_PARTS_PER_DECILITER "{e}"
   "{s}PMS5003 " D_PARTICALS_BEYOND " 10 " D_UNIT_MICROMETER "{m}%d " D_UNIT_PARTS_PER_DECILITER "{e}";      // {s} = <tr><th>, {m} = </th><td>, {e} = </td></tr>
-#endif  // PMS_MODEL_PMS3003
 #endif  // USE_WEBSERVER
 
 void PmsShow(bool json)
 {
-  if (Pms.valid) {
+  if (Pms.valid and pms_data.ok == 1) {
```
arendst, line 314: Did the compile work with the "and"?

I would expect an "&&"

briaeros: It did compile with and and works fine, but I agree it is subpar.
I've replaced it in the latest commit (allocated data structure).

briaeros:
> Hi @briaeros Could you take the opportunity of this refactoring to consolidate `pms_data` as a member of `Pms` and make this a dynamically allocated structure that would be allocated only if a PMS is used (after having tested for PinUse()) ? That would save memory when a driver is included into a build but not used. Thanks
Hi @barbudor ,
Thanks for the input.
I've done that in the new commit 'PMSX003 dynamically allocated data structure'.
Theses modifcation works fine in my test ring.

cordially

tasmota/xsns_18_pms5003.ino
barbudor: Just for our culture as I was very surprised too :
https://en.cppreference.com/w/cpp/keyword/and
Seems to be there since at least C++98
Every day brings an opportunity to learn something 🤣
When C++ becomes Pythonic 🤦‍♂️

ascillato2:
Bump. This PR seems ok to be merged.

PR #13438, opened 2021-10-22
feat: added GY-US42 I2C device

conrad784:
## Description:

I wanted to use this sensor for a project and did not find an implementation in tasmota, so I took some guidelines from other range sensor implementation and implemented the needed calls.

This is my first new sensor, so I don't know if I found all the right spots to adapt, but I tried to follow the official [guidelines](https://tasmota.github.io/docs/Sensor-API/).

I had it running for development on a Wemos D1, and now it runs on an ESP32 board, so both should be working. I had to disable the SHT3X sensor for it to work as it shares the same address (should this be part of the implementation?)
Like?
```
#if defined(USE_SHT3X)
  #ifdef USE_SHT3X
  #undef USE_SHT3X
  #warning **** Turned off conflicting drivers SHT3X
  #endif
#endif
```

## Checklist:

### DIFF
--- a/I2CDEVICES.md
+++ b/I2CDEVICES.md
@@ -97,3 +97,4 @@ Index | Define              | Driver  | Device   | Address(es) | Description
   61  | USE_T67XX           | xsns_89 | T67XX    | 0x15        | CO2 sensor
   62  | USE_SCD40           | xsns_92 | SCD40    | 0x62        | CO2 sensor Sensirion SCD40/SCD41
   63  | USE_HM330X          | xsns_93 | HM330X   | 0x40        | Particule sensor
+  64  | USE_GYUS42          | xsns_94 | GY-US-42 | 0x70        | Time-of-flight (ToF) distance sensor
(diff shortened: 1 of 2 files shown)


conrad784:
This device is still not in Tasmota, is there a reason why it should not be merged?

arendst:
Many reasons why it's not merged yet.

- Another TOF sensor
- As it stands it's not ready to merge as it needs better more "not-present" detection
- I haven't found the time to test it's overhead once merged and the hardware is not present.
- As can be seen from other PR's I'm swamped with sensors needing a dedicated sensor number from the (currently) max sensor list of 126.

conrad784:
Fully understand, you are right there are good alternatives already available.

I don't fully understand the device numbers. Could there be the option for compile-time only sensors, that they are tracked in this repository but only enabled if people really want to?

barbudor:
> Fully understand, you are right there are good alternatives already available.
> 
> I don't fully understand the device numbers. Could there be the option for compile-time only sensors, that they are tracked in this repository but only enabled if people really want to?

It's already the case but still each existing driver, compiled in or not, needs to have it's own unique id.
For example this allows commands like `Sensor<X>` or Ì2Cdriver<X>` to refer a unique ID which do not depend on which driver/sensor are compiled-in.

PR #13507, opened 2021-10-29
Reflect PVVX firmware reed switch status

Lenbok:
## Description:

The first bit of the flag field reflects reed switch status so this device can additionally be used to monitor the status of a door. 

See: https://github.com/pvvx/ATC_MiThermometer#reed-switch-on-gpio-pa6-label-on-the-p8-pin

Since existing door monitors seem to use the Btn field, this is what I used in this  case.

## Checklist:

### DIFF
--- a/tasmota/xsns_62_esp32_mi_ble.ino
+++ b/tasmota/xsns_62_esp32_mi_ble.ino
@@ -1586,6 +1586,10 @@ void MI32ParseATCPacket(const uint8_t * _buf, uint32_t length, const uint8_t *ad
         MIBLEsensors[_slot].bat = ppv_packet->battery_level;
         MIBLEsensors[_slot].eventType.bat  = 1;
 
+        MIBLEsensors[_slot].Btn = (ppv_packet->flags) & 0x1; // First bit is reed switch status
+        MIBLEsensors[_slot].eventType.Btn = 1;
+        MIBLEsensors[_slot].feature.Btn = 1;
+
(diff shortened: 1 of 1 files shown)


Lenbok:
Why are you using pvvx custom format? Is it to make use of the temp/hum trigger stuff instead of reed switch? I wondered about possibly adding support to reflect that too, but wasn't sure how to best reflect it in the status. 

If you're not using those then you should probably just turn off custom format (unless there's some other reason for it I am not aware of).

I think it's more useful to have pvvx custom format work out of the box with tasmota than worry about people who have actively enabled custom format getting additional JSON elements they are not using. There are probably plenty of people with sonoffs / shellys where they can't press the button due to mounting or lack of wiring either.  

Additionally, having it #ifdef'ed seems wrong, as it's the end-BLE device that may or may not have the reed switch attached, not the Tasmota gateway. E.g. you can have a bunch of these devices where some use and some do not use the reed switch (and temp/hum trigger if we add support for reflecting that), all reporting to the same gateway. Either the device should not report un-needed functionality in the transmission (e.g. turn off custom format), or handle it somewhere downstream that can apply per-device logic (home assistant, node red, etc).

Lenbok:
@barbudor (or @ascillato2) Do you have any further comments/clarifications on the above to help this PR progress?

barbudor:
Not from my side.
I see no problem if @arendst is OK with it.
I just gave my 2c but agrees it shouldn't be a problem

PR #13565, opened 2021-11-04
Add command TCPConfig to TCPBridge

barbudor:
## Description:

Adding Command `TCPConfig <configstring>` to change the mode from the default 8N1
The `<configstring>` is the standard 3 characters mode such as `8N1`, `7E1` etc ...
Supports:
- 5 to 8 bits
- N, E, O parity
- 1 or 2 stop bits

I haven't added the support for a numeric parameter (0..23) like `SerialConfig`...

To limit code size, I moved the parsing from `CmndSerialConfig()` into a new `ParseSerialConfig()` function in order to make it re-usable.
Also I changed `GetSerialConfig()` to support an optional parameter inn order to use it with other config than `Settings->serial_config`

In order to preserve backward compatibility with flash settings where the `tcp_config` field was not present (defaults to 0), I set the MSBit to 1 to state the value is valid. A value without this bit set is interpreted as default `8N1`.

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.1.0.7.5

### DIFF
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -733,8 +733,9 @@ typedef struct {
   uint8_t       shd_leading_edge;          // F5B
   uint16_t      shd_warmup_brightness;     // F5C
   uint8_t       shd_warmup_time;           // F5E
+  uint8_t       tcp_config;                // F5F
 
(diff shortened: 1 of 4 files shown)


arendst:
Nice reuse of available code!

PR #13583, opened 2021-11-06
Fix OneWire on ESP32-C3

arlosi:
## Description:

Fixes OneWire for ESP32-C3. Tested with an ESP32-C3 and two DS18B20 sensors on the bus. Bus is not using parasite power.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.1.0.7.5


s-hadinger:
Thanks for your PR

PR #13595, opened 2021-11-07
Add simple stairs effects to ws2812 driver

kruzer:
## Description:
This poc adds an effect switching light step by step usefull in the stairs intalations.
I've added one config variable: StepPixels - it defines the number of leds in each step, if equals 0 the effects are working as before.
Rotation - operates bitwise (b0 - changes scheme direction, b1 - changes effect direction)
Fade - generates gradual change in color (Width4 defines length)

an example, how it is working now:
https://youtu.be/oQxWOPwuRqw?t=217

## Checklist:


ascillato (approved):

blakadder:
very nice, its a very useful effect

curzon01:
there is probably an overflow when saving light_step_pixels, `uint8_t` is to small to store values > 255:

```C++
  uint8_t       light_step_pixels;				 // F60
```
```C++
const uint16_t WS2812_MAX_LEDS = 512;       // Max number of LEDs
```
```C++
void CmndStepPixels(void)
{
  if ((XdrvMailbox.payload >= 0) && (XdrvMailbox.payload <= WS2812_MAX_LEDS)) {
    Settings->light_step_pixels = XdrvMailbox.payload;
```

ascillato:
@curzon01 

Good eye!

PR: https://github.com/arendst/Tasmota/pull/13767

arendst:
No there is not. This value is supposed to be the number of pixels in ONE step only. So it would need to be very wide stairs when the overflow occurs.

NB I don't like the code as it breaks all conventions with regard to parameter use and functionality; scheme 13 overflows allother schemes.

ascillato:
Maybe we can limit the max quantity of pixels for a step to 255 so as to prevent a posible overflow?

```cpp
void CmndStepPixels(void)
{
  if ((XdrvMailbox.payload >= 0) && (XdrvMailbox.payload <= 255)) {
    Settings->light_step_pixels = XdrvMailbox.payload;
```

curzon01:
That was my understanding - if you enter `StepPixels 500` it stores 244. Is this the intention?

arendst:
I like Adrian's solution.

curzon01:
@ascillato to keep `decode-config` up-to-date I notice things like that 'cause I have to understand each change in settings with the corresponding new cmnd implementations...

KDMcMullan:
YES! This is what I want for a chain of LEDs, not on stairs, but in a series of wine bottles around my patio! You GENIUS!

blakadder:
> YES! This is what I want for a chain of LEDs, not on stairs, but in a series of wine bottles around my patio! You GENIUS!

I'd love to see a video of that

PR #13611, opened 2021-11-08
Fix compile error in xdrv39 with Arduino core 2.0.1

Jason2866:
## Checklist:


arendst:
Oh Oh. If core doesn't allow uint32_t here we have an issue. Just replacing ``uint32_t`` with ``int32_t`` is NOT the solution.

Jason2866:
Should i revert? What is the difference to the next lines?

arendst:
You're right. He is already using int32_t. So this is OK.

PR #13633, opened 2021-11-10
Add driver for Texas Instruments HDC2010

lucboudreau:
## Description:

Add driver for Texas Instruments HDC2010

works

## Checklist:


barbudor:
Nice job for 1st-timer 👍 
Just 2 comments:
- May be not all logs are usefull. It's good to keep logs that can help debug user experience (understand if device is found/connected, respond...), but logs that are more intended to debug your code should be commented as they use flash space with no value for the user.
- You don't have too many globals but the trend is to put all driver's data into a struct that will be allocated only if the driver is enabled (`I2CDriver<x> 1`) and the device is detected. So you just have a single pointer as global. That save memory, especially for builds that includes all if not all the drivers (like the Mega or Platinum).

I don't have the device to test.
Thanks for your contribution

lucboudreau:
@barbudor Thanks.

1) I'll cleanup the logs. Will comment out debug stuff and leave user info.
2) can you point me to a driver where the globals are in a struct done properly?

lucboudreau:
@barbudor All fixed. Thanks for the review.

PR #13695, opened 2021-11-16
sht3x support for 2 i2c busses

gemu2015:
## Description:

support for sht3x on both i2c busses

## Checklist:

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -2328,6 +2328,44 @@ bool I2cSetDevice(uint32_t addr)
   myWire.beginTransmission((uint8_t)addr);
   return (0 == myWire.endTransmission());
 }
+
+
+TwoWire *I2C_Check_Device(uint32_t addr) {
+TwoWire *wire = nullptr;
+
+#ifdef ESP8266
+  if (I2cActive(addr)) { return wire; }  // already active
+  if (!I2cSetDevice(addr)) { return wire; } // not found
+  wire = &Wire;
+#endif // ESP8266
+
+#ifdef ESP32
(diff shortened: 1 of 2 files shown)


arendst:
I/We have to find a way to service the second I2C channel correctly for all I2C drivers.

This implementation might be it but I haven't found the time yet to do an overall investigation if this can be implemented easily while keeping in mind to only support one I2C channel on the ESP8266 and two on the ESP32.

barbudor:
I was thinking that it could be interesting for tasmota to have it's own i2c API which could handle multiple buses either flat like the 2 buses on ESP32) or even hierarchic like if using a bus extender like TrDA was asking
A peripheral would have an address which would not only be 1 byte but also refer the it is connected to

All drivers would need to be rewritten to use this API so any peripheral could be on any of those  buses.
So no more direct usage of the standard wire library would be allowed

I think adafruit support this approach

gemu2015:
this approach works with esp8266 as the code is identical if you look at the sources.
it always returns &Wire
on esp32 it checks if bus 0 responds and then bus 1 and returns the appropriate &Wire
the modifications would require to modify the driver to a local TwoWire *pointer mode.
this is an easy task.
anyhow we could start with a single driver to check since at this time only epd47 would need this modification.
the current implementation rejects when the address is already in use. we would need another bitfield for bus 1 if we would allow equal addresses on both busses.

i think support for bus multiplexing would not be useful since only very few use cases exists.
these special cases could be better solved by berry or scripter i2c drivers instead.

remark: in the future there may be the possibility that also esp8266 allows multiple instances of TwoWire , a pull request is pending (https://github.com/esp8266/Arduino/pull/7911/files)

PR #13704, opened 2021-11-16
Update xdsp_17_universal.ino

mikep1998:
Compile error fix in  init_uDisplay

## Description:

## Checklist:

### DIFF
--- a/tasmota/xdrv_13_display.ino
+++ b/tasmota/xdrv_13_display.ino
@@ -67,7 +67,7 @@ struct MULTI_DISP {
   uint8_t used;
 } displays[MAX_MULTI_DISPLAYS];
 uint8_t cur_display;
-Renderer *Init_uDisplay(const char *desc, int8_t cs = -1);
+Renderer *Init_uDisplay(const char *desc, int8_t cs);
 
 void Set_display(uint8_t index) {
   displays[index].display = renderer;
@@ -631,7 +631,7 @@ void DisplayText(void)
                           srot = Settings->display_rotate;
                           Settings->display_rotate = rot;
                         }
-                        renderer = Init_uDisplay(fdesc);
+                        renderer = Init_uDisplay(fdesc, -1);
                         if (rot >= 0) {
                           Settings->display_rotate = srot;
                         }
--- a/tasmota/xdrv_52_3_berry_display.ino
+++ b/tasmota/xdrv_52_3_berry_display.ino
@@ -24,7 +24,7 @@
 #include <berry.h>
 
 #ifdef USE_UNIVERSAL_DISPLAY
-Renderer *Init_uDisplay(const char *desc, int8_t cs = -1);
+Renderer *Init_uDisplay(const char *desc, int8_t cs);
 #endif //  USE_UNIVERSAL_DISPLAY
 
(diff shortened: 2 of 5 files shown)


mikep1998:
Maybe there is a better way feel free to change.

Jason2866:
Which setup generates a compile error? I ask since i use Udisplay and i do not have.

mikep1998:
Sorry I should have included the setup in the description. 

ili9488 and universal display caused the error. 
Init_uDisplay was declared in ili9488 with the default parameter. But universal display was not and compile error was detected at current line 438     Init_uDisplay(nullptr);  did not have the second parameter defined.

Thanks

Jason2866:
Thx, using udisplay driver and still needing a driver for the ili9488 !?
@gemu2015 why is this neccessary?

s-hadinger:
Agreed. We should start removing libs if they are not used anymore and superseeded bynthe more general universal driver

mikep1998:
Sorry I'm not using ili9488.

I'm using xdrv_54_lvgl.ino and universal display.

mikep1998:
Is there any documentation on how to configure the display drivers which to use and not use?

Jason2866:
My question is why is `xdsp_08_ILI9488_UD.ino` in general there?
Universal display driver should not need a extra specifc display driver for a display.

Jason2866:
@mikep1998 let us move to Discord channel displays

barbudor:
@mikep1998 
There is a much simpler fix to your compile problem
In `xdrv_13_display.ino`
Move line 70:
```
Renderer *Init_uDisplay(const char *desc, int8_t cs = -1);
```
above line 52:
```
int16_t disp_ypos = 0;

Renderer *Init_uDisplay(const char *desc, int8_t cs = -1);

#ifdef USE_MULTI_DISPLAY
```
et voilà !
And we can keep calls with defaults

s-hadinger:
Thanks but finally we propose to simply remove ILI9488 driver since it's ni more needed and remove the second parameter.

mikep1998:
closed in favor of using ```Removed ILI9488 driver in favor of Universal Display Driver #13719```

gemu2015:
> Thx, using udisplay driver and still needing a driver for the ili9488 !?
> @gemu2015 why is this neccessary?

before universal display we could select a display driver by its appropriate chip select and the command "displaymodel"
this _ud driver is a template how the original drivers could be removed by maintaining the CS mechanism and use universal driver.

especially in 1M esp8266 devices without a filesystem this enables the use of display.bins with many drivers included.

in principle we could convert all drivers which have a display descriptor into such _UD driver versions and remove the original drivers.

there are however some limitations.

e.g. the original st7789 driver can select 3 different resolutions by software while the universal driver must switch descriptor for that.

i selected the ili9488 as an example because as far as i know there a no variants of this display with other resolutions on the market.

s-hadinger:
Thanks @gemu2015 
Still I think we should progressively move out of specific drivers and keep only universal driver, and switch templates if needed. It fits beautifully well with new Autoconfig feature.

I understand it may cause some difficulties for ESP8266 with 1MB Flash. But those devices are of the past and not a sufficient reason to keep too much legacy code.

PR #13750, opened 2021-11-20
InfluxDB: Add command IfxPeriod

barbudor:
## Description:

Discussion (https://github.com/arendst/Tasmota/discussions/13745) with @pkkrusty

Implementing command `IfxPeriod <value>` which allow to specify a period for InfluxDb publication different from `Teleperiod`.
- By default or by specifiying `IfxPeriod 0`, InfluxDb period is same as Teleperiod and will be changed if Teleperiod is changed.
- If set to value 10..3600, the publication period will follow this value. If Teleperiod is changed, it won't impact IfxPeriod.
- If set to value 1..9, the value is forced to 10 (just like Teleperiod)
- Value greater than 3600 are rejected (just like Teleperiod)

Note: As before, even when the InfluxDb publication period uses the Teleperiod value, it is not synced with Teleperiod.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.1

### DIFF
--- a/tasmota/settings.h
+++ b/tasmota/settings.h
@@ -598,7 +598,8 @@ typedef struct {
 
   int8_t        shutter_tilt_config[5][MAX_SHUTTERS];  //508
   int8_t        shutter_tilt_pos[MAX_SHUTTERS];        //51C
-  uint8_t       free_520[12];              // 520
+  uint16_t      influxdb_period;           // 520
+  uint8_t       free_522[10];              // 522
 
   uint16_t      mqtt_keepalive;            // 52C
   uint16_t      mqtt_socket_timeout;       // 52E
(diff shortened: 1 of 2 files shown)


pkkrusty:
You're the man! Will test as soon as it is merged!

arendst:
So why not change the teleperiod time as this ifxperiod does exactly the same. The only difference is that teleperiod messages are not changed

barbudor:
@arendst , @pkkrusty would like to increase the frequency of publishing to InfluxDB without increasing the frequency of MQTT.
If you want to save the flash settings, I could remove the flash settings and let anyone interrested use a boot rule to set that up.

arendst:
OK. I'll merge.

PR #13749, opened 2021-11-20
Odroid_go update to use autoconf

Jason2866:
with Universal Display Driver and LVGL

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [x] The code change is tested and works with Tasmota core ESP32 V.2.0.1
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #13752, opened 2021-11-21
Odroid_go update to use autoconf

Jason2866:
with Universal Display Driver and LVGL
Replaces #13749 

## Checklist:

### DIFF
--- a/tasmota/xdrv_55_touch.ino
+++ b/tasmota/xdrv_55_touch.ino
@@ -322,4 +322,4 @@ bool Xdrv55(uint8_t function) {
 uint32_t Touch_Status(uint32_t sel) {
 return 0;
 }
-#endif  // USE_TOUCH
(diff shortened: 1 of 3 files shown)


tasmota/xdrv_55_touch.ino
```diff
@@ -317,5 +317,9 @@ bool Xdrv55(uint8_t function) {
   }
   return result;
 }
-
+#else
+// dummy for LVGL without a touch controller
+uint32_t Touch_Status(uint32_t sel) {
+return 0;
+}
 #endif  // USE_TOUCH
```
arendst, line 325: Where is the beginning of this code? 

There are so many defines in this file that it is impossible to find the corresponding ``#ifdef``

arendst: Ah. Thanks. I now see it's a function. Stupid me.

PR #13814, opened 2021-11-26
add #define USE_WEBCLIENT_HTTPS

Jason2866:
- for some env (ODroid-Go M5 stack core 2 and lvgl). To make it possible to do OTA easily for own builded firmwares stored for example in github which needs https
- Enable Autoconfig for lvgl env
## Checklist:

### DIFF
--- a/platformio_tasmota_env32.ini
+++ b/platformio_tasmota_env32.ini
@@ -81,7 +81,7 @@ lib_extra_dirs          = lib/libesp32, lib/lib_basic, lib/lib_display
 
 [env:tasmota32-lvgl]
 extends                 = env:tasmota32_base
-build_flags             = ${env:tasmota32_base.build_flags} -DFIRMWARE_LVGL
+build_flags             = ${env:tasmota32_base.build_flags} -DFIRMWARE_LVGL -DUSE_AUTOCONF
 board_build.f_cpu       = 160000000L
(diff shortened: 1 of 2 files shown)


arendst:
Looking at the OTA code I conclude that by enabling this the "normal" OTA is disabled?

```
#if defined(ESP32) && defined(USE_WEBCLIENT_HTTPS)
          HTTPClientLight OTAclient;
          if (!OTAclient.begin(full_ota_url)) {
            AddLog(LOG_LEVEL_INFO, "OTA: unsupported protocol");
            ota_result = -999;
          } else {
            ota_result = (HTTP_UPDATE_FAILED != httpUpdateLight.update(OTAclient, version));
          }
#else // standard OTA over HTTP
          WiFiClient OTAclient;
          ota_result = (HTTP_UPDATE_FAILED != ESPhttpUpdate.update(OTAclient, full_ota_url, version));
#endif
```

If so, this is unwanted by me. If the reason for this https functionality is to have easy local OTA possibilities it is bogus. Since over two year the tasmota repository provide a simple python webserver able to provide a local http OTA server. I use a local apache server providing http OTA which also uses a python/php script as provided in the tasmota repository.

If this feature falls back to http when https fails I'm willing to allow it's presence.

s-hadinger:
Oh. I missed this behavior. The new client follows http/https directive of the URL. 

Do you mean in must first try https and then try http if https fails?

arendst:
> Why would the URL be https in the first place? I mean http still works

So why is the https OTA code implemented then?

Jason2866:
I tried http and https update and both do work.

arendst:
OK. Then it's fine.

PR #13837, opened 2021-11-28
Proposition: remove minimal from BUILDS table

barbudor:
## Description:

Leaving `minimal` in the builds table may still let people believe it is a variant like the others.
I'm suggesting to remove it from the table and adding a note that `minimal` shouldn't be used outside of the upgrading process.

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.1.1

### DIFF
--- a/BUILDS.md
+++ b/BUILDS.md
@@ -2,236 +2,238 @@
 
 ESP8266 / ESP32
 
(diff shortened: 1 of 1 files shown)


arendst:
Makes sense.

PR #13870, opened 2021-11-29
Tuya: add tuya_allow_dimmer_0 (SetOption131)

Tortue95:
## Description:

**with the original tuya firmware, when the light is OFF, it is possible to stay pressed on the hardware power button in order to adjust the dimmer from 0 and release to the desired power. Very useful in the bedroom so as not to light up loudly and then to adjust the power.**

with the current version of tasmota this is impossible, as soon as the MCU sends the information that it turns on, tasmota instantly sends the previous value of the dimmer saved because the dimmer value saved in tasmota is different from the last value sent by the MCU chip
`light_state.getDimmer() != Tuya.Levels`

my MCU chip (I don't know of any other hardware) keeps the previous value of the dimmer itself, even when the MCU chip sends dimmer 0 when it is OFF.

I therefore created a new option: SetOption131 in order to authorize the saving by tasmota of the value dimmer = 0 when it comes from the MCU chip.

I therefore use the options:
```
SetOption131 = On
SetOption54 = On
```

SetOption131: therefore allows you to save all the dimmer values ​​that come from the MCU (from 0 to 100)
SetOption54: (already present in the current release and not modified) allows to save the value of the dimmer from MCU even when the light is OFF (the current documentation is not clear about it) 

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.1.1

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -337,6 +337,7 @@
 #define RF_DATA_RADIX          false             // [SetOption28] RF receive data format (false = hexadecimal, true = decimal)
 #define IR_DATA_RADIX          false             // [SetOption29] IR receive data format (false = hexadecimal, true = decimal)
 #define TUYA_SETOPTION_20      false             // [SetOption54] Apply SetOption20 settings to Tuya device
+#define TUYA_ALLOW_DIMMER_0    false             // [SetOption131] Allow save dimmer = 0 receved by MCU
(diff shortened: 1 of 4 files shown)


Tortue95:
I hesitated to make the change in the same option as the SetOption54 but at least I am on the backward compatibility. you tell me

ascillato (approved):

PR #13907, opened 2021-12-03
read tariffs from ornowe517

woojtekk:
First of all I would like to thanks for provide module for this energy counter. Since I already have this energy meter this module really made my approach to energy monitoring and management easier and easier 
ORNO-WE-517 energy poses possibility to program 4 different tariff (at the moment only via original software via rs485). Therefore I would like to present some changes that allow to read the set tariffs and publish it via MQTT as well as directly on device webpage.

xnrg_17_ornowe517.ino
-	Added registers for tariffs 1 and 2. (The energy meter can be programmed into 4 different ones)
-	Modified switch case function to check register and not the position of register in array is much convenient and is more transparent in comparison to previous method (in my opinion)
-	I am not sure which method  is faster but both works ;)

### DIFF
--- a/tasmota/xnrg_17_ornowe517.ino
+++ b/tasmota/xnrg_17_ornowe517.ino
@@ -55,13 +55,29 @@ const uint16_t we517_start_addresses[] {
   /*  7  */ 0x0020,  //  +   -   +   kW   Phase 2 power
   /*  8  */ 0x0022,  //  +   -   -   kW   Phase 3 power
   /*  9  */ 0x0026,  //  +   -   +   VAr  Phase 1 volt amps reactive
-  /* 10  */ 0x0026,  //  +   -   -   VAr  Phase 2 volt amps reactive
+  /* 10  */ 0x0028,  //  +   -   -   VAr  Phase 2 volt amps reactive
   /* 11  */ 0x002A,  //  +   -   -   VAr  Phase 3 volt amps reactive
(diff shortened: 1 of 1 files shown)


woojtekk:
I'm just starting my adventure with GitHub and somehow miss to fill Checklist. Ima sorry for that. 
For all my changes checklist is the same as below: 

Checklist:

[X] The pull request is done against the latest development branch
[X] Only relevant files were touched
[X] Only one feature/fix was added per PR and the code change compiles without warnings
[X] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
[  ] The code change is tested and works with Tasmota core ESP32 V.2.0.1

[X]    I accept the CLA.

NOTE: The code change must pass CI tests. Your PR cannot be merged unless tests pass

barbudor:
Why have you split the PR in so many parts?
Can't you just make 1?
Are those PR independent?

woojtekk:
Those PR are related with the same. Thats my fault and I am sorry for this. To be honest it's my first request and I didn't known how to do it in proper way. that's why I am asking for understanding.

Proposed changes are working on esp8266 and since they are not depend to platform should also work with esp32.

Currently I am trying send data via morbus to energy meter in order to set time and program tariff periods. Next PR will be submitted in proper way.

Greetings

arendst:
Thx.

First - Have you looked at the already available tariff functionality in Tasmota. If so what makes this solution better.
Second - Clean up the four PR's and make one PR for review

PR #13917, opened 2021-12-04
Supports display of ascii text on MAX7219 based 8x8 LED dot matrix modules (1088AS)

sternegugger:
## Description:
This driver enables the display of ascii text on MAX7219 based 8x8 LED dot matrix modules (1088AS).

#define USE_DISPLAY_MAX7219_MATRIX

  Connect the MAX7219 display module's pins to any free GPIOs of the ESP8266 or ESP32 module.
  VCC should be 5V. Depending on the number of used modules and the brightness, the used current can be more than 500 mA.

  Connect the 5 outgoing pins (VCC, GND, DI, CS, CLK) of the first module to the next one. 
  With this you can connect up to 32 modules. 
  To extend the display hieght, multiple rows are supported. Each module row starts from left to right.

  Assign the pins as follows from Tasmota's GUI:

  DIN hardware pin --> "MAX7219 DIN"
  CS hardware pin --> "MAX7219 CS"
  CLK hardware pin --> "MAX7219 CLK"

  Once the GPIO configuration is saved and the ESP8266/ESP32 module restarts,
  set the Display Model to 19 and Display Mode to 0

  Depending on order of the wired 8x8 matrix modules you have got a display of size pixel_width x pixel_height.
  The size has to be set with the commands "DisplayWidth <pixel_width>" and "DisplayHeight <pixel_height>"

  After the ESP8266/ESP32 module restarts again, turn ON the display with the command "Power 1"

  Now, the following "Display" commands can be used:

  DisplayText  text 
    Sends the text to the display. 
    If the text fits into the display, it is shown in the center.
    Otherwise it scrolls to the left and repeats as long it is cleared or new "DisplayText" overwrites it.

  DisplayDimmer [0..100]
    Sets the intensity of the display.

  Power [ON|OFF]
    Sitches the display on or off. When "off", the display buffer is not cleared and will be shown again when after "Power ON". 
    Other display commands are still active when off.

  DisplayClear
    Clears the display

  DisplayScrollDelay [0..15]   // default = 0
    Sets the speed of text scroll. Smaller delay = faster scrolling.
    The maximum scroll speed is 50ms per pixel on DisplayScrollDelay 0.

  DisplayWidth [8..256]
    Sets the pixel width of the display (8x number of modules in a row)

  DisplayHeight [8..256]
    Sets the pixel height of the display (8x number of module rows)

  DisplayClock  [0|1|2]
    Displays a clock.
    Commands "DisplayClock 1"     // 12 hr format
            "DisplayClock 2"     // 24 hr format
            "DisplayClock 0"     // turn off clock

Constrains:
  Only one text row is supported
  Only standard ascii is supported, UTF8 or escape characters are not supported yet.
  It is not using "renderer" yet.
  #define USE_DISPLAY_MAX7219_MATRIX disables USE_DISPLAY_MAX7219 and  USE_DISPLAY_TM1637, cause it is using the same pin config.

## Checklist:


arendst:
Thx.

Can you give an estimate of code size increase? I see a table of 2k for fonts. I also see use of  libraries where Tasmota has it's own implementation (time.h) lowering overall code size.

In short, once you managed to provide your working code it's time to lower codesize by using standard Tasmota functions and stripping out non-used features which might save (a lot of) coed space

sternegugger:
> I see a table of 2k for font
I could remove the non printable characters to decrease it.
Does tasmota provide a font table for characters of size 6x8 which I could reuse?

> I also see use of libraries where Tasmota has it's own implementation (time.h) lowering overall code size.
Does tasmota povide functions to format time to string or do I have to format it manually in C code, like others did?

> Can you give an estimate of code size increase?
compiling tasmota-display with my code (#define USE_DISPLAY_MAX7219_MATRIX):
RAM:   [=====     ]  51.2% (used 41936 bytes from 81920 bytes)
Flash: [======    ]  59.6% (used 610520 bytes from 1023984 bytes)

compiling tasmota-display without my code:
RAM:   [=====     ]  47.6% (used 39008 bytes from 81920 bytes)
Flash: [======    ]  59.5% (used 609476 bytes from 1023984 bytes)

My code increases 
RAM by 2388 bytes and 
Flash by 1044 bytes.

Regards,
Michael

arendst:
Regarding your questions Tasmota most likely does; All matrix features are present in driver ``xdsp_03_matrix`` including fonts.

PR #13927, opened 2021-12-05
Change representation of uptime to ISO 8601 duration format

suhlig:
This PR changes the uptime format from the custom `128T14:35:44` to the ISO 8601 duration format `P128DT14H35M44S`.

The code was there already, but for some unknown reason it was commented out and a custom format was used instead.

I propose the use of the ISO 8601 duration format because if there is a standard it should be used. It is well-supported (e.g. for JSON) and is different enough from a timestamp to not be confused with it.

Affected by this change of format are:

- callers of `GetUptime()`
- `tasmota/support_wifi.ino` (for `Wifi.downtime`)
- `tasmota/xsns_83_neopool.ino` (for `D_NEOPOOL_JSON_CELL_RUNTIME`)
- `tasmota/xdrv_12_home_assistant.ino` (for `WifiDowntime` in `D_RSLT_HASS_STATE`)

### DIFF
--- a/tasmota/support_rtc.ino
+++ b/tasmota/support_rtc.ino
@@ -140,13 +140,9 @@ String GetDuration(uint32_t time)
   BreakTime(time, ut);
 
   // "P128DT14H35M44S" - ISO8601:2004 - https://en.wikipedia.org/wiki/ISO_8601 Durations
-//  snprintf_P(dt, sizeof(dt), PSTR("P%dDT%02dH%02dM%02dS"), ut.days, ut.hour, ut.minute, ut.second);
+  snprintf_P(dt, sizeof(dt), PSTR("P%dDT%02dH%02dM%02dS"), ut.days, ut.hour, ut.minute, ut.second);
 
-  // "128 14:35:44" - OpenVMS
-  // "128T14:35:44" - Tasmota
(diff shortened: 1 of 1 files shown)


suhlig:
I ran `platformio run -e tasmota` locally and it was green.

barbudor:
I'm not saying that this would be worth a SetOption, but this is definitively a breaking change for anyone parsing this string and less human readable, but if it's a standard ...

s-hadinger:
Agree with @barbudor 
This is a breaking change so we can't just 'replace' it without giving a (default) option of not changing it. Additionally, this is used in the Web UI but the standard form is not user friendly.
I understand the value of standards when it comes to machine-to-machine communication, but definitely not for UIs.

Jason2866:
If I remember correctly we discussed this a while ago and Theo decided against.
The reasons was readability.

sfromis:
For a machine readable uptime format, just use UptimeSec instead. The string format is mainly for human consumption, therefore readability matters more than following some existing standard, which is not really useful here due to the availability of seconds instead of a formatted string.

When I also noticed the nonstandard format a long time ago, I saw no good arguments for change.

arendst:
So let's keep it as it as. Thx.

suhlig:
Thanks for the feedback.

PR #13928, opened 2021-12-05
Add RFC2217, add 300bd, add option for TCP Nagle

netadair:
## Description:

This PR adds some features to the tcpbridge, namely 
- 300 baud as a special case (was: only 1200-115200 in 1200 increments)
- optionally disable TCP no delay resp enable Nagle algorithm, as an argument to TCPStart <port>,<ipfilter>,<0|1>
- most important, add rudimentary telnet protocol with auto-detection, and RFC2217 on top; ie a telnet client can change the baud rate and serial settings on the fly.

Regarding the check list, it is not exactly "only one feature", but it is in close proximity and more like "one use case".

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.1.1

### DIFF
--- a/tasmota/xdrv_41_tcp_bridge.ino
+++ b/tasmota/xdrv_41_tcp_bridge.ino
@@ -339,7 +339,7 @@ void TCPInit(void) {
     serial_config_rfc2217 = Settings->tcp_config & 0x7F;
 
     TCPSerial = new TasmotaSerial(Pin(GPIO_TCP_RX), Pin(GPIO_TCP_TX), TasmotaGlobal.seriallog_level ? 1 : 2, 0, TCP_BRIDGE_BUF_SIZE);   // set a receive buffer of 256 bytes
-    TCPSerial->begin(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0), ConvertSerialConfig(0x7F & Settings->tcp_config));
+    TCPSerial->begin(Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300, ConvertSerialConfig(0x7F & Settings->tcp_config));
     if (TCPSerial->hardwareSerial()) {
       ClaimSerial();
     }
@@ -404,10 +404,10 @@ void CmndTCPBaudrate(void) {
     XdrvMailbox.payload /= 1200;  // Make it a valid baudrate, implicitely handles special case 0 -> 300Bd
     Settings->tcp_baudrate = XdrvMailbox.payload;
     if(TCPSerial) {
-      TCPSerial->begin(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0), ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new baud rate
+      TCPSerial->begin(Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300, ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new baud rate
     }
   }
-  ResponseCmndNumber(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0));
+  ResponseCmndNumber(Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300);
 }
 
 void CmndTCPConfig(void) {
@@ -417,7 +417,7 @@ void CmndTCPConfig(void) {
       Settings->tcp_config = 0x80 | serial_config; // default 0x00 should be 8N1
       serial_config_rfc2217 = serial_config;
       if(TCPSerial) {
-        TCPSerial->begin(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0), ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new config
+        TCPSerial->begin(Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300, ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new config
       }
 
     }
--- a/tasmota/xdrv_41_tcp_bridge.ino
+++ b/tasmota/xdrv_41_tcp_bridge.ino
@@ -15,6 +15,13 @@
 
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  --------------------------------------------------------------------------------------------
+  Version yyyymmdd  Action    Description
+  --------------------------------------------------------------------------------------------
+  0.0.0.1 20211202  started - add 300 baud special case, use tcp nagle as option
+                              add (auto) RFC2217 support by Michael Rausch <mrvz@netadair.de>
+                    forked  - from arendst/tasmota            - https://github.com/arendst/Tasmota
 */
 
 #ifdef USE_TCP_BRIDGE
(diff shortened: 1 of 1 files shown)


tasmota/xdrv_41_tcp_bridge.ino
```diff
@@ -118,6 +174,142 @@ void TCPLoop(void)
...
+                }
+                if (specialseq[i][1] == 4) { // SET-STOPSIZE
+                  if (specialseq[i][2] >= 1 and specialseq[i][2] <=2 ) { // skip 3=1.5
+                    new_serialconfig = (new_serialconfig & 0xfb) | ((specialseq[i][2] - 1)<<2);
+                  }
+                }
+
+                if( new_baudrate != GetSerialBaudrate() || new_serialconfig != serial_config_rfc2217 ) { // only if changed
+                  serial_config_rfc2217 = new_serialconfig;
+                  TCPSerial->begin(new_baudrate, ConvertSerialConfig(new_serialconfig));  // Reinitialize serial port with new config
```
barbudor, line 285: So the new baudrate and config received from RFC2217 will not be saved as Settings ?
Just for the current session ?

tasmota/xdrv_41_tcp_bridge.ino
```diff
@@ -136,18 +328,21 @@ void TCPLoop(void)
...
+      if (!Settings->tcp_baudrate)  { Settings->tcp_baudrate = 115200 / 1200; } // only when uninitialized
+    }
     tcp_buf = (uint8_t*) malloc(TCP_BRIDGE_BUF_SIZE);
     if (!tcp_buf) { AddLog(LOG_LEVEL_ERROR, PSTR(D_LOG_TCP "could not allocate buffer")); return; }
+    serial_config_rfc2217 = Settings->tcp_config & 0x7F;
 
-    if (!Settings->tcp_baudrate)  { Settings->tcp_baudrate = 115200 / 1200; }
     TCPSerial = new TasmotaSerial(Pin(GPIO_TCP_RX), Pin(GPIO_TCP_TX), TasmotaGlobal.seriallog_level ? 1 : 2, 0, TCP_BRIDGE_BUF_SIZE);   // set a receive buffer of 256 bytes
-    TCPSerial->begin(Settings->tcp_baudrate * 1200, ConvertSerialConfig(0x7F & Settings->tcp_config));
+    TCPSerial->begin(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0), ConvertSerialConfig(0x7F & Settings->tcp_config));
```
barbudor, line 342: So Settings->tcp_baudrate == 0 means 300, 1 means 1200 etc ... ?

This would be clearer I think:
`Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300`

tasmota/xdrv_41_tcp_bridge.ino
```diff
@@ -184,33 +385,41 @@ void CmndTCPStart(void) {
...
 
 void CmndTCPBaudrate(void) {
-  if ((XdrvMailbox.payload >= 1200) && (XdrvMailbox.payload <= 115200)) {
-    XdrvMailbox.payload /= 1200;  // Make it a valid baudrate
+  if (XdrvMailbox.payload == 300 || (XdrvMailbox.payload >= 1200) && (XdrvMailbox.payload <= 115200)) {
+    XdrvMailbox.payload /= 1200;  // Make it a valid baudrate, implicitely handles special case 0 -> 300Bd
     Settings->tcp_baudrate = XdrvMailbox.payload;
-    TCPSerial->begin(Settings->tcp_baudrate * 1200, ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new baud rate
+    if(TCPSerial) {
+      TCPSerial->begin(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0), ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new baud rate
```
barbudor, line 407: `Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300`

tasmota/xdrv_41_tcp_bridge.ino
```diff
@@ -184,33 +385,41 @@ void CmndTCPStart(void) {
...
 
 void CmndTCPConfig(void) {
   if (XdrvMailbox.data_len > 0) {
     uint8_t serial_config = ParseSerialConfig(XdrvMailbox.data);
     if (serial_config >= 0) {
       Settings->tcp_config = 0x80 | serial_config; // default 0x00 should be 8N1
-      TCPSerial->begin(Settings->tcp_baudrate * 1200, ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new config
+      serial_config_rfc2217 = serial_config;
+      if(TCPSerial) {
+        TCPSerial->begin(Settings->tcp_baudrate * 1200 + (!Settings->tcp_baudrate ? 300 : 0), ConvertSerialConfig(0x7F & Settings->tcp_config));  // Reinitialize serial port with new config
```
barbudor, line 420: `Settings->tcp_baudrate ? Settings->tcp_baudrate * 1200 : 300`

barbudor:
Hi @netadair 
Nice PR
What terminal software (win or linux) would you suggest to test the RFC2217 part ?

tasmota/xdrv_41_tcp_bridge.ino
netadair: Exactly. Imho with RFC2217, that is for the use cases I know, the required parameters are set/changed on-the-fly.

netadair:
For testing RFC2217, I used my own client in Lua (meter reading for IEC 62056-21 mode E, that is switching from 300 7E1 to 9600 7E1 within the conversation). In turns, this is proven to work against my own RFC2217 implementation in Lua on NodeMCU and against ser2net on OpenWRT boxes.
For testing telnet protocol in general, merely a no-op because the test result is "no telnet protocol mess" on the serial line, I used Kitty. This is used to capture telnet protocol with it's options on connection and sending OOB info, also for escaping a value of 255 in both directions in case telnet protocol is used .

tasmota/xdrv_41_tcp_bridge.ino
netadair: Absolutely right. Changed it in the second commit.

s-hadinger:
Thanks for the PR. I'm not a big fan of auto-detection of serial configuration. It adds significant code and ram consumption.

TCPbridge intent is to provide TCP access to a MCU, not to support any telnet connection (which generally don't make sense for a MCU). And there is always the risk of a false positive.

netadair:
The auto-detection work as: if the very first byte of the first data transmission client to server is a telnet IAC (255), then assume telnet protocol, otherwise proceed as raw and don't do anything to the data. I think this is fairly reasonable for pure text/ascii data transmitted, for possible binary transfer there is a slight chance of misdetection, indeed. This could be made an explicit configuration though, if you prefer that to avoid false positives in any case.
The telnet protocol itself is only a transport means of dynamic serial configuration as described in RFC2217, to let the client change baud rates etc. on the fly. It is not to be confused with a telnet daemon or the like, presenting a shell on login.

barbudor:
I think it would be nice to have.
May be making it #ifdef dependant ?

Jason2866:
By design it is for binary data transfer. What is the use case for this?
We do not need telnet in Tasmota.

s-hadinger:
@netadair Hi, I really appreciate your work here. However we need to manage scarcity on ESP especially when it comes to memory and Flash.

I'm still unclear about the use-case here. Knowing that you anyways need to configure the Tasmota device (Template...) and that there are already 3 ways to configure serial (console, HTTP, MQTT), I'm not sure why we need a 4th way. To my knowledge, MCUs do not change their serial settings - ever (or?).

The risk of collision with binary protocols is real. For example, with Zigbee CC2530, you typically start sending a series of 256 bytes of value 0xFF to make sure that the device cleared its buffer. This would be wrongly interpreted as serial configuration.

Could you please separate features in separate PRs. RFC2217 if maintained would not be included by default and guarded with `#ifdef` so that people can self-compile with this feature included.

PR #13937, opened 2021-12-06
Adds DDP as new Scheme for RGB+ Lights

tony-fav:
## Description:

This PR adds a new Scheme that has the Light listen to a DDP stream (UDP on port 4048). A similar scheme should be possible for addressable lights. This allows the light to be controlled by external sources like xLights, LEDFx, WLED, etc. This works quite well for smoothly varying effects but not as well for fast stroke-like effects due to not using something like AsyncUDP and callbacks, but that type of implementation seems to me to be ill-suited for behaving well in Tasmota. My use case is just wanting to add some bulbs to ambient effects.

I realize this would be a breaking change with Scheme numbering and am open for ideas on how to address that. Perhaps introducing a fixed offset for addressable Schemes rather than PWM+1 provides a long term solution.

## Checklist:


tony-fav:
Closing in favor of #14017

PR #13982, opened 2021-12-09
Berry added `tcpclient`

s-hadinger:
## Description:

Berry added simple tcp client supporting string and binary transfers:
- create an instance of the client with `var tcp = tcpclient()`
- connect to the server `tcp.connect(address:string, port:int [, timeout_ms:int]) -> bool` Address can be numerical IPv4 or domain name. Returns `true` if the connection succeeded. Optional `timeout` in milliseconds. The default timeout is `USE_BERRY_WEBCLIENT_TIMEOUT` (2 seconds).
- check if the socket is connected with `tcp.connected()`
- send content with `tcp.write(content:string or bytes) -> int`. Accepts either a string or a bytes buffer, returns the number of bytes sent. It's you responsibility to resend the missing bytes
- check if bytes are available for reading `tcp.available() -> int`. Returns `0` if nothing was received. This is the call you should make in loops for polling.
- read incoming content as string `tcp.read() -> string` or as bytes `tcp.readbytes() -> bytes`. It is best to call `tcp.available()` first to avoid creating empty response objects when not needed
- close the socket with `tcp.close()`

Full example:

```
tcp = tcpclient()
tcp.connect("192.168.2.204", 80)
print("connected:", tcp.connected())
s= "GET / HTTP/1.0\r\n\r\n"
tcp.write(s)
print("available1:", tcp.available())
tasmota.delay(100)
print("available1:", tcp.available())
r = tcp.read()
tcp.close()
print(r)
```

## Checklist:


peepshow-21:
@s-hadinger not sure this is working ` tcp.available()`

When I call the on my NSPpanel, they whole unit reboots. I'm using latest 2022.01.2

s-hadinger:
Can you please provide the full code?

peepshow-21:
Sure, basically running the exmaple in the docs;
```
tcp = tcpclient()
tcp.connect("172.17.20.5", 8080)
print("connected:", tcp.connected())
s= "GET / HTTP/1.0\r\n\r\n"
tcp.write(s)
print("available1:", tcp.available())
tasmota.delay(100)
print("available1:", tcp.available())
r = tcp.read()
tcp.close()
print(r)
```
If I comment lines and put some output in I can see it's the call to available that crashes it.

As soon as I run that and look in the Console log, I can see tas has rebooted...

```
00:00:00.004 HDW: ESP32-D0WD-V3 (PSRAM)
00:00:00.061 UFS: FlashFS mounted with 264 kB free
00:00:00.148 CFG: Loaded from File, Count 267
00:00:00.165 QPC: Count 1
00:00:00.576 BRY: Berry initialized, RAM used=3958
00:00:00.612 Project tasmota - NS Panel Version 2022.01.2(tasmota)-2_0_2(2022-01-11T17:27:34)
00:00:01.376 NSP: Initializing Driver
00:00:01.388 BRY: Successfully loaded 'autoexec.be'
00:00:01.404 RSL: RESULT = {"NSPanel":{"Nextion":"bytes('1A')"}}
00:00:02.411 WIF: Connecting to AP1 Error-404s Channel 10 BSSId 00:22:33:44:AA:A0 in mode 11n as nspanel-1336...
00:00:04.449 WIF: Connected
00:00:04.711 HTP: Web server active on nspanel-1336 with IP address 172.17.0.110
15:14:00.666 MQT: Attempting connection...
15:14:00.706 MQT: Connected
15:14:00.716 MQT: tele/nspanel/LWT = Online (retained)
```

s-hadinger:
Thanks for reporting, this is now fixed in development branch.

peepshow-21:
Is there a bin build to test?

s-hadinger:
On this page http://ota.tasmota.com/tasmota/release/

barbudor:
I think you meant there http://ota.tasmota.com/tasmota/

ascillato:
Berry is available only in ESP32 Builds.
For downloading ESP32 builds of the development branch, please go to http://ota.tasmota.com/tasmota32/
Thanks.

barbudor:
Ah ah both wrong 🤣
Thanks @ascillato

Jason2866:
The NSPanel build is here https://github.com/tasmota/install

s-hadinger:
> Ah ah both wrong 🤣
> Thanks @ascillato

Ha ha. That's the consequence of late responses

peepshow-21:
Not sure it's working still. Trying to use a test case to count size of a 1.25mb file. The readbytes seems to often read 5744 bytes, sometimes less, which is fine depending of what's in buffer I assume. But I can't get it to count the whole file, I can get it to do about 200 or so calls to readbytes, but after that it just dies and I get no output.

See attached case. You just need to change the http:// at the top to point to a file. My was locally served.

After running once, you can just call `tcptest.check(100)` with different number. 200 iterations seems to count about 800k, 400 fails.

I also get this error after, which I don't understand;
image


Here is the test case;

```
class TCPTest

    static flash_url = "http://172.17.20.5:8080/static/chunks/nxpanel.tft"

    var flash_size
    var tcp

    def openUrl()

        import string
        var host
        var port
        var s1 = string.split(self.flash_url,7)[1]
        var i = string.find(s1,":")
        var sa
        if i<0
            port = 80
            i = string.find(s1,"/")
            sa = string.split(s1,i)
            host = sa[0]
        else
            sa = string.split(s1,i)
            host = sa[0]
            s1 = string.split(sa[1],1)[1]
            i = string.find(s1,"/")
            sa = string.split(s1,i)
            port = int(sa[0])
        end
        var get = sa[1]
        print(host,port,get)
        self.tcp = tcpclient()
        self.tcp.connect(host,port)
        self.tcp.write("GET "+get+" HTTP/1.0\r\n\r\n")

    end

    def check_int(max)

        self.openUrl()

        var tot = 0;
        for i: 1..max
            var a = self.tcp.available()
            if a>0
                var b = self.tcp.readbytes()
                print ("read", b.size())
                tot += b.size()
                tasmota.yield()
            else
                print ("not available: ", a)
                tasmota.delay(100)
            end
        end
        print("done: ", tot)

        self.tcp.close()

    end

    def check(max)

        var i = max
        tasmota.set_timer(0, self.check_int(i))

    end


end
 
tcptest = TCPTest()
tcptest.check(100)

```

s-hadinger:
This looks like a memory problem. Such errors tend to append when memory is saturated. Do you have PSRAM on your device?
You can also force a garbage collection asap with `tasmota.gc()`. So in your loop I would try adding:

```
            if a>0
                var b = self.tcp.readbytes()
                print ("read", b.size())
                tot += b.size()
                b = nil                # make the buffer gc-able
                print("gc=", tasmota.gc())    # force a gc now
                tasmota.yield()
```

peepshow-21:
Nope... same issue;
image

200 works, comes back with a count. 400. just comes back with the error (no output)

It's running on a NSPanel. Not sure what RAM it has.

peepshow-21:
I think it's  working. It seems to be if the output was too much, you get none at all. I didn't know that.
also, the other error seems to related to calling it via the timer. it doesn't seem to cause any issue.

So I'd say... working! 

Thank you :)

peepshow-21:
Pretty sure it's wrong. I  can read the same file, well a bigger version bc it's that file expanding to have not binary data, using the httpclient. and it works file. It's rock solid, but not ideal. I need to slide the file into lots of smaller files, bc you can't control the read size with that. but I'd rather use this as I had to fudge things a lot to use the http one, when available() didn't work.

But this one works partly. when you start to do loads of calls to readbytes, it starts to fail for sure.

s-hadinger:
It's hard to understand what file you have and what you try to achieve. Parsing 1mb file with so little memory is a challenge and not what tcp was meant for. It's a minimal version for simple uses. 

Why don't you use http client?

peepshow-21:
I'm using berry to control the screen on the Nexion in the NSPanel. I've
written a firmware uploader to upload new screen firmware. These files are
too big to copy to the file system on the NSPanel so I need to read via
network and stream via serial to the nextion, and it's binary. http will
not support binary, nor can you read part of the file with it. so it's not
really fit for the job. so I need to write a pre-processor on the PC to
split the file into non-binary, 4096 byte files. so I can read each via the
httpcleint and send to the serial channel. like I say, very clunky. it
works, but I need the pre-processing part. If tcp worked, I could pull 4096
chunks (of binary) from the file and send like that, no messing about to ge
the file httpclient friendly. I dont really need much memory as the process
is just transient, pulling 4k then sending. so it shouldn't be an issue on
resources. it's just that it doesn't seem to work.

s-hadinger:
Ok, got it. Thanks. Would that help to be able to specify a maximum size when reading from the tcp stream? Currently Berry probes the tcp stream for bytes in buffer, and allocates a buffer big enough for the entire chunk. I could add a parameter to limit to 4096 bytes for example.

peepshow-21:
I was able to work around that in code pretty easy by brining back 5477 or
less and either taking the first 4096 of that and leaving rest, or if less
doing another read. The main issue was it just seeming to die after a
number of iterations for no reason. not sure setting a buffer limit would
fix that, as it should have worked just fine the way it was.

To be honest a much better fix would be to implement an readbytes(max) on
the httpclient. That would be more useful all round i think. A lot of
content will be on https, which would be too much for tcpclient. For me, I
could put the file on http server, but it would not work for all. so the
readbyte on httpcleint does seem optimal.

But i'm open to either, just need an option to process in incoming file
directly.

s-hadinger:
Having a readbytes() on httpClient is more tricky than tcp client. The problem is that HTTP can have chunked encoding so the stream of data received from the network needs to be processed separately.

peepshow-21:
tcpclient would be fine. but I think there is some internal issues with it
and it's just not working. I don't think it can be mem as such because I
call gc() all through the reads. The working httpclient one that does work
fine, has no gc() anywhere and uses much more mem in how it processes it
too. could it just be a bug that makes it crash?  I don't see any output to
help know.

s-hadinger:
The one thing you're of course missing with TCP is TLS. Would you be kind enough to share the file? Or should I try with a random large file?

peepshow-21:
I pasted a link to the script and the binary file i was using in the post above.
Here it is;
https://github.com/peepshow-21/ns-flash/releases/download/v0.1.1-alpha/nxpanel.tft

peepshow-21:
@s-hadinger , did you manage to get anywhere seeing what was going wrong?

s-hadinger:
I don't know what went wrong in your code. I have uploaded your file, and set a loop to read it. I didn't see any issue.

Quick & dirty code:
```
var tcp = tcpclient()
tcp.connect("demo-hadinger.s3.eu-west-3.amazonaws.com",80)
tcp.write("GET http://demo-hadinger.s3.eu-west-3.amazonaws.com/nxpanel.tft HTTP/1.0\r\n\r\n")

var s = 0

def rr()
  if tcp.connected()
    while tcp.available() > 0
      var inc = tcp.readbytes()
      s += size(inc)
      print("inc=", size(inc), s)
      inc = nil
    end
    tasmota.set_timer(50, rr)
  else
    print("TOTAL=", s)
  end
end
tasmota.set_timer(50, rr)
```

The total number of bytes read is `1985749` while the file is `1985360`. The difference are the headers of the response.

peepshow-21:
@s-hadinger - Struggling with this.. I don't understand the error. Just can't find online help for Berry. Would you mind telling me what's wrong?

I run this;

```
class TCPTest

    #static flash_url = "http://demo-hadinger.s3.eu-west-3.amazonaws.com:80/nxpanel.tft"
    static flash_url = "http://172.17.20.5:8080/static/chunks/nxpanel.tft"
    static max_block = 4096

    var flash_size
    var flash_count
    var flash_mode
    var flash_complete
    var flash_buff

    var really_read    
    var tcp
    var loop_max

    def open_url()

        import string
        var host
        var port
        var s1 = string.split(self.flash_url,7)[1]
        var i = string.find(s1,":")
        var sa
        if i<0
            port = 80
            i = string.find(s1,"/")
            sa = string.split(s1,i)
            host = sa[0]
        else
            sa = string.split(s1,i)
            host = sa[0]
            s1 = string.split(sa[1],1)[1]
            i = string.find(s1,"/")
            sa = string.split(s1,i)
            port = int(sa[0])
        end
        var get = sa[1]
        print(host,port,get)
        self.tcp = tcpclient()
        self.tcp.connect(host,port)
        self.tcp.write("GET "+self.flash_url+" HTTP/1.0\r\n\r\n")
        tasmota.delay(500)

    end

    def proc_block(b)
    
        self.flash_count += size(b)
        #print("read", size(b), self.flash_count)

    end
                    
    def loop_till_done()

        while self.tcp.available()>0
            var b = self.tcp.readbytes();
            self.flash_buff += b
            while size(self.flash_buff)>self.max_block
                self.proc_block(self.flash_buff[1..self.max_block])
                self.flash_buff = self.flash_buff[self.max_block..]
            end
        end
        if self.tcp.connected()
            tasmota.set_timer(50,self.loop_till_done())
        else
            if size(self.flash_buff)>0
                self.proc_block(self.flash_buff)
            end
            print("complete")
        end

    end

    def start_flash()

        self.flash_count = 0
        self.flash_buff = bytes()
        self.open_url()
        print("url opened")
        self.loop_till_done()
        print("done", self.flash_url)

    end
                        
end
 
tcptest = TCPTest()

tcptest.start_flash()

```

When I run with my URL it works, but I get these errors, which I don't understand ...

```
172.17.20.5 8080 /static/chunks/nxpanel.tft
url opened
complete
done http://172.17.20.5:8080/static/chunks/nxpanel.tft
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
BRY: Exception> 'type_error' - 'nil' value is not callable
stack traceback:
	<unknown source>: in function `run_deferred`
	<unknown source>: in function `event`
```

If I comment out my URL and use yours, it doesn't work, but I get ever more odd errors?

```
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	...
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `loop_till_done`
	<unknown source>: in function `start_flash`
	<unknown source>: in function `main`
```
If I can get past this I think I'm sorted!!!

s-hadinger:
Your problem is in set_timer, you don't register the function but the result of the function which is faulty

It should be
```
tasmota.set_timer(50, /-> self.loop_till_done())
```

peepshow-21:
@s-hadinger - Thank you for the help. It's all working a treat now.

PR #14004, opened 2021-12-11
Add Linkind dimmer support

pcdiem:
## Description:

This PR adds support for Linkind (https://www.amazon.com/dp/B08DRF58T1) ESP32-based dimmer that uses I2C to communicate with the dimmer hardware.

Please check the changes to the build and template files and let me know if anything was not done the way you want them implemented.

## Checklist:

### DIFF
--- a/tasmota/support.ino
+++ b/tasmota/support.ino
@@ -1525,6 +1525,14 @@ void ModuleDefault(uint32_t module)
 void SetModuleType(void)
 {
   TasmotaGlobal.module_type = (USER_MODULE == Settings->module) ? Settings->user_template_base : Settings->module;
+
+  TasmotaGlobal.use_pwm_dimmer = false;
+#ifdef ESP8266
+  if (PWM_DIMMER == TasmotaGlobal.module_type) TasmotaGlobal.use_pwm_dimmer = true;
+#endif  // ESP8266
+#ifdef USE_LINKIND
+  if (LINKIND == TasmotaGlobal.module_type) TasmotaGlobal.use_pwm_dimmer = true;
+#endif  // USE_LINKIND
 }
 
 bool FlashPin(uint32_t pin)
--- a/.github/workflows/Tasmota_build_devel.yml
+++ b/.github/workflows/Tasmota_build_devel.yml
@@ -694,6 +694,44 @@ jobs:
         path: ./build_output
 
 
+  tasmota32solo1-linkind:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
+    - name: Install dependencies
+      run: |
+        pip install -U platformio
+    - name: Run PlatformIO
+      run: |
+        platformio run -e tasmota32solo1-linkind
+    - uses: actions/upload-artifact@v2
+      with:
+        name: firmware
+        path: ./build_output
+
+
+  tasmota32solo1-linkind-ble:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
+    - name: Install dependencies
+      run: |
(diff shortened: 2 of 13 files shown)


Jason2866:
Why are new build env added? The linkind dimmer can be added to the solo1 env with a ifdef in `user_config_override.h`. Tasmota env are not ment to be used just for a single device.

pcdiem:
I thought I needed to add the device in tasmota_template.h because the PWM dimmer module does not exist for ESP32. Maybe the right thing is to add the PWM Dimmer module as an ESP32 template?

Jason2866:
Why adding a template in general? Is there a specific function called which is not possible to call with a usual configuration Template provided?
So adding the needed changes in code (#ifdef...) and that's all.
Do I overlook something?

tasmota/support.ino
```diff
@@ -1525,6 +1525,14 @@ void ModuleDefault(uint32_t module)
...
   TasmotaGlobal.module_type = (USER_MODULE == Settings->module) ? Settings->user_template_base : Settings->module;
+
+  TasmotaGlobal.use_pwm_dimmer = false;
+#ifdef ESP8266
+  if (PWM_DIMMER == TasmotaGlobal.module_type) TasmotaGlobal.use_pwm_dimmer = true;
+#endif  // ESP8266
+#ifdef USE_LINKIND
+  if (LINKIND == TasmotaGlobal.module_type) TasmotaGlobal.use_pwm_dimmer = true;
+#endif  // USE_LINKIND
 }
```
arendst, line 1536: This won't work as module_types will be duplicated between esp8266 and esp32. I strongly suggest NOT to use the module_type for ESP32 specific code.

Your overall implemntation is much too invasive. There must be a better way to implement this specific hardware.

A possible solution is the way I implemented the Sonoff SPM hardware; it only uses one file and acts based on it's template name. The most preferred way is/was select functionality based on GPIO selection but for a POC using the template name is a much cleaner approach as it doesn't need small changes to many files.

pcdiem:
The only real extra code that is necessary to support the Linkind dimmer is the few lines of code in the light module that write the dimmer level through I2C instead of through analogWrite. There also needs to be a way to indicate that this device should use the PWM Module logic to respond to the buttons. Please let me know how you would like me to implement this so it fits in with your vision.

arendst:
Will have a look how to implement this and come back here.

pcdiem:
Theo was going to look in to how to implement this. The code for the actual Linkind dimmer support is simply using I2C in the light module instead of setting the PWM value. In addition, there's the template. The issue that needs to be worked out is how to enable the PWM Dimmer code.

arendst:
Will revisit ASAP

arendst:
I've been trying to implement this by using two new GPIO_OPTIONS.

During this I expect this LINKID device is not using ALL USE_PWM_DIMMER code so in addition to it's I2C code is does NOT use all PWM_DIMMER code. Am I correct.

If so we need to assume the LINKID device is a new device with some overlap of USE_PWM_DIMMER but definitly NOT full USE_PWM_DIMMER plus I2C code.

Pls enlighten me.

tony-fav:
The device has 2 buttons, a red led and a green led (located in the same place), and a secondary chip that manages the physical dimming and received the dimming value over I2C. So it is useful to use essentially the PWM Dimmer functionality for managing the buttons, the dimming, and the brightness of the green LED.

I suspect there may be 2 changes represented by this PR. The first change is enabling PWM Dimmer for ESP32 (could be useful to upgrade existing PWM Dimmer devices to Hardware PWM with a C3 for example). The second change is for the Linkind Dimmer device specifically, enabling the I2C write in place of the analogWrite.

arendst:
Merged my way.

For use on ESP32 Linkind load following template:
```
{"NAME":"ESP32-Linkind","GPIO":[6213,8448,0,0,640,0,0,0,0,288,0,0,0,0,0,0,0,608,0,0,0,544,0,0,0,0,0,0,33,32,0,0,0,0,0,0],"FLAG":0,"BASE":1}
```

PR #14078, opened 2021-12-17
Add OpenTherm Boiler Lockout Reset command

avp-avp:
## Description:

Add new OpenTherm command BLOR (Boiler Lockout Reset). This command reset boiler error state. 

## Checklist:
  - [ ?] The code change is tested and works with Tasmota core ESP32 V.2.0.1.1


avp-avp:
@yury-sannikov check it please

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -243,6 +244,14 @@ OpenThermCommand sns_opentherm_commands[] = {
      .m_ot_make_request = sns_opentherm_get_generic_u16,
      .m_ot_parse_response = sns_opentherm_parse_generic_u16,
      .m_ot_appent_telemetry = sns_opentherm_tele_generic_u16},
+    {// Boiler Lock-out Reset command
+     .m_command_name = "BLOR",
+     .m_command_code = (uint8_t)OpenThermMessageID::Command,
+     .m_flags = {.manual = 1},
```
yury-sannikov, line 250: I like the idea of putting these one-shot commands into the list because if the send fails, the logic can retry this command and this will guarantee that BLOR will be sent when the connection is restored.

One caveat with the telemetry. Telemetry will send 0..127 if the command failed or 128..255 in case of success. Though there is no way to say if the BLOR was ever initiated or not since 0 falls under the failure scenario.

I would suggest creating a custom version of the parser. The parser might put the result code to the lower u8 and 0xFF to the upper. Also, in the case of 128..255 response range value, just replace it with 255. 

Then the stats matrix would look like this:
low:0, high: 0 - never invoked
low:0..7F, high: FF - invoked and failed, lower bit contains the boiler specific error code (I guess)
low:FF, high: FF - invoked and succeeded

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -431,6 +440,26 @@ void sns_opentherm_tele_oem_diag(struct OpenThermCommandT *self)
...
+    self->m_flags.notSupported = true; // Disable future calls of this command
+
+    unsigned int data = 1; //1 : “BLOR”= Boiler Lock-out Reset command
+    data <<= 8;
+    return OpenTherm::buildRequest(OpenThermMessageType::OPTH_WRITE_DATA, OpenThermMessageID::Command, data);
+}
+
+bool sns_opentherm_call_blor() 
+{
+    OpenThermCommandT *cmd = &sns_opentherm_commands[(sizeof(sns_opentherm_commands) / sizeof(OpenThermCommand))-1];
```
yury-sannikov, line 457: here you're effectively doing `&sns_opentherm_commands[SNS_OT_COMMANDS_COUNT - 1]` thus you rely on the fact that BLOR is the last command in the list. If someone else will add another command, it will break. I would suggest doing a simple loop to find it like
```
    for (int i = 0; i < SNS_OT_COMMANDS_COUNT; ++i)
    {
        struct OpenThermCommandT *cmd = &sns_opentherm_commands[I];
        if (!strcmp(cmd->m_command_name, "BLOR") && cmd->manual) {
          cmd->m_flags.notSupported = false;
        }
    }
```

yury-sannikov: or rename the function to something like `sns_opentherm_schedule_manual(const char* command_name)` to make it generic for one-off commands.

tasmota/xsns_69_opentherm_protocol.ino
yury-sannikov: Currently `sns_opentherm_get_next_request` does not check the `manual` flag and BLOR will always be transmitted after the Tasmota re/starts. This is probably not what you want.

Instead, you can set both `notSupported` and `manual`. to True. That way `sns_opentherm_get_next_request` will always skip the BLOR.

Also, you could update `sns_opentherm_process_success_response` to have something like:

```
void sns_opentherm_process_success_response(struct OT_BOILER_STATUS_T *boilerStatus, unsigned long response)
{
    if (sns_opentherm_current_command >= SNS_OT_COMMANDS_COUNT)
    {
        return;
    }
    struct OpenThermCommandT *cmd = &sns_opentherm_commands[sns_opentherm_current_command];
    
    if (cmd->m_flags.manual) {
      // stop one-off commands from firing after confirming the request was delivered
      cmd->m_flags.notSupported = true;
    } else {
      // mark command as supported
      cmd->m_flags.supported = true;
    }
    cmd->m_ot_parse_response(cmd, boilerStatus, response);
}

```

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -568,7 +597,8 @@ void sns_opentherm_protocol_reset()
     for (int i = 0; i < SNS_OT_COMMANDS_COUNT; ++i)
     {
         struct OpenThermCommandT *cmd = &sns_opentherm_commands[i];
-        cmd->m_flags.m_flags = 0;
+        cmd->m_flags.notSupported = cmd->m_flags.manual;
```
yury-sannikov, line 600: I know you covered all use cases here, however, if someone else will add another flag in the future and forget to change the logic here, he might end up with a buggy code.
I would suggest doing something like:
```
bool isManual = cmd->m_flags.manual;
cmd->m_flags.m_flags = 0;
if (isManual) {
  cmd->m_flags.manual = 1;
  cmd->m_flags.notSupported = 1;
}
```

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -32,6 +32,7 @@ typedef union {
         uint8_t notSupported : 1; // If set, boiler does not support this command
```
yury-sannikov, line 32: Since `notSupported` is used to skip one-off commands as well, I think it worth renaming it into something like
```
uint8_t skip : 1; // If set, the command won't be transmitted to the boiler.
```

yury-sannikov (changes requested):
~Hate to reject PR's~ (it seems I can't do that, lol), but the main issue is that the BLOR will be transmitted on each Tasmota re/start.
Minor things are:
- BLOR logic rely on the fact that its the last command in the list
- BLOR transmitted only once, and if the transmission failed, BLOR will never retry.

Anyway, thank you for the PR, I think with some of the tweaks it will make a lot of sense.

avp-avp:
> but the main issue is that the BLOR will be transmitted on each Tasmota re/start.
As I undestand sns_opentherm_protocol_reset() called on each restart before command cycle

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -431,6 +440,26 @@ void sns_opentherm_tele_oem_diag(struct OpenThermCommandT *self)
     ResponseAppend_P(PSTR("%d"), (int)self->m_results[0].m_u16);
 }
 
+/////////////////////////////////// Boiler Boiler Lock-out Reset  //////////////////////////////////////////////////
+unsigned long sns_opentherm_send_blor(struct OpenThermCommandT *self, struct OT_BOILER_STATUS_T *status)
+{
+    AddLog(LOG_LEVEL_ERROR, PSTR("[OTH]: Call Boiler Lock-out Reset"));
+    
+    self->m_flags.notSupported = true; // Disable future calls of this command
```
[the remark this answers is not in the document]
avp-avp: I think its OK. It's dangerous command and all call should be manual

tasmota/xsns_69_opentherm_protocol.ino
avp-avp: on every restart called sns_opentherm_protocol_reset(). It set notSupported to 1. 
I didn't see BLOR data on telemetry before manual call of BLOR

tasmota/xsns_69_opentherm_protocol.ino
avp-avp: SNS_OT_COMMANDS_COUNT declared below this function. I tried keep other code

tasmota/xsns_69_opentherm_protocol.ino
avp-avp: cmd->m_flags.m_flags = 0;

will reset supported flag to 0. I think it's a bug

tasmota/xsns_69_opentherm_protocol.ino
avp-avp: may be its not a bad idea

yury-sannikov:
> > but the main issue is that the BLOR will be transmitted on each Tasmota re/start.
> 
> As I undestand sns_opentherm_protocol_reset() called on each restart before command cycle

It gets called from the `sns_ot_start_handshake` as a result of the OpenTherm connection reset. This usually happens when the boiler lost power or restarts. Also, it gets called as a result of the `FUNC_SAVE_AT_MIDNIGHT`. Although I'm not 100% sure about the necessity of a midnight reset. The idea was to implement some sort of Soft Reset and try again OT commands that have been marked as not supported.

avp-avp (commented):
I fixed PL review issues 
@yury-sannikov please look it again

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -552,7 +587,7 @@ void sns_opentherm_dump_telemetry()
     for (int i = 0; i < SNS_OT_COMMANDS_COUNT; ++i)
     {
         struct OpenThermCommandT *cmd = &sns_opentherm_commands[i];
-        if (!cmd->m_flags.supported)
+        if (!cmd->m_flags.supported || cmd->m_flags.skip)
```
yury-sannikov, line 590: Are you sure you want to skip it? You're probably interested in the result of the BLOR call and it's available only through Telemetry.

tasmota/xsns_69_opentherm_protocol.ino
```diff
@@ -431,6 +441,32 @@ void sns_opentherm_tele_oem_diag(struct OpenThermCommandT *self)
     ResponseAppend_P(PSTR("%d"), (int)self->m_results[0].m_u16);
 }
 
+/////////////////////////////////// Boiler Boiler Lock-out Reset  //////////////////////////////////////////////////
+unsigned long sns_opentherm_send_blor(struct OpenThermCommandT *self, struct OT_BOILER_STATUS_T *status)
+{
+    AddLog(LOG_LEVEL_ERROR, PSTR("[OTH]: Call Boiler Lock-out Reset"));
+    
+    self->m_flags.skip = true; // Disable future calls of this command
```
yury-sannikov, line 449: As I mentioned earlier, in case of OT comm failure your BLOR command will not be delivered to the boiler. If you want to keep it as-is, it's okay. Just make sure your logic will periodically make attempts to make an `ot_blor` calls.

yury-sannikov (approved):
This code should be okay, though I see 2 caveats:
- you won't be able to get telemetry to check whether your BLOR command succeeded or not. I think you can safely remove the `|| cmd->m_flags.skip)` check in the `sns_opentherm_dump_telemetry()` but it's up to you.
- you should be aware, that in rare cases your BLOR might not be delivered, and you should build your automation in such a way that it can re-issue the BLOR command after some timeout if the boiler is still locked.

tasmota/xsns_69_opentherm_protocol.ino
avp-avp: I think that only manual controlled call of BLUR is valid. You should never use this command in automation scripts. I don't know any commercial OT adapters that support BLOR. Support team usual answer that you should't remotely reset boiler....

tasmota/xsns_69_opentherm_protocol.ino
avp-avp: Thank you. It's bug.

 I will test code and push a new version.  (Remote update boiler controller is not a good idea)

avp-avp: now it return 
"BLOR":{"lo":255,"hi":0} - it's OK

avp-avp:
@yury-sannikov what's next
I checked last version. It's work 
I don't wan't add any retry to BLOR command. Baxi have a limited reactions to RESET. After several unsuccessful retry it block BLOR and reset button.

yury-sannikov:
@avp-avp looks good for me and I approved the PR. I do not have rights to merge the PRs, I assume maintainers will do it at their own pace. I would suggest adding a PR against the documentation as well.

avp-avp:
> @avp-avp looks good for me and I approved the PR. I do not have rights to merge the PRs, I assume maintainers will do it at their own pace. I would suggest adding a PR against the documentation as well.

Thank you!

avp-avp:
How I can add BLOR command to documentation?

PR #14106, opened 2021-12-19
Tuya esp32c3

barbudor:
## Description:

As disccussed in https://github.com/arendst/Tasmota/discussions/14086
esp32C3 is a good alternative to esp8266 for Tuya frankesteinization
This PR brings Tuya_MCU to C3

TUYA_MCU is module 2 on C3 vs 54 on ESP8266
So templates need to be manually adapted

Note that it also brings SK03_TUYA in order to minimize code change.
Alternatively, SK03_TUYA could be excluded with #if

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.1.1

### DIFF
--- a/tasmota/tasmota_globals.h
+++ b/tasmota/tasmota_globals.h
@@ -133,7 +133,7 @@ String EthernetMacAddress(void);
 #undef FIRMWARE_MINIMAL                            // Minimal is not supported as not needed
 
 // Hardware has no ESP32
-#undef USE_TUYA_DIMMER
+//#undef USE_TUYA_DIMMER
 #undef USE_PWM_DIMMER
 #undef USE_EXS_DIMMER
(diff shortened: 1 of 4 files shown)


arendst:
Thx.

I see the issue with modulenames and I don't like it. Shouldn't we enable this tuya thing not also for ESP32 and ESP-S2? If so I will think about a solution for all ESP32 without using the modulename as leading for enabling the code.

barbudor:
I think there is definitely a use case for 32C3 especially in case of brain transplant, less with 32 and 32S2
However if this ease the maintenance of code base then it's surely a good move

s-hadinger:
Agree with Theo, it should be managed with autoconf, i.e. Template and Setoption, instead of a module

arendst:
Refactored differently just now.

PR #14135, opened 2021-12-21
Add support for TMP117 temperature sensor?

atifan:
Hi, it would be nice if there would be added support for the i2c temperature sensor TMP117.
It is a high precision temperature sensor, not cheap but good.

Infos here:
https://www.ti.com/lit/ds/symlink/tmp117.pdf?ts=1639991283833&ref_url=https%253A%252F%252Fwww.google.com%252F

Is it possible to add this? Or to complicated? :D


ascillato2:
Invalid Pull Request. You are trying to merge 2 different branchs of Tasmota. Closing.

Please, use discussions instead.

PR #14153, opened 2021-12-22
Fixed non-8N1 serial issues. 

wir3z:
TasmotaSerial:
- Fixed issue where Serial.begin for the ESP8266 was not passing the UART SerialConfig compatible values.

support.ino
- Cleanup to use already present ConvertSerialConfig API.

tasmota.ino:
- Force the baudrate + serial config settings after boot.  Previously, the baudrate would change, but any non 8N1 settings were not applied.

xdrv_08_serial_bridge.ino:
- Increased the software serial bridge buffer size and changed type to prevent overflow.
- Added missing serial config settings to the serial begin.  Previously was forcing 8N1.

xdrv_10_scripter.ino
- Cleanup to use already present ConvertSerialConfig API.

## Description:

## Checklist:


arendst:
Thx.

Those fake parity bite were not represented for a reason: TIMING. Did you check your changes in the real world for all supported baudrates and different ESP8266 clock rates?

wir3z:
Great question.  Might be best if I roll those changes out.  The change on line 141/143 needs to be there for the settings application.

arendst:
Yeah, that change I don't like as the library needs to refererence a function externally.

I propose you do some rigourous testing before I merge this (knowing the hell we had to go through to get software serial kind of stable).

wir3z:
Understood.  I removed the external function call and replaced with the direct call.  When running the serial bridge, it failed to operate in anything by 8N1 mode until these changes.

Jason2866:
Don't try to include like `#include "..\..\..\..`

wir3z:
> Don't try to include like `#include "..\..\..\..`

Suggestions?  Seems the library cannot see back to the base tasmota folder.  (I don't like the tree depth either since it isn't portable).

lib/default/TasmotaSerial-3.3.0/src/TasmotaSerial.cpp
```diff
@@ -29,6 +29,11 @@ extern "C" {
 
 #ifdef ESP8266
 
+// added to access kTasmotaSerialConfig
+#include "../../../../tasmota/tasmota_compat.h"
+#include "../../../../tasmota/tasmota.h"
```
s-hadinger, line 34: Use extern definition instead.

wir3z: Apparently you cannot extern a const since it's not a variable so the linker cannot find it.

I moved kTasmotaSerialConfig from Tasmota.h to TasmotaSerial.h to address this.

s-hadinger (approved):

arendst:
Well let's hope we don't introduce new ghost issues related to serial timing.

wir3z:
> Well let's hope we don't introduce new ghost issues related to serial timing.

Shouldn't (yeh, I know, famous last words!) be any timing issues.  After rolling back the bit bang changes, all changes committed are just around pushing in the Settings->serial_config values into the pre-exisiting serial API's during the serial.begin's.  Previously it was defaulting to class overloads to push in SERIAL_8N1 by default.

Jason2866:
Murphys law... `Shouldnt` 👻 Merry Christmas

wir3z:
What, ghosts are for Halloween!  :)

What was the issue with this pull request?

Jason2866:
There is something wrong with serial config parameter, the change performs the config conversion twice resulting in a bogus config.

wir3z:
> There is something wrong with serial config parameter, the change performs the config conversion twice resulting in a bogus config.

Was this on the ESP8266 device?  All devices or just the serial bridge?   I'm thinking I should have headed Theo's warning and left the library alone.  :)  Thinking about it it should have gotten inherited through the classes.

I can retest this all in a couple days.  I know the other changes for the ESP32 are fully working on the serial bridge, but the changes didn't force the config settings on that platform.  I had did brief testing the ESP8266 for the serial bridge (hardware mode) and it "seemed" to be fine (but now I'm questioning the changes in the library on that platform).

PR #14157, opened 2021-12-23
Fix changeUIntScale to be stable

s-hadinger:
## Description:

I redid the math and changed slightly the computation. Now instead of converting a number to another number, it converts a range bucket to another range bucket. For value `n`, its bucket if `[n,n+1[`. So now I'm taking the value of the center of the bucket which is `n+1/2`

The new algorithm is now proven to be stable when going from `0..100` to `0..255` and back to `0..100`

``` ruby
for i: 0..100
  var d = tasmota.scale_uint(i,0,100,0,255)
  var j = tasmota.scale_uint(d,0,255,0,100)
  print(string.format("i=%i d=%i j=%i", i, d, j))
  if i != j print("BLAH ----------------") end
end
```

## Checklist:

### DIFF
--- a/tasmota/support_float.ino
+++ b/tasmota/support_float.ino
@@ -407,8 +407,19 @@ uint16_t changeUIntScale(uint16_t inum, uint16_t ifrom_min, uint16_t ifrom_max,
     to_max = ito_min;
   }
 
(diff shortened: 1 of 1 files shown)


Tortue95:
Youhou ! Thanks a lot. everything is working properly now.

Issue #14165, opened 2021-12-25
Problem with Sensor DHT22 when Temperature is below 0 degrees Celsius


flinty32:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._
Using the ESP 8266, WEMOS D1 and other ESP with ESP8266EX on Tasmota 10.1.0 (sensors)
If Temperatur is greater then 0 degrees Celsius the Tasmota Displays the correct Temp. with DHT22 (AM2301) It works fine.
But if the Temperature is below 0 Gegrees C It Display wrong Temperature : AM2301 Temperature -3276.3 °C

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [X] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [X] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [X] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [X] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [X] Device used (e.g., Sonoff Basic): ESP8266EX
- [X] Tasmota binary firmware version number used: 10.1.0
  - [X] Pre-compiled
  - [ ] Self-compiled
- [X] Flashing tools used: NodeMCU-PyFlasher-4.0-x86.exe
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:


```
- [ ] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```lua
  Rules output here:


```
- [X] Provide the output of this command: `Status 0`:
```lua
  STATUS 0 output here:
  00:25:07.421 RSL: SENSOR = {"Time":"1970-01-01T00:25:07","AM2301":{"Temperature":-3276.2,"Humidity":95.7,"DewPoint":-3373.4},"BME280":{"Temperature":0.9,"Humidity":100.0,"DewPoint":0.9,"Pressure":980.1,"SeaPressure":1004.2},"BH1750":{"Illuminance":0},"PressureUnit":"hPa","TempUnit":"C"}


```
[...]

PR #14173, opened 2021-12-27
DHT support both type of negative temperature

barbudor:
## Description:

https://github.com/arendst/Tasmota/issues/14165
There seems to be 2 variants of DHT22: 
- one with temperature provided as signed 16 bits (2's complement)
- one with only bit 31 set to indicate a negative temperature (more standard model)

This patch seamlessly handles both case by comparing high nibble to 0x8

## Checklist:


arendst:
Wonderfully done!! Nice teamwork.

PR #14212, opened 2021-12-29
Bug fix in serial library for hardware parity applications.

wir3z:
TasmotaSerial.cpp
- Fixed issue where fake parity for software serial was being called on software and hardware serial, and was overwriting parity settings for hardware configurations.

tasmota.ino:
- Force the baudrate + serial config settings after boot. Previously, the baudrate would change, but any non-8N1 settings were not applied if using the debug hardware serial port.

support.ino
- Cleanup to use already present ConvertSerialConfig API.

xdrv_10_scripter.ino
- Cleanup to use already present ConvertSerialConfig API.

## Description:

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.2

### DIFF
--- a/tasmota/tasmota.ino
+++ b/tasmota/tasmota.ino
@@ -337,7 +337,7 @@ void setup(void) {
     Settings->baudrate = APP_BAUDRATE / 300;
     Settings->serial_config = TS_SERIAL_8N1;
   }
-  SetSerialBaudrate(Settings->baudrate * 300);  // Reset serial interface if current baudrate is different from requested baudrate
+  SetSerialBegin();                            // Reset serial interface if current serial settings are different from requested serial settings
 
   if (1 == RtcReboot.fast_reboot_count) {      // Allow setting override only when all is well
     UpdateQuickPowerCycle(true);
--- a/tasmota/tasmota.ino
+++ b/tasmota/tasmota.ino
@@ -337,7 +337,7 @@ void setup(void) {
     Settings->baudrate = APP_BAUDRATE / 300;
     Settings->serial_config = TS_SERIAL_8N1;
   }
-  SetSerialBegin();                            // Reset serial interface if current serial settings are different from requested serial settings
+  SetSerialBaudrate(Settings->baudrate * 300);  // Reset serial interface if current baudrate is different from requested baudrate
 
   if (1 == RtcReboot.fast_reboot_count) {      // Allow setting override only when all is well
     UpdateQuickPowerCycle(true);
--- a/lib/default/TasmotaSerial-3.3.0/src/TasmotaSerial.cpp
+++ b/lib/default/TasmotaSerial-3.3.0/src/TasmotaSerial.cpp
@@ -121,21 +121,6 @@ bool TasmotaSerial::isValidGPIOpin(int pin) {
 
 bool TasmotaSerial::begin(uint32_t speed, uint32_t config) {
   if (!m_valid) { return false; }
-  if (config > 2) {
-    // Legacy support where software serial fakes two stop bits if either stop bits is 2 or parity is not None
-    m_stop_bits = ((config &0x30) >> 5) +1;
-    if ((1 == m_stop_bits) && (config &0x03)) {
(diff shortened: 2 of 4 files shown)


wir3z:
Ran tests on the USB serial analyzer to confirm:
ESP8266:
- hardware and software serial bridge 9600 8N1 and 9600 8E1

ESP32
- hardware and software serial bridge 9600 8N1 and 9600 8E1

tasmota/tasmota.ino
```diff
@@ -337,7 +337,7 @@ void setup(void) {
     Settings->baudrate = APP_BAUDRATE / 300;
     Settings->serial_config = TS_SERIAL_8N1;
   }
-  SetSerialBaudrate(Settings->baudrate * 300);  // Reset serial interface if current baudrate is different from requested baudrate
+  SetSerialBegin();                            // Reset serial interface if current serial settings are different from requested serial settings
```
arendst, line 340: Another try do ya. I tried to explain why this is wrong in a fix after your first pr. This breaks any which is in quick power cycle mode as it will corrupt flash. Don't do this.

wir3z: Whoops... didn't realize that.  Reversing changes on that hunk.

arendst:
Thx. I see what you try to achieve.

I'll refactor the code an remove the "old" begin() config of 1 or 2. So onve that is done there should be no more problem with regards to correct config for both software and hardware serial.

I'll let you know when I'm done.

arendst:
Again Thx.

Your suggestions have been merged.

wir3z:
> Thx. I see what you try to achieve.
> 
> I'll refactor the code an remove the "old" begin() config of 1 or 2. So once that is done there should be no more problem with regards to correct config for both software and hardware serial.
> 
> I'll let you know when I'm done.

No worries, and and thanks for such a great project.  Just had an "angry old alarm panel" that needed functional parity.  :)

sfromis:
Got serial trouble with a SenseAir S8 sensor on ESP8266 again. Using precompiled tasmota-sensors.bin.gz builds from tasmota/install, I zoomed in to this change being the culprit. Last build before this worked, and after this one, up to current, fails.

```
"StatusFWR":{"Version":"10.1.0.3(sensors)","BuildDateTime":"2021-12-30T13:40:12"  fails
"StatusFWR":{"Version":"10.1.0.3(sensors)","BuildDateTime":"2021-12-30T12:27:54"  works
```

wir3z:
Which baudrate are you using?  8N1 as well?  Can you let me know what IO pins you have configured for this?

sfromis:
I'm simply using standard hardware serial, with the [xsns_17_senseair.ino](https://github.com/arendst/Tasmota/blob/development/tasmota/xsns_17_senseair.ino) driver, using TasmotaModbus, and a Begin call including the constant `SENSEAIR_MODBUS_SPEED` with a value of 9600.

The reason for my "again" was due to me already experiencing same symptoms with another refactoring of serial, basically borking all serial on ESP8266. This looks different, and also hits ESP32 this time.
https://github.com/arendst/Tasmota/discussions/14062#discussioncomment-1868566

wir3z:
I think this is a case of "two wrongs made a right previously".  It looks like a bug in the Tasmota ModBus driver:
int TasmotaModbus::Begin(long speed, int stop_bits)

int stop_bits  cannot be just stop bits, it needs to be uint32_t config (IE SERIAL_8N1 not just a '1') since it's inheriting this class:
bool TasmotaSerial::begin(uint32_t speed, uint32_t config)

sfromis:
Wherever that fix needs to be done, refactoring of serial to fix other issues should ensure existing use of the driver still working.

arendst:
I see what is wrong. My error. Will fix now.

arendst:
Latest dev commit should solve ALL modbus library using drivers.

sfromis:
Confirmed that current dev precompiled tasmota-sensors.bin.gz works with my S8 again :-)

PR #14315, opened 2022-01-07
[Solax X1] Resolve RTS-timing problem

SteWers:
## Description:
On some devices there is a timing problem with the RTS line, when not using flush. It seems that it isn't only a send buffer flush.
flush does also wait until all data is send: https://www.arduino.cc/reference/en/language/functions/communication/serial/flush/

## Checklist:


arendst:
You're right.

PR #14322, opened 2022-01-07
Add favicon on browser

### LINKED ISSUE #4830: favicon in Tasmota
Hello, i am new to esp8266 and tasmota !!
i have a question can i add favicon in html of tasmota ?

barbudor:
## Description:

[**Related issue (if applicable):** fixes #4830](https://github.com/arendst/Tasmota/issues/4830)

Add a favicon directly encoded in base64 in the HTML header (not served through the webserver)
May be not the most pretty tasmo'art but compact (16x16, 16 grey levels).
Only cost 256 bytes of flash due to unishox compression (🙏 @s-hadinger )

Needed a small change in html compression tool to support string with `//` inside. => @s-hadinger 
(replaced the split() with a simple state-machine parser)

@gemu2015 could you please check that I've not broken SCRIPT_FULL_WEBPAGE please ?

image

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.2


arendst:
Thx. As the icon is actually black and white do you see a chance to lower it's color depth from 16 gray scale to 1 gray scale?

I don't use compression so for me it adds 400+ bytes bogus information.

s-hadinger:
Awesome!

s-hadinger:
Black and white would make it ugly since there would be no dithering. 
What about a #ifdef?

arendst:
I love #ifdef.

barbudor:
Hi @arendst 
Reduced to 1 bit which move the HTML cost from 488 down to 328 (-160 bytes)
No gain on compressed, still 256 bytes.
Aliaising is not as bad as I would have expected
image
Indeed the browser or windows is doing some antialiaising on its own
image

I added `#ifdef USE_WEB_FAVICON`
But because compressor script ignore preprocessing it will work only for uncompressed.
When using unishox, favicon will always be included.

arendst:
Thx. Will chk today

InHi66:
i do it this way:

in xdrv_01_webserver.ino

i add at "const char HTTP_HEAD_STYLE3[]"

following 

</style>""" & vblf & "  ""<link rel='icon' type='image/gif' href='data:image/gif;base64,R0lGODlhEAAQAPABAAAAAAAAACH5BAUAAAEALAAAAAAQABAAAAIrjB8AyKjaGEwrUldlPpNv7BnPIy4ThFbdqX5JKDbZKsOvVuayyVsv6rMhCgA7'>""" & vblf & vblf

PR #14347, opened 2022-01-09
Automatic adjustment of file size in xdrv_50_filesystem.ino

Staars:
Makes it easier to load a file of unknown size into a buffer.

## Description:

Example use case:
Using a JSON-file to store a config, where the resulting file size will be unknown, when it will be loaded into memory again. Only important thing now is, that the load memory buffer is big enough, otherwise the JSON would be incomplete/invalid, which will be seen at parse time, but will not lead to a crash in the load function.

So for unknown file sizes call `TfsLoadFile(fname, buf, sizeof(buf))`, which will work for known file sizes too. 
It is still possible to call `TfsLoadFile(fname, buf, fileSize)` as before, so no breaking change.

## Checklist:

### DIFF
--- a/tasmota/xdrv_50_filesystem.ino
+++ b/tasmota/xdrv_50_filesystem.ino
@@ -325,6 +325,11 @@ bool TfsLoadFile(const char *fname, uint8_t *buf, uint32_t len) {
     return false;
   }
 
+  size_t flen = file.size();
+  if (len > flen){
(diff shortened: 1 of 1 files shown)


arendst:
Pls build agains development NOT master.

Staars:
Upps …

Staars:
I'll start again.

PR #14350, opened 2022-01-09
Berry fast loop

s-hadinger:
## Description:

Enable fast loop mode for Berry. This features allows to call a Berry function at each Tasmota loop iteration. This is useful for special cases where you need near realtime control, like audio or animations. For example, with `Sleep 10`, Berry callbacks will be called every 10ms; even more often with `Sleep 0`.

Special attention is made so that there is no or very little impact on performance. Until a first callback is registered, performance is not impacted and Berry is not called. This protects any current use from any performance impact.

Once a callback is registered, it is called separately from Berry drivers to ensure minimal overhead.

`tasmota.add_fast_loop(cl:function) -> nil` registers a callback to be called in fast loop mode.

The callback is called without any parameter and does not need to return anything.

Warning, if you need to register a method from an instance, you need a closure:

```
class my_driver
  def every_100ms()
    # called every 100ms via normal way
  end

  def fast_loop()
    # called at each iteration, and needs to be registered separately and explicitly
  end

  def init()
    # register fast_loop method
    tasmota.add_fast_loop(/-> self.fast_loop())
    # variant:
    # tasmota.add_fast_loop(def () self.fast_loop() end)
  end
end

tasmota.add_driver(my_driver())    # register driver
```

## Checklist:


ortegafernando:
Hi, just in case, I have checked docs and this change is not documented. 
https://tasmota.github.io/docs/Berry/#creating-a-tasmota-driver
will be better to update also the docs when a PR is merged?
Sorry if I am too fast (this PR has been merged only 2 hours ago)

s-hadinger:
Yes you are a little too fast... If you want to use it now, refer at the above as documentation. It will be copied to the Tasmota doc soon.

PR #14401, opened 2022-01-12
Berry Partition Manager v2

s-hadinger:
## Description:

New version of Partition Manager to allow for different sizes for `app0` and `app1` partitions.

Usage: copy `partition.be` in file-system, in the Berry console type `import partition` and click on the `Partition Manager` button on the `Consoles` page.

For LVGL with 4MB flash, one interesting setup will be:
- `app0` 1984KB for optimal code size
- `app1` 896KB which allows for a Tasmota32-minimal firmware
- file-system with 1152KB of free space

part_mgr_2_1

And new UI for setting the sizes:
part_mgr_2_2

## Checklist:


tony-fav:
Wow! This is an excellent tool!

digiblur:
Excellent work @s-hadinger !!  I had a C3 with a screwed up partition table.  I was able to fix it in a couple steps and didn't need to swap it or pull the chip off the board again.

s-hadinger:
@digiblur Thanks for your kind feedback. 

Is there any major feature missing? Although I generally don't expect users to mess around with partition table, unless they need it and they know what they're doing.

digiblur:
I didn't see anything really, I was just winging it on a ESP32-C3 I would have had to pull since it wasn't able to be flashed in place due to the pulldowns/ups on the device.  

I opened a freshly flashed C3 to see what values it had and modified the SPIFFS first as it had unallocated space, sent over a bin and I saw the partition part at the bottom had options now.  I made it match the freshly flashed one and called it a day.  Seemed simple enough.  Only thing I could recommend would be an easy reference of standard sizes for folks to look at based on the device (if there are differences).

tony-fav:
As I understand it the ESP32 OTA process is basically uploading to the opposite app partition that's currently in use then telling the ESP to boot from that one, right? Is there some way to execute that upload without the switch? I noticed you cannot resize things completely if you are not on app0? Is there some way to ensure app0 has a Tasmota build on it to switch to (some magic bytes or something) or is the suggestion just to OTA to get it to switch?

s-hadinger:
I haven't explored all possibilities but if the current partition is not app0, I advise to re-upload the firmware once again.

Jason2866:
Minicustoms (minimal32) ESP32 ESP32-C3 and ESP32-S2 are available from https://github.com/tasmota/install/tree/main/firmware/unofficial
The support update via mqtt, mqtt via TLS, http and https.

PR #14405, opened 2022-01-12
Fix for SHT mode 1 with pulsed output: Stop pulse must be at same dir…

rakaiser:
…ection as start pulse.

## Description:

## Checklist:

### DIFF
--- a/tasmota/xdrv_27_shutter.ino
+++ b/tasmota/xdrv_27_shutter.ino
@@ -476,7 +476,8 @@ void ShutterPowerOff(uint8_t i)
 {
   AddLog(LOG_LEVEL_DEBUG, PSTR("SHT: Stop %d. Mode %d, time:%d"), i+1,Shutter[i].switch_mode, Shutter[i].time); // fix log to indicate correct shutter number
   ShutterDecellerateForStop(i);
-  if (Shutter[i].direction !=0) {
+  uint8_t current_direction = Shutter[i].direction;
+  if (current_direction !=0) {
     Shutter[i].direction = 0;
   }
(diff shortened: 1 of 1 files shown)


arendst:
@stefanbode pls advise.

stefanbode:
I do not see any benefit of the change. Please explain what it will change and why this is required.

stefanbode:
I agree that on pulse motors the pulse should be on the same relay. But specially with multiple shutter assigned I assume this will not work. I will check why the original code flip the wrong relay. We have the movement direction already. Currently I assume the fix must be a little different.

stefanbode:
Ok, I understand now the problem. Let me check later if a movement of the direction=0 if clause behind the switch is possible. I agree setting this to 0 at that point create a static value for the switch. And this is for sure wrong.

stefanbode:
I will bring up line 496 up after line 478. Then the relay is fine. Currently only on my mobile with no edit options to do so

arendst:
Thx. I will make the suggested line change today.

arendst:
Merged as requested.

rakaiser:
Thx for fixing that obvious flaw. Works ok with my modified belt winder. Mode SHT_TIME combined with SHT_PULSE obviously not yet used.

stefanbode:
Thanks for fixing this. Still on "holiday" after a motorbike accident in Thailand. I assume no other side effects. The issue came up when optimizing the operation time to ensure accurate movement. Automate testing is a little bit tricky because you do not get back easily feedback on the standard procedures. e.g. if the right relay way ON for 0.2 seconds. Anyhow @rakaiser thanks for bringing up any issue and we are happy to solve this asap.

PR #14424, opened 2022-01-13
Inconsistencies in decode-status.py feature flags

sfromis:
Not all flag names were updated to match updates to Tasmota source in `support_features.ino`. I assume the changes to be uncontroversial, with the possible exception of:

While `USE_TUYA_DIMMER` arguably is a more descriptive name than `USE_TUYA_MCU` the former name is not found anywhere in Tasmota code, which has 
```
#if defined(USE_LIGHT) && defined(USE_TUYA_MCU)
    feature2 |= 0x00008000;  // xdrv_16_tuyadimmer.ino
#endif
```
My assumption is that the actually used names "wins", to avoid user confusion, even if you could say that the name `USE_TUYA_MCU` is too broad for the actual feature. In principle, it would be more logical to create a new "real" `#define` name, but backwards compatibility and all that.... Makes sense? 
I'm ok either way....

## Description:

## Checklist:


barbudor:
Isn't the official repo for decode-config here ? 
https://github.com/tasmota/decode-config

sfromis:
> Isn't the official repo for decode-config here ?
> https://github.com/tasmota/decode-config

When I asked that question in [#12090](https://github.com/arendst/Tasmota/pull/12090), I was reassured that the correct location is here, not in the the other repository used to package the tool for user download, like in .exe files.

barbudor:
Thanks. Good to know.
I openned an issue on the other side but that is related to Windows EXE packaging, so that was the right place.
However I have an automated script to update the py file on my server from the other side. So I will need to get this one instead.

sfromis:
To nitpick, the other side for this tool would instead be https://github.com/tasmota/decode-status :-D 

When using the tool for both execution and parsing, some time ago I changed to let it be based on the main repository, picked up from when I download Tasmota source for builds.

arendst:
Thx.

This is the only location where I update this python script. All others should be copies of this.

PR #14458, opened 2022-01-15
INA28022 support in INA219 driver

barbudor:
## Description:

**Related discussion :** fixes #14356

It appears that ISL28022 is very similar to INA219 with identical register map but supports up to 60V.
Because of that , it's not worth a dedicated driver.

This PR makes the existing ina219_13_xsns driver seamlessly compatible with both chips  with auto detection.

I took also the opportunity to refactor the driver, removing all static data in favor of dynamic allocation if at least 1 chip is discovered.

## Checklist:


barbudor:
tagging @mhardy6351

barbudor:
Thanks Theo

PR #14485, opened 2022-01-17
BLE_EQ3: Fix and enhance the JSON response

obrain17:
## Description:

Fix and enhance the JSON response  after command execution.

a) Add

-   "windowdur"
-   "windowtemp"
-   "day"
-   "night"
-   "offset"

to "ResponseAppend_P" (json response)

b) Assign cmdtype = 23 to "lock" so the correct `"cmd":"lock"` will show in "ResponseAppend_P

**Info:** I have also re-worked the documentation and will create  a separate PR for it the next days

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9  (ESP8266 has no Bluetooth)

### DIFF
--- a/tasmota/xdrv_85_BLE_EQ3_TRV.ino
+++ b/tasmota/xdrv_85_BLE_EQ3_TRV.ino
@@ -188,7 +188,8 @@ const char *cmdnames[] = {
   "day",
   "night",
   "reqprofile",
-  "setprofile"
+  "setprofile",
+  "lock",
 };
 
 const uint8_t *macprefixes[1] = {
@@ -206,7 +207,7 @@ struct eq3_device_tag{
   int8_t RSSI;
   uint64_t timeoutTime;
   uint8_t pairing;
-  uint8_t lastStatus[10]; // last received 02 stat
+  uint8_t lastStatus[16]; // last received 02 stat
   uint8_t lastStatusLen;
   uint32_t lastStatusTime; // in utc
   uint8_t nextDiscoveryData;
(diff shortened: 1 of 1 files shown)


barbudor:
@btsimonh @ysbrand
Could you please kindly review this PR and if possible test ?
Thanks

btsimonh:
it looks fine to me.  Basically, it's a bugfix - when a lock command is sent, the response is currently reported as a poll response, this PR reports it correctly as a 'lock' response.
Ref the additions to the JSON response, there was something funny about the EQ3 response length, but I believe the implementation is correct ref responses >10 bytes.
I don't think the addition of variables in the JSON should cause anyone any problems.
Have to say that lock was probably not too heavily tested in the original.

I have not built or tested but 95% positive this will not cause any problems.

Ysbrand:
I'm currently reviewing this with @btsimonh . In principle the changes look ok to us.
Not sure how we ever missed the "lock" status message during testing (altough it's not really impacting operations, but it is indeed not correct).
I'm not sure how much these extra json fields will be used, but they won't hurt either.

I will attempt to build this version and test.

obrain17:
Probably the extra json fields  might not be used regularly. But you might want to know e.g. the temperature it will switch to, when you call `day` or `night`.  Now you always have a complete status of the TRV reported with every `poll`  and kept updated in your home automation system.

Ysbrand:
I've build the binary and did upload it to my test rig. All looks as documented and I do not see any issues.

2 minor comments:

- Can you update the response example in the code starting at line 70 to match the current json (it was already outdated anyway, but it just to avoid confusion when someone looks at the code)
- Not sure if it makes sense but when one sets the windowtemp and duration,the order is 'temp, duration', I would prefer if we keep that order in the json as well.

Other than that, thanks for the effort making those improvements!

barbudor:
@obrain17 Are you ok to adjust your PR following @Ysbrand comment ?

obrain17:
Yes I am ok with it - and already did it:
see latest commit

barbudor:
Thanks
@arendst Looks like this is ok for merging at your convenience

PR #14491, opened 2022-01-17
Adding HomeKit to legacy MI32 driver

Staars:
## Description:

Besides adding a new GUI, configuration file, Berry mapping and new sensors. Tested with ESP32 and ESP32C3.
Preliminary documentation here: https://staars.github.io/site/Bluetooth_MI32/

To compile new environments in platform_tasmota.cenv.ini could be added like so:
```
[env:tasmota32-mi32-homebridge]
extends                 = env:tasmota32_base

build_flags             = ${env:tasmota32_base.build_flags}
                          -DFIRMWARE_BLUETOOTH 
                          -DUSE_MI_HOMEKIT=1
lib_extra_dirs          = lib/libesp32, lib/libesp32_div, lib/lib_ssl, lib/lib_basic
lib_ignore              = ESP8266Audio
                          ESP8266SAM
                          TTGO TWatch Library
                          Micro-RTSP
                          epdiy
                          esp32-camera

[env:tasmota32c3-mi32-homebridge]
extends                 = env:tasmota32c3

build_flags             = ${env:tasmota32_base.build_flags}
                          -DFIRMWARE_BLUETOOTH 
                          -DUSE_MI_HOMEKIT=1
lib_extra_dirs          = lib/libesp32, lib/libesp32_div, lib/lib_ssl, lib/lib_basic
lib_ignore              = ESP8266Audio
                          ESP8266SAM
                          TTGO TWatch Library
                          Micro-RTSP
                          epdiy
                          esp32-camera
```

This is not a replacement for the current BLE driver and is intended for custom builds. 

## Checklist:

### DIFF
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -48,6 +48,11 @@ be_extern_native_module(lv_extra);
 be_extern_native_module(lv_tasmota);
 #endif // USE_LVGL
 
+#ifdef USE_MI_ESP32
+extern void be_load_MI32_class(bvm *vm);
+extern void be_load_BLE_class(bvm *vm);
+#endif //USE_MI_ESP32
+
 /* user-defined modules declare start */
 
 /* user-defined modules declare end */
@@ -229,5 +234,9 @@ BERRY_API void be_load_custom_libs(bvm *vm)
     be_load_lv_wifi_arcs_icon_class(vm);
     be_load_lv_clock_icon_class(vm);
 #endif // USE_LVGL
+#ifdef USE_MI_ESP32
+    be_load_MI32_class(vm);
+    be_load_BLE_class(vm);
+#endif //USE_MI_ESP32
 }
 #endif
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -48,7 +48,7 @@ be_extern_native_module(lv_extra);
 be_extern_native_module(lv_tasmota);
 #endif // USE_LVGL
 
-#ifdef USE_MI_ESP32
+#if defined(USE_MI_ESP32) && !defined(USE_BLE_ESP32)
 extern void be_load_MI32_class(bvm *vm);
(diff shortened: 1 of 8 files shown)


tony-fav:
Hey @Staars loving this driver with Berry mappings! Any chance or plans for options to enable active scan and adjust BLE duty cycle or for Berry methods to subscribe to notifications?

Again, great work!

tony-fav:
Awesome! I am not totally invested into what possible data an Xbox controller is spitting out! 😅

My primary interest is that this drivers seems way more stable than the typical one but if some of these features start pushing towards instability, I understand. I've been supporting as many devices as I can via berry scripting and rule callbacks here: https://github.com/tony-fav/tasmota-blerry/tree/dev-support-NewBLEDriver#sensors and that WP6003 air box was my next up when I realized it was a notify not an advertisement.

And just to clarify, I am looking for active scanning not active connection, just to get the "full" advertisement packet from some devices. I've been able to do the momentary connection using `BLEOp` already for control with things like Switchbot.

Staars:
Of course I would never suggest to write a c/cpp driver for a xbox controller for Tasmota, but with Berry it does not matter so much. If someone needs/wants an exotic solution, then he can try out whatever he wants in Berry. The main code base does not grow and we can concentrate on Tasmotas core stability.
BTW, I finally got my controller example working. The pairing procedure is totally cumbersome, awfully slow and  the typical point of failure, but after that it works surprisingly well, which makes me very optimistic that it could be usable for more realistic use cases.
I will try to clean up the things and do a PR in the next few days. Maybe after that, a reevaluation of the API makes sense, because I had zero external input for this so far and it is possible, that there is room for optimization.

Yes, I did recognize, that you were talking about active scanning and not connections, but I do not like the idea, that a BLE device forces every other BLE device in sight to give a scan response 24/7. I know, that the added battery drain and data traffic is pretty small but I just do not like it.
As long as this is not the default setting, "nothing bad will happen" and if it is used with purpose, it totally makes sense. My current Berry solution for the returned byte buffer was only my first attempt and I already had changes for it in mind, which then could also handle active scan responses. It should technically be no big deal to implement it, I just want to think a bit about.

PR #14494, opened 2022-01-18
fix ambiguous error in knx

Jason2866:
## Description:

when compile with upstream Arduino core ESP32
Thx @Staars 

## Checklist:


ascillato:
Thanks.

For anyone reaching this, this change does not affect Tasmota KNX in any way, since that file is not used in Tasmota KNX driver. It is there only as part of the library. This change was required just for compilation but it is not included in the final firmware.

PR #14531, opened 2022-01-20
[SML] Serial activity monitor via LED

SteWers:
## Description:

Visualize the activity of one or all serial meters via LED. It is helpful for debugging and also for viewing serial activity easily. The configuration is done via console commands:

- LED-GPIO for monitoring serial activity => sensor53 l2, l13, l15 ... or l255 for turn off (default)
- meter number for monitoring serial activity => sensor53 m1, m2, m3 ... or m0 for all (default)

To start at boot-time, add the entries in the boot-section (>B) of the script.

## Checklist:

### DIFF
--- a/tasmota/xsns_53_sml.ino
+++ b/tasmota/xsns_53_sml.ino
@@ -1041,6 +1041,9 @@ void ADS1115_init(void) {
 char sml_start;
 uint8_t dump2log=0;
 
+uint8_t ser_act_LED_pin=255;
+uint8_t ser_act_meter_num=0;
+
 #define SML_SAVAILABLE Serial_available()
 #define SML_SREAD Serial_read()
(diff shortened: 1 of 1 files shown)


barbudor:
It doesn't feel natural for me to define a Led GPIO from a command.
Why not defining a dedicated "SML Led" to be used in template/configuration ?

SteWers:
In general I agree with you. I thought about an extra pin via template. The cause, because I decided for the other way, is because all other configuration for SML (counter/meter) is done via script. So there are no different points, where the configuration has to be done.

barbudor:
I'm not really versed into Scripting so I trust on that 😀

gemu2015:
i agree with @SteWers .
SML is descriptor based and completely decoupled from Tasmota GPIO definition.
so this solution is feasible.

PR #14547, opened 2022-01-21
esp8266 email, scripter on 1m devices

gemu2015:
## Description:

fixes esp8266 send_mail missing bearssl
fixes compile error scripter on devices without UFILESYS

## Checklist:

### DIFF
--- a/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
+++ b/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
@@ -27,7 +27,7 @@
 
 #include "tasmota_options.h"
 
-#if defined(ESP8266) && defined(USE_TLS)
+#if defined(ESP8266) && (defined(USE_TLS) || defined(USE_SENDMAIL))
 #include <stdint.h>
 #include <stdlib.h>
 #include "StackThunk_light.h"
--- a/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
+++ b/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
@@ -27,7 +27,7 @@
 
 #include "tasmota_options.h"
 
-#if defined(ESP8266) && (defined(USE_TLS) || defined(USE_SENDMAIL))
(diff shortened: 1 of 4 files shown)


lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
```diff
@@ -27,7 +27,7 @@
 
 #include "tasmota_options.h"
 
-#if defined(ESP8266) && defined(USE_TLS)
+#if defined(ESP8266) && (defined(USE_TLS) || defined(USE_SENDMAIL))
```
arendst, line 30: Isn't this solved with the latest re-organization of config header files?

gemu2015: yes it should but at least it does not compile (sometimes) strange ?  platformio bug?
ok, reverted this as it SHOULD work

barbudor: May be tasmota_configurations.h is missing something like ?
```
#ifdef USE_SENDMAIL
#define USE_TLS
#endif
```

gemu2015: this should set TLS but obviously doesn't do it (always)

myuserconfig =>

#if defined(USE_MQTT_TLS) || defined(USE_SENDMAIL) || defined(USE_TELEGRAM) || defined(USE_WEBCLIENT_HTTPS) || defined(USE_ALEXA_AVS)
  #define USE_TLS                                  // flag indicates we need to include TLS code
#endif

barbudor: THis should be at the end of tasmota_configurations
The compiler will not see that is USE_SENDMAIL is added in user_config_override.h as the order is
- my_user_config
- user_config_override
- tasmota_configurations

Fixing the dependencies should be done in the latest

gemu2015: i have a slightly outdated version. this must have been corrected very recently
with the updated version it works now

PR #14555, opened 2022-01-21
Add setting for IRremoteESP8266 tolerance

hr-kapanakov:
## Description:
Add setting that allow changing IRremoteESP8266 tolerance.
Discussion as ref: https://github.com/arendst/Tasmota/discussions/14435

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -858,6 +858,7 @@
     #define IR_RCV_TIMEOUT          15           // Number of milli-Seconds of no-more-data before we consider a message ended (default 15)
     #define IR_RCV_MIN_UNKNOWN_SIZE 6            // Set the smallest sized "UNKNOWN" message packets we actually care about (default 6, max 255)
     #define IR_RCV_WHILE_SENDING    0            // Turns on receiver while sending messages, i.e. receive your own. This is unreliable and can cause IR timing issues
+    #define IR_RCV_TOLERANCE        25           // Base tolerance percentage for matching incoming IR messages (default 25, max 100)
 
 // -- Zigbee interface ----------------------------
 //#define USE_ZIGBEE                                // Enable serial communication with Zigbee CC2530/CC2652 flashed with ZNP or EFR32 flashed with EZSP (+49k code, +3k mem)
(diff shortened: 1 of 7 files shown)


s-hadinger:
You beat me on this one. It was on my todo list, well done. Theo was also proposing to reuse `P_ex_TUYA_VOLTAGE_ID` as you did.

I'm just worried about leftovers in `P_ex_TUYA_VOLTAGE_ID`. Also if a user updates with OTA, the value of the option will be zero which translates to zero tolerance; and things will break. I.e. the options equals zero should not have any regression.

Is Tolerance==0 even valid?

hr-kapanakov:
Yes, I am agree about the leftovers. About zero tolerance it seems ok based on the IRremoteESP8266 docs - https://crankyoldgit.github.io/IRremoteESP8266/doxygen/html/classIRrecv.html#aa091c449db70c65fd0221669df7438ea. I also checked the code and also seems fine to be zero: https://github.com/crankyoldgit/IRremoteESP8266/blob/93af543dcb849918b0c342ed1b6af68953a61d06/src/IRrecv.cpp#L1123
The only problem is that in such case the IR signal should match at 100% which most probably is not possible in real world.

hr-kapanakov:
Yeah, it seems a good compromise. I'll make the change.

s-hadinger (approved):

arendst:
Thx. I need to make one final chk if ex tuya voltage id isn't used for version delta.

once chkd today i will merge.

sfromis:
Looks like the "sacrifice" of the zero value to mean default of 25 did not work for existing configs. When upgrading firmware on an IR puck, it stopped working, with `SetOption44` being zero. Only after manually setting the tolerance, I could use a remote control again. 

While zero for default works when issuing the command `SetOption44 0`, it becomes a breaking change for existing configs, where the existing value of the config option byte will usually be zero.

s-hadinger:
@sfromis You are right, `IrReceiveUpdateTolerance()` is never called when doing an update. I suggest it is called in `IrReceiveInit()` too

s-hadinger:
@sfromis It should be fixed now

PR #14581, opened 2022-01-23
Supports for MQ analog sensor for air quality

francescoadriani:
## Description:
It's support the analog sensor for quality air (also for esp-32)
-added on list of assignable pins in module
-with adcparams can choose whic mqx you have with seconds param ex: adcparam 10 4 (you have the mq-04)
immagine_2022-01-23_172341

## Checklist:


tasmota/xsns_02_analog.ino
```diff
@@ -334,6 +370,32 @@ uint16_t AdcGetLux(uint32_t idx) {
...
+  float _RL = 10; //Value in KiloOhms
+  float _R0 = 10;
+  int _adc = AdcRead(Adc[idx].pin, 2);
+  avg += _adc;
+  for (int i = 0; i < retries-1; i ++) {
+    delay(retries_delay);
+    _adc = AdcRead(Adc[idx].pin, 2);
+    avg += _adc;
+  }
+  float voltage = (avg/ retries) * ANALOG_V33 / ((pow(2, ANALOG_RESOLUTION)) - 1);
```
barbudor, line 386: We are rtying to avoid using the heavy `pow(..)`standard function.
Please use Tasmota's `FastPrecisePow(...)` instead

tasmota/xsns_02_analog.ino
```diff
@@ -334,6 +370,32 @@ uint16_t AdcGetLux(uint32_t idx) {
...
+
+  float _RS_Calc = ((ANALOG_V33 * _RL) / voltage) -_RL; //Get value of RS in a gas
+  if (_RS_Calc < 0)  _RS_Calc = 0; //No negative values accepted.
+  float _ratio = _RS_Calc / _R0;   // Get ratio RS_gas/RS_air
+  float ppm= Adc[idx].param2/ANALOG_MQ_DECIMAL_MULTIPLIER*pow(_ratio, Adc[idx].param3/ANALOG_MQ_DECIMAL_MULTIPLIER); // <- Source excel analisis https://github.com/miguel5612/MQSensorsLib_Docs/tree/master/Internal_design_documents
```
barbudor, line 391: Same as above

tasmota/xsns_02_analog.ino
```diff
@@ -36,6 +37,8 @@
 #define ANALOG_RANGE                  4095             // 4095 = 12, 2047 = 11, 1023 = 10
 #undef ANALOG_PERCENT
 #define ANALOG_PERCENT                ((ANALOG_RANGE + 50) / 100)  // approximation to 1% ADC range
+#undef ANALOG_BOARD
+#define ANALOG_BOARD                  "ESP-32"
```
barbudor, line 41: What is this #define ANALOG_BOARD used for ?

tasmota/xsns_02_analog.ino
barbudor: You are assuming a 3V3 range which is true on a ESP8266 Wemos D1 / NodeMCU due to the existing voltage divider.
Will be false on an different board with different consideration such as a bare ESP module.

I'm not sure that the range is the same on ESP32 DevKit or various ESP32 based boards.

francescoadriani: You right, sorry!
DONE

tasmota/xsns_02_analog.ino
francescoadriani: You right, sorry!
DONE

tasmota/xsns_02_analog.ino
francescoadriani: I'm used library MQUnifiedsensor-2.0.1 but there it isn't used... I'm solving

tasmota/xsns_02_analog.ino
francescoadriani: Suggest?

francescoadriani: ANALOG_V33 const is also used for other analog reader

barbudor: There is 1 use case in `AdcEverySecond()` where there is a #ifdef ESP8266 ...

You have checked both box that the PR has been tested on ESP8266 and on ESP-32
Is that the case ?
On which boards have you tested ?
Can you garantue that this PR will work exactly the same on standard ESP8266 and ESP-32 boards ?

tasmota/xsns_02_analog.ino
```diff
@@ -334,6 +367,32 @@ uint16_t AdcGetLux(uint32_t idx) {
...
+  float _RL = 10; //Value in KiloOhms
+  float _R0 = 10;
+  int _adc = AdcRead(Adc[idx].pin, 2);
+  avg += _adc;
+  for (int i = 0; i < retries-1; i ++) {
+    delay(retries_delay);
+    _adc = AdcRead(Adc[idx].pin, 2);
+    avg += _adc;
+  }
+  float voltage = (avg/ retries) * ANALOG_V33 / ((FastPrecisePow(2, ANALOG_RESOLUTION)) - 1);
```
barbudor, line 383: As per my other suggestion, you don't need to use a function to calculate this !
Use the integer left shift.

tasmota/xsns_02_analog.ino
francescoadriani: i tried with wemos d1 mini and esp32 dev kit

francescoadriani: Wemos d1 mini supports 3.2v max voltage
https://www.wemos.cc/en/latest/d1/d1_mini.html

francescoadriani: I don't know esp board that works with max voltage other than 3,3v

tasmota/xsns_02_analog.ino
francescoadriani: yes could be better, off course

tasmota/xsns_02_analog.ino
francescoadriani: Esp8266.... 0..1V

francescoadriani: Ok how could i fix them?

francescoadriani: THe problem is you can't use mq if you have max voltage of 1V!

francescoadriani: @barbudor how could we do with 1V ADC?

barbudor: It's ok, most people will use a Wemos or NodeMCU
Just add in the beginning of the code some pseudo wiring diagram like for the other sensors
And when you will PR the docs, don't forget to add that it needs a 3V3 input like NodeMCU/WemosD1
Thanks

francescoadriani: Done, thanks

erdinguma:
Issuing command: adcparam 10 4 is not changing the sensor from the default MQ2 to MQ4. Any guidance?

barbudor:
@erdinguma Did you correctly configured your ADC input in the GPIO configuration ?

erdinguma:
@barbudor Yes. Selected “ADC MQ” and I am reading values. Just need to change it to MQ 4.

rice103:
https://github.com/tasmota/docs/blob/development/docs/ADC.md

rice103:
Yeah!

erdinguma:
For future enhancement, considering high sensitivity of the sensor to humidity and temperature, I inferred from documentation the following relationship of ratio (rs/r0) to temp and humidity.

image

Ratio adjustment factor (F) for Temp T and Humdity H:
F = 5.693977721259878 + -0.023280722319645328 * T + -0.24767739454737786 * H + 0.00054602964962393 * T^2 + -0.0001297777486611509 * T * H + 0.004485775052457577 * H^2 + -7.564523748530513e-06 * T^3 + 1.277287227509287e-06 * T^2 * H + 1.2323178573075948e-06 * T * H^2 + -2.5805390257858728e-05 * H^3

rice103:
Is not simple link two sensor value.. But could be a starting point for future developments

erdinguma:
Yes. I am thinking to implement something similar to Sensirion SGP40 or SGP41 which reads another sensor, SHT4x to compensate VOC sensing for temp and humidity. (I realize SGP41 itself hasn't even been ported to Tasmota yet)

In the interim, I thought to post my compensation formula because it can be used in post processing of data after received from MQTT and it would make a significant difference in results even if not processed at firmware level in real time.

francescoadriani:
@Links2004 https://github.com/arendst/Tasmota/pull/14581#issuecomment-1066853932

PR #14590, opened 2022-01-24
PWM auto-phasing for lights by default (new behavior) unless ``SetOption134 1``

s-hadinger:
## Description:

**ESP32 only** (including ESP32C3/S2)

You can revert to previous behavior (synced PWM) with `SetOption134 1`

Change in behavior for PWM. Until now all PWM phases were synced and started at the same time. This could cause increased stress on the power supply because the instantaneous power is high, even if average is low. This was also blocking from using H-bridge.

IMG_2062

Now the default behavior is that each PWM pulse starts at the end of the previous pulse (yellow is PWM1, blue is PWM2). This also makes H-bridge possible as long as PWM1+PWM2 do not exceed 100% (possibly use `LedTable 0` to disable gamma correction)

IMG_2061

This feature also works for PWM inverted `PWMi`, the phase is based on the active part of the pulse (high for `PWM`, low for `PWMi`).

This feature has currently no impact on `PWM` (when `SetOption15 0`).

This PR also includes minor refactoring of PWM ESP32 functions to remove inlines.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - Command ``SspmLog<relay> [x]`` to retrieve relay power state change and cause logging
 - Command ``SspmScan`` to rescan Sonoff SPM modbus
 - Support for MQ analog sensor for air quality by Francesco Adriani (#14581)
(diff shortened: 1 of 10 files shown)


arendst:
Nice pictures!!

Thx.

PR #14595, opened 2022-01-25
include DeviceName in retained telemetry mqtt message

AlfaBravoX:
I have build an android App that is heavily using Tasmota devices mqtt data. The app is made so, it creates GUI only using Tasmota retain mqtt messages. This is big advantage, because other gui apps usually needs to modify GUI config when new element is added. This is not necessary here. One limitation is missing DeviceName in Telemetry STATE message. IThis patch may help others who are building GUI just based on MQTT messages. to keep my app simple I want to avoid Using discovery Tasmota mqtt messages and deal with multiple topics correlating them later

## Description:

## Checklist:

### DIFF
--- a/tasmota/support_tasmota.ino
+++ b/tasmota/support_tasmota.ino
@@ -762,9 +762,9 @@ void MqttShowState(void)
 #endif  // USE_ADC_VCC
 #endif  // ESP8266
 
-  ResponseAppend_P(PSTR(",\"" D_JSON_HEAPSIZE "\":%d,\"SleepMode\":\"%s\",\"Sleep\":%u,\"LoadAvg\":%u,\"MqttCount\":%u"),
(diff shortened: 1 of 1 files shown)


AlfaBravoX:
compiled on gitpod and tested on ESP device `Environment    Status    Duration
-------------  --------  ------------
tasmota        SUCCESS   00:00:31.801
================================================= 1 succeeded in 00:00:31.801 =================================================
gitpod /workspace/Tasmota $ `

barbudor:
I don't think it makes sense to repeat the DeviceName in a message that is sent periodically as it is a constant that will not change.

INFO1 message that is sent at boot seems more suitable.
You can make INFO1..3 messages retained with command `InfoRetain 1`

AlfaBravoX:
ok, it absolutely make sense what you are saying, however complexity on frontend side (gui)will raise. PR would be accepted if DeviceName will be part of INFO1 ?

Jason2866:
Closing since duplicated, replaced from #14596

ascillato:
>PR would be accepted if DeviceName will be part of INFO1 ?

Having device name in INFO1 message at boot makes sense.

PR #14628, opened 2022-01-27
Increase PWM channels to 16 (Esp32 only)

s-hadinger:
## Description:

- Increase PWM channels to 16 (ESP32), 8 (ESP32S2), 6 (ESP32C3).
- various cleaning in Esp8266toEsp32
- Backlighting of displays is now using 10 bits resolution

## PWM

**ESP8266** has only software and supports 5 PWM channels. `PWM` and `PWMi` GPIOs are used in two modes depending on `SetOption15`: either as lights or as pure PWM.

**ESP32** has hardware PWM support, named `ledc`, up to 16 channels depending on CPU type. You can mix lights and pure PWM channels. The first 5 PWM are reserved for lights, unless `SetOption15 0`. For pure PWM GPIOs, you can assign any PWM number, they don't need to be continuous. For example you can use `PWM 1/2/3` for a 3-channel RGB light, and `PWM 6` & `PWM 10` for pure PWM at the same time.

CPU type|PWM channels
:---|:---
ESP32|16 channels
ESP32S2|8 channels
ESP32C3|6 channels

Channels are assigned to GPIOs in a first-in-first-serve way, and PWM GPIOs are assigned first. If `ledc` channel are exhausted, an error will appear in logs.

The following GPIOs use `ledc` PWM channels:

GPIO type|Description
:---|:---
`PWM` or `PWMi`|`PWM 1..5` are used for lights, `PWM O6..11` are general purpose PWM.
`LedPwmMode`|Assigns a `Led` GPIO to a PWM channel
`Buzzer`|If `BuzzerPwm` is used
`Backlight`|PWM backlighting for displays
`XCLK`|Used as a clock generator for webcam

Example of `PWM` console output with 16 PWM assigned. By default PWM range is 0..1023.

```
RSL: RESULT = {"PWM":{"PWM1":410,"PWM2":286,"PWM3":286,"PWM4":0,"PWM5":0,"PWM6":0,"PWM7":0,"PWM8":0,"PWM9":0,"PWM10":0,"PWM11":0,"PWM12":0,"PWM13":0,"PWM14":0,"PWM15":0,"PWM16":0}}
```

### Auto-phasing of PWM

(ESP32 only) By default, phases of consecutive PWM are disaligned so that a PWM pulses starts when the pulse of the previous PWM channels ends. This helps in distributing over time all pulses and have a smoother effect on power supply.

You can revert this with `SetOption134 1`; all phases are synces and all pulses start at the same moment.

### H-bridge

H-bridge is an electronic circuit that switches the polarity of a voltage applied to a load. It uses 2 PWM outputs to control the current sent to each polarity.

When auto-phasing is enabled, you can use 2 consecutive PWM to drive a H-bridge siunce PWM phases don't overlap - under the condition that the sum of both PWM don't exceed `1023`.

**Important**: you must always ensure that the sum of both PWM channels is less or equal than `1023`. Values over this threshold can damage the circuit.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Command ``SspmScan`` to rescan Sonoff SPM modbus
 - Support for MQ analog sensor for air quality by Francesco Adriani (#14581)
 - Command ``SetOption134 1`` to disable PWM auto-phasing for lights by default (new behavior) (#14590)
+- Increase PWM channels to 16 (Esp32 only)
 
 ### Changed
 - BME68x-Sensor-API library from v3.5.9 to v4.4.7
--- a/lib/lib_display/Arduino_ST7789-gemu-1.0/Arduino_ST7789.cpp
+++ b/lib/lib_display/Arduino_ST7789-gemu-1.0/Arduino_ST7789.cpp
@@ -18,6 +18,7 @@ const uint16_t ST7789_colors[]={ST7789_BLACK,ST7789_WHITE,ST7789_RED,ST7789_GREE
 ST7789_LIGHTGREY,ST7789_DARKGREY,ST7789_ORANGE,ST7789_GREENYELLOW,ST7789_PINK};
 
 #ifdef ESP32
+#include "esp8266toEsp32.h"
 #define ST7789_DIMMER
 #endif
 
@@ -258,11 +259,9 @@ void Arduino_ST7789::commonInit(const uint8_t *cmdList) {
   }
 
   if (_bp>=0) {
-#define ESP32_PWM_CHANNEL 1
+// #define ESP32_PWM_CHANNEL 1
 #ifdef ST7789_DIMMER
-    ledcSetup(ESP32_PWM_CHANNEL,4000,8);
-    ledcAttachPin(_bp,ESP32_PWM_CHANNEL);
-    ledcWrite(ESP32_PWM_CHANNEL,128);
+    analogWrite(_bp, 128);
 #else
     pinMode(_bp, OUTPUT);
 #endif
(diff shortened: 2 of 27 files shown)


s-hadinger:
@gemu2015 this PR allows for up to 16 pwm channels. `analogAttach` is not mandatory anymore, a channel is assigned at first call to `analogWrite`. 

I have fixes all the code impacted but I have seen that scripter directly allocates ledc channels. This might conflict with GPIOs

gemu2015:
@s-hadinger 
thanks for the info. will check how to adapt when it is merged.

arendst:
Wow. This is major! I'll merge and monitor.

blakadder:
Is there interest to code in PWM6+ as a proper power/dimmer output, at least for using MQTT commands for easier integration?

sfromis:
You mean using the Dimmer command instead of the PWM command for a similar effect, but also getting gamma correction?

blakadder:
UI should not be considered, that's way too much work for no gain. 

The idea is that if you make a PWM6 you get a new Power output and Dimmer command in succession. It should also be picked up by Tasmota discovery as a new light, that is all

s-hadinger:
Oh, you mean that any additional PWM would be treated like they are with `SetOption68 1`

That should be doable.

s-hadinger:
One problem though is that there is no room in settings to store the Dimmer value for PWM 6-31

blakadder:
then its a no go, forget it

barbudor:
in persist.json ?

sfromis:
In principle in `persist.json` I suppose, but that's a major detour from how the Dimmer command and persistence works. So major that a different command would make sense. Like using PWM which already is there......

Issue #14483, opened 2022-01-17
DeepSleepTime":28800 does not work correct


ManuelF001:
### PROBLEM DESCRIPTION
Why the DeepSleep is only to 2022-01-17T01:00:00 with "DeepSleepTime":28800   ?
I think with DeepSleepTime":28800 it must be in deepsleep until 2022-01-17T08:00:00?
```haskell
00:00:00.001 HDW: ESP8266EX
00:00:00.049 CFG: Loaded from flash at F4, Count 324
00:00:00.064 Project tasmota - Tasmota Version 10.1.0(tasmota)-2_7_4_9(2021-12-08T14:47:33)
00:00:00.554 WIF: Connecting to AP1 Home1 Channel 5 BSSId 10:FE:ED:AF:61:83 in mode 11n as tasmota-C0A179-0377...
00:00:02.820 WIF: Connected
00:00:03.072 HTP: Web server active on tasmota-C0A179-0377 with IP address 192.168.1.110
20:59:58.101 MQT: Attempting connection...
20:59:58.125 MQT: Connected
20:59:58.129 MQT: tele/tasmota_C0A179/LWT = Online (retained)
20:59:58.130 MQT: cmnd/tasmota_C0A179/POWER = 
20:59:58.184 RUL: SYSTEM#BOOT performs "ruletimer1 65"
20:59:58.191 MQT: stat/tasmota_C0A179/RESULT = {"T1":65,"T2":0,"T3":0,"T4":0,"T5":0,"T6":0,"T7":0,"T8":0}
21:00:02.446 MQT: tele/tasmota_C0A179/STATE = {"Time":"2022-01-16T21:00:02","Uptime":"0T00:00:10","UptimeSec":10,"Heap":28,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":43,"MqttCount":1,"Wifi":{"AP":1,"SSId":"Home1","BSSId":"10:FE:ED:AF:61:83","Channel":5,"Mode":"11n","RSSI":50,"Signal":-75,"LinkCount":1,"Downtime":"0T00:00:04"}}
21:00:02.514 MQT: tele/tasmota_C0A179/SENSOR = {"Time":"2022-01-16T21:00:02","ANALOG":{"A0":1024},"DS18B20":{"Id":"3C01D60797D2","Temperature":16.2},"TempUnit":"C"}
21:00:53.410 RUL: TIME#MINUTE=1260 performs "deepsleeptime 28800"
[...]

PR #14633, opened 2022-01-28
Added WIFI check to avoid extensive battery usage

stefanbode:
The device is sent to an additional deepsleep cycle if there could not establish a WIFI connection with IP in the first 15 seconds after startup. The normal connection time is 6-7 seconds. Address #14483

## Description:
#14483 send the device back into deepsleep if WIFI does not work for whatever reason

## Checklist:

### DIFF
--- a/tasmota/xdrv_29_deepsleep.ino
+++ b/tasmota/xdrv_29_deepsleep.ino
@@ -32,7 +32,9 @@
 
 #define D_PRFX_DEEPSLEEP "DeepSleep"
 #define D_CMND_DEEPSLEEP_TIME "Time"
-#define DEEPSLEEP_NETWORK_TIMEOUT 15
+#ifndef DEEPSLEEP_NETWORK_TIMEOUT
+  #define DEEPSLEEP_NETWORK_TIMEOUT 15
+#endif
 
 const uint32_t DEEPSLEEP_MAX = 10 * 366 * 24 * 60 * 60;  // Allow max 10 years sleep
 const uint32_t DEEPSLEEP_MAX_CYCLE = 60 * 60;            // Maximum time for a deepsleep as defined by chip hardware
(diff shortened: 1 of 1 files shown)


tasmota/xdrv_29_deepsleep.ino
```diff
@@ -154,9 +154,13 @@ void DeepSleepStart(void)
 }
 
 void DeepSleepEverySecond(void)
-{
+{  
+  //AddLog(LOG_LEVEL_DEBUG_MORE, PSTR("Wifi Info: up %d, wifidown %d, wifistatus %d, flag %d"),TasmotaGlobal.uptime, TasmotaGlobal.global_state.wifi_down, Wifi.status , deepsleep_flag);
+  if (TasmotaGlobal.uptime > 15 && Wifi.status != WL_CONNECTED && !deepsleep_flag && DeepSleepEnabled()) {
```
barbudor, line 159: @stefanbode Could you please make this value with a #define that can be overridable from user_config_override ?
I see cases where connection with the router is difficult on an outdoor sensor and being able to adjust that time could be usefull.
Thanks

stefanbode: Yes, makes absolute sense. You can now define in seconds or totally disable with "0".

stefanbode: Are you fine with the default? Or should we increase it a bit?

arendst: I think it's fine.

I suggest to put the define between 
```
#ifndef ...

#endif 
```
so users can change the default from the user_config.h file.

stefanbode: yep, done

stefanbode: @arendst : Again you're absolutely right. Did not work the way I did it. Now it should be ok. Thanks for the hint.

barbudor: Many thanks @stefanbode

Issue #14568, opened 2022-01-22
Problem parsing HSBColor JSON payload


RyanYuckx:
### PROBLEM DESCRIPTION
Tasmota seems to have problems parsing JSON payload for HDBColor. I hope this excerpt from the console log spaks for itself.

````
10:27:09.295 CMD: dimmer 100
10:27:09.320 MQT: stat/bulb02/RESULT = {"POWER":"ON","Dimmer":100,"Color":"0,255,56,0,0","HSBColor":"133,100,100","White":0,"CT":153,"Channel":[0,100,22,0,0]}
10:28:05.732 CMD: HSBColor {"HSBColor":"153,84,100"}
10:28:05.752 MQT: stat/bulb02/RESULT = {"POWER":"ON","Dimmer":100,"Color":"255,41,41,0,0","HSBColor":"0,84,100","White":0,"CT":153,"Channel":[100,16,16,0,0]}
10:33:39.136 CMD: HSBColor {"HSBColor":"153,84,70"}
10:33:39.157 MQT: stat/bulb02/RESULT = {"POWER":"ON","Dimmer":70,"Color":"178,29,29,0,0","HSBColor":"0,84,70","White":0,"CT":153,"Channel":[70,11,11,0,0]}
14:17:40.478 CMD: HSBColor {"HSBColor":"60,70,80"}
14:17:40.553 MQT: stat/bulb02/RESULT = {"POWER":"ON","Dimmer":80,"Color":"204,62,62,0,0","HSBColor":"0,70,80","White":0,"CT":153,"Channel":[80,24,24,0,0]}
14:18:09.808 CMD: HSBColor {"HSBColor":"50,90"}
14:18:09.868 MQT: stat/bulb02/RESULT = {"POWER":"ON","Dimmer":80,"Color":"204,21,21,0,0","HSBColor":"0,90,80","White":0,"CT":153,"Channel":[80,8,8,0,0]}
14:28:54.754 CMD: HSBColor {"Dimmer":70,"HSBColor":"40,50,60"}
14:28:54.779 MQT: stat/bulb02/RESULT = {"POWER":"ON","Dimmer":50,"Color":"128,128,128,0,0","HSBColor":"0,0,50","White":0,"CT":153,"Channel":[50,50,50,0,0]}
````

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

[...]

PR #14641, opened 2022-01-28
Make light commands support json

barbudor:
## Description:

**Related issue: https://github.com/arendst/Tasmota/issues/14568

Make all light commands to respond to json payload by using CmndJason()

```
HSBColor  {"HSBColor": "hhh, sss, bbb"}
Scheme {"Scheme": "sss"}
CT {"CT": "ccc"}
Power {"Power<x>": "ppp"}
Dimmer {"Dimmer": "ddd"}
```

The effective changes are much lighter than what the diff highlights.
It's just encapsulating each existing command code between:
```c
void CmndXxxxx(void)
{
  if ('{' == XdrvMailbox.data[0]) {
    CmndJson();
  } else {
  .... original code ....
  }
}
```

## Checklist:

### DIFF
--- a/tasmota/support_command.ino
+++ b/tasmota/support_command.ino
@@ -471,23 +471,27 @@ void CmndDelay(void)
 
 void CmndPower(void)
 {
-  if ((XdrvMailbox.index > 0) && (XdrvMailbox.index <= TasmotaGlobal.devices_present)) {
-    if ((XdrvMailbox.payload < POWER_OFF) || (XdrvMailbox.payload > POWER_BLINK_STOP)) {
-      XdrvMailbox.payload = POWER_SHOW_STATE;
+  if ('{' == XdrvMailbox.data[0]) {
+    CmndJson();
+  } else {
+    if ((XdrvMailbox.index > 0) && (XdrvMailbox.index <= TasmotaGlobal.devices_present)) {
+      if ((XdrvMailbox.payload < POWER_OFF) || (XdrvMailbox.payload > POWER_BLINK_STOP)) {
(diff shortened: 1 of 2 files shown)


arendst:
Pls pay attention to the way CmndJson operates: all commands are parsed as backlog commands.

I suggest you do some testing with this PR and chk what happens when users start using this json command line prepended with a backlog command. I expect some great fun but it may also pass unnoticed.

Pls chk.

There was a reason I introduced a new command instead of just adding JSON parsing to every command (which in addition adds to the codebase already very tight on ESP8266).

barbudor:
@arendst 
Alternative is to change the HA discovery message to refer the JSON command instead of still refering the original one that do not support json input. That would justify the introduction of the JSON command.
As of today JSON command cannot be used.

Do you prefer I change the discovery message to use cmnd/.../json ?

arendst:
In the past there was no issue wirh HA and the command structure; The legacy HA code provides HA with all information it needed.

Then a new approach was introduced with two retained MQTT messages at restart. The information is decoded by a HA add-on providing HA with basic information.

The Domoticz people expect the same not knowing there is a Teasmota add-on in HA providing the information. Domoticz has another tool to talk to tasmota which was mentionend here before.

Then the JSON command line option: as it stands it does nothing more than wrapping the JSON command and send it to backlog. It will behave differently then a "normal" command if there are already commands in the backlog; the backlog commands will be processed first before the intended direct JSON command is performed. 

The functionality is different and could work well as long as the user is aware.

barbudor:
let's drop this
I don't care neither of HA or Domoticz anyhow

PR #14663, opened 2022-01-29
i2cdevices in simpler english

fabian727:
Hi guys,
I had problems understanding, what was written; so here my proposal.

I recommend also to change line 43 and 44, as address 0x27 is also supported:
https://tasmota.github.io/docs/MCP230xx/#usage-of-the-driver

It may not be a good idea, but i have a working example on my desk.

## Description:

## Checklist:

### DIFF
--- a/I2CDEVICES.md
+++ b/I2CDEVICES.md
@@ -1,6 +1,5 @@
 # I2C devices
-
-Tasmota supports several I2C devices but to use them they mostly need to be enabled at compile time to solve possible address conflicts.
+Tasmota supports several I2C devices. To use them I2C and the device need to be enabled at compile time. I2C and some devices are supported also in the official releases. Devices can be de/-actived on runtime to solve possible address conflicts. (e.g. address 0x27 is used by multiple devices)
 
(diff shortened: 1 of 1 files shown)


barbudor:
https://github.com/arendst/Tasmota/blob/f6df3b7c9f1669dbc3ccd898205816d48223653f/tasmota/xsns_29_mcp230xx.ino#L22-L29

Jason2866:
It is a good idea to use addresses from 0x20 to 0x26 to avoid conflicts
https://tasmota.github.io/docs/MCP230xx/#usage-of-the-driver
The change you suggest is not recommended

fabian727:
as stated in my first post: i do have a working example here. So it is supported. I guess it is not a good idea, but it is working and therefore supported.
I would suggest to either fail the compilation with this address or to state that it is possible, as it is, and recommend to a different address range.
But I was unsure if the board would work with this address, as I read different information in the wiki.

barbudor:
As I already told you on your PR in docs, working and supported is not the same.
We just don't want to hear about people complaining about addresses conflicts while the MCP has already a wide choice of address.
It's working for you, fine but the doc shouldn't contradict what is written in the code.

PR #14683, opened 2022-01-30
Refactor Deepsleep to new uint64 capability

stefanbode:
removed 71min limit and simplify driver.

## Description:

## Checklist:


Jason2866:
So there is no 71 min limit for the esp8266 from hardware side?
It is just software because of using 32 bits?

stefanbode:
Yes it was a 32bit int for a long time but now it is int64 in the core. Therefore no need anymore for fancy wake ups.

PR #14807, opened 2022-02-11
add yeelight dimmer to legacy Mi32

Staars:
## Description:

Support for YLKG08YL (and YLKG07YL which is the otherwise identical version with connection to mains,), which provides a button press(single and double), a hold operation and an incremental rotary encode (pressed and non-pressed).
Very limited support for HomeKit as this type of device is not really supported (only button press single/double/hold)

YLKG08YL and YLKG07YL are named YLKG08 as both have the same product ID.
Renaming of the (badly chosen name) YEERC to YLYK01.

Button presses are now more in line with Tasmotas naming conventions in the form of `Buttonx` followed by a number 1,2 or 3 for single, double and hold (like in RULES).
Hold operation is triggered after about 2 seconds and reports an up counting number x `"Hold":x`.
Rotation of the dimmer knob reports steps a signed integer like `"Dimmer":-1`(one step counter clockwise not pressed)  or `"Dimmer_P":5` (five steps clockwise pressed in a quick motion).

## Checklist:

### DIFF
--- a/tasmota/xsns_62_esp32_mi.h
+++ b/tasmota/xsns_62_esp32_mi.h
@@ -236,6 +236,7 @@ struct mi_sensor_t{
       uint32_t NMT:1;
       uint32_t motion:1;
       uint32_t Btn:1;
+      uint32_t knob:1;
       uint32_t door:1;
       uint32_t leak:1;
     };
@@ -254,6 +255,8 @@ struct mi_sensor_t{
       uint32_t motion:1;
       uint32_t noMotion:1;
       uint32_t Btn:1;
+      uint32_t knob:1;
+      uint32_t longpress:1; //needs no extra feature bit, because knob is sufficient
       uint32_t door:1;
       uint32_t leak:1;
     };
@@ -285,11 +288,15 @@ struct mi_sensor_t{
     }; // MJ_HT_V1, LYWSD0x
     struct {
       uint16_t events; //"alarms" since boot
-      uint32_t NMT;    // no motion time in seconds for the MJYD2S
+      uint32_t NMT;    // no motion time in seconds for the MJYD2S and NLIGHT
     };
     struct {
-      uint16_t Btn;
-      uint8_t leak;
+      uint8_t Btn;     // number starting with 0
+      uint8_t BtnType; // 0 -single, 1 - double, 2 - hold
+      uint8_t leak; // the leak sensor is the only non-RC device so far with a button fuctionality, so we handle it here
+      int8_t dimmer;
+      uint8_t pressed; // dimmer knob pressed while rotating
+      uint8_t longpress; // dimmer knob pressed without rotating
     };
     uint8_t door;
   };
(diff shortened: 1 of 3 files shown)


Staars:
The dimmers send encrypted data and needs a "bind_key" (12 bytes long), which can be generated and retrieved using this script:
https://github.com/custom-components/ble_monitor/blob/master/custom_components/ble_monitor/ble_parser/get_beacon_key.py

I had to port this python script to another BLE library to get it to work an MacOS using `bleak`:
https://gist.github.com/Staars/9cd9bbaf9c5bec68d46187bd67399093
(this may or may not work on Windows and Linux)

sfromis:
I'm trying to make YLYK01YL and YLKG08YL work with Tasmota, having gotten those with Tasmotized Yeelight lamps.

While the small YLYK01YL seems to work, I was a bit surprised to see it detected as `M32: MI32: new YEERC at slot: 1` when this thread suggests that the type should instead be YLYK01. My precompiled build should be new enough, but to be extra sure I just upgraded to most recent tasmota32-bluetooth.bin, no difference seen.
`"StatusFWR":{"Version":"11.0.0.1(bluetooth)","BuildDateTime":"2022-02-13T10:49:31"`

Scanning Tasmota source, I did see `YEERC` in [xsns_62_esp32_mi_ble.ino](https://github.com/arendst/Tasmota/blob/development/tasmota/xsns_62_esp32_mi_ble.ino#L459)
`const char kMI32DeviceType9[] PROGMEM = "YEERC";` 
The similar table in `xsns_62_esp32_mi.h` does have the change.

YLKG08YL does not work great, but I assume that this is due to me not having generated bind_key. Will I need to get a BLE dongle to work with a Linux box (or similar) to have a chance?

Jason2866:
@sfromis  Just to ne sure. Latest version needs to be compiled as described in #14818

Staars:
@sfromis Jason8266 is right here, you have built the the main driver BLE driver without dimmer support.
The bind_key is mandatory and the two linked utilities above are proven to work on Linux (the first one) and on MacOS (the second one which is a port of the first one by me). 
The Bleak library is an attempt to provide a cross platform solution for Python and thus it might work on Windows and Linux too, but I could not test this. The detour in my python script is necessary because of Apple's super nice decision to hide MAC addresses in the underlying framework (Core Bluetooth).

Staars:
Not these scripts, but I did some experiments with our Wiki and embedded some JS-scripts directly into the documentation ink my fork. This is primarily for my driver in order to get a ready-to-use mi32cfg file more easily, but the generated bind_key can be used anywhere. I did all the key generations of my sensors with it, so it seems to work to some extent.
You need Chrome (or Edge) and must(!) enable experimental platform features in the browser.
Feel free to test this here:
https://staars.github.io/site/Bluetooth_MI32/#bind-key-generator-web-app
(It might be necessary to reload the page after initial loading, if I remember ...)

If I find the time I will do a port of the bind_key generator for the dimmer (which is Xiaomis legacy protocol) for this page too.

PR #14808, opened 2022-02-11
Add support for digital Poti DS3502

f-reiling:
## Description:

** This PR adds support for digital poti DS3502 via I2C
* Possible commands:
* driver65 poti <value> - set poti wiper to <value> = 0-127
* driver65 status - get current wiper setting
* driver65 reset - reset digital poti

## Checklist:


tasmota/xdrv_65_ds3502.ino
```diff
@@ -0,0 +1,153 @@
...
+  }
+  return serviced;
+}
+
+void DS3502_OutputTelemetry(bool telemetry)
+{
+  ResponseTime_P(PSTR(",\"DS3502\":{\"POTI\":%i}}"),DS3502_potiValue);
+  if (telemetry) {
+    MqttPublishTeleSensor();
+  }
```
barbudor, line 123: This shouldn't be called from within a driver
If you want to consider your poti as a sensor (would be strange) you should implement as a xsns and handle the FUNC_JSON_APPEND function

However I don't see why a poti (which is an output device) should provide telemetry

tasmota/xdrv_65_ds3502.ino
```diff
@@ -0,0 +1,153 @@
...
+
+  if (FUNC_INIT == function) {
+    DS3502_Detect();
+  }
+  else if (DS3502_detected) {
+    switch (function) {
+      case FUNC_EVERY_SECOND:
+        if (TasmotaGlobal.tele_period == 0) {
+          DS3502_OutputTelemetry(true);
+        }
```
barbudor, line 140: linked ot the above, a driver shouldn't output it's own telemetry but contribute to the common JSON

tasmota/xdrv_65_ds3502.ino
f-reiling: Good point, telemetry is not required. Current state can be checked using status command.

PR #14854, opened 2022-02-15
Added support for BL6523 chipset based Energy Monitors

jvstch:
## Description:

**Added support for Bellini BL6523 chipset based Watt meter/Energy monitors.** 

## Checklist:
  - [*] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [*] The code change is tested and works with Tasmota core ESP32 V.2.0.2.1


jvstch:
Some details available here: https://youtu.be/s3rMb4opuDg

tasmota/tasmota_template.h
```diff
@@ -172,6 +172,7 @@ enum UserSelectablePins {
   GPIO_MCP2515_CS,                     // MCP2515 Chip Select
   GPIO_HRG15_TX, GPIO_HRG15_RX,        // Hydreon RG-15 rain sensor serial interface
   GPIO_VINDRIKTNING_RX,                // IKEA VINDRIKTNING Serial interface
+  GPIO_BL6523_TX, GPIO_BL6523_RX,                      // BL6523 based Watt meter Serial interface
```
barbudor, line 175: This is wrong location
All new GPIO _MUST_ be added at the end of the list, right before GIO_SENSOR_END

tasmota/tasmota_template.h
```diff
@@ -393,6 +394,7 @@ const char kSensorNames[] PROGMEM =
   D_SENSOR_MCP2515_CS "|"
   D_SENSOR_HRG15_TX "|" D_SENSOR_HRG15_RX "|"
   D_SENSOR_VINDRIKTNING_RX "|"
+  D_SENSOR_BL6523_TX "|" D_SENSOR_BL6523_RX "|"
```
barbudor, line 397: Same as above, you need to add those at the end of the array

tasmota/xsns_96_bl6523.ino
```diff
@@ -0,0 +1,353 @@
...
+  uint32_t freq = 0;
+  uint32_t watts = 0;
+  uint32_t powf = 0;
+  uint32_t watthr = 0;
+
+  uint8_t type = 1;
+  uint8_t valid = 0;
+  uint8_t got_data_stone = 0;
+  bool discovery_triggered = false;
+} Bl6523;
```
barbudor, line 86: INstead of statically allocate memory, please make that a pointer to the structure and allocate memory only if the driver is in use
That will save RAM for when the driver is compiled in a large build but not used.

tasmota/xsns_96_bl6523.ino
```diff
@@ -0,0 +1,353 @@
...
+    if (json)
+    {
+      ResponseAppend_P(PSTR(",\"BL6523\":{"));
+      ResponseAppend_P(PSTR("\"Amps\":%s,"), amps_str);
+      ResponseAppend_P(PSTR("\"Volts\":%s,"), volts_str);
+      ResponseAppend_P(PSTR("\"Freq\":%s,"), freq_str);
+      ResponseAppend_P(PSTR("\"Watts\":%s,"), watts_str);
+      ResponseAppend_P(PSTR("\"Powf\":%s,"), powf_str);
+      ResponseAppend_P(PSTR("\"WattHr\":%s"), watthr_str);
+      ResponseJsonEnd();
```
barbudor, line 294: Don't you your own name strings for reporting the measurments.
There are standard Tasmota name available in `i18n.h` such as D_JSON_CURRENT, D_JSON_VOLTAGE, D_JSON_FREQUENCY, D_JSON_POWERUSAGE, D_JSON_POWERFACTOR, D_JSON_ENERGY

Moreover, shouldn't this driver be a XNRG driver instead of XSNS and use the standard Tasmota ENERGY reporting model ?

tasmota/xsns_96_bl6523.ino
```diff
@@ -0,0 +1,353 @@
...
+    else
+    {
+      WSContentSend_PD(HTTP_BL6523_SNM);
+      WSContentSend_PD(HTTP_BL6523_SNS, PSTR("Amps:"), amps_str);
+      WSContentSend_PD(HTTP_BL6523_SNS, PSTR("Volts:"), volts_str);
+      WSContentSend_PD(HTTP_BL6523_SNS, PSTR("Freq:"), freq_str);
+      WSContentSend_PD(HTTP_BL6523_SNS, PSTR("Watts:"), watts_str);
+      WSContentSend_PD(HTTP_BL6523_SNS, PSTR("PowF:"), powf_str);
+      WSContentSend_PD(HTTP_BL6523_SNS, PSTR("WattHr:"), watthr_str);
+#endif // USE_WEBSERVER
```
barbudor, line 317: Same as above, don't use your own strings

tasmota/xsns_96_bl6523.ino
```diff
@@ -0,0 +1,353 @@
...
+      Bl6523Show(0);
+      break;
+#endif // USE_WEBSERVER
+    case FUNC_INIT:
+      Bl6523Init();
+      break;
+    }
+  }
+  return result;
+}
```
barbudor, line 351: This structure is not correct
It should be:
```C++
if ( FUNC_INIT == function )
{
   BL6523Init();
}
else if ( Bl6523.type ) {  // Bl6523.type must be defaulting to 0 and set to 1 in INIT is the driver is enabled only
    switch( funcstion ) {
    ....
}
```

barbudor:
Hi @jvstch 
Thanks but a few things need to be reviewed before it can be considered
The main question is : Shouldn't this be an ENERGY driver instead of a SENSOR driver ?

arendst:
Indeed this needs to be nrg_ driver using many existing features of the energy driver.

Pls go back to the drawing board and try to achieve the same with much less code size increase. Take a look the available nrg_ drivers for inspiration.

jvstch:
@arendst  I considered that initially but for the features that I target looking at nrg drivers didn't inspire a lot. Also didn't see any clear explanation on where the lines are drawn on deciding an nrg vs sns. Can you point me to some good doc or help clarify what else I should be doing bare minimum other than renaming to make it nrg to get the same functionality as right now?

jvstch:
@arendst  and @barbudor  Could you guys please take a fresh look again. Just converted to xnrg as mentioned.

tasmota/xsns_96_bl6523.ino
jvstch: I changed as per comment. But I can see that many places in existing code uses the previous structure as well.

jvstch (commented):
Converted to Xnrg and addressed some earlier comments. Request to take fresh look.

jvstch:
@arendst  Also could you please help to underline my understanding:

Any sensor dealing with Voltage/Current/PowF/Amp/Watt/Watthr related(subset) irrespective of AC/DC should be coded as xnrg even if particular sensor also includes sensors like temperature, Light and any other.

Is that the only criteria or are there any minimum set of compulsory requirements for xnrg devices?

Jason2866:
Yes, devices dealing with measuring "Power" at AC mains level are xnrg
In many rare cases Power monitoring devices have sensors connected (because of the electrical design).

jvstch:
> Yes, devices dealing with measuring "Power" at AC mains level are xnrg In many rare cases Power monitoring devices have sensors connected (because of the electrical design).

Thanks @Jason2866. So a DC Volt/Amp/Watt/Watthr/meter like Sonoff USB and similar at ~approx 20-30v is an xsns and only AC mains(90v and above) meters are xnrg. Is that correct?

arendst:
As suggested earlier, take a look at how the xnrg_ drivers are written.

A good example is the xnrg_19_cse7761.ino driver. You won't find any representation code simply because it is handled by the xdrv_energy driver which calls your driver in turn.

Your energy driver ONLY supports the BL6523 energy monitoring function. All other sensors are either already supported ot need a dedicated driver to be used by other devices too.

Your latest changes do NOT make this a valid xnrg driver.

jvstch:
@arendst Thanks,that is very helpful and interesting. It makes a lot more sense now . Will rewrite the code to use the TasmotaGlobal.energy_driver and see how it goes.

jvstch:
@arendst Could you please take a look whenever you can. Thanks

jvstch:
image

image

tasmota/xnrg_22_bl6523.ino
```diff
@@ -0,0 +1,291 @@
...
+/* No idea how to derive human readable units from the byte stream.
+For now  dividing the 24-bit values with below constants seems to yield something sane
+that matches whatever displayed in the screen of my 240v model. Probably it would be possible
+to extract these values from the write register commands (0xCA).*/ 
+#define BL6523_DIV_AMPS 297899.4f
+#define BL6523_DIV_VOLTS 13304.0f
+#define BL6523_DIV_FREQ  3907.0f
+#define BL6523_DIV_WATTS 707.0f
+#define BL6523_DIV_WATTHR 674.0f
+
```
arendst, line 72: These values are typical calibration values. See some other drivers how they handle these calibration defaults.

tasmota/xnrg_22_bl6523.ino
```diff
@@ -0,0 +1,291 @@
...
+TasmotaSerial *Bl6523RxSerial;
+TasmotaSerial *Bl6523TxSerial;
+
```
arendst, line 75: Do we really need two serial ports? From the datasheet I thought the chip uses SPI to communicate.

tasmota/xnrg_22_bl6523.ino
jvstch: Any examples? I derived this by doing multiple reverse maths comparing with whatever displayed on screen.

arendst: As said, see other drivers. Look for ``Settings->energy_voltage_calibration`` and commands like ``VoltageCal``

tasmota/xnrg_22_bl6523.ino
```diff
@@ -0,0 +1,291 @@
...
+bool Bl6523ReadData(void)
+{
+   uint32_t powf_word = 0, powf_buf = 0;
+   float powf = 0.0f;
+
+  if (!Bl6523RxSerial->available())
+  {
+    AddLog(LOG_LEVEL_DEBUG, PSTR("BL6523 No Rx Data available " ));
+    return false;
```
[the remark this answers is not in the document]
jvstch: Done

tasmota/xnrg_22_bl6523.ino
jvstch: I think for now this can stay as it is unless the same exact values present somewhere else which can be reused.

jvstch:
@arendst Can you take a look again please.

Regarding the settings I am not fully sure and can be handled as a future enhancement. The reason is that in this particular case we are not doing any actual calibration. We are only trying to derive the division factor to get the actual units. 

It would be nice to have if we have these settings pre-populated by default and an additional option for the user to change it.

I don't have already supported device to visualize how Settings->energy_calibration looks and works. 

I have 5 different constants to take care:
#define BL6523_DIV_AMPS 297899.4f
#define BL6523_DIV_VOLTS 13304.0f
#define BL6523_DIV_FREQ  3907.0f
#define BL6523_DIV_WATTS 707.0f
#define BL6523_DIV_WATTHR 674.0f

And we have only 3 mappable fields in Settings as below:

  Settings->energy_power_calibration 
  Settings->energy_voltage_calibration
  Settings->energy_current_calibration

I want 5 different settings which get's pre-assigned with the above derived constants and gets used by default. If someone changes the default value from web ui and saves it those saved values will be used for the calculation.

Can you provide a little more insight into this.

arendst:
No additions to Settings are supported for this driver but you can use the already available `` Settings->energy_frequency_calibration`` and invent a formula for using the same ``Settings->energy_power_calibration`` for both power and energy.

jvstch:
Thanks @arendst. That seems pretty doable. Do you think that can go as an enhancement or along with this? 

Also any reason why Voltage and Wattage are whole numbers? Voltage is fine but Wattage can be made to atleast accomodate two decimal points in my humble opinion.

arendst:
Dive into the docs and learn commands like ``wattres``, ``voltres`` and ``ampres``

jvstch:
Sorry. I should have typed default wattage resolution.

arendst:
> Sorry. I should have typed default wattage resolution.

???? That's different for every user. Once set using the command ``WattRes 2`` it stays set. Or if you insist a user is not allowed to change it use code: ``Settings->flag2.wattage_resolution = 2`` but I hate these forced settings.

jvstch:
All I'm saying is changing the default wattres from 0 to 2 considering the fact that a large group will be using these with low power devices.

jvstch:
image
image

tasmota/xnrg_22_bl6523.ino
```diff
@@ -0,0 +1,362 @@
...
+  Bl6523RxSerial->flush(); // Make room for another burst
+
+  AddLogBuffer(LOG_LEVEL_DEBUG_MORE, rx_buffer, BL6523_RX_DATASET_SIZE);
+  
+  while (Bl6523TxSerial->available() < BL6523_TX_DATASET_SIZE)
+  {
+    // sleep till TX buffer is full
+     unsigned long timeout = millis() + 10;
+     while (millis() < timeout) {}
+  }
```
arendst, line 123: Do a ``delay(10);`` here instead of both ``timeout`` lines. This allows for background wifi updates.

Make an abort from the while loop too in case serial data is never received.

jvstch: Indeed a good catch. Thanks

tasmota/xnrg_22_bl6523.ino
```diff
@@ -0,0 +1,356 @@
...
+
+  AddLogBuffer(LOG_LEVEL_DEBUG_MORE, rx_buffer, BL6523_RX_DATASET_SIZE);
+  
+  i=0;
+  while (Bl6523TxSerial->available() < BL6523_TX_DATASET_SIZE)
+  {
+    // sleep till TX buffer is full
+    delay(10);
+    if ( i++ > RX_WAIT ){
+      abort();
```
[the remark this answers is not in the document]
jvstch: Done :)

arendst:
Whow what a work for a sensor we will probably never see.

jvstch:
@arendst Probably true. Let's see if atleast one person uses this :). Thanks a lot for your support. Ciao until time for another two nighter.:)

PR #14981, opened 2022-02-26
Allow dynamic reconfiguration of TasmotaSerial

barbudor:
## Description:

SerialBridge, TCPBridge, Teleinfo have commands that tries to change the serial configuration dynnamically by calling `TasmotaSerial::begin()` while the port was already allocated.
Unfortunately up to now, on ESP32, this was trying to reallocate an new UART. Which quickly comes to an end if a few are in use...

This PR bypass the UART allocation if TasmotaSerial already has an UART assigned.

@arendst Should I change the version number to 3.4.1 ?

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.2.3

### DIFF
--- a/lib/default/TasmotaSerial-3.4.0/src/TasmotaSerial.cpp
+++ b/lib/default/TasmotaSerial-3.4.0/src/TasmotaSerial.cpp
@@ -87,6 +87,7 @@ TasmotaSerial::TasmotaSerial(int receive_pin, int transmit_pin, int hardware_fal
   if ((receive_pin >= 0) && !GPIO_IS_VALID_GPIO(receive_pin)) { return; }
   if ((transmit_pin >= 0) && !GPIO_IS_VALID_OUTPUT_GPIO(transmit_pin)) { return; }
   m_hardserial = true;
+  TSerial = nullptr;
 #endif  // ESP32
   m_valid = true;
 }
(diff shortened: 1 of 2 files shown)


arendst:
I haven't looked at your solution but do they perform a serial end() before a new begin? That should work without changing tasmotaserial.

barbudor:
@arendst
The problem is that the current TasmotaSerial::begin() code will always reassign another UART. If you already have 2 drivers using UARTs, this will steal the console and have unwanted side effects. If you have 3, then it will fails.

I don't think there a definitive need for dynamic configuration per se, but the 1st configuration of SerialBridge/TCPBridge or Teleinfo is by default a reconfiguration as the driver has already been initialized with some values. We could alternatively force a restart but I think this is nicer to allow setting baudrate/config without rebooting.

There is no TasmotaSerial::end(), only a destructor. But as said in the comment, it doesn't work because it just re-increment tasmota_serial_index. If we want the destructor to work, we should manage a bit field for assignement, not just a counter.
I don't think we need a proper destructor and UART alloc/free mecanism. There is no real need to enabling/disabling drivers dynamically. Once a driver is there, it should be able to keep the same TasmotaSerial and underlying UART.

Note that when calling TasmotaSeriall:begin() on already assigned UART, HardwareSerial:begin() automatically does an `end()`. So it is not explicitly needed:
https://github.com/espressif/arduino-esp32/blob/b254765ef81a2d80f498a8c22c7750152e3e68df/cores/esp32/HardwareSerial.cpp#L152-L156

Note that my other PR on Teleinfo is dependant on this. If you decide to reject this PR, I will need to change the Teleinfo one in order to force restart in case of parameter change.

Take your time, no emergency.

Thanks

arendst:
I see the problem.

1) At first it was never intended to open another softwareserial occasion or re-assign a uart.
2) The SR04 sensor is the first to allocate a serial connection and if not needed it releases it. I never tested if it really released the software serial or the uart but hence the increase in ``tasmota_serial_index``. From your experience it seems that doesn't work so that issue has to be addresses.
3) Then there are drivers that want to change baudrates and configs. That was also never anticipated but it could have worked as bullet 2) had worked but then it would fail on the esp32 using uarts if another uart would have been assigned after the one that needs re-config.

So I think we have two issues to solve while supporting "standard" re-config procedures.

1) Fix de-allocation of an allocated uart (or software serial)
2) Add uart mapping bitmap (as you suggested) to keep score of used and released uarts to be used for assignment in favour of uart0.

I think that is the best approach as it is in line with standard Arduino practices instead of using tests for non-null pointers.

This will hold the teleinfo driver for now. I will have a look (and test with SR04) how to implement release after I added the uart mapping.

arendst:
While adding the mapping I see your implementation is great for detecting dynamic changes.

I'll use most of this code and release TasmotaSerial 3.5.0 in a minute.

arendst:
Implemented as suggested and added usaed uart bitmap too.

Thx for the idea.

barbudor:
😀🙏

PR #15061, opened 2022-03-08
Add command to change timeout in RfReceive

DarkPh0eNixKrg:
## Description:

Add command to change timeout in RfReceive
My door bell transmiter (Quhwa/1byone) sending ~70 payload (protocol 22).
I make rule to send mqtt status, but mqtt send 2-3 times payload.
To send 1 mqtt payload i was change "RF_TIME_AVOID_DUPLICATE" and add command to can change it.
But i dont know how to save timeout. After reboot timeout change to default.

I newbee in Tasmota (((
Sorry for my English

## Checklist:

### DIFF
--- a/tasmota/xdrv_17_rcswitch.ino
+++ b/tasmota/xdrv_17_rcswitch.ino
@@ -1,18 +1,18 @@
 /*
   xdrv_17_rcswitch.ino - RF transceiver using RcSwitch library for Tasmota
-  
+
   Copyright (C) 2021  Theo Arends
-  
+
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation, either version 3 of the License, or
   (at your option) any later version.
-  
+
   This program is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   GNU General Public License for more details.
-  
+
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
--- a/tasmota/xdrv_17_rcswitch.ino
+++ b/tasmota/xdrv_17_rcswitch.ino
@@ -1,18 +1,14 @@
 /*
   xdrv_17_rcswitch.ino - RF transceiver using RcSwitch library for Tasmota
-
   Copyright (C) 2021  Theo Arends
-
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation, either version 3 of the License, or
   (at your option) any later version.
-
   This program is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   GNU General Public License for more details.
-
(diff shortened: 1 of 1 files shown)


tasmota/xdrv_17_rcswitch.ino
```diff
@@ -1,18 +1,14 @@
 /*
   xdrv_17_rcswitch.ino - RF transceiver using RcSwitch library for Tasmota
-
```
barbudor, line 3: Do not delete lines from the header
This is the standard Tasmota header.
Please set them back

tasmota/xdrv_17_rcswitch.ino
```diff
@@ -27,24 +23,26 @@
...
-  &CmndRfSend, &CmndRfProtocol };
+  &CmndRfSend, &CmndRfProtocol, &CmndRfTimeOut};
 
 #include <RCSwitch.h>
 
 RCSwitch mySwitch = RCSwitch();
 
-#define RF_TIME_AVOID_DUPLICATE 1000  // Milliseconds
+int RF_TIME_AVOID_DUPLICATE = 3000;  // Use variable to change "RfTimeOut" to get new data from RfReceive
+//#define RF_TIME_AVOID_DUPLICATE 1000  // Milliseconds
```
barbudor, line 45: You can remove that line

tasmota/xdrv_17_rcswitch.ino
```diff
@@ -27,24 +23,26 @@
...
 void (* const RfCommands[])(void) PROGMEM = {
-  &CmndRfSend, &CmndRfProtocol };
+  &CmndRfSend, &CmndRfProtocol, &CmndRfTimeOut};
 
 #include <RCSwitch.h>
 
 RCSwitch mySwitch = RCSwitch();
 
-#define RF_TIME_AVOID_DUPLICATE 1000  // Milliseconds
+int RF_TIME_AVOID_DUPLICATE = 3000;  // Use variable to change "RfTimeOut" to get new data from RfReceive
```
barbudor, line 44: Do not change the default value.
3000 may be the one that suits your specific use-case but in many other case this will break user's expectation as it will prevent users to send the same command within a short time. For exemple for a dimmer or a sound volume you need to press the key multiple time. Waiting 3 seconds between each press is not acceptable.

Also do not keep the name as fully capitalized. `RfTimeToAvoidDupplicate` would be better for a variable. Captalized should be only for preprocessors macros/constants.

barbudor (commented):
You can set the value at each reboot using a boot rule such as `ON System#Boot DO RFTImeout 3000 ENDON`

tasmota/xdrv_17_rcswitch.ino
DarkPh0eNixKrg: Corrected

tasmota/xdrv_17_rcswitch.ino
DarkPh0eNixKrg: Corrected

tasmota/xdrv_17_rcswitch.ino
DarkPh0eNixKrg: Corrected

Jason2866:
Please remove formatting style changes.

DarkPh0eNixKrg:
> c

I`m newbee, pls show or write what is formatting style need remove.

arendst:
I've seen your effort and will make a more permament solution. Hold on...

arendst:
Implemented my way. Thx.

PR #15099, opened 2022-03-10
MI32: active BLE connections can remain

Staars:
## Description:

This allows subscribing to notifications for a longer period of time and does not disconnect after every operation. No breaking changes but some additions to the API:

New operations:  
(<10 does not disconnect)
1 - read
2 - write
3- subscribe
5- disconnect
(the old 11 - read, 12 - write, 13 - subscribe_once are still there and will disconnect after completion)

Additions to the Berry connection callback function in the form of
```
    def cb(error,op,uuid)
      # error - code number
      # op - number 1 -3
      # uuid - 16-bit characteristic UUID (from read or notification)
    end
```

I tested this with an LYWSD03 (reading the history data from flash), a Govee desk lamp and a Xbox-S controller. It seems to work, but of course it will not be perfect. For IOT I still prefer passive sensor readings. But it is fun to play with Berry and BLE!

### DIFF
--- a/tasmota/xsns_62_esp32_mi.h
+++ b/tasmota/xsns_62_esp32_mi.h
@@ -150,7 +150,8 @@ struct ATCPacket_t{ //and PVVX
 struct MI32connectionContextBerry_t{
   NimBLEUUID serviceUUID;
   NimBLEUUID charUUID;
-  uint8_t * MAC;
+  uint16_t returnCharUUID;
+  uint8_t MAC[6];
   uint8_t * buffer;
   uint8_t operation;
   uint8_t addrType;
@@ -177,7 +178,6 @@ struct {
 
       uint32_t shallTriggerTele:1;
       uint32_t triggeredTele:1;
-      uint32_t shallClearResults:1;   // BLE scan results
       uint32_t shallShowStatusInfo:1; // react to amount of found sensors via RULES
       uint32_t didGetConfig:1;
       uint32_t didStartHAP:1;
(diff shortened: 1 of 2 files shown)


Staars:
An example to illustrate the communication with the xbox controller (must be re-paired on every connect):
```
class XBOX : Driver
    var ble, cbp, buf

    def init()
        self.cbp = tasmota.gen_cb(/e,o,u-> self.cb(e,o,u))
        self.buf = bytes(-256)
        self.ble = BLE()
        self.ble.conn_cb(self.cbp,self.buf)
        var _mac ="112233445566" # xbox controller MAC
        self.ble.set_MAC(bytes(_mac),0)
        self.connect()
    end

    def connect() # separated to call it from the berry console if needed
        self.ble.set_svc("1812")
        self.ble.set_chr("2a4a") # the first characteristic we have to read
        self.ble.run(1) # read
    end

    def handle_read_CB(uuid) # uuid is the notifying characteristic
    # we just have to read these characteristics before we can finally subscribe
        if uuid == 0x2a4a # ignore data
            self.ble.set_chr("2a4b")
            self.ble.run(1) # read next characteristic 
        end
        if uuid == 0x2a4b # ignore data
            self.ble.set_chr("2a4d")
            self.ble.run(1) # read next characteristic 
        end
        if uuid == 0x2a4d # ignore data
            self.ble.set_chr("2a4d")
            self.ble.run(3) # start notify
        end
    end

    def handle_HID_notifiaction() # a very incomplete parser
        if self.buf[14] == 1
            print("ButtonA") # a MQTT message could actually trigger something
        end
        if self.buf[14] == 2
            print("ButtonB")
        end
        if self.buf[14] == 8
            print("ButtonX")
        end
        if self.buf[14] == 16
            print("ButtonY")
        end
    end

    def cb(error,op,uuid)
        if error == 0
            if op == 1
                print(op,uuid)
                self.handle_read_CB(uuid)
            end
            if op == 3
               self.handle_HID_notification()
            end
            return
        else
            print(error)
        end
    end

end

xbox = XBOX()
tasmota.add_driver(xbox)

```

tony-fav:
Great work @Staars !!

Staars:
Hope it is not too unstable.
I will add the option for active scanning soon, maybe even tomorrow (after minimal testings).
After that I will probably do some minor changes to the advertisement buffer, which might require very little changes to the way of parsing it. I am still scratching my head, if it would help to pass some data as arguments to the callback function in order to have them easily accessible as variables. But on the other hand you have shown in Blerry, that it is possible to do a lot of work in Berry. Nice job, BTW!

tony-fav:
Not sure about stability yet! But I got a working prototype for the WP6003 air box: https://github.com/tony-fav/tasmota-blerry/blob/dev-support-NewBLEDriver/one-shots/blerry_WP6003.be all thanks to this new driver.

Staars:
Cool!
I am sure you already know this, but what I like about the current Berry implementation in Tasmota is, that for debugging purposes you can interact with the driver in the web console, like:
`wp6003.ble.run(5)` to disconnect by hand and so on …
BTW, your approach in Blerry to pass the MAC address to the init function of a driver is nicer than my example and I will do it too in the future.

PR #15141, opened 2022-03-14
Revert "MI32: active BLE connections can remain"

Jason2866:
Reverts arendst/Tasmota#15099

since it crashes.

@Staars please have a look


Staars:
These additions were not as harmless as I thought and currently I do not know, why it was completely trouble free in the two weeks of testing:
```    
NimBLEDevice::setOwnAddrType(BLE_OWN_ADDR_RANDOM); //seems to be important for i.e. xbox controller, hopefully not breaking other things
NimBLEDevice::setSecurityAuth(true, true, true);
```

I will prepare a new PR (after some more testing).

PR #14619, opened 2022-01-27
Added %timer1% to %timer16% variables to rules

alexasf:
## Description:

**Related issue (if applicable):** fixes #<Tasmota issue number goes here>

## Checklist:
  - [ ] The pull request is done against the latest development branch
  - [ ] Only relevant files were touched
  - [ ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.2.1
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #15224, opened 2022-03-25
Fix %timerx% variables for negative values

barbudor:
## Description:

**Related issue (if applicable):** fixes https://github.com/arendst/Tasmota/discussions/15205 from @rapi3

When timer uses mode 1 (relative to sunrise) or mode 2 (relative to sunset), the "negative" value stored in settings is "encoded" as > 12:00. This was not taken into account in previous PR https://github.com/arendst/Tasmota/pull/14619 by @alexasf

I've chosen to return the value as negative offset.

## Possible alternative 
Instead of returning the offset, for mode 1 and 2, it could return the absolute time %sunset%+offset or %sunrise%+offset 
This would return a value that is valid at this instant and not the value of the setting.

Let me know if you prefer this way, I will change the PR.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.2.3

### DIFF
--- a/tasmota/xdrv_10_rules.ino
+++ b/tasmota/xdrv_10_rules.ino
@@ -477,7 +477,11 @@ bool RulesRuleMatch(uint8_t rule_set, String &event, String &rule, bool stop_all
       if ((index > 0) && (index <= MAX_TIMERS)) {
         snprintf_P(stemp, sizeof(stemp), PSTR("%%TIMER%d%%"), index);
         if (rule_param.startsWith(stemp)) {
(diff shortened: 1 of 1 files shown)


rapi3:
I think it will be perfect to return the absolute time %sunset%+offset or %sunrise%+offset value Instead of returning the offset, for mode 1 and 2.
More useful when used in rule because it will save a lot of space in rules so I don't have to use another variables to calculate the sunset/sunrise value+offset...

Thank you very much !

barbudor:
Will redo the PR with absolute time

PR #15228, opened 2022-03-26
MI32: refactoring conn task, bugfix, add response to op

Staars:
## Description:

This PR is pretty much the result of my attempt to port the legacy bind_key generator to Berry inside Tasmota, which seems to work now.
Changes:
- the connection task stays alive for asynchronous operations
- added optional `response`-flag to the `run`-method   in Berry -> subscribe and/or write in BLE w/o response from BLE device
i.e. `ble.run(2,1)`- means write with response 
- introduced return op of 103 in the callback to report a notification, that came asynchronously
- fixed a really stupid bug for command `MI32Key`, which also correctly expands the 12-byte-legacy key to a 16-byte-key.
- less LOG_LEVEL_INFO and more LOG_LEVEL_DEBUG, there were a bit to many messages for a more complex Berry project 
- use NimBLEAttValue

Here is the very unpolished version of the legacy bind_key generator:
https://gist.github.com/Staars/8f87799cc9f4fcb5a535d7c5623d671f

Possible usage:
Have the MAC of the dimmer in the format AABBCCDDEEFF (for instance from the dashboard).
Paste the berry script into the Berry web console and run the code.
In the Tasmota console type `pair AABBCCDDEEFF`, then long press the pair button of the dimmer.
After completion copy-paste the key from the console and merge with the MAC to a key_mac_string.
Add key with `MI32key KEYMAC`.

## Checklist:

### DIFF
--- a/tasmota/xsns_62_esp32_mi.ino
+++ b/tasmota/xsns_62_esp32_mi.ino
@@ -22,6 +22,9 @@
   --------------------------------------------------------------------------------------------
   Version yyyymmdd  Action    Description
   --------------------------------------------------------------------------------------------
+  0.9.5.5 20220326  changed - refactored connection task for asynchronous op, add response option,
+                              fixed MI32Key command
+  -------  
   0.9.5.4 20220325  changed - add Berry adv_watch and adv_block to BLE class
   -------  
   0.9.5.3 20220315  changed - reworked Berry part, active scanning and holding active connections possible, new format of advertisement buffer
@@ -92,7 +95,9 @@ class MI32SensorCallback : public NimBLEClientCallbacks {
   }
   void onDisconnect(NimBLEClient* pclient) {
     MI32.mode.connected = 0;
-     MI32.infoMsg = MI32_DID_DISCONNECT;
+    MI32.infoMsg = MI32_DID_DISCONNECT;
+    MI32.conCtx->error = MI32_CONN_DID_DISCCONNECT;
+    MI32.mode.triggerBerryConnCB = 1; //mainly for unexpected or requested disconnects
     //AddLog(LOG_LEVEL_DEBUG,PSTR("disconnected"));
   }
   bool onConnParamsUpdateRequest(NimBLEClient* MI32Client, const ble_gap_upd_params* params) {
@@ -169,12 +174,15 @@ void MI32scanEndedCB(NimBLEScanResults results){
 
 void MI32notifyCB(NimBLERemoteCharacteristic* pRemoteCharacteristic, uint8_t* pData, size_t length, bool isNotify){
     if(MI32.mode.triggerBerryConnCB) return; //discard data, if we did not pass the old to Berry yet
-    MI32.infoMsg = MI32_GOT_NOTIFICATION;
-    MI32.conCtx->buffer[0] = (uint8_t)length;
-    memcpy(MI32.conCtx->buffer + 1, pData, length);
-    MI32.conCtx->returnCharUUID = pRemoteCharacteristic->getUUID().getNative()->u16.value;
-    MI32.mode.triggerBerryConnCB = 1;
-    MI32.mode.readingDone = 1;
+    if(isNotify){
+      MI32.infoMsg = MI32_GOT_NOTIFICATION;
+      MI32.conCtx->buffer[0] = (uint8_t)length;
+      memcpy(MI32.conCtx->buffer + 1, pData, length);
+      MI32.conCtx->returnCharUUID = pRemoteCharacteristic->getUUID().getNative()->u16.value;
+      MI32.conCtx->operation = 103;
+      MI32.mode.triggerBerryConnCB = 1;
+      MI32.mode.readingDone = 1;
+    }
 }
 /*********************************************************************************************\
  * Helper functions
@@ -238,15 +246,13 @@ void MI32_ReverseMAC(uint8_t _mac[]){
 
 void MI32AddKey(mi_bindKey_t keyMAC){
   bool unknownMAC = true;
-  for(auto _sensor : MIBLEsensors){
+  for(auto &_sensor : MIBLEsensors){
     if(memcmp(keyMAC.MAC,_sensor.MAC,sizeof(keyMAC.MAC))==0){
-      AddLog(LOG_LEVEL_DEBUG,PSTR("new key"));
-      uint8_t* _key = (uint8_t*) malloc(16);
-      memcpy(_key,keyMAC.key,16);
-      _sensor.key = _key;
+      _sensor.key = new uint8_t[16];
+      memcpy(_sensor.key,keyMAC.key,16);
       unknownMAC=false;
       _sensor.status.hasWrongKey = 0;
-      AddLogBuffer(LOG_LEVEL_DEBUG,(uint8_t*) _sensor.key, 16);
+      AddLog(LOG_LEVEL_INFO,PSTR("add key to %s"),kMI32DeviceType[_sensor.type-1]);
     }
   }
   if(unknownMAC){
@@ -638,6 +644,7 @@ void MI32Init(void) {
     NimBLEDevice::init("");
     AddLog(LOG_LEVEL_INFO,PSTR("M32: Init BLE device"));
     MI32.mode.init = 1;
+    MI32.mode.readyForNextConnJob = 1;
     MI32StartTask(MI32_TASK_SCAN); // Let's get started !!
   }
 #ifdef USE_MI_EXT_GUI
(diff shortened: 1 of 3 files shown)


tony-fav:
Great update as always @Staars. In the framework so far, is there an equivalent of a command like `BLEOp1 m:AABBCCDDEEFF s:fff0 c:fff1 n:fff4 w:ab go` ? Something that writes to one characteristic but at the same time subscribes for notification on another?

Staars:
No, it has to be done step by step.
If I understand you correctly, that would mean first subscribe to a chr, then wait for the ack callback, then write to the other chr, but have your callback prepared to wait for the notification from the first chr.
Please keep in mind, that I already expanded this „connection part“ of the driver way beyond my initial plans. But as it worked out a lot better than expected and things like the key generator are possible, I still try to keep it somehow slim and move as much work to Berry as possible.

Staars:
You may be right and perhaps it could be added that way.
But wouldn’t it be okay in your example to say something like (pseudocode)
`DidWriteToFFF1 = true` after write
And then later in the callback
`If uuid == 0xfff4 && DidWriteToFFF1`
I mean, the notification does come from that characteristic and not from the one, which was written to even though it did trigger the notification.
Besides how can the internal notification callback know, that a user wants a user defined uuid in one case and the real notification uuid in another one?

tony-fav:
I think I've decided that `BLEOp` command above isn't even working right. This device (WP6003 air box) periodically notifies on that characteristic (`fff4`) but you can write `ab` to `fff1` to get an immediate response (say with the NRF Connect app). The `BLEOp` command above was inconsistent and I believe because it wasn't actually working it was just catching the periodic notification. So, your first interpretation is correct. Order of operations being 1) subscribe to notifications on `fff4`, 2) write `ab` to `fff1` without unsubscribing or changing callbacks. 

If outside your scope, that's no problem. I'm going to see next if I can reliably time listening for the notification, then I won't need the write.

Staars:
Thanks for reporting, this is very helpful!

So did it actually work or not? In theory your example should be doable with the current driver.

I know my key_gen example is quite ugly, but at the moment I have no better idea, how to deal with the problem of blocking and  asynchronous  operations and without that ping-pong style.

BTW, I updated my preliminary doc page at https://staars.github.io/site/Bluetooth_MI32/#ble-class
I did not find the time yet to update all the examples at the bottom, but the rest should be in sync with the current driver.

tony-fav:
Timing the reconnection seems to work

```lua
var ble = BLE()
var buf = bytes(-64)
var wp6003_notify_delta = 5*60*1000 # every 5 minutes
var wp6003_buffer_time = 5*1000 # 5 seconds

def time()
  return tasmota.time_str(tasmota.rtc()['local'])
end

def subscribe()
  print("CONN:", time(), "Subscribing...")
  ble.set_svc("fff0")
  ble.set_chr("fff4")
  ble.run(3)
end

def disconnect()
  print("DCON:", time(), "Disconnecting...")
  ble.run(5)
end

def cb(err, op, uuid)
  if op == 3
    if err == 0
      print("CALB:", time(), "Subscribed Successfully")
    else
      print("CALB:", time(), "Failed to Subscribe, Re-Trying in 1s")
      tasmota.set_timer(1000, subscribe)
    end
  elif op == 5 && err == 2
    print("CALB:", time(), "Disconnected Successfully")
    print('------------------------------------------')
  else
    print("CALB:", time(), err, op, uuid, buf)
  end
  
  if op == 103
    if buf[0..1] == bytes('120A')
      disconnect()
      tasmota.set_timer(wp6003_notify_delta - wp6003_buffer_time, subscribe)
    end
  end 
end
var cbp = tasmota.gen_cb(/e,o,u->cb(e,o,u))

ble.conn_cb(cbp, buf)
ble.set_MAC(bytes('60030394342A'), 0)

subscribe()
```
Output:
```
CONN: 2022-03-29T17:03:11 Subscribing...
CALB: 2022-03-29T17:03:13 Subscribed Successfully
CALB: 2022-03-29T17:05:25 0 103 65524 bytes('120A160316081E00DE58020050000B0100024D00000000000000000000000000...')
DCON: 2022-03-29T17:05:25 Disconnecting...
CALB: 2022-03-29T17:05:25 Disconnected Successfully
------------------------------------------
CONN: 2022-03-29T17:10:20 Subscribing...
CALB: 2022-03-29T17:10:22 Subscribed Successfully
CALB: 2022-03-29T17:10:25 0 103 65524 bytes('120A160316082300DE5802004C000B0100024C00000000000000000000000000...')
DCON: 2022-03-29T17:10:25 Disconnecting...
CALB: 2022-03-29T17:10:25 Disconnected Successfully
------------------------------------------
CONN: 2022-03-29T17:15:20 Subscribing...
CALB: 2022-03-29T17:15:22 Subscribed Successfully
CALB: 2022-03-29T17:15:25 0 103 65524 bytes('120A160316082800E15802005B000D0100024F00000000000000000000000000...')
DCON: 2022-03-29T17:15:25 Disconnecting...
CALB: 2022-03-29T17:15:25 Disconnected Successfully
```

Still can't figure out why the writes always say unsuccessful. Will have to pull out the NRF sniffer and/or whip up a minimal ESP32 sketch to figure out what's going on.

Staars:
Did you try w/o write response?

tony-fav:
Yup. Tried quite a few variations. In NRF connect app, writing only causes a notification if a "command" (no response). Fails/no notification happens if it's a request (with response).

tasmota/xsns_62_esp32_mi.ino
```diff
@@ -1139,83 +1160,110 @@ void MI32ConnectionTask(void *pvParameters){
...
-            }
-            else{
-              MI32.conCtx->error = MI32_CONN_CAN_NOT_NOTIFY;
-            }
-            break;
-        default:
-            break;
+              break;
+            case 2:
+              if(pChr->canWrite()) {
```
tony-fav, line 1202: `              if(pChr->canWrite() || (pChr->canWriteNoResponse() && !MI32.conCtx->response)) {`

This change fixed my write issue!

tony-fav: I would have thought canWrite covered with response and without response and canWriteNoResponse would help you differentiate if needed but it seems they are disjoint. canWrite == canWriteWithResponse only.

Staars:
Ah, nice!
Obviously I missed that part the API.

I would like to try that with my examples too, but I am not able to test before tomorrow.
But it looks like a useful change and we should change this part in the driver.

tony-fav:
Everything works exactly as expected now: subscribe --> subscribe success --> write --> get notification --> get temperature in my house yet another way!

tony-fav:
I've updated my test build to have 
```c
              if(pChr->canWrite() || pChr->canWriteNoResponse()) {
                uint8_t len = MI32.conCtx->buffer[0];
                if(pChr->writeValue(MI32.conCtx->buffer + 1,len,MI32.conCtx->response & !pChr->canWriteNoResponse())) { // true is important !
```

Slight change but is a bit more robust I think. I also PR'd the same change to the other driver and my `BLEOp` command works flawlessly with that driver now too (https://github.com/arendst/Tasmota/pull/15254).

Staars:
Sorry for late responding, I had some busy days.
As I understand the BLE standard it should be possible, that the server allows write with and without response, so IMHO your proposed changes, would turn off the write response, if a write response was requested and the server also allows writing without response.
I think the following tiny change, should create a fallback to write without response, if at least the latter is available:
```
if(pChr->canWrite() || pChr->canWriteNoResponse()) {
  uint8_t len = MI32.conCtx->buffer[0];
  if(pChr->writeValue(MI32.conCtx->buffer + 1,len,MI32.conCtx->response & pChr->canWrite())) {
```
(I hope, I am not writing crap here  😃 )

BTW, this `// true is important !` was indeed important due to my faulty implementation, but is not necessary
anymore.
If you can agree with this feel free to do a PR or let me know, if I should do it.

tony-fav:
From my understanding of not BLE standard but the NimBLE functions specifically:

- canWrite = can be written to with or without response request
- canWriteNoResponse = can only be written to without response request

Once `pChr->canWrite() || pChr->canWriteNoResponse() == true`, we have `pChr->canWrite() == !pChr->canWriteNoResponse()` So, I think our solutions are functionally equivalent. PR at will!

I am currently having the issue that advertisements never return after a `ble.run(5)`. They seem to return when doing a single op but not after a sequence finished with a disconnect.

Staars:
I am sure, both solution lead to the same real world result in 99,9999% (or higher) of the cases, as it normally makes no sense for a GATT server to provide both options. But at least I found some hints, that it is theoretically possible (like in the Chapter: 'Writing to a characteristic' in https://punchthrough.com/android-ble-guide/).
Anyway, I am fine with both implementations.

It seems, I forgot to add some logic to restart the scan after adding `ble.run(5)`. I will have to fix this.

arendst:
OMG. Punchthrough! I still have four beans on my desk. History repeats itself.

PR #15278, opened 2022-04-03
console: keep cursor at the end of command input

schlimmchen:
## Description:

move the text cursor to the end of the command input when filling the console command from history using JavsScript. this hopefully is more convenient to most users and is similar to the behavior of a desktop terminal.

## Checklist:

I have no access to an ESP32 device. However, I argue that it is irrelevant for this PR.


s-hadinger (approved):

arendst:
Thx. Pity the commands are so long. I see at least two spaces which can be removed.

I will do some tests before merge.

schlimmchen:
> I see at least two spaces which can be removed.

Three. And two semicolons :wink: 

More effective is moving the cursor to a position assumed to be farther to the right than ever required but using a constant position that can be expressed with a small amound of characters. This trick could be used at least three times more when I look at the JS code block of the console HTML.

Also the use of the lambda function (arrow syntax) is more effective than the spaces and semicolons. However (1) this breaks compatibility with IE (if that is something you think is desirable), see https://caniuse.com/arrow-functions, and (2) it should not be used once for anonymous functions but for all if it is to be embraced. Which one should it be?

arendst:
If the gain (less code size) is significant it could be an option to choose the most gaining solution. If not then leave it be.

Thx for your insights.

For now I'll merge this.

PR #15280, opened 2022-04-03
Downsizing JavaScript

schlimmchen:
## Description:

Using lambdas (arrow functions) and using scientific notation for large integers we can save 58 Bytes of JavaScript code.

Not sure it is really worth a lot, but for fun I provide this PR :wink: 

## Checklist:


arendst:
Thx.

I just remembered these lambda functions failed on IE and I even have a define for them: USE_JAVASCRIPT_ES6

I see it's disabled in ``my_user_config.h`` so the latest PR will only work on non-IE browsers. Not a big deal as IE is phased out anyway.

For now thx for the effort. I need to decide if I want to go all-in on lambda functions.

s-hadinger:
I love it. Nice work.

I don't think it's a big deal to lose IE compatibility.

Jason2866:
100% agree to drop IE support. Microsoft has done too.

PR #15292, opened 2022-04-04
Add HX711 absolute weight conversion

ThomDietrich:
## Description:

**Related issue (if applicable):** fixes #13983

Implemented as discussed in issue ticket. Tested and working as expected.

## Checklist:


tasmota/settings.h
```diff
@@ -796,8 +798,8 @@ typedef struct {
   uint16_t      pulse_counter_debounce_high;  // FBA
   uint32_t      keeloq_master_msb;         // FBC
   uint32_t      keeloq_master_lsb;         // FC0
-  uint32_t      keeloq_serial;             // FC4
-  uint32_t      keeloq_count;              // FC8
+  //uint32_t      keeloq_serial;             // FC4 TODO discuss
+  //uint32_t      keeloq_count;              // FC8 TODO discuss
```
ThomDietrich, line 802: This is not finished yet. Waiting for advise how to proceed.

This new functions needs two variables in permanent memory. The absolute weight is calculated from raw value following the formula

```
     weight = (float)Hx.absconv_a / 1e9 * Hx.raw + (float)Hx.absconv_b / 1e6;
```

Typical config values are:
* absconv_a = 953499
* absconv_b = 15418390

The end result of the computation is a weight with a few decimals. The size of those variables can therefore certainly be reduced!
I was not sure about other builds and configs of hx711 use cases though. Another user could end with raw_weight values in different ranges and need to apply a different parameters a and b.

tasmota/xsns_34_hx711.ino
```diff
@@ -293,11 +314,13 @@ void HxInit(void)
...
     SetWeightDelta();
 
-    if (HxIsReady(8 * HX_TIMEOUT)) {                 // Can take 600 milliseconds after power on
+    if (HxIsReady(8 * HX_TIMEOUT)) {  // Can take 600 milliseconds after power on
       if (!Settings->weight_max) { Settings->weight_max = HX_MAX_WEIGHT / 1000; }
       if (!Settings->weight_calibration) { Settings->weight_calibration = HX_SCALE; }
       if (!Settings->weight_reference) { Settings->weight_reference = HX_REFERENCE; }
       Hx.scale = Settings->weight_calibration;
+      if (Settings->weight_absconv_a) { Hx.absconv_a = Settings->weight_absconv_a; }
+      if (Settings->weight_absconv_b) { Hx.absconv_b = Settings->weight_absconv_b; }
```
ThomDietrich, line 323: Not sure if conditions needed.

tasmota/settings.h
arendst: I just made these variables available in settings.h

I think you need to find some kind of scaling to keep the values within the space of a uint32_t being 0 to 4.294.967.295 while keeping in mind results of calculation being also within this range. Using floats makes things worse as the float range is a lot smaller.

tasmota/xsns_34_hx711.ino
```diff
@@ -441,6 +464,9 @@ void HxShow(bool json)
     }
     weight = (float)Hx.weight / 1000;                // kilograms
   }
+  if (Hx.absconv_a != 0 && Hx.absconv_b != 0) {
+    weight = (float)Hx.absconv_a / 1e9 * Hx.raw + (float)Hx.absconv_b / 1e6;
+  }
```
arendst, line 469: Am I right in the conclusion that using the raw value here makes current calibration efforts useless? How do you define the absconv parameters? By trial and error?

ThomDietrich: No you are correct. No offense, I didn't find the implemented calibration combined with taring the weight upon startup useful for my use case (and wonder in which use case it is). The strategy I follow is to take a few raw weights for known masses, then run a simple linear regression, e.g. here: https://mycurvefit.com
It's not as integrated but is a good deterministic process. Let me know if you believe we should have a discussion about a different path!

tasmota/settings.h
ThomDietrich: Thanks!

Okay let me think of a reasonable scaling!

ThomDietrich: Having some difficulties wrapping my head around this.

- `raw` is currently of type `long`
- `absconv_a` and `absconv_b` are of type long as well (meaning they are signed, which contradicts uint32_t)
  - `absconv_b` can be positive or negative, therefore signed long makes sense
- the `weight` result realistically lies below 200kg and could also be negative when the setup is changed, lets say [-400..400]

Therefore, result of the calculation should always be within range, under reasonable parameters. How much sanity checking of these parameters do we need to implement? 


----

For reference, some of my setup configurations:

image

arendst: Just changed both absconv to int32_t (long)

tasmota/xsns_34_hx711.ino
arendst: My implemented use case is based on a commercial litchen scale  like the Ikea one.

After power on the user adds an empty dish, selects the reset button to zero the result and then is able to weigh anything he adds to the dish. Hence the zero button in the GUI.

There are ofcourse other uses which may use your suggested solution.

As your implementation doesn't interfere with mine I'll add it now so users can get some experience with it.

arendst:
Merged for testing

tasmota/settings.h
arendst: I've played around with this and I wonder why we need such an elaborate formula.

Considering my situation. I have a loadcell which, when no load is attached results in a raw value of 1253. When I load a weight of known 280grams it returns a value of 1533. So all I need is a possibility to decrease the raw value with 1253 and I have the weigth.

What am I missing here?

ThomDietrich: Hey @arendst,
Did you go through the internal calibration process prior to this experiment?
Asking because I mostly ignore this function and use standard values for all my devices. In combination with the calibration formula (like examples given above) I have a deterministic approach to build reproducible devices. I have a dozend of these devices operational, number growing.

```
{"Sensor34":{"WeightRef":5000,"WeightCal":22,"WeightMax":200000,"WeightItem":0.0,"WeightChange":ON,"WeightDelta":10}}
```

I understand that all of this could feel like a solution that is not in line with functionality that is already there. My use case is to build devices which are calibrated and produce absolute weight measurements (10-90kg) across reboots, in permanent installations over years. The approach as described works quite satisfying for allmost two years now. However, so far the `y=ax+b` conversion is done in Home Assistant and I wanted to change that.

Should we discuss a different (more Tasmota integrated) approach?

tasmota/xsns_34_hx711.ino
ThomDietrich: Interesting element you brought up!

`USE_HX711_GUI` is not defined for `FIRMWARE_SENSORS`. As a result I did never see this "zero button in GUI" you refered to. Maybe this should be changed?

tasmota/settings.h
arendst: Let's continue here https://github.com/arendst/Tasmota/discussions/13983#discussioncomment-2530483

tasmota/xsns_34_hx711.ino
ThomDietrich: @arendst ping in case you missed this.

PR #15324, opened 2022-04-09
Add Rule Trigger for WebQueryResponse of WebQuery Command

ascillato:
## Description:

The `WebQuery` command has the optional **WebQueryResponse feature**. This feature allows Tasmota to see in the console the response of a WebQuery. This PR adds a trigger to use this response in rules.

For example, if querying a weather API and if **WebQueryResponse** is enabled, you will see in the console:

```haskell
07:18:03.079 CMD: webquery http://goweather.herokuapp.com/weather/RiodeJaneiro GET
07:18:04.435 MQT: stat/tasmota_84B598/RESULT = {"temperature":"20 °C","wind":"4 km/h","description":"Light rain shower","forecast":[{"day":"1","temperature":"+28 °C","wind":"5 km/h"},{"day":"2","temperature":"+26 °C","wind":"11 km/h"},{"day":"3","temperature":"+26 °C","wind":"8 km/h"}]}
```

That response can't trigger any rule.

With this PR, the response is changed to:

```haskell
07:18:03.079 CMD: webquery http://goweather.herokuapp.com/weather/RiodeJaneiro GET
07:18:04.435 MQT: stat/tasmota_84B598/RESULT = {"WebQuery":{"temperature":"20 °C","wind":"4 km/h","description":"Light rain shower","forecast":[{"day":"1","temperature":"+28 °C","wind":"5 km/h"},{"day":"2","temperature":"+26 °C","wind":"11 km/h"},{"day":"3","temperature":"+26 °C","wind":"8 km/h"}]}}
```

Now, this new response can trigger rules like the following:

```JS
Rule1
  ON WebQuery#Temperature DO Var1 %value% ENDON
  ON WebQuery#Description DO Var2 %value% ENDON
  ON WebQuery#Forecast[1]#Temperature DO Var3 %value% ENDON
```

_Note:_ In order to use the **WebQueryResponse feature**, it must be added at compile time by adding `#define USE_WEBSEND_RESPONSE` to _**user_config_override.h**_ file.
_Note 2:_ If using an ESP8266, the WebQuery command can be used only with HTTP URLs. If using an ESP32, the WebQuery command can be used with HTTP and HTTPS URLs.

## Checklist:

### DIFF
--- a/tasmota/xdrv_01_webserver.ino
+++ b/tasmota/xdrv_01_webserver.ino
@@ -3164,15 +3164,22 @@ int WebQuery(char *buffer)
           // Return received data to the user - Adds 900+ bytes to the code
           const char* read = http.getString().c_str();  // File found at server - may need lot of ram or trigger out of memory!
           ResponseClear();
-          char text[2] = { 0 };
-          text[0] = '.';
           Response_P(PSTR("{\"" D_CMND_WEBQUERY "\":"));
+          char text[2] = { 0 };
+          text[0] = *read++;
+          bool assume_json = (text[0] == '{') || (text[0] == '[');
+          if (!assume_json) { ResponseAppend_P(PSTR("\"")); }
           while (text[0] != '\0') {
-            text[0] = *read++;
             if (text[0] > 31) {               // Remove control characters like linefeed
-              if (ResponseAppend_P(text) == ResponseSize()) { break; };
+              if (assume_json) {
+                if (ResponseAppend_P(text) == ResponseSize()) { break; };
+              } else {
+                if (ResponseAppend_P(EscapeJSONString(text).c_str()) == ResponseSize()) { break; };
+              }
             }
+            text[0] = *read++;
           }
+          if (!assume_json) { ResponseAppend_P(PSTR("\"")); }
           ResponseJsonEnd();
 #ifdef USE_SCRIPT
           extern uint8_t tasm_cmd_activ;
--- a/tasmota/xdrv_01_webserver.ino
+++ b/tasmota/xdrv_01_webserver.ino
@@ -3166,12 +3166,14 @@ int WebQuery(char *buffer)
           ResponseClear();
           char text[2] = { 0 };
           text[0] = '.';
+          Response_P(PSTR("{\"" D_CMND_WEBQUERY "\":"));
           while (text[0] != '\0') {
             text[0] = *read++;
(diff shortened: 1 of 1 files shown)


arendst:
Hope the result is valid json. In that case you may need to escape some characters too.

ascillato:
Added Support to non JSON answers for `WebQuery` command.
Also, in this case received characters are JSON-escaped.
The procedure applied is the same as the one in SerialReceive.

barbudor:
I'm surprised this need a PR
I'm already using a WebQuery to retrieve my ISP Public IP address and it's working fine

barbudor:
So I see that you are introduding an additional level 
What was 
```
sitename/gateway/stat/WEBQUERY = {"myip":"2.7.146.133"}
```
becomes
```
sitename/gateway/stat/WEBQUERY = {"WebQuery":{"myip":"2.7.146.133"}}
```

arendst:
The advantage of this pr is better distinguishing between rules. It also breaks your rules.....

What to do? (another SO.....)

barbudor:
Oh no please not another So 🤣
My rule is not a problem, I can change it

The question would mainly be: shouldn't it become the standard for all message? To have a 1st json level by the name of the command?
For external mqtt processor, so4 which use the command name instead of result is pretty convenient

Or should the rule mecanism be amended to always include this commandname# level without changing the json published?
So that for any command result, whether it is published under result or command name topic the rule trigger would be comlandname# ?

arendst:
> The question would mainly be: shouldn't it become the standard for all message?

Major rewrite. It would add an extra level to current "correct" JSON unless elaborate tests are implemented. Also see below.

> Or should the rule mecanism be amended to always include this commandname# level without changing the json published?

The rule has no idea where the JSON trigger comes from. No option.

So I'll merge this.

ascillato:
> I'm surprised this need a PR
> I'm already using a WebQuery to retrieve my ISP Public IP address and it's working fine

Yes, if you receive a JSON with just one data, it worked without this PR, but if you receive 2 or more data in the initial level `{"myip":"2.7.146.133","something":"extra_data"}`, the triggers `myip#data` or `something#data` won't do anything. That is why, an extra level solves that.

Now that you mention it, maybe we can modify the rules driver to accept those JSONs triggers, so in that case, we can revert this PR as it won't be needed anymore.

PR #15328, opened 2022-04-09
Update italian language

bovirus:
Please check and merge

## Description:

## Checklist:

### DIFF
--- a/tasmota/language/it_IT.h
+++ b/tasmota/language/it_IT.h
@@ -28,7 +28,7 @@
  * Use online command StateText to translate ON, OFF, HOLD and TOGGLE.
  * Use online command Prefix to translate cmnd, stat and tele.
  *
- * Updated until v10.1.0.6 - Last update 28.03.2022
+ * Updated until v10.1.0.6 - Last update 09.04.2022
 \*********************************************************************/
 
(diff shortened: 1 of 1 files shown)


curzon01 (commented):
The devices use italien terms that the user of these devices can no longer assign. UV and Aux is abbreviated to UV/Aux there, nobody can find the system setting for "Ultravioletti/Ausiliario". The polarization is abbreviated as "Pol" throughout the device, also the terms used on the WebUI are compounded hugely long and produce a very wide web page.

Blind translation into Italian is sometimes suboptimal.

bovirus:
I believe that is not so clear that POL = Polariazation.
If you have the space in the web GUI is better to have a clear (and not abbreviated) defintition to avoid misunderstanding.
If there is a limit about webgui etxt lenght should be posted as comment near to the definition
Other definition in the same section are not sumemrized.
Then why someone are summerized and other not?

curzon01:
This is due to how the device processes information and represents it in the device condition itself. These differ simply due to the blind translation, so that the user loses the relationship between the values on the WebUI and the values within the device condition and configuration menus.

I just think that with a translation you also have to deal a little with the technical equipment even it's not a simple device. NeoPool are complex control units with a large number of menu and konfiguration items. These are consistent within the device, so they lose their reference.

Now there are different names for one and the same thing. Those in the device and those on the WebUI. I don't think it's good if you simply translate them blindly.

PR #15332, opened 2022-04-09
Enhance embedding of local root certificate

aweatherguy:
Embedding a local root certificate for MQTT TLS currently requires modifying the source distribution (tasmota_ca.ino). This change eliminates the need to do this, providing the option through user_config_override.h. Some un-referenced data structures in tasmota_ca.ino have also been removed.  A pull request has also been made on Tasmota/docs to update docs/Self-signed-Mosquitto.md to include a description of the new process for embedding a root certificate.

## Checklist:

### DIFF
--- a/tasmota/local_ca_data_sample.h
+++ b/tasmota/local_ca_data_sample.h
@@ -1,0 +1,51 @@
+/*
+  local_ca_sample.h - sample file for embedding a local CA certificate
+
+  Copyright (C) 2021  Theo Arends
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


s-hadinger:
Thanks. If you're using esp32 we have more memory available and could include something easier

aweatherguy:
Yes -- the code I added to optionally omit the two commercial certs is really not necessary -- the main goal here is to allow including a local cert w/o modifying the source distribution. I am fine with removing the two macros which can omit the commercial certs.

Let me know if that's the best way to go and I'll remove them...

redhat421:
I like including the option to drop the predefined certs.

Dropping the AWS and LE certs will save a bit of space.

Thanks!

arendst:
On hold waiting for final decision by @s-hadinger

PR #15335, opened 2022-04-09
Added support for Sonoff MS01 soil moisture sensor

minovap:
## Description:

## Checklist:


arendst:
Before ever merge:

- Do NOT change global code for a local feature. Add the presentation to the DHT driver only.
- Do NOT use function pow() but use FastPrecisePowf() saving lots of code space

arendst:
I ordered an MS01 and will play with it. I plan to use your code for a more simpler implementation.

So no action on your part needed for now.

minovap:
Ok, thanks for looking into it! I was just about to dive in and see if I could rewrite my implementation as a new driver that is compatible with ESP32, but I haven’t worked in C++ before so I am glad you can help out with finishing up 👍

I added the datapoints in a comment for reference. I obtained the dataset by connecting a MS01 to a Sonoff TH10 and then ran a script that listened for a signal with a digital analyzer, read the data and waited for Sonoff to give their soil percentage value.  

I tried fitting the value conversion as a single third polynomial function, but it I could not get close to anything like linear correlation with the moisture reported by the sensor and how many % of the sensor was being dipped into water. I used a salt water mix with a concentration that gave me 100% soil moisture at the white line indicating soil level in the sensor, so I had some kind of reference values - dry air 0% and water up to the line 100%.

Anyway Sonoffs implementation is of course not a “correct” value anyway, but I wanted to reverse engineer it as they implemented it anyway.

I think the best implementation for this sensor would have some kind of sane default voltage values for 0% and 100%, and then let the user calibrate by changing those values according to their soil properties.

I could look into a calibration process and a function for calculating moisture based on calibration values if you are willing to add user inputs to the driver.

arendst:
I merged it in my way (less code changes and kept it close to the source).

Does the original sonoff app also need calibrating? I sure hope not.

Let's wait until I receive mine (should be next tuesday) and verify my results.

For now pls test if my merge works for you.

minovap:
I tested and I can't get it to work in the (voltage < 15037) range. The calculations looks correct. All I get is -0.1% humidity.

A couple of suggestions:
- Voltage is nice to also include, so the end user can interpret it themselves if they wish to (the voltage is the actual value and the calculation we do is just a rough estimate)
- I think 0% looks better as a lower limit than 0.1% - it is not a problem on normal air sensors since they practically never have to show the lower limit
- Also I would suggest going with zero decimals for humidity

The Sonoff app doesn't have calibration, however in their firmware I found api endpoints for getting soil type tables for voltage to humidity conversion. I never figured out the authentication method so I don't know exactly what the look like.

Also, read this to get a better idea of this sensor does and why calibration would be a good addition (or at least sending voltage so the end user can implement their own calibration):
https://makersportal.com/blog/2020/5/26/capacitive-soil-moisture-calibration-with-arduino

arendst:
The voltage < 15037 is caused by uin32_t. I already changed it to int32_t in latest commit.

I'll add voltage.

barbudor:
> Also I would suggest going with zero decimals for humidity
Your choice: [HumRes](https://tasmota.github.io/docs/Commands/#humres)

minovap:
By the way, before going the MS01 route I tried Gardernas solution but it had a really messy and unreliable integration to Home Assistant so I returned them. Anyway, I have attached a couple of screenshots with what they ask their customers to do in order to calibrate their capacitive sensor.
IMG_5998
IMG_5997
IMG_5996
.

PR #15349, opened 2022-04-11
Add support for Sensirion's SHT4x sensor

stibus:
## Description:

Add support for Sensirion's SHT4x temperature and humidity sensors. Tested with SHT40.

## Checklist:


arendst:
As this driver looks much as the SHT3x driver perhaps you can find a way to distinguish the SHT3x hardware from the SHT4x hardware (using specific I2C commands) and then, adding some minor timing/calculation differences, support both devices in one driver (xsns_14_sht3x.ino, well known to you ;-))

stibus:
Interesting idea. I wasn't sure if this would be desired or separate drivers are preferred. The command codes are different, so trying to read a register or result and evaluating the CRC might be good for identification. However, I'm not familiar with the details, e.g. error handling, of the I2C protocol and don't have an SHT30 for testing. That was another reason to go the easy way first, but I'm considering the challenge ;-)

stibus:
This is working on my side with a SHTC3 and a SHT40 or both at the same time. Originally, I probed the sensors by reading a serial number or an id register. As such is not documented for the SHT30 I then switched to probing by reading measurement data for all of them. By checking the CRC it should be quite robust I think. As stated before I could not test the SHT30.
Consulting the datasheets I noticed that the driver should work with the SHTC1 as well. Compared to the SHTC3 it did not support sleep, wakeup and low power mode yet.

stibus:
Hi @arendst, I guess my comment when updating the PR was a bit vague. I had modified the PR such that it now extends the old SHT3x driver to support the SHT40 as well. Please have a look at it whether it matches your expection.

arendst:
Thx.

Just merged my way saving some code space.

stibus:
Hi Theo. Thanks for merging. However, it somehow fails to detect the SHT40 on my Shelly1. I did not spot the mistake yet but will keep looking into this.

PR #15404, opened 2022-04-19
Add SDCard SDIO support

s-hadinger:
## Description:

Add SDCard SDIO support, which is the only supported way on Olimes ESP32 POE, and support 4-bit mode which is 4 times faster.

Now there are 3 ways to configure SDCard support:
- SPI mode: configure `SPI CLK`, `SPI MISO`, `SPI MOSI` and `SDCARD CS`
- SDIO 1-bit mode: configure `SDIO CLK`, `SDIO CMD`, `SDIO D0`
- SDIO 4-bit mode: configure `SDIO CLK`, `SDIO CMD`, `SDIO D0`, `SDIO D1`, `SDIO D2`, `SDIO D3`

How to choose?
- If ESP32 has only 3 GPIOs connected to SDCARD 2/5/7 connectors, only SDIO 1-bit mode will work
- If the SDCard shares GPIOs with other SPI devices (like M5Stack) then you need to use SPI mode
- If the SDCard does not GPIOs with any other SPI device, and if all SDCard connectors 2/3/5/7/8/1 are connected to GPIOs, you can use SDIO 4-bit mode which is faster

Default layout for ESP32:

GPIO|SPI|SDIO 1-bit|SDIO 4-bit
:---|:---|:---|:---
2|SPI MISO|SDIO D0|SDIO D0
4|||SDIO D1
12|||SDIO D2
13|SDCARD CS||SDIO D3
14|SPI CLK|SDIO CLK|SDIO CLK
15|SPI MOSI|SDIO CMD|SDIO CMD

## Checklist:


tasmota/language/af_AF.h
```diff
@@ -651,7 +651,13 @@
 #define D_SENSOR_SPI_MISO      "SPI MISO"
 #define D_SENSOR_SPI_MOSI      "SPI MOSI"
 #define D_SENSOR_SPI_CLK       "SPI CLK"
-#define D_SENSOR_BACKLIGHT     "Backlight"
+#define D_SENSOR_SDIO_CMD      "SDIO CMD"
+#define D_SENSOR_SDIO_CLK      "SDIO CLK"
+#define D_SENSOR_SDIO_D0       "SDIO D0"
+#define D_SENSOR_SDIO_D1       "SDIO D1"
+#define D_SENSOR_SDIO_D2       "SDIO D2"
+#define D_SENSOR_SDIO_D3       "SDIO D3"
```
arendst, line 659: Can't we use "SDIO D" and then index by number like we do for relays and buttons?

s-hadinger: I thought about it at first, but it has two major drawbacks:
- it would prevent having a second SDMMC connected. ESP32 has 2 slots for SDMMC
- GPIO numbers starts at `1` in Tasmota but the naming in SDIO is `D0`. It would be very confusing to have `SDIO D-1` mapped to SDIO D0.

arendst: Convinced

PR #15474, opened 2022-04-27
Add flowratemeter

curzon01:
## Description:

Add flow rate meter support like YF-DN50 and similar that provide the flow rate as a frequency.
Up to two meters providing flow rates from 1 Hz to 1 kHz are supported.
Uses a single GPIO for each meter.

## Checklist:


s-hadinger (approved):

jeroenst:
Can I use this for smaller flow meters? Is there any documentation on this PR?

sfromis:
While I have no real knowledge of the sensor or driver, I can see that the source code does have a bit of docs on commands:
https://github.com/arendst/Tasmota/blob/9c3588256addece4d59d9ef800e71c32607874c3/tasmota/tasmota_xsns_sensor/xsns_96_flowratemeter.ino#L268-L290

ALso, you need a build with `USE_FLOWRATEMETER` and one (or two) gpio configured as `Flowrate`. After that, the general sensor concept is that you get data from the sensor(s) on the web UI front page, and in the regular JSON payloads for sensors.

jeroenst:
Thank you, I probably going to use this and I will update documentation accordingly.

PR #15515, opened 2022-05-02
Firmware upload using factory partition

s-hadinger:
## Description:

When running a single OTA partition and a factory partition is present, the new Firmware File Upload does the following:
1/ restarts Tasmota to `factory` partition (request "/u4?u4=fct&api=")
2/ polls the same URL until the factory has connected to wifi/eth and confirms that factory is running (returning "true")
3/ sends the firmware to be flashed to OTA partition
4/ restarts on the new firmware from OTA partition

In this whole process, the factory partition is mostly invisible.

Behavior is unchanged for dual OTA partitions on ESP32, or ESP8266.

## Checklist:


s-hadinger:
The one thing missing is that I can't distinguish from Tasmota firmware upgrade and MCU upgrade. So with this PR, when in single OTA, the MCU flashing is temporarily broken.

s-hadinger:
New commit adds filtering based on first byte of file.

If first byte is 0xE9, apply the scheme above.

If not, upload as before (it is a file for MCU)

PR #15531, opened 2022-05-03
Support all Features of OV2640 Webcam

philrich:
## Description:

Support all OV2640 Camera Features, like SpecialEffect,(Auto) White Balance, (Auto) Exposure Control, (Auto) Gain Control, White/Black Pixel Correct, DCW, Gamma Correction, Lens Correction

Additionally there is a WcFeature Command to support some extra Features of the OV2640 Camera:
- Reduce FPS: Reduce Framerate which also increases Exposure Time. This causes better low light performance
- Nightmode: Further increase Exposure Time/lower Framerate in dark environments

Framerate is shown in Webgui
 
## Checklist:


tasmota/settings.h
```diff
@@ -309,28 +309,43 @@ typedef union {                            // Restricted by MISRA-C Rule 18.4 bu
...
+    uint32_t raw_gma : 1;
+    uint32_t lenc : 1;
+    uint32_t colorbar : 1;
+    uint32_t wpc : 1;
+    uint32_t dcw : 1;
+    uint32_t bpc : 1;
     uint32_t spare15 : 1;
     uint32_t spare16 : 1;
-    uint32_t spare17 : 1;
-    uint32_t spare18 : 1;
```
s-hadinger, line 326: This is a very high price in terms of Settings storage. Ideally I would have preferred a Berry driver complement to store all these in `_persist.json` which gives you much more flexibility. But the job is done.

Looking at the values below, most are just a single bit, or 2. Using 32 bits value for them is a complete waste. If Theo is ok to use Settings storage for these, I would request to use a bit-field instead. After a quick count, I believe they can all fit in a single 32 bits value.

s-hadinger:
Thanks for your PR. As mentioned the cost in Settings is currently too high. Can you squeeze everything in a bit-field of 4 bytes?

tasmota/settings.h
barbudor: @s-hadinger They are indeed bitfields mapped inside a single uint32_t.
Same as the SetOption bitfields for example
So the total cost of the PR is 11 bit inside an already allocated uint32_t + 1 new uint32_t

barbudor: Things being said t is always questionable if a driver should always store its settings in Settings or have it initialized by a boot-rule or autoexec.bat
Most of Mi32 options for example are not saved in settings and must be re-initialiazed at each boot

s-hadinger: Oh this is embarrassing. I missed the fact it was already a bit field. Never look at code before having a first coffee

Jason2866:
Thank you for the PR! 
In general changes in Webcam code have always a high risk to make it instable.
Webcam code is a "fragile" construct.

philrich:
Thanks for the comments. I have 3 AI-Thinker ESP32-CAM Boards running with my changes since over 1 week without any problems.

All of the settings I implemented are using the official esp32-camera api (also used in the Arduino WebCam Sketch which every Website mentions regarding this camera, or the ESPHome Camera Component implementation), so i don't think this makes the cam code more unstable. All Settings are done in the WcApplySettings function. Exception maybe is the WcFeature command which directly alters registers (as described in the pdf here:  https://hobbylad.files.wordpress.com/2020/02/ov2640-camera-module-software-application-notes.pdf) to cause a better low light performance. But i think this feature could be benificial for a lot of people who want to use the cam in low light environments (like me). Generally the picture quality can be much improved with the settings covered in this PR.

@s-hadinger regarding the settings size: I tried to change as little as possible so i stuck with the already implemented 32bit field and used a second one for the values which didn't fit. The manual exposure value alone needs 11 Bits. Of course i tried to use as few bits as possible. Maybe the colorbar settings doesn't need to get saved as ist is used to generate a test picture, but this is only 1 bit.

If it is really required to stick with one 32bit field to get this merged i will of course remove the second one. But then again, contrast, brightness and saturation are already saved, i think it would be confusing for users if some settings are saved and some not.

Regarding the Settings size mismatch, i don't yet understand why, all i did was to change this
```
-  uint8_t       free_esp32_448[4];         // 448
+  WebCamCfg2    webcam_config2;            // 448
```
It compiles cleanly for esp32s2 if i also change this:
```
 #ifdef CONFIG_IDF_TARGET_ESP32S2
-  uint8_t       free_esp32s2_456[2];       // 456 - fix 32-bit offset for WebCamCfg
+  //uint8_t       free_esp32s2_456[2];       // 456 - fix 32-bit offset for WebCamCfg
 #endif
```
Is there some kind of alignment i am overlooking?

Jason2866:
@arendst 
```
In file included from /home/runner/work/Tasmota/Tasmota/tasmota/tasmota.ino:86:
tasmota/settings.h:825:33: error: static assertion failed: TSettings Size is not correct
 static_assert(sizeof(TSettings) == 4096, "TSettings Size is not correct");
               ~~~~~~~~~~~~~~~~~~^~~~~~~
*** [.pio/build/tasmota32s2-safeboot/src/tasmota.ino.cpp.o] Error 1
```

arendst:
I'll have a chk.

arendst:
> Is there some kind of alignment i am overlooking?

Yes! I overlooked too.

It's caused by the ESP32-S2 much larger GPIO table overlaying the area where we planned the second webconfig parameters.

Anyway it's fixed now by moving the parameters down.

PR #15535, opened 2022-05-04
safemode for all 4MB devices

Jason2866:
## Description:

since it is NOT a breaking change. OTA does work for devices with legacy partition scheme.

**PR #15532 needs to be merged before!!**

@Staars as already discussed for solo1 the Python script needs a change to work 

## Checklist:


arendst:
So with this change all factory compiles are now using the new partition layout?

If so we might be to quick as safemode still needs some tweaks.

Jason2866:
Safemode does work. User feedback about details is not bad at all.

PR #15544, opened 2022-05-05
Fix broken js introduced in #15515

s-hadinger:
## Description:

**Related issue (if applicable):** fixes #15515 #15543

## Checklist:


arendst:
Now the ESP32 will fail don't you think as it doesn't have a ``<script>`` start anymore....

arendst:
Ah I see it's duplicating the ESP8266 script. In that case we need #ifdef ESP32 around both versions in webserve.

s-hadinger:
I'm not sure to understand. Only part of the JS changes with ESP32

arendst:
Yes. Sorry My fault. I forgot where the #ifdefs were. No problem.

PR #15573, opened 2022-05-08
Berry: add content_response() to webserver

Staars:
## Description:

Adds the possibility to simply respond to XMLHttpRequest allowing AJAX based web pages.
For now only strings are supported, but that should be fine for a lot of use cases.

Code snippet for illustration:
```
  #######################################################################
  # Display the complete page
  #######################################################################
  def page_ble_scan()
    import webserver
    import string

    # AJAX response
    if webserver.has_arg("loop")      # reacting to something like 'xr.open('GET','/ble_scan?loop=1',true)' from the JS side
      webserver.content_response("Hello") # there may be better things to send
      return
    end

    # regular web page
    webserver.content_start("BLE scan")           #- title of the web page -#
    webserver.content_send_style()                #- send standard Tasmota styles -#
```

## Checklist:

### DIFF
--- a/lib/libesp32/berry/generate/be_const_strtab.h
+++ b/lib/libesp32/berry/generate/be_const_strtab.h
@@ -370,6 +370,7 @@ extern const bcstring be_const_str_consume_stereo;
 extern const bcstring be_const_str_contains;
 extern const bcstring be_const_str_content_button;
 extern const bcstring be_const_str_content_flush;
+extern const bcstring be_const_str_content_response;
 extern const bcstring be_const_str_content_send;
 extern const bcstring be_const_str_content_send_style;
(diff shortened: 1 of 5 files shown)


sfromis:
Great :-)

Some days ago, I was looking for a way to have a Berry web server simply have a response without any html headers/footers added, but found none. Closest was content_send, but that did not work without a content_stop also adding the footer.

For completeness, it would make sense to also accept a bytes argument, to provide for a web server to have responses including images and so on.

s-hadinger:
> For completeness, it would make sense to also accept a bytes argument, to provide for a web server to have responses including images and so on.

The Arduino implementation relies on `String` which means that NULL char is not allowed. Hence a binary version wouldn't work as expected.

PR #15674, opened 2022-05-23
Berry improve precompiled gen.sh

s-hadinger:
## Description:

`gen.sh` is used to generate pre-compiled classes and modules of Berry. This new version removes unused files, and could be eventually included in the build system.

## Checklist:


Jason2866:
@s-hadinger When should this script started in build process? Always or what should trigger the build?

s-hadinger:
I'm hesitant since things can always go wrong. If it were to be integrated in the build process, then it should run for every ESP32 build and `generate` library should be empty and included in `.gitignore`.

PR #15839, opened 2022-06-22
Sm2335 5-channel LED driver

Cossid:
## Description:

Adds an I2C-based light driver for the SM2335(EGH) controller. This chip is used by e.g. [SwitchBot Color Bulbs](https://us.switch-bot.com/pages/switchbot-color-bulb).

It was created based on chip datasheet from the original manufacturer. It supports (as far as I can tell) all features of the original driver.

Notes to reviewer:  This I2C has a large current range, and rather than hard-code values, I took an approach similar to SM2135's "model" method, but instead made it more of a linear "power factor" setting rather than specific model mapping, and left all color correction to the SetOption37 feature.  As a result, this will take up a total of **17** AGPIO mappings.  This approach will use the lowest and safest value by default while given users the ability to change current levels without code changes.  If there is a different/better way to do this, please advise.  Please verify I set the AGPIO mappings correctly to not break any others, as I am not 100% familiar with this process.  Additionally, there is a comment that notes the size added by the driver for each config, and I could not see how to calculate that (program size in platformio does not break out individual files).   Please correct changelog commit if this does not go out in the next release or it was inappropriate to add.

## Checklist:


arendst:
Thx for your input.

I guess the driver works for you. As my goal is to not only provide working drivers but also make them codewise as small as possible I suggest you go over the code and remove the many redundant code. Look at for next loops and use arrays. At first look you might save at least 500 bytes.

Also do not enable this driver as default. The ESP8266 image sizes are already reaching the high watermark so I would like to keep control over this.

Cossid:
Changes have been pushed which:

- Reduce code by replacing redundant calls with for loops
- Disable by default on ESP8266 and enable by default on ESP32

@arendst please re-review.

arendst:
Perfect Thx.

PR #15867, opened 2022-06-25
Improve button debouncing

lugerun:
Currently, there is a good debouncing implementation for the switches.
But this does not apply anymore when configuring an input as button
for the usage at a blind. Then, the current implemntation works by
just checking the input once per debounce interval. At some devices,
this causes ghost switching. As a result, this change checks the
input twice per devounce interval and uses the old values if the two
read attempts reported different values.

## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_support/support_button.ino
+++ b/tasmota/tasmota_support/support_button.ino
@@ -43,7 +43,9 @@ struct BUTTON {
   uint16_t hold_timer[MAX_KEYS] = { 0 };     // Timer for button hold
   uint16_t dual_code = 0;                    // Sonoff dual received code
 
+  uint8_t debounce_phase = 0;                // 0 = regular iteration; 1 = read data for debouncing only
   uint8_t last_state[MAX_KEYS];              // Last button states
+  uint8_t last_state_between[MAX_KEYS];      // Last button states at the half time read
   uint8_t window_timer[MAX_KEYS] = { 0 };    // Max time between button presses to record press count
   uint8_t press_counter[MAX_KEYS] = { 0 };   // Number of button presses within Button.window_timer
 
@@ -91,6 +93,7 @@ void ButtonInit(void) {
 #endif  // ESP8266
   for (uint32_t i = 0; i < MAX_KEYS; i++) {
     Button.last_state[i] = NOT_PRESSED;
+    Button.last_state_between[i] = NOT_PRESSED;
     if (PinUsed(GPIO_KEY1, i)) {
       Button.present++;
 #ifdef ESP8266
@@ -204,6 +207,18 @@ void ButtonHandler(void) {
       button = AdcGetButton(Pin(GPIO_ADC_BUTTON_INV, button_index));
     }
 #endif  // USE_ADC
+    if (Button.debounce_phase == 1) {
+      // adjust values
+      Button.last_state_between[button_index] = button;
+
+      continue;
+    }
+
+    if (button != Button.last_state_between[button_index]) {
+       // no clean signal, just keep the previous one
+       button = Button.last_state[button_index];
+    }
+
     if (button_present) {
       XdrvMailbox.index = button_index;
       XdrvMailbox.payload = button;
@@ -374,6 +389,8 @@ void ButtonHandler(void) {
     }
     Button.last_state[button_index] = button;
   }
+
+  Button.debounce_phase = (Button.debounce_phase + 1) & 1;
 }
 
 /*
@@ -409,7 +426,7 @@ void MqttButtonTopic(uint32_t button_id, uint32_t action, uint32_t hold) {
 void ButtonLoop(void) {
   if (Button.present) {
     if (TimeReached(Button.debounce)) {
-      SetNextTimeInterval(Button.debounce, Settings->button_debounce);  // ButtonDebounce (50)
+      SetNextTimeInterval(Button.debounce, Settings->button_debounce / 2);
       ButtonHandler();
     }
   }


arendst:
Did playing with command ``ButtonDebounce`` not solve your issue?

lugerun:
No, because it will still falsely trigger at some point. The `ButtonDebounce` would only increase the time before it happens, because it would only check the state of the Button less frequent. It is only a matter of time when the false trigger impulse and checking of the button state will happen at the same time.
There was no real debounce implementation for buttons like for switches, which is confusing because you would expect `ButtonDebounce` & `SwitchDebounce` to do the same thing. Therefore it worked when configured as a switch but not as a button, which is used for the roller/shutter mode to enable doulbleclick and so on.

l-jonas:
This is still current, please reopen

sfromis:
Looks like this PR had a bad side effect. After this, a simple switch-type device goes from a delay from button push to relay action of ~0.5 seconds to ~1.0 second, from felling "a bit sluggish", to WTF. Activating `SetOption13` for immediate action works, but disables multi/long press features. This should be corrected.

Here's the long delay from button press to the relay toggling, 1.053 seconds, which looks like 1 second + the default ButtonDebounce time of 50 milliseconds.
```
00:56:44.069 APP: Button1 multi-press 1
00:56:45.122 SRC: Button
00:56:45.130 MQT: xxxxxxxx/stat/POWER = {"POWER":"ON"}
```

arendst:
I noticed something too. I hate those insisting PR providers in ancient code 

I'll revert today

l-jonas:
> I hate those insisting PR providers in ancient code

And I hate not clean code that for example makes a big difference between a button and a switch. Moreover, this code is not ancient at all and fixes a important issue for other users.

> After this, a simple switch-type device goes from a delay from button push to relay action of ~0.5 seconds to ~1.0 second

And you verified (not only tested) that this is caused by this, right? It could introduce a delay of up to the debounce period, but not any longer (assuming that the button does not bounce for a long time).

sfromis:
My verification was by flashing the immediately previous development iteration, seeing the expected latency of 0.5 seconds, and then updating to this PR, seeing the latency go to > 1 second. Standard debounce time of 50 milliseconds, and I tried several button presses both with working and failing code, with very consistent results. If your testing for debounce processing was using `SetOption13`, you'd not have noticed the flaw introduced.

l-jonas:
Which module are you using? The Sonoff Dual and CH4 are not handled like other modules and thus the behavior could be different. The same applies for the ESP32 with touch buttons.

It could be that the ButtonLoop is not called often enough. In this case it would be worth a try to double the button debounce time (where it expects as many calls as in the previous version).

sfromis:
I'm using a very plain Sonoff S20 (basically the same as a Sonoff Basic, but in plug format). You are right that `ButtonDebounce 100` can be a workaround, getting down to 600 ms instead of 1050 ms. Still slower.
```
13:33:31.094 APP: Button1 multi-press 1
13:33:31.697 SRC: Button
```

Jason2866:
>And you verified (not only tested) that this is caused by this, right? It could introduce a delay of up to the debounce period, but not any longer (assuming that the button does not bounce for a long time).

You are mixing who has to verify that the PR is working as described. See the PR checklist. If after reverting a PR everything is fine the PR is responsible. Thats it. If you want to get this feature again in, come with a new PR do all needed changes and it is considered to  merge the PR. Interpreting the comment of the project owner, the second attempt of a PR related to `Improve button debouncing` will need many test of the PR maker.

l-jonas:
> Still slower.

Slower by 50 ms which is exactly the amount of increased debouncing. Of course, it would be possible to subtract this time of the debouncing from the delay in the other code path, while that would make it more difficult to understand the code than it is already.

The problem could be related to the deep sleeping that maybe prevents more than one iteration per 50 ms - but that's something that the project owner should know and thus should trigger a comment during the code review and before merging a PR. (And it would be possible to disable the deep sleep when a button state change was detected so that the debounce could work faster - but that would be very complicated just to satisfy someone insisting on 25 ms faster reactions.)

> You are mixing who has to verify that the PR is working as described.

You are mixing verification and testing. Moreover, environments and the perception of slowness can be different.

> will need many test of the PR maker

I am not the PR maker and still it solves a problem for me and does not create one for me. I prefer a good working upstream instead of fixes that are only part of otherwise unmaintained forks. I assume that the project owner has the same opinion.

Jason2866:
>I am not the PR maker and still it solves a problem for me and does not create one for me. I prefer a good working upstream instead of fixes that are only part of otherwise unmaintained forks. I assume that the project owner has the same opinion.

For some a problem solved for many a problem introduced. So the project owner decided and removed this code.

sfromis:
Of course a good working upstream is the goal, with the focus on **good**. This PR was not.

With the code being pretty complex, with many use cases, a lot of testing is needed when trying to make "small" improvements in core parts. Most of the testing is not of the change, but to verify that nothing changed.

I do not consider artificially high debounce time to be feasible as part of a "solution", when there is no electrical need for it.

l-jonas:
> "small" improvements

Would it better if this would be marketed as bugfix for ghost switching? That's the actual problem. It's not just some small improvement.

> when there is no electrical need for it

To prevent ghost switching, there is the electrical need to read the button status twice. If the sleep is the cause, then there is no way to prevent a high debounce time (without decreasing the sleep time).

Jason2866:
I think we agree it is NOT simple to solve.

l-jonas:
That depends on what detail you are talking about. The only thing that is not simple for me is that I think that you don't understand the ghost switching problem and that I think that you don't understand the PR.

When looking at the switch debouncing, it uses a Ticker to run more than once per 50 ms to query the switch. Why did the owner not say how it would work better? Does he not know his code base?

The question is if that just works randomly because the ESP is not yet sleeping again or if there is some hidden background that I don't see.

sfromis:
What I meant is that in the cases where there is no electrical need for a long debounce time, such a config should not be made. Hence this is not a useful workaround for the badly increased latency.

Of course, I fully recognized that there are very real cases of troublesome electrics where debounce is useful, but in most cases there is not much of a need. And yes, we fully understand that ghost switching due to "less than ideal" electrics is an issue to some, but that does not change the very simple fact that the PR submitted introduced a bad error for other users.

No one would object to an improved and well-tested PR without negative side effects for other users.

lugerun:
I would consider Shutters randomly moving without input a more severe issue than using Setoption13 or changing debounce time if the delay feels too long on some devices. I don't think you can expect someone to buy and test against every supported device, just to make a PR request. It's hard to make sense of all the dependencies of this code snippet, due to the lack of a proper debounce function which should be used for switches **&** buttons.

sfromis:
We *can* expect people to make the most basic testing, which does not require buying "every supported device", my test was not an even slightly unusual setup.

No one is arguing against ghost switching being severe when it happens, but it only happens in a minority of cases (I'd first look at the questionable electrics anyway), but people using switches is a major use case which should not be impeded by an unintentionally breaking change.

And no one is arguing against improved debouncing, as long as it works as it should, without creating issues for other users. We are not trying to "attack" you, merely pointing out that there was a bad PR which deserved to be rolled back due to lack of reasonable testing.

An improved and well-tested PR would likely be a good enhancement of Tasmota.

arendst:
I released an updated button driver based on the switch debounce code.

Let's continue discussion in issue #16339

PR #15886, opened 2022-06-28
Fix missing sign of current power value on smart meter AS2020

KG3RK3N:
## Description:

Currently, the Smart Meter Honeywell AS2020 returns only positive current power values. Its like the ED300L, the status for the power direction is contained in OBIS 1.8.0. But the difference is, that the direction is on Bit 11 (not Bit 5) and the current power value has a different OBIS code.

The SML decoding discussion for the AS2020 is here: https://www.mikrocontroller.net/topic/538936

To enable the fix, its necessary to add the following to the `user_config_override.h`:
```
#ifndef AS2020
#define AS2020
#endif
```

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_10_scripter.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_10_scripter.ino
@@ -3993,7 +3993,7 @@ extern char *SML_GetSVal(uint32_t index);
           } else {
 
 
-#ifdef ED300L
+#if defined(ED300L) || defined(AS2020)
             fvar = SML_Status(fvar1);
 #else
             fvar = 0;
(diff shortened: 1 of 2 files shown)


chrissi5120:
HI, 

i got an holley ehz meter suffering from this problem.
currently i am running tasmota 12.1.1 and my script for current power looks like this:

1,77070100100700ff@1,Consumption (Current),W,curr_w,0

is there a way to read the negative value for me?

output of the meter looks like this:

77 07 01 00 10 07 00 ff 01 01 62 1b 52 00 52 1a 01 01 01 63 1a 86 00 76 04 00 00 03 62 00 62 00 72 65 00 00 02 01 71 01 

help would be really appreciated!

PR #15915, opened 2022-07-02
Analog TDS Sensos

JanezKr:
Support for Analog TDS sensor (water quality sensor)

## Description:

## Checklist:


tasmota/include/tasmota_template.h
```diff
@@ -181,6 +181,7 @@ enum UserSelectablePins {
   GPIO_OPTION_E,                       // Emulated module
   GPIO_SDM230_TX, GPIO_SDM230_RX,      // SDM230 Serial interface
   GPIO_ADC_MQ,                         // Analog MQ Sensor
+  GPIO_ADC_TDS,                         // Analog TDS Sensor
```
barbudor, line 184: New GPIOs must be added at the end of the list as stated in the comment lines 23-24

tasmota/include/tasmota_template.h
```diff
@@ -411,6 +412,8 @@ const char kSensorNames[] PROGMEM =
   D_SENSOR_OPTION " E|"
   D_SENSOR_SDM230_TX "|" D_SENSOR_SDM230_RX "|"
   D_SENSOR_ADC_MQ "|"
+  D_SENSOR_ADC_TDS "|"
```
barbudor, line 415: Same for this one, at the end of the list please

platformio_tasmota_env32.ini
```diff
@@ -64,6 +64,15 @@ extends                 = env:tasmota32_base
 board                   = esp32_16M
 build_flags             = ${env:tasmota32_base.build_flags} -DFIRMWARE_TASMOTA32
 
+
+[env:tasmota32-sensors]
```
barbudor, line 68: There are no plans to introduce a tasmota32-sensors, especially like this one which does nothing differently than tasmota32
As of today Tasmota32 covers the sensors use case
Please remove

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
```diff
@@ -164,6 +167,7 @@ struct {
   uint8_t type = 0;
   uint8_t pin = 0;
   float mq_samples[ANALOG_MQ_SAMPLES];
+  int tds_samples[ANALOG_TDS_SAMPLES];
```
barbudor, line 170: This adds up memory which is useless when TDS is not in use
You would save Ram by making a Union of both arrays as they are not used simultaneously

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
```diff
@@ -417,6 +431,76 @@ float AdcGetMq(uint32_t idx) {
...
+  AddLog(LOG_LEVEL_DEBUG, PSTR(D_LOG_APPLICATION "Getting value for tds-sensor "));
+  
+  
+  float temp = (float)Adc[idx].param1;
+
+  temp = ANALOG_TDS_COMPENSATION_TEMPERATURE; // TODO:  Adc[idx].param1 is not init with ANALOG_TDS_COMPENSATION_TEMPERATURE
+  // it would be better if we can read temperature sensor directly for compensation
+
+  int median = getMedianNum(Adc[idx].tds_samples, ANALOG_TDS_SAMPLES);
+  float voltage = median * ANALOG_V33 / ((FastPrecisePow(2, ANALOG_RESOLUTION)) - 1);
```
barbudor, line 487: It is inefficient to use Pow function for a power of 2. Especially for a constant result
Please use (1 << ANALOG_RESOLUTION)

barbudor (commented):
Please see my comments
Thanks

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
sfromis: Also beware that `FastPrecisePow` is horribly imprecise outside of intended use cases. 
> 2 to the power of 2 is 3.884

tasmota/include/tasmota_template.h
JanezKr: Ok

platformio_tasmota_env32.ini
JanezKr: I removed tasmota32-sensor.

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
JanezKr: I merged mq_samples and tds_samples arrays in one.

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
JanezKr: Thanks for info. I use shift left operation in this case.

I need little improvment of this sensor. Actualy TDS need temperature compensaton for precise value.  It would be the best that we read temperature from sensor connected on device.  

What is best practice to read read temperature sensor data?

arendst: Use global temperature variable set by attached temp sensor or by mqtt message.

Jason2866:
Please do not use a common term `samples` for a name or variable.
There is a high chance of generating issues...

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
JanezKr: Thanks Theo for very good hint. 

The version which read global temperature is commited.
But with global temperature variable are some issues.
In my case global temperature variable switching between external temperature sensor and CPU temperature sensor.

Here is log export from my device:
00:02:16.846 APP: Ppm read. ADC-RAW: 0.00, ppm: 0.00, temp: 29.12
00:02:17.081 APP: Getting value for tds-sensor 
00:02:17.086 APP: Ppm read. ADC-RAW: 0.00, ppm: 0.00, temp: 53.33


First entry with value 29.12 is external temperature sensor. The last value (just next tick) 53.33 is the CPU teprature. 


I found that is this declared in file xsns_127_esp32_sensors.ino (line 64). There are variable add_global_temp 
which check in json string exist any temperature parameter. This solution is unreliable.
Maybe is good enough that we never define CPU temperature as global temperature variable (We more offen need external temperature in sensor API). 
Or better and more comprehensive solution is make UI where user select Global temperature sensor name.

Can you fix this issiue with global variable?

barbudor: If température is needed, I would imagine that the water temperature is needed, not the ambiant air.

arendst: Was fixed recently.

Latest global temp/hum/press change makes it possible for a user to select a temperature from the JSON message. In your case you insert a manual temp. Global temperatures timeout after 5 minutes and become unavailable unless a new update is provided (either by a local temp sensor or externally by MQTT).

So to fix your issue you'll need to insert a real temperature at least once every 5 minutes. If you do not do this the next available temperature sensor is used. In your case the internal ESP32 one. Just disable the ESP32 sensor with command ``setsensor127 0``.

Best approach is to have a local temp sensor which will provide temp in the JSON message which in turn can be used as global temp with command ``GlobalTemp2 1`` signalling the first JSON temperature in the message.

Heritafita:
@JanezKr 
I gave it a try with by using a KS0429 TDS meter.
II think there is something wrong here : 
`float voltage = median * ANALOG_V33 / ((ANALOG_RESOLUTION << 1) - 1);` (file : xsns_02_analog.ino)

I replaced by this and it seems to work better now : 
`float voltage = median * ANALOG_V33 / ANALOG_RANGE;`

barbudor:
@JanezKr Your comment would be correct execpt that I can't find anywhere this line of code in the source file
The line I see is:
``` C
float voltage = (avg / ANALOG_MQ_SAMPLES) * ANALOG_V33 / ((FastPrecisePow(2, ANALOG_RESOLUTION)) - 1);
```
which definitively could be optimized using ANALOG_RANGE as you pointed

dingausmwald:
Any news on this? Would love to see this implemented

Jason2866:
@dingausmwald See https://github.com/arendst/Tasmota/pull/16287 it is implemented

dingausmwald:
@Jason2866 this started with a water quality sensor. The MQs are air quality sensors. Both measure tds in ppm, right? I am talking about the water quality sensor.

PR #15962, opened 2022-07-11
Audio i2s update

gemu2015:
## Description:

refactor i2s audio
select pins via tasmota GPIO gui
add codecs for s3box and s3box lite
add stereo microphone recording for s3box

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.2.0.4

### DIFF
--- a/lib/lib_audio/es7210/library.json
+++ b/lib/lib_audio/es7210/library.json
@@ -1,0 +1,7 @@
+{
+    "name": "ES7210",
+    "description": "Audio codec",
+    "keywords": "ESP8266, ESP32, MP3, AAC, WAV, MOD, FLAC, RTTTL, MIDI, I2S, DAC, Delta-Sigma, TTS",
+    "version": "1.0.0",
+    "frameworks": "Arduino"
+}
--- a/lib/lib_audio/es7210/library.properties
+++ b/lib/lib_audio/es7210/library.properties
@@ -1,4 +1,4 @@
-name=ES8156
+name=ES7210
 version=1.0
 author=
 maintainer=
--- a/lib/lib_audio/es7210/src/es7210.cpp
+++ b/lib/lib_audio/es7210/src/es7210.cpp
@@ -1,0 +1,549 @@
+/*
+ * ESPRESSIF MIT License
+ *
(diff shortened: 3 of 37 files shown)


s-hadinger:
I2S GPIOs were already defined. Can't you use the existing ones?

arendst:
I agree

gemu2015:
these are not compatible with esp idf audio i2s and mclk is missing

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/i2s.html#

An I2S bus that communicate in Standard or TDM mode consists of the following lines:

MCLK: Master clock line. It’s an optional signal depends on slave side, mainly used for offering a reference clock to the I2S slave device.
BCLK: Bit clock line. The bit clock for data line.
WS: Word(Slot) select line. It is usually used to identify the vocal tract except PDM mode.
DIN/DOUT: Serial data input/output line. (Data will loopback internally if din and dout are set to a same GPIO)

arendst:
But what are the chances of having both the "old" and your new GPIOs on one device?

If zero than the GPIO names can be reused by your driver without any problem. It's just naming for GPIO, there is no code behind it unless you maske the relation between the GPIO and the code.

arendst:
What if we can match the old naming to the new ones? Looking at it I think we cannot match all of them...

s-hadinger:
There are different naming in I2s. Fell free to rename with the new one

s-hadinger:
Do you need MCLK? It's used only when the external device is sending the clock?

gemu2015:
the s3 box uses mclk

gemu2015:
why not leaving both definitions ?
the old one is enabled by USE_I2S
the new one is enabled by USE_I2S_AUDIO

it well rarely be that both are used together

gemu2015:
yes of course

s-hadinger:
> why not leaving both definitions ?
> the old one is enabled by USE_I2S
> the new one is enabled by USE_I2S_AUDIO

I've been the only one using the previous definitions in a project still in development. I'm ok to align to any other #fidef and having the old ones depend on `USE_I2S_AUDIO`

arendst:
I'll do some changes now.

gemu2015:
Thx

PR #16003, opened 2022-07-16
Zigbee RGB Automatic color change mode

BassT23:
Hi

is it possible to make a automatic color change mode for RGB Lights within Zigbee?


s-hadinger:
Please open a discussion, not a Pull Request

PR #16275, opened 2022-08-21
Fix wrong flash_chip_id on ESP32, when DOUT is not used

Staars:
This also fixes the wrong reported flash size for flash modes: DIO, QUOUT and QIO. Now a global struct is used, which gets populated by the boot loader.

@Jason2866 Please test.

## Checklist:


Jason2866:
@Staars  Great find! Finally "just" a different routine to call.

PR #16314, opened 2022-08-23
all ESP32 boards do support `dio`

Jason2866:
## Description:

`dout` was used since it was and is necessary for some ESP8266 devices. The settings for ESP32 where just adopted from the ESP8266 settings.
Since core 2.0.4 finally solved the bootloader issues which some cores had before, i consider it is now safe to do this change. Arduino IDE has since a long time `dio` as default for the boards.

When using `dio` as flash mode the 2nd stage bootloader does a check if switching to a faster mode (`qio`) is even possible. There are plans for the future from Arduino espressif crew to remove the `dout` mode in the future.
Using mode `dout` prevents the 2nd stage bootloader to do this test/switching.

Tested on all my ESP32 devices i have. None did fail with OTA upgrade nor wired flash.
**But there is always a chance for a corner case where thing can go wrong ;-)**

I would appreciate more tests from other contributors before merging this PR

@arendst It is all up to you  if and when you merge ;-)

## Checklist:

### DIFF
--- a/boards/esp32-cam.json
+++ b/boards/esp32-cam.json
@@ -7,7 +7,7 @@
     "extra_flags": "-DARDUINO_ESP32_DEV -DBOARD_HAS_PSRAM -DHAS_PSRAM_FIX -mfix-esp32-psram-cache-issue -mfix-esp32-psram-cache-strategy=memw -DARDUINO_USB_CDC_ON_BOOT=0 -DESP32_4M",
     "f_cpu": "240000000L",
     "f_flash": "80000000L",
-    "flash_mode": "dout",
+    "flash_mode": "dio",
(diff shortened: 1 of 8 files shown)


arendst:
How do you test ifit works?

I've recompild for a c3 and after ota upload it still reports as DOUT....

Jason2866:
You are right. Since the bootloader is the old (configured `dout`) one.
As long this bootloader is there it will no change happen!  OTA updates with "old" bootloader do "just" update the firmware. Nothing else is changed. ONLY when you flash a new bootloader (with faster mode) the 2nd stage bootloader will check the settings in the magic bytes of the new OTA firmware and kicks in

PR #16346, opened 2022-08-26
Balanced energy consumption and export values over all phases.

areicherdt:
## Description:

I use the 3EM as a grid meter. And I am interested in the consumption and export values, balanced over all phases. To explain, I run a pv plant, which only feeds via one phase. The meter of my provider also balances over all phases, I want to emulate this with the 3EM.
So far I have not found a switch or value in tasmota that allows me to do this. For this reason, I have made a small PoC, which provides the values. Maybe one or the other has similar requirements and can do something with it.

## Checklist:

### DIFF
--- a/tasmota/include/i18n.h
+++ b/tasmota/include/i18n.h
@@ -26,6 +26,11 @@
  * (At this location they can/may be overruled by language files using undefs)
 \*********************************************************************************************/
 
+#define D_JSON_DAILY_SUM_CON_BAL "DailySumConsumptionBalanced"
+#define D_JSON_DAILY_SUM_EXP_BAL "DailySumExportBalanced"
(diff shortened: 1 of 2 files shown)


arendst:
What's the difference between using ``SO129 0`` or ``SO129 1`` other than this accumulates totals instead of daily values?

areicherdt:
`SO129 0/1` also allows negative values for total consumption. 
This does not reflect the behavior of my meter. The phases are summed together, but a negative value is NOT subtracted from the daily total. The negative value is added as an absolute value to the actual export.
So I get the values of the purchased electricity from the grid and what I have really exported.

arendst:
Implemented just now.

PR #16448, opened 2022-09-06
CI PR test refactored

Jason2866:
## Description:

PRs are now tested if the are working with the OS Linux, Windows and macOS.
Update action workflows to latest (security fix)

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.4.1

### DIFF
--- a/.github/workflows/Tasmota_build_devel.yml
+++ b/.github/workflows/Tasmota_build_devel.yml
@@ -59,7 +59,7 @@ jobs:
         with:
           ref: development
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
       - name: Install dependencies
         run: |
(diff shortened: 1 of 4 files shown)


Jason2866:
@arendst Maybe checking only one Tasmota32x version with the different OS?
Seems time consuming

arendst:
What are the criteria for testing different OSes? 

If things are OS specific then the PR writer should test this first. I think it's a waste of time and resources to test every Tasmota PR as most are OS independant.

Jason2866:
@arendst  The Platformio python scripts can be OS dependent. But yes it is the part the dev only do changes and if a contributer does changes we can keep a eye on.
I will check if i find a way to manually start a OS CI check workflow for a PR.
So we can do if we want/need.

PR #16464, opened 2022-09-08
Add option FORCE_USE_ETHERNET for single core MCUs that support it

ma-lalonde:
## Description:

Enables users to forcibly enable Ethernet for ESP32-MINI-1 MCUs made before 2022 and some versions of the ESP32-SOLO-1.

To use this feature, users only need to add `#define FORCE_USE_ETHERNET` to their `user_config_override.h` file.

The way it is implemented should not change the way users currently use Tasmota, but would enable this specific option in TasmoCompiler. I think this is the least invasive way to add this functionality.

## Checklist:

### DIFF
--- a/tasmota/include/tasmota_globals.h
+++ b/tasmota/include/tasmota_globals.h
@@ -107,9 +107,11 @@ String EthernetMacAddress(void);
 
 #if CONFIG_IDF_TARGET_ESP32
 
-#ifdef CORE32SOLO1
+#ifdef CORE32SOLO1                                 // ESP32-SOLO-1 OEM variant (Xiaomi devices) does not support ethernet...
+#ifndef FORCE_USE_ETHERNET                         // ...but ESP32-MINI-1 single core supports ethernet.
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks for the PR. Naive question, does it harm to always enable Ethernet on solo1?

ma-lalonde:
@s-hadinger Quoting Jason2866 on Discord:
> The solo1 OEM variant (Xiaomi devices) does not support Ethernet. Since we do not want build many frameworks and Tasmota variants we dropped Ethernet support for solo1.

The current PR is an option to bypass this "feature", adding more flexibility while not disturbing current users.

Jason2866:
Ethernet is disabled for solo1 builds since there was no known solo1 device which uses Ethernet.
@s-hadinger It does no harm. I vote for always enable to avoid not needed diffs.

Jason2866:
Thx, for the PR. We will close because it is against branch "master". PRs have to be against "development"
We will add "Ethernet" for solo1 build always.

PR #16645, opened 2022-09-24
Software Serial buffer to MIN_INPUT_BUFFER_SIZE

ortegafernando:
Increase software serial buffer to MIN_INPUT_BUFFER_SIZE.

## Description:

I have found this pull request that was accepted https://github.com/arendst/Tasmota/commit/b8752b7272495ebbc485ee4eca140aae0b4184f0

But it was revert afterwards
https://github.com/arendst/Tasmota/commit/aaa87a037a113dd7b4b006500d545a57bf85bbeb
https://github.com/arendst/Tasmota/pull/14153

The reason seems to be about: _"There is something wrong with serial config parameter, the change performs the config conversion twice resulting in a bogus config."_

Nothing about, buffer size.

I ask you for increasing the buffer of software serial connection from 130 to 200 or to MIN_INPUT_BUFFER_SIZE. May be a constant declared here is considered better option
https://github.com/arendst/Tasmota/blob/6bfb9d6aa4a8c0170581850cf5f66db909dc01f7/tasmota/include/tasmota.h#L200

It will not waste RAM if you dont use software serial.

## Checklist:


Jason2866:
One thing, why?

ortegafernando:
> One thing, why?

Because I (and probably some people) need more buffer in software serial. In this case, for me is enoght about 200 bytes, but searchin code I have found that a constant like MIN_INPUT_BUFFER_SIZE = 256; may be is better.

Also, I could change source code and make a PR to have a configurable buffer for software serial (it will need a restart after updating it). It will only "cost" a byte in flash memory like other settings (default 130, but user selectable).

This last option is better of course, but we have plenty of RAM for upgrading 130 bytes to 256.

Do you prefer the "configurable" option like "serialbuffer" ---> "sserialbuffer" command?

Jason2866:
@arendst Whats your opinion?

arendst:
Let's have it fixed to 256 for now.

ortegafernando:
I will make a PR in docs. @barbudor you can reject it if you dont want to give this value in wiki, thanks.

barbudor:
@ortegafernando 
No problems as far as you make it in docs/development 😄

PR #16649, opened 2022-09-25
Add support alias for ds18x20 sensors. 

### LINKED ISSUE #6970: Rename Sensors / give them friendly names
Hello, I have three DS18B20 connected and working perfectly.

However, the names are DS18B20 -1, DS18B20-2, DS18B20-3

How can these be renamed to friendly names? I am assuming they cant be because I was not able to find any info about this what-so-ever.

Might the only way this is possible is compiling my own build with the names embedded?

I think the feature request in that case is obvious.
Thank you
Ryan

_A clear and concise description of what the problem is._

_A clear and concise description of what you want to happen._

**Additional context**  

md5sum-as:
## Description:

**Related issue (if applicable):** fixes #6970
ESP8266 and ESP32 support now.
Added command
`DS18Alias XXXXXXXXXXXXXXXX,N`
where XXXXXXXXXXXXXXXX **full** SN of sensor and N - number 1..255
If use command without arguments in result printing current **full** SN and name in JSON
After set aliases you can use this in scripts or rules.
Initial settings can be make from autoexec.bat or MQTT

## Checklist:

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
@@ -28,8 +28,8 @@
 //#define USE_DS18x20_RECONFIGURE    // When sensor is lost keep retrying or re-configure
 //#define DS18x20_USE_ID_AS_NAME      // Use last 3 bytes for naming of sensors
       
-#define DS18x20_USE_ID_ALIAS
-/* Use alias for fixed sensor name in scripts by autoexec. Command: DS18Alias XXXXXXXXXXXXXXXX,N where XXXXXXXXXXXXXXXX full serial and N number 1-255
+/* #define DS18x20_USE_ID_ALIAS in my_user_config.h or user_config_override.h
+  * Use alias for fixed sensor name in scripts by autoexec. Command: DS18Alias XXXXXXXXXXXXXXXX,N where XXXXXXXXXXXXXXXX full serial and N number 1-255
   * Result in JSON:  "DS18Alias_2":{"Id":"000003287CD8","Temperature":26.3} (example with N=2)
   * add 8 bytes used memory
 */
--- a/tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
@@ -27,6 +27,12 @@
 
 //#define USE_DS18x20_RECONFIGURE    // When sensor is lost keep retrying or re-configure
 //#define DS18x20_USE_ID_AS_NAME      // Use last 3 bytes for naming of sensors
+      
+#define DS18x20_USE_ID_ALIAS
+/* Use alias for fixed sensor name in scripts by autoexec. Command: DS18Alias XXXXXXXXXXXXXXXX,N where XXXXXXXXXXXXXXXX full serial and N number 1-255
+  * Result in JSON:  "DS18Alias_2":{"Id":"000003287CD8","Temperature":26.3} (example with N=2)
+  * add 8 bytes used memory
+*/
 
 #define DS18S20_CHIPID       0x10  // +/-0.5C 9-bit
 #define DS1822_CHIPID        0x22  // +/-2C 12-bit
@@ -54,6 +60,9 @@ struct {
   uint8_t address[8];
   uint8_t index;
   uint8_t valid;
+#ifdef DS18x20_USE_ID_ALIAS
+  uint8_t alias;
+#endif //DS18x20_USE_ID_ALIAS  
 } ds18x20_sensor[DS18X20_MAX_SENSORS];
 
 struct {
@@ -325,6 +334,9 @@ void Ds18x20Init(void) {
       for (uint32_t j = 6; j > 0; j--) {
         ids[DS18X20Data.sensors] = ids[DS18X20Data.sensors] << 8 | ds18x20_sensor[DS18X20Data.sensors].address[j];
       }
+#ifdef DS18x20_USE_ID_ALIAS      
+      ds18x20_sensor[DS18X20Data.sensors].alias=0;
+#endif
       DS18X20Data.sensors++;
     }
   }
@@ -434,7 +446,15 @@ void Ds18x20Name(uint8_t sensor) {
     }
     snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("%s%c%s"), DS18X20Data.name, IndexSeparator(), address);
 #else
+#ifdef DS18x20_USE_ID_ALIAS
+    if (ds18x20_sensor[sensor].alias) {
+      snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("DS18Alias%c%d"), IndexSeparator(), ds18x20_sensor[sensor].alias);
+    } else {
+#endif
     snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("%s%c%d"), DS18X20Data.name, IndexSeparator(), sensor +1);
+#ifdef DS18x20_USE_ID_ALIAS
+    }
+#endif
 #endif
   }
 }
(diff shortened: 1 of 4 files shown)


tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
```diff
@@ -27,6 +27,12 @@
 
 //#define USE_DS18x20_RECONFIGURE    // When sensor is lost keep retrying or re-configure
 //#define DS18x20_USE_ID_AS_NAME      // Use last 3 bytes for naming of sensors
+      
+#define DS18x20_USE_ID_ALIAS
```
barbudor, line 31: I would suggest to move this outside so that
1. It is not defined by default (no impact of existing builds)
2. It can be defined in user_config_override.h

You may also add a commented (not enabled) line in user_config.h as a reference

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
```diff
@@ -434,7 +446,15 @@ void Ds18x20Name(uint8_t sensor) {
     }
     snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("%s%c%s"), DS18X20Data.name, IndexSeparator(), address);
 #else
+#ifdef DS18x20_USE_ID_ALIAS
+    if (ds18x20_sensor[sensor].alias) {
+      snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("DS18Alias%c%d"), IndexSeparator(), ds18x20_sensor[sensor].alias);
```
barbudor, line 451: Why not keep the existing standard syntax ? name+separator+index ?

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
md5sum-as: NNot now, because I'm not ready for the esp 32 board. If this function is in demand, then after being added to xsns_05_esp32_ds18x20.ino, we can take it outside.

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
md5sum-as: DS18X20Data.name not used for sensor type independence.

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
md5sum-as: Add support for esp32 and move DS18x20_USE_ID_ALIAS outsde

Jason2866:
As long the sn is different from the used sensors. Since ds18x20 is one of the most cloned / counterfeit devices this approach will not statisfy. I do have just a few of that sensors and the have all the same sn.

md5sum-as:
Two different sensors with the same SN will not work correctly anyway.

md5sum-as:
The user has two sensors installed: at the entrance and exit of the heating system. A script is configured that uses a standard name. The output temperature is transmitted in the DS18B20-1 value. When replacing the sensor, its number may change (features of the 1-wire search) and you will need to change the script. Or change alias in autoexec. The situation is in demand when using more than one sensor.

Jason2866:
> Two different sensors with the same SN will not work correctly anyway.

Nope, since the are not used together.

sfromis:
The existing feature for making the sensor name in the JSON be based on id instead of a not very useful sequence number, should cover your case, or...?

More generally, a Tasmota feature to assign friendly names to sensors should really not be limited to one sensor type, same arguments could be made for all other sensors, but Tasmota support for wired sensors is based on technical names being the general principle, with the backend supposed to take care of presentation with more user-friendly naming. According to this principle, you already have sufficient data with both ID available always, and also the option to get rid of the sequence number. 

IMO, the option for using ID-based sensor name instead of the sequence number should really be the default, but breaking changes and all that....

md5sum-as:
Current naming implementations are not productive in terms of hardware repairs. When naming a serial number, the number can change according to the 1-wire search algorithm. When using DS18x20_USE_ID_AS_NAME, the name will change anyway. In this case, it will be necessary to change the settings of ALL applications subscribed to MQTT topic. When using aliases MQTT subscribers operate with names like DS18Alias-1. These names can be assigned in autoexec.bat and when changing the sensor number or SN, you do not need to reconfigure all subscribers.

ouinouin:
looking this subject with interrest, i actually have one setup with 6 sensors, does it adds some friendly name abilty ? 
naming with a fixed number seems a first step, but when you have a lot of them, it would remain nice to be able to give  them friendly name.

sfromis:
The PR was merged, if you select the feature when building a binary, you should be able to use the new command.

barbudor:
@ouinouin No it doesn't add a friendly name but just an index that is reproducable
So a "friendly number"

krystof-k:
@md5sum-as, I bumped into an issue today. When rebooting, I guess one sensor was not found for a while, which resulted in the sensors exchanging between aliases. This is how it looks in Home Assistant:

image

Also, one sensor was removed from HA (probably because of the auto discovery), I have to add it (rename it) in HA manually again.

But according to this:

> When naming a serial number, the number can change according to the 1-wire search algorithm.

I guess there is no way to solve this. Or do you have any idea?

sfromis:
You could use the older feature for having DS18x20 sensors named according to the ID, instead of some sort of numbering.

barbudor:
> > When naming a serial number, the number can change according to the 1-wire search algorithm.
> I guess there is no way to solve this. Or do you have any idea?

This comment is related to the original Tasmota algorithm
If you have build with `DS18x20_USE_ID_ALIAS` and used the new command `DS18Alias XXXXXXXXXXXXXXXX,N` to map a ID to a number, this should be repetitive across reboot

krystof-k:
Yes, I have, [as I documented here](https://github.com/tasmota/docs/pull/1103/files?short_path=1d04d57#diff-1d04d57e87cf9e9bc9b36bbb5a0d9cee449a0c1ed83ee898bfda12e4a341f4f6), but the behavior I described still happened.

I understand it as the addresses are static only for a certain HW setup, when you add or remove a sensor, they (can) change.

krystof-k:
I just tried to reboot it again and the same thing happened again.

I use 7 sensors in total on 2 buses (6 and 1). The separate sensor is placed on a very long, 2 × 1.5 mm cable, in parasitic mode. After a restart, it doesn't get found sometimes, which is causing these troubles. However, when it gets found, it is working pretty OK, just sometimes reporting 85°C.

I tried to swap the buses and it renumbered the sensors completely.

md5sum-as:
For long lines, parasite powering, etc: [GUIDELINES FOR RELIABLE LONG LINE 1-WIRE NETWORKS](https://pdfserv.maximintegrated.com/en/an/AN148.pdf)

sfromis:
Without careful tweaking, multiple parasite power sensors have a hard time. Parasite power is not recommended "just so".

krystof-k:
There is just one. But I'm not complaining it is not working perfectly, it is an old cable and it is a miracle it is working at all :)

Just trying to find out if there is a way to assign the aliases more reliably.

barbudor:
> I understand it as the addresses are static only for a certain HW setup

The ID are hard coded in the silicon during the manufacturing process. They won't ever change.

krystof-k:
Hmm. Yesterday, the outdoor parasite-powered sensor reported -3064.3 °C. Very weird…

audax1764:
While trying to use aliases in a setup with 6 sensors, it seems to me that the labels are applied in the wrong order. 

While the labels are written into the correct entries in ds18x20sensors[], the name lookup in Ds18x20Name(uint8_t sensor) is not correct for the label case. The function argument "sensor" seems to refer to the desired printing order, which is different from the array index. 

After fixing this, the usage in CmndDSAlias() needs to be corrected as well (passing print index instead of array index) in order to make the output of DS18Alias command correct as well.

Here's a diff of the changes I've made. Probably the esp32 version needs fixing as well (can't test that on my hardware).
[tasmota_label_fix.txt](https://github.com/arendst/Tasmota/files/10393385/tasmota_label_fix.txt)

PR #16684, opened 2022-09-29
Adding CanSniffer based on MCP2515

kwiatek6324:
## Description:

PR add functionality of CAN-bus sniffer to tasmota based ESP devices.
Hardware need is MCP2515 connected via SPI.
Need customization in 
user_config_override.h

```
#define USE_SPI
//#define USE_MCP2515
#define USE_CANSNIFFER
```

Received CAN MQTT message looks like:
/tasmota/stat/tasmota_xxxxxx/CAN {"ID":51,"LEN":5,"D0":255,"D1":71,"D2":0,"D3":0,"D4":227}

Sending CAN frames via mqtt: 
/tasmota/cmnd/tasmota_xxxxxx/CANSEND {"Id":"11","Len":6,"Data":["121",20,"7","255","0","0"]}

## Checklist:
  - [ x] The code change is tested and works with Tasmota core ESP32 V.2.0.5

### DIFF
--- a/tasmota/include/tasmota_template.h
+++ b/tasmota/include/tasmota_template.h
@@ -556,6 +556,9 @@ const uint16_t kGpioNiceList[] PROGMEM = {
 #ifdef USE_MCP2515
   AGPIO(GPIO_MCP2515_CS),
 #endif  // USE_MCP2515
+#ifdef USE_CANSNIFFER
+  AGPIO(GPIO_MCP2515_CS),
+#endif  // USE_MCP2515
 #endif  // USE_SPI
 
 #if defined(USE_SDCARD) && defined(ESP32)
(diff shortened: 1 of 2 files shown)


arendst:
Thx. I'm doing a rewrite to better integrate with Tasmota (and current MCP2515 driver) and merge.

RcRaCk2k:
Did you correct the wrong JSON implementation? ASCII BIN was enclosured in "" instead of being a 8-bit integer. Also the JSON Receive should use Data Array instead of d0, d1, d2, d3 ....

sfromis:
Enclosing in quotes as string values would not be "wrong", just more generalized than needed. Anyway, when `xsns_87_can_sniffer.ino` generates JSON, integer values are not quoted as strings.

Again, implementation details like using an array is no "should", but about preference, and use cases. What I see in the driver code is using the format with separate fields `D1` etc.

goodkiller:
Related with #18287

PR #16706, opened 2022-10-01
Remove whitespaces

yuvalabou:
## Description:

Remove all trailing whitespaces

## Checklist:


s-hadinger:
Can you share the tool you used? This would allow us to automate. 

I'm not a big fan of such a huge change, there can be some hidden breaking changes

yuvalabou:
Used this terminal command
```
git grep -I --name-only -z -e '' | xargs -0 sed -i -e 's/[ \t]\+\(\r\?\)$/\1/
```

s-hadinger:
How do you know you didn't accidentally changed a binary file that would contain a sequence that looks like white spaces at end of file?

s-hadinger:
Also I wouldn't change the `lib` directories that contain libraries unchanged

barbudor:
I believe too there are less risks by doing it one by one while doing other changes

sfromis:
Yeah, while detecting/trimming trailing whitespace is a relevant linter feature, I do not see it as worth the effort and potential risk to go through all files in a mass operation. Of course, risks could be reduced by only going for specific file types, instead of all files, but still - just not important.

Jason2866:
Closing since to many files touched without knowing what is changed.

PR #16714, opened 2022-10-03
Add QMC5883L driver

Performance20:
## Description:

## Checklist:


tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,397 @@
...
+int16_t   MX = 0, MY = 0, MZ = 0, HG = 0;
+int16_t   xhigh = 0, yhigh = 0, xlow = 0, ylow = 0;
+int16_t   temp = 0;
+uint16_t  scalar = 0;
+uint8_t   status;
+bool      ready = false;
+uint8_t   i2c_address = QMC5883L_ADDR;
+} QMC5883L;
+
+void writeRegister(uint8_t reg, uint8_t val)
```
arendst, line 233: As these functions are global there is a good chance they are already used somewhere. I propose you dive into the Tasmota I2C functions and use them. See ``support.ino`` and search for ``I2CWrite8`` and pick whatever you need.

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,397 @@
...
+{
+  Wire.beginTransmission(QMC5883L.i2c_address); // start talking
+  Wire.write(reg);
+  Wire.write(val);
+  Wire.endTransmission();
+}
+
+int readRegister(uint8_t reg, uint8_t count)
```
arendst, line 241: Ditto

arendst (commented):
See comments

tasmota/my_user_config.h
```diff
@@ -715,6 +715,8 @@
                                                  // Reference: https://cdn-learn.adafruit.com/downloads/pdf/adafruit-led-backpack.pdf
     // #define SEVENSEG_ADDRESS1     0x70      // No longer used.  Use MTX_ADDRESS1 - MTX_ADDRESS8 instead to specify I2C address of sevenseg displays
 //    #define USE_DISPLAY_SH1106                   // [DisplayModel 7] [I2cDriver6] Enable SH1106 Oled 128x64 display (I2C addresses 0x3C and 0x3D)
+    #define USE_QMC5883L                            // USE_QMC5883L compass sensor
+      #define USE_QMC5883L_Temp 22                  // compass sensor temperatur are not calibrated (only relativ measurement) and need an absolute ground value in °C (see datasheet)
```
barbudor, line 719: Please comment those lines as we don't want this sensor to be enabled by default in all variants

However it is good to have them here commented so people can copy from here and paste in their `user_config_override.h`  what they want.
Which means that for your testing you also need to have those in your own `user_config_override.h`  that will not be part of the PR.

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,385 @@
...
+struct {
+int16_t   MX = 0, MY = 0, MZ = 0, HG = 0;
+int16_t   xhigh = 0, yhigh = 0, xlow = 0, ylow = 0;
+int16_t   temp = 0;
+uint16_t  scalar = 0;
+uint8_t   status;
+bool      ready = false;
+uint8_t   i2c_address = QMC5883L_ADDR;
+} QMC5883L;
+
```
barbudor, line 232: This is using quite some RAM even if the driver is not in use
I suggest another approcha using dynamically allocated memory
You can change that 
- removing every initializations
- define as a pointer:
```
struct QMC5883L_s {
...
} *QMC5883L = nullptr;
```
In your code, replace every `QMC5883L.` (dot) by `QMC5883L->`

See below changes to QMC5883L_Init() function

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,385 @@
...
+  // reset QMC5883L
+  if (I2cWrite8(QMC5883L.i2c_address, QMC5883L_CONFIG2, QMC5883L_CONFIG2_RESET) == false) return; // Software Reset
+  // writeRegister(QMC5883L_CONFIG2,QMC5883L_CONFIG2_RESET);  // Software Reset
+  //writeRegister(QMC5883L_RESET, 0x01);
+  if (I2cWrite8(QMC5883L.i2c_address, QMC5883L_RESET, 0x01) == false) return;
+  // write config
+  //writeRegister(QMC5883L_CONFIG, QMC5883L_CONFIG_OS256 | QMC5883L_CONFIG_8GAUSS | QMC5883L_CONFIG_10HZ | QMC5883L_CONFIG_CONT);
+  if (I2cWrite8(QMC5883L.i2c_address, QMC5883L_CONFIG, QMC5883L_CONFIG_OS256 | QMC5883L_CONFIG_8GAUSS | QMC5883L_CONFIG_10HZ | QMC5883L_CONFIG_CONT) == false) return;
+  QMC5883L.ready = true;
+}
```
barbudor, line 247: Don't use the i2c_address from QMC5883L as it wouldn't have been initialized but use directly the constant `QMC5883L_ADDR`

At the end of the function, if the init is ok, instead of
```
QMC5883L.ready = true;
```
use
```
QMC5883L = calloc(1, sizeof(struct QMC5883L_s));
```
This will initialize all members of the structure to 0 (calloc does that but malloc doesn't)
Now you don't need anymore the `ready` field as you can replace every `if (QMC5883L.ready)`  with `if (QMC5883L)`

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,385 @@
...
+
+//Read the magnetic data
+void QMC5883L_read_data(void)
+{
+  if(QMC5883L.ready != true) return;
```
barbudor, line 252: As per above comment this becomes
```
if(!QMC5883L) return;
```

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,385 @@
...
+const char HTTP_SNS_QMC5883L_ERROR[] PROGMEM =
+    "{s}QMC5883L {m} %s {e}";
+#endif
+
+
+void QMC5883L_Show(uint8_t json)
+{
+  if (json) 
+  {
+    if (QMC5883L.ready != true)
```
barbudor, line 325: You don't need this test as you should never come here is the init hasnt been successfull

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,385 @@
...
+    }
+    else 
+    {
+      ResponseAppend_P(PSTR(",\"QMC5883L\":{\"" D_JSON_MX "\":%d,\"" D_JSON_MY "\":%d,\"" D_JSON_MZ "\":%d,\"" D_JSON_MAGNETICFLD "\":%u,\"" D_JSON_HEADING "\":%d,\"" D_JSON_TEMPERATURE "\":%d}"), QMC5883L.MX, QMC5883L.MY, QMC5883L.MZ, QMC5883L.scalar, QMC5883L.HG, QMC5883L.temp);
+    }
+  }
+#ifdef USE_WEBSERVER
+  else 
+  {
+    if (QMC5883L.ready != true)
```
barbudor, line 338: same as above

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,385 @@
...
+  if (!I2cEnabled(XI2C_71))
+  {
+    return false;
+  }
+
+  if (FUNC_INIT == function) {
+    QMC5883L_Init();
+  }
+  else 
+  if (QMC5883L.ready == true)
```
barbudor, line 365:
```
if (QMC5883L)
```
and please put the if on the same line as the else

barbudor (commented):
@Performance20 
As to improve the RAM footprint of your driver, please find in my review some recommendations
Let me know if you have any questions about them
Thanks

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
barbudor: Overall this also mean you son't need anymore the `ready` and the `i2c_address` of the structure
Even more RAM saving 😄

arendst:
Thx. With your latest changes you introduced an issue when the hardware is not attached.

Originally you performed some tests during the init phase and set a flag if the hardware was found. That's great. This test is gone and now every second a call is made to read the sensor while it is not initialised.

I see where the problem comes from; introducing variables on the heap. I suggest you re-introduce a global test flag to be used during init phase and add it again to the main interface to stop unnecassary calls to unsafe functions.

Performance20:
I hope, I undestand what you mean. I added the test aigain.

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,331 @@
...
+  int16_t   temp;
+  uint16_t  scalar;
+  bool      ready;
+} *QMC5883L = nullptr;
+
+
+// Initialise the device
+void QMC5883L_Init()
+{
+  QMC5883L = (QMC5883L_s *)calloc(1, sizeof(struct QMC5883L_s));
```
barbudor, line 230: Wrong too
I suggested to allocate the memory if and ONLY if the device is detected
Please read gain my previous comments

barbudor:
You can check this driver which implements it correctly : xsns_93_hm330x.ino

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
```diff
@@ -0,0 +1,331 @@
...
+{
+  if (!I2cEnabled(XI2C_71))
+  {
+    return false;
+  }
+  if (FUNC_INIT == function)
+  {
+    QMC5883L_Init();
+  }
+  else if (QMC5883L->ready == true)
```
barbudor, line 311: Wrong
I suggested a test : `if (QMC5883L)`
With your implementation if you dereference QMC5883L as a nullptr you read random value for `ready`
You MUST test QMC5883L as a non nullptr and this is what my suggestion is about from hte beginning

tasmota/tasmota_xsns_sensor/xsns_33_qmc5883l.ino
Performance20: Yes but Theo suggest the ready flag to avoid calling of functions in  Xsns33, but if the init went wrong I must have the the stuct available before I leave the Initi funtion with return.

barbudor (approved):
All good for me
Sorry for bothering you but that's the cleanest way of doing it
Thanks for your contribution
Just put a QMC5883L in my Ali cart for the fun

arendst:
I've changed your sqrt() to tasmotas SqrtInt() to safe 650 bytes.

Pls test latest dev release with the QMC5883L driver.

Performance20:
have tested, works as expected

arendst:
Thx.

sugus6:
Thank you for the QMC5883L driver. I got my pcb running.
Is it possible to add the "Compass Heading" to the Tasmota UI?
QMC5883L Tasmota UI

PR #16728, opened 2022-10-05
size option on picture download

gemu2015:
## Description:

allows to download different picture size than video.
ip/wc.jpg?s=X   (1 ..13)

+ adds camera chip id info message

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
@@ -109,12 +109,16 @@
 #include "esp_camera.h"
 #include "sensor.h"
 #include "fb_gfx.h"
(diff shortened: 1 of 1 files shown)


Jason2866:
@gemu2015  What was the reason for closing the PR?

gemu2015:
it worked only with esp32-s3 not with standard esp32
and after checking the webcam sources i realized that there would be much more work to get it working on all  variants, even on esp32-s3 there were too much restrictions.
my goal was to not disturb streaming video.

it would however be easy to implement this by stopping streaming
deiniz the driver
init with new resolution
get the picture 
deiniz  the driver again
and reinit with original resolution.
this would work with all variants
but would however stop the video stream for about a second

i will probably implement this in the future when i am back
i will be traveling for several weeks and be back in end of november

PR #16762, opened 2022-10-08
Update bg_BG.h

StoyanDimitrov:
## Description:
Translation update


StoyanDimitrov:
What is the meaning of D_CURRENT_NEUTRAL and D_CT_POWER?

arendst:
> What is the meaning of D_CURRENT_NEUTRAL and D_CT_POWER

D_CURRENT_NEUTRAL  = "Current Neutral". The current flowing to the neutral power line. Text needs to be short as it is displayed in the GUI. See below.
D_CT_POWER = "CT Power" where CT is a "Current Transformer" but do not translate this to a long text form as it will ruin the GUI display. Hence short form.

sfromis:
Ideally, well-balanced 3-phase would have no current on the neutral wire (which then would not be needed), but the 3 phases may be in use for different purposes.

StoyanDimitrov:
I'll keep than in mind for the next iteration.

PR #16772, opened 2022-10-09
Epoch in Berry time map

sfromis:
When handling time and intervals in Berry, I found no feature for getting from parsing a string timestamp to the calculation-friendly epoch value, only to the map of broken-down time fields. When getting data from the outside, like a MQTT subscription, time stamps are usually in human-readable text format, which is great, but when doing programming, like a simple duration between two timestamps, having the epoch is highly useful.

@s-hadinger - Do you agree with this enhancement, or does it "need improvement"?

## Description:

No issue fixed, but I suggested this small enhancement on Discord some time ago.
https://discord.com/channels/479389167382691863/826169659811168276/984930495848665119

## Checklist:
  - [*] The code change is tested and works with Tasmota core ESP32 V.2.0.5


s-hadinger:
Good with me, thankg

PR #16802, opened 2022-10-12
 fix stack overflow vulnerability

winmin:
## Description:

fix stack overflow vulnerability exists in the rtsp module in libesp32


WinMin:
swing of Chaitin Security Research Lab

PR #15230, opened 2022-03-26
Fix %TIMERx% variables for negative values

barbudor:
## Description:

**Related issue (if applicable):** fixes https://github.com/arendst/Tasmota/discussions/15205 from @rapi3

When timer uses mode 1 (relative to sunrise) or mode 2 (relative to sunset), the "negative" offset stored in settings is "encoded" as > 12:00. This was not taken into account in previous PR https://github.com/arendst/Tasmota/pull/14619 by @alexasf

This PR provides a coherent value with mode 0, where `%TIMERx%` is the time of day at which the event has/will happen so it can be compared to `%time%` in rules.

Example with `%sunrise%` as 06:10 (400) and `%sunset%` as 19:11 (1151):
```
14:13:56.914 CMD: backlog rule3;timers;event test
14:13:56.992 MQT: stat/nodemcu/RULE = {"Rule3":{"State":"ON","Once":"OFF","StopOnError":"OFF","Length":204,"Free":307,"Rules":"ON event#test DO backlog var1 %timer1%;var2 %timer2%; var3 %timer3%;var4 %timer4%;var5 %timer5%;var6 %timer6%;var7 %timer7%;var8 %timer8%;var9 %timer9%;var10 %timer10%;var11 %sunrise%;var12 %sunset% endon"}}
14:13:57.219 MQT: stat/nodemcu/TIMERS = {"Timers":"ON",
  "Timer1":{"Enable":1,"Mode":1,"Time":"00:10","Window":0,"Days":"1111111","Repeat":1,"Action":3},
  "Timer2":{"Enable":1,"Mode":1,"Time":"-00:10","Window":0,"Days":"1111111","Repeat":1,"Action":3},
  "Timer3":{"Enable":1,"Mode":1,"Time":"11:59","Window":0,"Days":"1111111","Repeat":1,"Action":3},
  "Timer4":{"Enable":1,"Mode":1,"Time":"-11:59","Window":0,"Days":"1111111","Repeat":1,"Action":3},
  "Timer5":{"Enable":1,"Mode":2,"Time":"00:10","Window":0,"Days":"1111111","Repeat":1,"Action":3},
  "Timer6":{"Enable":1,"Mode":2,"Time":"-00:10","Window":0,"Days":"1111111","Repeat":1,"Action":3},
  "Timer7":{"Enable":1,"Mode":2,"Time":"11:59","Window":0,"Days":"1111111","Repeat":1,"Action":3},
[...]

PR #9543, opened 2020-10-14
Timer support for no-sunset permanent day

cybermaus:
## Description:

**Related issue (if applicable):** fixes #9539 

## Checklist:
  - [X] The pull request is done against the latest dev branch
  - [X] Only relevant files were touched
  - [X] Only one feature/fix was added per PR.
  - [X] The code change is tested and works on Tasmota core ESP8266 V.2.7.4.3
  - [ ] The code change is tested and works on core ESP32 V.1.12.2
  - [X] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #16807, opened 2022-10-13
Fix %timerN% rule variables for SunRise/Sunset timers

cybermaus:
Enhance %timerN% rule variables to also work for timers with SunRise/SunSet configuration

## Description:

The %timerN% variables added by #14619 only work with hard-scheduled timers. For sunrise/sunset, it is incorrectly interpreting the offset time as real time. This fix corrects that.

## Checklist:

## Use case

It allows for a very nice paired-timer setup that takes "what if sunrise before on" and "what if sunset after off" into account.
As well as recalculate correct time of day on/off state after power cycle.

While it does require some static rules, it allows to set the desired times using the UI, for causal use by less savvy family members. 

```
# Program the schedule timers in UI with your desired times
# Odd timers 1,3 interpreted by the rules as ON, even timers 2,4 are OFF
# Example times:
#       Timer1 ON  at 06h00   (unless if past timer 2)
#       Timer2 OFF at SunUp   (always)
#       Timer3 ON  at SunDown (unless if past timer 4)
#       Timer4 OFF at 23h00   (always)
# Note, this fails if SunDown past midnight, hence of SunDown <06h00 then SunDown=SunDown+24h00
# Further rules will then calculate the "unless if" part and switch accordingly

Timers 1
Timer1 {"Enable":1,"Mode":0,"Time":"06:00","Window":0,"Days":"1111111","Repeat":1,"Output":2,"Action":3}
Timer2 {"Enable":1,"Mode":1,"Time":"00:00","Window":0,"Days":"1111111","Repeat":1,"Output":2,"Action":3}
Timer3 {"Enable":1,"Mode":2,"Time":"00:00","Window":0,"Days":"1111111","Repeat":1,"Output":2,"Action":3}
Timer4 {"Enable":1,"Mode":0,"Time":"23:00","Window":0,"Days":"1111111","Repeat":1,"Output":2,"Action":3}

Rule2
ON System#Init DO backlog Power1 0; Power1 0 ENDON
ON Time#Initialized DO Backlog var1 0; event checktime=%time% ENDON
ON Clock#Timer DO Backlog var1 0; event checktime=%time% ENDON
ON event#checktime>=%timer1% DO var1 1 ENDON
ON event#checktime>=%timer2% DO var1 0 ENDON
ON event#checktime>=%timer3% DO var1 1 ENDON
ON event#checktime>=%timer4% DO var1 0 ENDON
ON event#checktime DO Power1 %var1% ENDON
Rule2 1
```

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_10_rules.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_10_rules.ino
@@ -1459,6 +1459,18 @@ bool findNextVariableValue(char * &pVarname, float &value)
     uint32_t index = sVarName.substring(5).toInt();
     if (index > 0 && index <= MAX_TIMERS) {
       value = Settings->timer[index -1].time;
+#if defined(USE_SUNRISE)
+      // Correct %timerN% values for sunrise/sunset timers
+      if (Settings->timer[index -1].mode == 1 || Settings->timer[index -1].mode == 2) {
+        // in this context, time variable itself is merely an offset, with <720 being negative
+        value += -720 + SunMinutes(Settings->timer[index -1].mode -1);
+      }
+      if (Settings->timer[index -1].mode == 2) {
+        // To aid rule comparative statements, sunset past midnight (high lattitudes) is expressed past 24h00
+        // So sunset at 00h45 is at 24h45
+        if (value<360) { value += 1440; }
+      }
+#endif  // USE_SUNRISE
     }
 #if defined(USE_SUNRISE)
   } else if (sVarName.equals(F("SUNRISE"))) {


cybermaus:
Oh? I was still modding it with your suggested change. Took a bit of time because I wanted to test on actual device too (even though trivial change)  Does that mean I do not need to commit that change anymore?

arendst:
Yep. Sorry. Thx. All done.

barbudor:
I don't understand
Wasn't this already addressed by my https://github.com/arendst/Tasmota/pull/15230 ?

cybermaus:
Well, during my testing, I did not get any sensible values. And I was not aware of that issue #15230

Reading the code for that issue, it also looks as if (as per title) it does address negative values (it modulates it back it within bounds). And it does seem to add the correct SunMinutes(). 

But it does not interpret the 720 minutes offset that the .time has when it is a sunrise/sunset value, so that would explain why I would not get any sensible values.

barbudor:
ok, it was 6 months ago but I remember spending quite some time to find a good way to handle the way time was coded
thanks

cybermaus:
Well, I do not want to go and upset anyone, but while utterly unaware of your code, I am quite sure I did not get good values before. I will do some more testing, With both the pre and post version

barbudor:
not upset, just surprised as I thought it was well tested 😄

barbudor:
Your changes are in a different part of the rules. Still don't get the difference between those 2 sections

It looks like the code you introduces is very related to the same that I added in `xdrv_09_timer:TimerGetTimeOfDay()`
There is probably some optimization to do here re-using that function to avoid doing the same calculation in 2 different locations.

cybermaus:
Right now, I am on something else. But when I checked earlier, an important difference was that in your code, there was no mention whatsoever of the 720 minutes offset that the .time value has when it is a sun value. Instead, it just wraps around at 0 and 1440. 

I need to check better, but while this may under some circumstanced yield the same (as 720 is mid between) I think it is not always so. But as stated, busy on somewhere else now, will check better later.

cybermaus:
Ok, @barbudor, I have to apologise. Your code works.  Your testing was correct.

_Maybe not relevant, but why did I miss it during testing ... Originally I was on 11.0.0.3, where your code was not in yet. I checked sources, found why I had the issue. Next I did in fact (as per requirement & good practice) upgraded and checked against the latest version, but as the source did not change, I "knew" what I was going to find, and in retrospect I must have retested sloppily. Classic, almost textbook, example of confirmation bias. Sorry._

However....

There is one aspect of your code that is not in line with my use case. And arguable even breaks pre-existing functionality. 
You correct negative numbers, and calculate the correct time. Ok. But you also prevent times higher then 24h00 (1440).

This is relevant as I use these devices up Sweden. Where the sun sets very late. Past midnight. Sometimes not at all 
Your code converts very late (24h45) to very early (00h45). Which makes ">=" comparing very difficult (see use case all way above)

Also, strictly speaking this breaks the already present "permanent daytime" provision from #9543 (admittedly also made by me, but already a while ago). As you can see permanent day is expressed as 2047, which now is shifted to a valid sunset time of 10h07. There was a functional reason why permanent day was shifted to a high number, so >= would never be true.

So I (respectfully, as you are a appreciated contributor, and I am nobody) would argue the code should allow for sunset late night, rather then shifting it to early morning.
At the very least the "permanent daytime" value should be retained, and not converted to 10h07

barbudor:
Once again, there was no critism in my comment (sorry if the shortness appeared as aggressive) and your contributions are as valuable as mine. Even more as they address very specific case where you have experience and I don't.
I was more wondering why my change was not good enough and you gave a perfectly valid explanation and I thank you for that.
I still don't understand yet why the variable replacement is performed in 2 places and why I missed the location where you did your changes (again mainly wondering why my PR was not good enough, nothing to do with yours). I need to dig a little further.

If I undersyand you well, my `xdrv_09_timer:TimerGetTimeOfDay()` is buggy and should be fixed.
May be it would be great that you fox it so that it can be used in both parts of the code and avoid dupplicating thoses calculation ?
What do you think ?
Thanks again

cybermaus:
It should definitely be de-duplicated. And I have no problem taking that on me, given the whole suntime timer thing is important to me.  But I need to study it all a bit better to see how this locations work together.

barbudor:
You can go ahead with the PR and I think it can all go in the same PR

PR #16833, opened 2022-10-15
Adding support multiple GPIO for DS18x20 sensors. Aliases of DS18x20 optimization.

md5sum-as:
## Description:
Adding support multiple GPIO for DS18x20 sensors. 
To enable this function, use the DS18x20_MULTI_GPIOs flag in my_user_config.h or user_config_override.h
Discussion: #16811
Optimization a code in xsns_05_ds18x20.ino and xsns_05_esp32_ds18x20.ino in function Ds18x20Name()

## Checklist:

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
@@ -2,8 +2,7 @@
   xsns_05_ds18x20.ino - DS18x20 temperature sensor support for Tasmota
 
   Copyright (C) 2021  Theo Arends
-  Supplement by md5sum-as (https://github.com/md5sum-as)
-
+ 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
   the Free Software Foundation, either version 3 of the License, or
@@ -16,6 +15,8 @@
 
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+  Updated by md5sum-as (https://github.com/md5sum-as)
 */
 
 #ifdef ESP8266
@@ -59,14 +60,14 @@ struct {
   float temp_sum;
   uint16_t numread;
   uint8_t address[8];
+  uint8_t index;
+  uint8_t valid;
 #ifdef DS18x20_USE_ID_ALIAS
   uint8_t alias;
 #endif //DS18x20_USE_ID_ALIAS  
 #ifdef DS18x20_MULTI_GPIOs
   int8_t pins_id = 0; 
 #endif //DS18x20_MULTI_GPIOs
-  uint8_t index;
-  uint8_t valid;
 } ds18x20_sensor[DS18X20_MAX_SENSORS];
 
 #ifdef DS18x20_MULTI_GPIOs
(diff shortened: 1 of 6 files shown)


tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
```diff
@@ -446,15 +510,14 @@ void Ds18x20Name(uint8_t sensor) {
...
+uint8_t print_ind = sensor +1;
 #ifdef DS18x20_USE_ID_ALIAS
     if (ds18x20_sensor[sensor].alias) {
-      snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("DS18Alias%c%d"), IndexSeparator(), ds18x20_sensor[sensor].alias);
-    } else {
-#endif
-    snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("%s%c%d"), DS18X20Data.name, IndexSeparator(), sensor +1);
-#ifdef DS18x20_USE_ID_ALIAS
+      snprintf_P(DS18X20Data.name, sizeof(DS18X20Data.name), PSTR("DS18Sens"));
+      print_ind = ds18x20_sensor[sensor].alias;
```
barbudor, line 517: Changing from "DS18Alias" to "DS18Sens" is now a breaking change
I know you introduced that recently so probably not much is use yet but just raising the point

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
```diff
@@ -63,8 +65,20 @@ struct {
...
 
+#ifdef DS18x20_MULTI_GPIOs
+struct {
+  int8_t pin = 0;            // Shelly GPIO3 input only
+  int8_t pin_out = 0;        // Shelly GPIO00 output only
+  bool dual_mode = false;    // Single pin mode
+} ds18x20_gpios[MAX_DSB];
+uint8_t ds18x20_ngpio = 0;    // Count of GPIO found
+#endif
+
```
barbudor, line 81: Why introducing a new data structure instead of modulating the size of the existing `DS18X20Data` between `[1]` and `[MAX_DSB]` 
That should simplify the rest of the code

barbudor (commented):
I've put some comments because I think this could be optimized and impact on code size reduced

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
md5sum-as: Yes, I recently added this functionality and it hasn't been released yet. DS18Sens is a more correct name. I will try to avoid such edits. Also fixed aliases output to the accepted style.

tasmota/tasmota_xsns_sensor/xsns_05_ds18x20.ino
md5sum-as: The new structure uses only 12 bytes of RAM. DS18X20Data uses 22 bytes and increasing to 4 copies will add another 66 bytes.
And will also have to pass GPIO numbers to all low-level functions: OneWire1ReadBit(void), OneWire2ReadBit(void), OneWireWriteBit(uint8_t v), OneWireReset(void). Such changes can lead to a significant increase in the final code.

barbudor: That's what I was wondering : passing some parameters to function vs copying

md5sum-as: I made a test module, replaced copying with transferring the index of the gpio array. 
With copying use 687300 bytes of flash.
With passing the parameter to the low-level functions 687364 bytes of flash. 
Loss of 64 bytes.

barbudor: Thanks for trying

arendst:
Set to on hold to merge after next release.

PR #16837, opened 2022-10-15
INA219 enhancement : any shunt value

barbudor:
## Description:

This improve the INA219 driver by allowing any value for the shunt resistor as well as different value for each of the 4 possible sensors. With a bit of refactoring the code increase is limited to 276 bytes.
No impact on settings but a system#boot rule is required for the new feature.
No breaking change for existing settings.

Original legacy configuration (saved in setting):
`Sensor13 0..2` : Default behavior using default shunt resistor of 0.1 ohm, same shunt for every INA219
`Sensor13 10..255` : Allow some other values but limited by a 1-byte encoding, same shunt for every INA219
Improved configuration:
`Sensor13 0, rshunt1, rshunt2, rshunt3, rshunt4` : Allow to specify the shunt resistor in ohm for each INA219
Note: The shunt are NOT saved in flash settings (no cost) so they must be set at each boot using an `on system#boot` rule.

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -598,6 +598,7 @@
     #define USE_VEML6070_SHOW_RAW                // VEML6070, shows the raw value of UV-A
 //  #define USE_ADS1115                            // [I2cDriver13] Enable ADS1115 16 bit A/D converter (I2C address 0x48, 0x49, 0x4A or 0x4B) based on Adafruit ADS1x15 library (no library needed) (+0k7 code)
 //  #define USE_INA219                             // [I2cDriver14] Enable INA219 (I2C address 0x40, 0x41 0x44 or 0x45) Low voltage and current sensor (+1k code)
+  //  #define INA219_SHUNT_RESISTOR (0.100)        // 0.1 Ohm default shunt resistor, can be overriden in user_config_override or using Sensor13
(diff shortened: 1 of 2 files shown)


Longlife4ever:
Hi, is there a way to use a shunt with 0.5 mOhm?
"Sensor 13 10" is the smallest value for 1 mOhm.
I have two nice resistors here that I would like to use.

barbudor:
Yes
Using the improved configuration from this PR
For 0.5 mOhm you need to use:
```
sensor13 0, 0.0005
```
As said this is NOT saved in flash so you must have a boot rule such as
```
Rule1 on system#init do sensor13 0, 0.0005 endon
Rule1 1
```

Looks like I forgot to update the doc. Will that now

Longlife4ever:
ok, thank you very much.
I couldn't find it's the documentary. There was only the possibility over the values

10..255: Define custom shunt resistor encoded as a decimal number RRM such that Rshunt = RR * 10^M milliohm

I guess I overlooked it in this PR. But it's good that it was noticed.
Would be good to describe it at https://tasmota.github.io/docs/Commands/#sensors as well.
I`ll try it 👍

barbudor:
Docs is now updated. See "3)" in https://tasmota.github.io/docs/Commands/#sensors

PR #16848, opened 2022-10-17
Reduce Boards manifests

Jason2866:
## Description:

by using the 4MB default safeboot partition scheme layout. Tasmota will increase the FS partition with the new Autoresize feature introduced with PR #16838 #16842

So it is possible to reduce the boards mainifest since it is no more needed to have different ones for boards with different flash sizes.
Added generic S3 Flash / PSRAM boards variants. Removed all S3 boards specific variants. Configuration is done via Autoconfig.
For a few special use cases maybe an [env] needs to be defined. Examples for are in file `platformio_tasmota_cenv_sample.ini`

@s-hadinger @arendst @staars and all others contributors who wants to please test.
Since it is a major change how the firmwares are build, there is maybe something i overlooked. So i do not check all below.

## Checklist:


Jason2866:
@arendst I use since yesterday and did not run in any error. Imho good to merge.

arendst:
@Jason2866 thx. Works well after changing my board definitions (as you suggested) in my private ``platformio_tasmota_theo_env.ini``.

PR #16863, opened 2022-10-19
Devcontainer

reimerp:
## Description:

Maybe you like to use platformio inside a devcontainer to not mess up local environment
- vscode is req
- Docker is req -- also devcontainer plugin. I thnk if this is not installed nothing will change at all
- untested for windows -- but it should work since devcontainer plugin should handle this
- platformio should not be installed locally (it is installed automatically in container)
- upload from inside container to device is working (if COM5 is removed, its working automatically)
- debug should be possible (its currently blocked by platformio issues) but I will try soon with ftdi

## Checklist:


barbudor:
Already existing
https://github.com/tasmota/docker-tasmota

reimerp:
docker-tasmota is just a pio build insider a docker container, it is cli based. With devcontainer you use everything inside vscode. just press build button and pio is building inside your devcontainer. same for upload.
You can even do source level debug if you add an env for this. I haven't tried it yet for espressif but other platforms. it seems to work currently only with esp32 (and ftdi device, like ft2232)

Jason2866:
This is a no go
`platformio should not be installed locally (it is installed automatically in container)`

reimerp:
Source level debugging / Intellisense / clean local environment might be arguments.
I started building CLI in docker but you degrade an IDE to an editor, there is no intellisense support, since libraries are not added. Its done automatically (yes it should be if you install pio locally -- but then you loose advantage from container support.
Its getting worse if e.g. you needed tools, depending on different python versions ( even 2.7 is needed if you wand to debug esp32 code...
I did not notice speed decrease while compiling, there is some time needed idf containers are outdated.

you can have a local pio installation BUT there is some race condition between plugins. PIO is installing things before the container is ready

In the end its yout choice

Jason2866:
> Its getting worse if e.g. you needed tools, depending on different python versions ( even 2.7 is needed if you wand to debug esp32 code...
I did not notice speed decrease while compiling, there is some time needed idf containers are outdated.

Wrong. I have no Python 2.7 on my system. Debugging with OpenOCD does work for ESP32, ESP32-S2, ESP32-S3 and ESP32-C3. I use the ft2232 and the inbuilt jtag.

And we forked and modified IDF 4.4 to build the Tasmota Arduino framework (changes too). The platformio framework espressif32 is forked and heavily changed too.
Every build component we need is changed.

If you take a closer look you will notice Tasmota uses espressif toolchain patch5.
This new toolchains are not (yet) supported from platformio.

So long story short, we want and need the possibility to change anything from one moment to the other.

PR #16875, opened 2022-10-22
Set accepted filename extensions for firmware uploads

scop:
## Description:

Makes it easier to locate appropriate files to upload.

https://html.spec.whatwg.org/multipage/input.html#attr-input-accept

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
@@ -341,6 +341,14 @@ const char HTTP_FORM_END[] PROGMEM =
   "<button name='save' type='submit' class='button bgrn'>" D_SAVE "</button>"
   "</form></fieldset>";
 
+#ifdef ESP8266
+#define D_UPLOAD_ACCEPT " accept='.bin,.bin.gz,.hex,.ota'"
+#elif defined(ESP32)
+#define D_UPLOAD_ACCEPT " accept='.bin,.hex'"
+#else
+#define D_UPLOAD_ACCEPT ""
+#endif
+
 const char HTTP_FORM_RST[] PROGMEM =
   "<div id='f1' style='display:block;'>"
   "<fieldset><legend><b>&nbsp;" D_RESTORE_CONFIGURATION "&nbsp;</b></legend>";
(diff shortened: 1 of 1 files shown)


s-hadinger:
Nice one, I didn't know this option. However we need to have different values for ESP32 which accepts only `.bin`.

Can you please gate with `#ifdef ESP8266` with the current value `#else` only `.bin`

arendst:
The same upload mechanism is used by other file types like .hex (EFM8BB1 RF-bridge and TasmotaClientfiles) and .ota (Zigbee bridge).

scop:
Ok, maybe something like e6f12ae4db0c272b3e6a45721e77e4120cd0aa68? Limited to ESP8266 and ESP32 there, didn't find `define`s to use for the other mentioned scenarios. And I don't know if this code is being used in too many different scenarios for something like this being easily feasible in the first place.

arendst:
But both ESP8266 and ESP32 need to support .hex too and the ESP32 needs .ota.

sfromis:
Looks like the support for Tube'z Zigbee EFR32 includes uploading a `.gbl` file. Other features might come along with further extensions....

s-hadinger:
> Looks like the support for [Tube'z Zigbee EFR32](https://github.com/arendst/Tasmota/tree/master/tools/fw_TubeZigbee_efr32) includes uploading a `.gbl` file. Other features might come along with further extensions....

Indeed. Actually `.gbl` is the official extension. We renamed as `.ota` for Zbbridge as a convenience. We can use `.ota` for TubeZB as well and drop `.gbl`

scop:
I don't know how the mentioned devices (EFM8BB1 RF-bridge and TasmotaClient, Zigbee bridge, Tube'z Zigbee EFR32) translate to `#define`s we could use here, but 857b3be1c733fa7fa57a637fc290b9b3230f65f5 updates to follow the note in https://github.com/arendst/Tasmota/pull/16875#issuecomment-1288076012.

If this still isn't quite right, perhaps it would be more productive if someone who actually knows what they're doing could either enumerate the complete set of defines and corresponding extensions for me to add, or go ahead and just push the fixes to my branch here (it's open for contributors), rather than me continuing with the guesswork :)

arendst:
unstale

arendst:
Thx!

Just merged an adapted version of your code.

PR #16941, opened 2022-10-28
Use accumulating energy instead of instant power for ADE7953

Flavien:
## Description:

Instead of using the instant power registers ([`AWATT`, `AVA`, `AVAR`](https://github.com/arendst/Tasmota/blob/c7a864231ab2af3597f58bcbe188a54495bbc4e9/tasmota/tasmota_xnrg_energy/xnrg_07_ade7953.ino#L169), etc.) to determine active, reactive and apparent power:
- We use accumulating energy registers (`AENERGYA`, `APENERGYA`, `RENERGYA`, etc.).
- These registers are read once per second inside the `Ade7953EnergyEverySecond` > `Ade7953GetData` function.
- The energy accumulated over one second (in watt second) is converted to watts to obtain **the average power over the last second**.
- These registers are read-with-reset registers. This means that the contents of these registers are reset to 0 after a read operation.
- This will allow the accumulated energy variable reported by Tasmota to be a lot more accurate because the energy registers have a 4.83 μs sampling time - handled in hardware (versus a 1 second sampling time in the current approach based on accumulating instant power every second).
- This will allow a much more precise calculation of the total energy consumed especially in situations where current varies rapidly.
- I have tested this build over the last couple of days and there was no issue.

### Caveat:
Users on the current version of Tasmota will have to recalibrate their power.
- They can use `PowerSet` with a known load to recalibrate.
- Or they can divide their `PowerCal` by `23.41494` (this was determined empirically and may vary slightly case by case).

## Checklist:

### DIFF
--- a/tasmota/tasmota_xnrg_energy/xnrg_07_ade7953.ino
+++ b/tasmota/tasmota_xnrg_energy/xnrg_07_ade7953.ino
@@ -274,7 +274,7 @@ void Ade7953Write(uint16_t reg, uint32_t val) {
 }
 
 int32_t Ade7953Read(uint16_t reg) {
-	uint32_t response = 0;
+  uint32_t response = 0;
 
   int size = Ade7953RegSize(reg);
   if (size) {
@@ -306,7 +306,7 @@ int32_t Ade7953Read(uint16_t reg) {
     }
 #endif  // USE_ESP32_SPI
   }
-	return response;
+  return response;
 }
 
 #ifdef ADE7953_DUMP_REGS
@@ -441,6 +441,11 @@ void Ade7953GetData(void) {
     acc_mode, reg[0][4], reg[1][4], reg[0][5], reg[1][5],
     reg[0][0], reg[1][0], reg[0][1], reg[1][1], reg[0][2], reg[1][2], reg[0][3], reg[1][3]);
 
+  // If the device is initializing, we read the energy registers to reset them, but don't report the values as the first read may be inaccurate
+  if (Ade7953.init_step) {
+    return;
+  }
+
   uint32_t apparent_power[2] = { 0, 0 };
   uint32_t reactive_power[2] = { 0, 0 };
 
@@ -495,14 +500,15 @@ void Ade7953GetData(void) {
 }
 
 void Ade7953EnergyEverySecond(void) {
-	if (Ade7953.init_step) {
+  if (Ade7953.init_step) {
     if (1 == Ade7953.init_step) {
       Ade7953Init();
-	  }
+      Ade7953GetData();
+    }
     Ade7953.init_step--;
-	}	else {
-		Ade7953GetData();
-	}
+  } else {
+    Ade7953GetData();
+  }
 }
 
 /*********************************************************************************************/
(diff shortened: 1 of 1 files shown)


Flavien:
**Note**: I am not able to test on ESP32 V.2.0.5, I would greatly appreciate if someone is able to do it.

arendst:
Thx. Will do some testing before merge.

isabido:
Thanks Flavien for this improvement...of course this Shelly EM with TASMOTA is at a much higher level than Shelly's original firmware.

At the moment, after several weeks of accumulated active energy, I continue with the tasmota energy totalizer vs. the electric company with the same value.

arendst:
Implemented just now in latest dev release with some adjustements. Thx.

Flavien:
> No need to re-calibrate as I calculate the difference in the code.

@arendst I had a look at your changes. Are you sure this is done correctly? You only make the adjustment in the `PowerSet` command but not in the `Ade7953GetData` function, so whatever value the users have configured with `PowerSet` prior to the upgrade is based on the old LSB ratio and will be used as is in the `Ade7953GetData` function.

arendst:
It works great. See line 485 which calculates the correct value.

Tried on all known shelly ade9753 devices and works as expected.

Flavien:
Yep, I missed that line, thanks @arendst, that looks great.

isabido:
Hello colleagues!, Since this last improvement, I am observing that the active energy totalizer is out of phase (accumulates less, tasmota) with respect to that of the electric company. I have to give it more days of accumulation, but previously it was perfect.

Please Flavian, watch it carefully. I have not changed anything in the configuration, only update.

Flavien:
@isabido If you use the Tasmota calibration commands, the calibration should continue to work seamlessly, however if you use `Rule3` to calibrate using the gain registers, you will need to multiply your power gain registers by `1.335296202`.

For reference, the math of this is: `4,862,401 / 206.9 kHz * (2.5 / 44) = 1.335296202`

isabido:
> 
I am indeed using rule3, here is the current value.

Rules":"on file#calib.dat do {\"rms\":{\"current_a\":4800000,\"current_b\":4869356,\"voltage\":1596500},\"angles\":{\"angle0\":200,\"angle1\":200},\"powers\":{\"totactive\":{\"a\":3100000,\"b\":3100000},\"apparent\":{\"a\":3100000,\"b\":3100000},\"reactive\":{\"a\":3100000,\"b\":3100000}}} endon"}}

If I have not misunderstood you, would it be like this?

3100000 x 1,335296202 = 4139418

Thanks for your answer!

Flavien:
Yes that's correct @isabido.

isabido:
> Yes that's correct @isabido.

Perfect Flavien, I have already made the change, we are going to observe it for a few days to see how the totalizers evolve, previously after weeks of accumulation the result was precise.

In your case, how are you working with rule3 or tasmota commands?

Flavien:
I am using `Rule3` as it lets me calibrate everything separately. I am calibrating using my utility meter as a reference which gives be both active and reactive energy.

isabido:
Thank you Flavian for your answer. You are lucky to have a company meter that registers the reactive energy in my case alone, it registers positive values of reactive (inductive) and my reactive loads are capacitive.

image

image

In any case, now with the new applied factor that you have recommended to me, the instantaneous values of the three powers W VA VAR do not coincide according to the mathematical calculations, it gives me the feeling... that it will not total correctly.

PR #16961, opened 2022-10-31
Berry: Expose the ULP on ESP32S2 and ESP32S3 (RISCV version)

Staars:
## Description:

Initially I thought about adding the FSM version too, but after playing with some examples I think, this not useful and would create (even more) confusion about the ULP usage.
This RISCV ULP is programmed in C, which unsurprisingly is a ton easier to do than the assembly code on the FSM versions of the ULP.
Binaries can only be created by building an ESP-IDF project and then using https://github.com/Staars/berry-examples/blob/main/ulp_helper/binS2Berry.py or the JS app in https://tasmota.github.io/docs/ULP/#export-from-esp-idf-project.
There are no real API changes on the Tasmota/Berry side, but in the ESP-IDF we have some differences between ESP32,..S2 and ..S3, which beside other major things include some enums. The major changes are abstracted away in the Berry bindings.
In my tests "everything" like ADC, GPIO, (bitbanged) I2C and OneWire did work as expected.

## Checklist:

### DIFF
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -166,7 +166,7 @@ BERRY_LOCAL const bntvmodule* const be_module_table[] = {
 #ifdef USE_ALEXA_AVS
     &be_native_module(crypto),
 #endif
-#if defined(USE_BERRY_ULP) && defined(CONFIG_IDF_TARGET_ESP32)
+#if defined(USE_BERRY_ULP)
     &be_native_module(ULP),
 #endif // USE_BERRY_ULP
 #if defined(USE_MI_ESP32) && !defined(USE_BLE_ESP32)
--- a/lib/libesp32/berry_tasmota/src/be_ULP_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_ULP_lib.c
@@ -6,9 +6,9 @@
 #include "be_constobj.h"
 #include "be_mapping.h"
 
-#if defined(USE_BERRY_ULP) && defined(CONFIG_IDF_TARGET_ESP32)
+#if defined(USE_BERRY_ULP) &&  (defined(CONFIG_IDF_TARGET_ESP32) || defined(CONFIG_IDF_TARGET_ESP32S2) || defined(CONFIG_IDF_TARGET_ESP32S3))
 
-#include "esp32/ulp.h"
+// #include "esp32/ulp.h"
 #include "driver/rtc_io.h"
 #include "driver/gpio.h"
 #include "driver/adc.h"
@@ -37,8 +37,6 @@ BE_FUNC_CTYPE_DECLARE(be_ULP_sleep, "", "[i]");   // optional int arg
 extern void be_ULP_load(struct bvm *vm, const uint8_t *buf, size_t size);
 BE_FUNC_CTYPE_DECLARE(be_ULP_load, "", "@(bytes)~");   // pass: 1/ vm, 2/ bytes point, 3/ bytes size
 
-#include "be_fixed_ULP.h"
-
 /* @const_object_info_begin
 module ULP (scope: global) {
   run,          ctype_func(be_ULP_run)
@@ -51,5 +49,6 @@ module ULP (scope: global) {
   adc_config,   ctype_func(be_ULP_adc_config)
 }
 @const_object_info_end */
+#include "be_fixed_ULP.h"
 
 #endif // USE_BERRY_ULP
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_ulp.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_ulp.ino
@@ -1,5 +1,5 @@
 /*
-  xdrv_52_3_berry_ulp.ino - Berry scripting language, ULP support for ESP32
+  xdrv_52_3_berry_ulp.ino - Berry scripting language, ULP support for ESP32, ESP32S2, ESP32S3
 
   Copyright (C) 2021 Stephan Hadinger & Christian Baars, Berry language by Guan Wenliang https://github.com/Skiars/berry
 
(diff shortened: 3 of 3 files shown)


s-hadinger:
Awesome job!

Staars:
Here is an example of board specific base driver, which uses the ULP to offload some work to the ULP, by reading and decoding an analog button array.
Sadly on the ESP32-S3-Eye the accelerometer shares the I2c bus with the camera, so it is not possible to also handle the accelerometer completely on the ULP while also using the cam. But I have a driver version, that can handle the button array and the accelerometer outside the main cores - without the camera.
```
#-
 - S3-Eye
 - Example of a board specific driver written in Berry
 -
 - Support for the analog button array and the QMA7981 accelerometer
 - Button readings completely done on the RISCV ULP
 - I2C via ULP would only work without the cam, because pin access must be exlusive :(
 -#

 class S3Eye : Driver
    var wire          #- if wire == nil then the module is not initialized -#
    static button = ["UP","DOWN","PLAY","MENU"]
    var accel
  
    def init()
        self.wire = tasmota.wire_scan(0x12)
        if self.wire
            print("EYE: init")
            var v = self.wire.read(0x12,0,1)
            if v != 231 return end  #- wrong device -#
            # activate accelerometer
            self.wire.write(0x12,0x36,0xb6,1)
            self.wire.write(0x12,0x36,0,1)
            var data = self.wire.read(0x12,0x11,1)
            data |= 128;
            self.wire.write(0x12,0x11,data,1)
            self.accel = [0,0,0]
            # init button reader for an analog array of 4 button, that change the voltage on GPIO 1 in 4 steps
            # the ULP will measure the voltage and calculate the button index including a flag for a change of the button value
            import ULP
            ULP.wake_period(0,100 * 1000)
            var c = bytes().fromb64("bwDgARMAAAATAAAAEwAAAIKAAAAAAAAAAAAAAAAAFxEAABMBIf69IKEgvSgBoEERgUUBRQbGZSAjLKAWEwcQHIVHY1OnAhMHUDuJR2NepwAFZ5MG95WNR2PYpgATB1e4gUdjQ6cAkUeyQD6FQQGCgEERBsZtPyMuoBYDJ0AYYwflACMioBgFRyMg4BiyQAFFQQGCgKFmg6dGEDcHwP19F/mPI6L2EIKAoWcDp0cQt0bA//0WdY+3xg8AVY8joucQA6dHELcGAAJVjyOi5xADp0cQtwZAAFWPI6LnEAGghUezlbcAhWf9F/2NwgW1Z8GBQRGThweAzgUV6dhHtwYIgP0WdY/ZjczHuEM6whJHWYMTd/cPdfvYR4F2/RZ1j9jH2Ee3BgIAVY/YxxWgmFu3BgiA/RZ1j9mNjNuYW4F2/RZ1j5jbmFu3BgIAVY+Y2xHt2EdBgwWLZd8Z6dxHPsQDVYEAQgVBgUEBgoCYW923nFs+xgNVwQD1tw==")
            # Length in bytes: 376
            ULP.load(c)
            ULP.run()
            ULP.adc_config(0,3,3) # did not work, when applied before ULP.run()
        else
          print("EYE: I2C error. Pin config uses I2C bus 2 for cam?")
        end
    end

    def read_accel()
        if !self.wire return end  #- exit if not initialized -#
        var b = self.wire.read_bytes(0x12,0x1,6)
        b[0] = b[0] & 252
        b[2] = b[2] & 252
        b[4] = b[4] & 252
        self.accel[0] = (b.geti(0,2))/16
        self.accel[1] = (b.geti(2,2))/16
        self.accel[2] = (b.geti(4,2))/16
    end


    #- trigger a read every second -#
    def every_second()
      self.read_accel()
    end

    def every_100ms()
        import ULP
            if ULP.get_mem(96)               # button updated?
                var b = ULP.get_mem(95)      # button index calculated by the ULP 1-4, 0 is no button pressed
                if b > 0
                    print(self.button[b-1])  # trigger something from here
                end
                ULP.set_mem(96,0)
            end
    end
  
    #- display sensor value in the web UI -#
    def web_sensor()
      if !self.wire return nil end  #- exit if not initialized -#
      import string
      var msg = string.format(
               "{s}QMA7981 acc_x{m}%.3f G{e}"..
               "{s}QMA7981 acc_y{m}%.3f G{e}"..
               "{s}QMA7981 acc_z{m}%.3f G{e}",
                self.accel[0]/1000.0, self.accel[1]/1000.0, self.accel[2]/1000.0)
      tasmota.web_send_decimal(msg)
    end
  
    #- add sensor value to teleperiod -#
    def json_append()
      if !self.wire return nil end  #- exit if not initialized -#
      import string
      var msg = string.format(",\"QMA7981\":{\"AX\":%i,\"AY\":%i,\"AZ\":%i}",
      self.accel[0], self.accel[1], self.accel[2])
      tasmota.response_append(msg)
    end
  
  end
  s3eye = S3Eye()
  tasmota.add_driver(s3eye)
  ```

**Comment by s-hadinger** [comment]
We'll need to find an alternative way to load the code. The following line

```
            var c = bytes().fromb64("bwDgARMAAAATAAAAEwAAAIKAAAAAAAAAAAAAAAAAFxEAABMBIf69IKEgvSgBoEERgUUBRQbGZSAjLKAWEwcQHIVHY1OnAhMHUDuJR2NepwAFZ5MG95WNR2PYpgATB1e4gUdjQ6cAkUeyQD6FQQGCgEERBsZtPyMuoBYDJ0AYYwflACMioBgFRyMg4BiyQAFFQQGCgKFmg6dGEDcHwP19F/mPI6L2EIKAoWcDp0cQt0bA//0WdY+3xg8AVY8joucQA6dHELcGAAJVjyOi5xADp0cQtwZAAFWPI6LnEAGghUezlbcAhWf9F/2NwgW1Z8GBQRGThweAzgUV6dhHtwYIgP0WdY/ZjczHuEM6whJHWYMTd/cPdfvYR4F2/RZ1j9jH2Ee3BgIAVY/YxxWgmFu3BgiA/RZ1j9mNjNuYW4F2/RZ1j5jbmFu3BgIAVY+Y2xHt2EdBgwWLZd8Z6dxHPsQDVYEAQgVBgUEBgoCYW923nFs+xgNVwQD1tw==")
```

will unfortunately force the Berry runtime to keep the string in memory forever, while we only need the binary version.

Moving to a static class attribute should do the trick, the long string should be garbage collected quite soon

Staars:
GrußChristianAm 31.10.2022 um 10:22 schrieb s-hadinger ***@***.***>:﻿
We'll need to find an alternative way to load the code. The following line
            var c = bytes().fromb64("bwDgARMAAAATAAAAEwAAAIKAAAAAAAAAAAAAAAAAFxEAABMBIf69IKEgvSgBoEERgUUBRQbGZSAjLKAWEwcQHIVHY1OnAhMHUDuJR2NepwAFZ5MG95WNR2PYpgATB1e4gUdjQ6cAkUeyQD6FQQGCgEERBsZtPyMuoBYDJ0AYYwflACMioBgFRyMg4BiyQAFFQQGCgKFmg6dGEDcHwP19F/mPI6L2EIKAoWcDp0cQt0bA//0WdY+3xg8AVY8joucQA6dHELcGAAJVjyOi5xADp0cQtwZAAFWPI6LnEAGghUezlbcAhWf9F/2NwgW1Z8GBQRGThweAzgUV6dhHtwYIgP0WdY/ZjczHuEM6whJHWYMTd/cPdfvYR4F2/RZ1j9jH2Ee3BgIAVY/YxxWgmFu3BgiA/RZ1j9mNjNuYW4F2/RZ1j5jbmFu3BgIAVY+Y2xHt2EdBgwWLZd8Z6dxHPsQDVYEAQgVBgUEBgoCYW923nFs+xgNVwQD1tw==")

will unfortunately force the Berry runtime to keep the string in memory forever, while we only need the binary version.
Moving to a static class attribute should do the trick, the long string should be garbage collected quite soon

Staars:
Good to know 👍 .
I guess, I have some work to do regarding docs and axamples 😀 .

s-hadinger:
Would it be possible to have non-ULP compile fine even if `#define USE_BERRY_ULP` is enabled? This would simplify compilation options across multiple targets.

Currently it fails for ESP32C3

PR #17016, opened 2022-11-07
fix for deepsleep first calculation #13955

reimerp:
Signed-off-by: Reimer Prochnow <reimer-github@ideenhal.de>

## Description:

## Checklist:

I made a fix for the first calculation of the deepsleep time, which was discussed in #13955 and is working since a couple of month for me now.
The deepsleep interval is untouched, but the very first time is prolonged to ensure that it's always after the desired timestamp, eg: 3600 -> ~ 3 minutes after xx:00 and never before xx:00, since this is confusing almost any tool I used (homeassistant energy dashboard, grafana)
300 -> xx:01, xx:06, xx:11 ....

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_29_deepsleep.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_29_deepsleep.ino
@@ -118,11 +118,16 @@ void DeepSleepPrepare(void)
     RtcSettings.deepsleep_slip = tmin(tmax(RtcSettings.deepsleep_slip, 9000), 11000);
   }
 
+  AddLog(LOG_LEVEL_DEBUG, PSTR("DSPrep: time %ld, next %ld, slip %ld"),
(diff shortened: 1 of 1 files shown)


smhc:
FYI this change appears to result in a "DeepSleepTime 86400" to wake at 01:12 instead of 00:00 due to the 5% difference. This conflicts with the documentation (which is already a bit inconsistent stating that 10 years is the max, but 24 hours is also the max...) and also makes trying to calculate a "wake at this specific time" even more difficult.

reimerp:
I have not verified this since I don't use this max sleep times -- but at least with 300s / 2600s (which I commonly use) the first trigger time is after the desired time ( depending on when exactly the deepsleep was issued) -- but it is autocorrected for following intervals, so there is no summary of errors. But maybe the adjustment should be constant rather that a fraction on the time ?

PR #17092, opened 2022-11-15
add command RgxPort to setup port forwarding

joba-1:
## Description:

Add command RgxPort to setup port forwardings for range extender clients.
I.e. ports on the clients can be reached by using "gateway" ports on the range extender.
You can forward TCP and UDP ports: `RgxPort [tcp|udp], gateway_port, client_mac, client_port`
RgxPort can be used multiple times for different forwardings but obviously the gateway ports must be otherwise unused.

Example:
* Assume some client (tasmota or not) with mac 4CEBD677F760 runs a webserver on port 80 
* The client is connected to the AP of the tasmota range extender rex1 and got ip 192.168.4.2
* The range extender is accessible to the outside world with its ip, here 192.168.1.239
* On rex1 issue command `RgxPort TCP, 8080, 4cebd677f760, 80`
* The result is `OK TCP 192.168.1.239:8080 -> 192.168.4.2:80` or `ERROR` if something went wrong
* Now you can access the webserver from outside the nat network with the new url http://rex1:8080

Currently only ESP32 is supported

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9 (compile only)

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
@@ -94,6 +94,8 @@ const char kDrvRgxCommands[] PROGMEM = "Rgx|" // Prefix
 #ifdef USE_WIFI_RANGE_EXTENDER_NAPT
                                        "|"
                                        "NAPT"
+                                       "|"
+                                       "Port"
 #endif // USE_WIFI_RANGE_EXTENDER_NAPT
 #ifdef ESP32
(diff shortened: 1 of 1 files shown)


joba-1:
done. 

A nice next step would be if this command could be used without the device already connected.
But currently I have no good idea how to achieve this (how to store wanted forwards, how to detect new clients).
Local dynamic forwarding list? Some event callback for new clients available? Polling client list?

Issue #15785, opened 2022-06-12
Range Extender stops routing


dsteinkopf:
### PROBLEM DESCRIPTION
"Sometimes" the range extender stops routing. The Wifi is still open, clients stay connected, others can newly connect. But packets are not routed anymore.
I experienced some correlation between that problem and change of the bssid that tasmota hotspot is using: Shortly after it changed its bssid, other devices connected to this range extender lost tcp connection. But I am not able to reproduce the problem.
BTW: Number of clients connected to the range extender is less than 5-6.

### REQUESTED INFORMATION

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): Shelly 2.5
- [x] Tasmota binary firmware version number used: 11.1.0.3
  - [ ] Pre-compiled
  - [x] Self-compiled
- [x] Flashing tools used: OTA update (minimal then mine)
- [x] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
07:22:58.572 MQT: stat/kleinelampe_335E29/RESULT = {"NAME":"Shelly 2.5","GPIO":[320,0,32,0,224,193,0,0,640,192,608,225,3456,4736],"FLAG":0,"BASE":18}
07:22:58.792 MQT: stat/kleinelampe_335E29/RESULT = {"Module":{"0":"Shelly 2.5"}}
[...]

PR #17103, opened 2022-11-16
preserve AP connections if STA side reconnects

joba-1:
## Description:

Until now wifi mode was always reset to STA-only mode on STA reconnects.
This destroyed AP connections and port forwardings. Should be fixed now.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9 (compile only)


joba-1:
hm, ci error seems unrelated. How to fix/retry?

joba-1:
this probably fixes issue #15785 as well...

Jason2866:
@joba-1 
setHostname needs to be set after wifi mode is set. WifiRangeExtender sets wifi mode without setHostname. The hostname will probably not be set.

sfromis:
Specifically about setting Wifi mode vs naming, when I configure a device with ESP8266 as range extender, the AP name popping up is ESP-xxxxxx, instead of the configured range extender SSID. Works when it is ESP32 instead.

joba-1:
@Jason2866: haven't looked at hostname, I configure a static IP for the extender in my router along with a hostname and the device seems to pick this up. It ends up in my syslog messages. Where is it missing?

@sfromis: Must be something specific about your setup. I test with an ESP32 and an ESP8266. Both show up with the same AP SSID (configured once with RgxSsid) or my test device would not connect.

My intention was, not to change any existing code path, except not tearing down AP when it is already set up. So my expectation is, STA is setup as usual, including set hostname. I'll check again...

sfromis:
What I posted was details leading to the post by Jason2866, as a possible explanation for the issue I'm seeing recently, related to another issue with the ESP-xxxxxx AP name popping up when it shouldn't.

My setup is a very straightforward ESP8266 range extender build, with the range extender part configured with:
`Backlog RgxSSID xxxx; RgxPassword xxxx; RgxAddress 10.99.1.1; RgxSubnet 255.255.255.0; RgxState ON; RgxNAPT ON`

joba-1:
ok, but that was the only change that touched core tasmota parts - so pointing at this pr is at least plausible. 
Anyways, as you know, logs help :) But I will check myself anyways .

sfromis:
Well the as the range extender produces no logs when starting, there is basically nothing to report.
`{"Rgx":{"Valid":"true","SSId":"xxxx","Password":"xxxx","IPAddress":"10.99.1.1","Subnetmask":"255.255.255.0"}`
And scanning wifi only gives ESP-xxxxxx

Jason2866:
@joba-1  Your touch off the code introduced the sending out hostname esp-xxxx only,
beside the fact that this PR broke Tasmota in general. Sfromis noted and Theo fixed.
Just pointed out what needs to be done to send out a hostname different from esp-xxxx
Fix it or not, up to you. There are no logs needed, issue is clear and related to your change(s).

sfromis:
This is the commit which made non range extender ESP8266 builds work without opening up the ESP-xxxxxx AP. While this made such devices work normally again, it did not make the range extender builds work as they should. https://github.com/arendst/Tasmota/commit/69f903618da484b586860301aa1faeb5c494ea5a

joba-1:
@sfromis, I understand what this commit does (revert my change for all but range extenders by preprocessor commands). I also understand there are no logs, but it would help to understand how you get to the ESP* AP. I don't see it. I rebooted the device and it connects immediately to my SSID and opens an AP like setup with RgxSsid. I execute WifiScan and it just says: not started (probably also not ok). But I do not see an ESP* AP. 
I can imagine (but didn't follow relevant code paths yet) that my change also has an impact on switching APs or SSIDs if they need to go through a wifi disconnect to do their thing. Unfortunately that would mean they tear down any port forwarding, so for range extenders not ideal. On the other hand, a range extender that gets quietly torn down every now and then as it was before my change is also not ideal.

@Jason2866: you sound a bit pissed when you state it would be a fact that I broke tasmota in general (which I obviously did not, in general it just works fine, even better than before for my usecases). I can just offer my apologies, I did take care to think about all I could and tested quite a bit - but still, not enough. Sorry.
Wifi state handling in Tasmota is not trivial, lets just work together and make it work as good as possible for range extenders while not disturbing others. Theo's commit is a good starting point, but I hope, not the end.

Jason2866:
@joba-1  Sorry for sounding harsh, was not ment this way!

sfromis:
Well, I did nothing but what I already stated, do a build for ESP8266, and make the config I posted. Going from a 2022-11-09 build to 2022-11-17 build, I got the lots of bad AP names. I did not notice it for RGX at first, due to the flooding with masses of unexpected ESP8266 AP names popping up, and was suspecting that the device I used RGX for was so weak in signal strength that it was not a good test case. Hence I reconfigured without RGX to make it work, and only returned to checking what happened with RGX after the AP flooding issue got fixed.

(I've not got any experience with how the coding around Wifi works, or where the critical spots are).

My setup was deliberately pretty simple, and I chose to use NAPT mode to avoid any additional router config, and the D1 Mini I run on is doing nothing else. Did you test with NAPT both ON and OFF? Suggestion for things to try?

joba-1:
So far I only tested with NAPT ON. I figured it shouldn't make a difference (and it probably doesn't).

I think I have an improvement and do testing now.

My previous PR overlooked that WifiManager also uses AP_STA mode, not just AP (somehow I only found mesh as AP_STA user and assumed it is incompatible with range extender anyways and ignored it).

So, WifiBegin() needs to distinguish who started the AP_STA and I'm not sure how that should be done cleanly in tasmota.  Currently I use a function RgxApUp() added to xdrv_58_range_extender.ino and use it in support_wifi.ino.

```diff
diff --git a/tasmota/tasmota_support/support_wifi.ino b/tasmota/tasmota_support/support_wifi.ino
index 692822384..c294c4c48 100644
--- a/tasmota/tasmota_support/support_wifi.ino
+++ b/tasmota/tasmota_support/support_wifi.ino
@@ -221 +221 @@ void WifiBegin(uint8_t flag, uint8_t channel)
-  if (WiFi.getMode() != WIFI_AP_STA) {  // Preserve range extender connections (#17103)
+  if (WiFi.getMode() != WIFI_AP_STA || !RgxApUp()) {  // Preserve range extender connections (#17103)
diff --git a/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino b/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
index 88631bb27..3868dde46 100644
--- a/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
@@ -151,0 +152,6 @@ TRgxSettings RgxSettings;
+// externalize to be able to protect Rgx AP from teardown
+bool RgxApUp()
+{
+  return RgxSettings.status == RGX_CONFIGURED || RgxSettings.status == RGX_SETUP_NAPT;
+}
+
```

I'll make sure my testing includes using WifiManager this time...

PR #17112, opened 2022-11-17
Add DS18x20 parasitic power usage on ESP32

curzon01:
## Description:

DS18x20 parasitic power usage doesn't work on ESP32, sensors are detected but always showing invalid 85° C.

For using DS18x20 parasitic power on ESP32 use `#define W1_PARASITE_POWER` optimization like on ESP82xx in relation with `USE_DS18x20`.

Testet with multiple DS18B20/DS18S20

## Checklist:


arendst:
Thx.

PR #17124, opened 2022-11-17
Avoid crash if malloc fails

s-hadinger:
## Description:

Avoid crash in `ext_vsnprintf_P` when `malloc` fails.
**Related issue (if applicable):** fixes #17101

## Checklist:


sfromis:
Looks a lot like this commit breaks data from energy monitoring on ESP32. The front page has only the unit, all values are missing. And for two-channel devices, the formatting of the JSON is also broken, like `"Total":,11.07128]` with the first value missing. Reverting to the previous build, the values are intact, and presented correctly again.

Looks like a failing detection of fail is causing it to bail out when it should not.

s-hadinger:
Thanks, I will investigate. I have reverted for now.

PR #17140, opened 2022-11-19
Add SerialStart, SerialIgnore commands for Serial Bridge (both Hardware and Software)

ortegafernando:
## Description:

Hi, I have done two new commands that are very usefull with SerialBridge. (both commands work with hardware or software serial bridge)

If this PR is merged, I will add a new PR for "tasmota docs"

**SerialStart**: it is configured to work up to 3 bytes. Buffer will begin to record only when this 1, 2 or 3 bytes are recieved in that order

**SerialIgnore (0 .. 255)**: usefull when you have a serial TX that send date every X seconds (in my case a Victron VE.Direct equipment that sends data every second) and you want to exclude data not to saturate your system. 

Both commans' configuration are preserved between restarts. Configuration is saved in flash

Also, I have added code to distinguish between **D_JSON_SSERIALRECEIVED and D_JSON_SERIALRECEIVED** in SerialBridge (hardware vs software serial)
CAUTION: before this change, every message is recieved as D_JSON_SSERIALRECEIVED, so people using hardware serial must check if they manage well D_JSON_SERIALRECEIVED 

Compiled and checked with gitpod.io

Examples:
  // SerialStart 0 to disable
  // SerialStart C1 to start only with C1
  // SerialStart C1,C2 to start with C1, C2
  // SerialStart C1,C2,C3 to start with C1, C2, C3
  // SerialStart show current chars
  // C1, C2, C3 must me decimal numbers representing ascii code ("PID" = 80 73 68):   serialstart 80,73,68

  // SerialIgnore 0 to disable => process all recieved messages
  // SerialIgnore 1 is identical as SerialIgnore 0
  // SerialIgnore X to "ignore" X serialrecieved messages
  // SerialIgnore show current value

Output message from gitpod.io before and after this PR:
```
RAM:   [=====     ]  50.9% (used 41676 bytes from 81920 bytes)
Flash: [======    ]  62.9% (used 644360 bytes from 1023984 bytes)

RAM:   [=====     ]  50.9% (used 41676 bytes from 81920 bytes)
Flash: [======    ]  63.0% (used 644952 bytes from 1023984 bytes)
```

## Checklist:

### DIFF
--- a/tasmota/include/i18n.h
+++ b/tasmota/include/i18n.h
@@ -365,6 +365,8 @@
 #define D_CMND_SERIALBUFFER "SerialBuffer"
 #define D_CMND_SERIALSEND "SerialSend"
 #define D_CMND_SERIALDELIMITER "SerialDelimiter"
+#define D_CMND_SERIALSTART "SerialStart"
+#define D_CMND_SERIALIGNORE "SerialIgnore"
 #define D_CMND_BAUDRATE "Baudrate"
 #define D_CMND_SERIALCONFIG "SerialConfig"
 #define D_CMND_TEMPLATE "Template"
--- a/tasmota/include/tasmota_types.h
+++ b/tasmota/include/tasmota_types.h
@@ -731,7 +731,8 @@ typedef struct {
   uint16_t      artnet_universe;           // 734
   uint16_t      modbus_sbaudrate;          // 736
 
-  uint8_t       free_esp32_738[5];         // 738
+  char          serialstart[4];            // 738  Serial Start Chars
+  uint8_t       serialignore;              // 73C  Serial recieved messages to ignore
 
   uint8_t       novasds_startingoffset;    // 73D
   uint8_t       web_color[18][3];          // 73E
--- a/tasmota/include/tasmota_version.h
+++ b/tasmota/include/tasmota_version.h
@@ -20,6 +20,6 @@
 #ifndef _TASMOTA_VERSION_H_
 #define _TASMOTA_VERSION_H_
 
-const uint32_t VERSION = 0x0C020005;   // 12.2.0.5
+const uint32_t VERSION = 0x0C020006;   // 12.2.0.6
 
 #endif  // _TASMOTA_VERSION_H_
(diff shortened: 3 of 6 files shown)


arendst:
Thx.

As it stands this won't be merged as it changes default behaviour in case of hardware/software serial. This should not be distinguished here in the code. Although it uses hardware serial, it uses a different driver than the inbuilt serial driver and hence should still report as ``sserialreceived``.

Anyway, I keep this PR open as an idea of how to manage duplicate messages.

BTW how often are the serial messages being received? If it is every 5 seconds (as in your tests) there is no reason to handle it here. You should do it in your back ground code and not in Tasmota.

ortegafernando:
> Thx.
> 
> As it stands this won't be merged as it changes default behaviour in case of hardware/software serial. This should not be distinguished here in the code. Although it uses hardware serial, it uses a different driver than the inbuilt serial driver and hence should still report as `sserialreceived`.
> 
> Anyway, I keep this PR open as an idea of how to manage duplicate messages.
> 
> BTW how often are the serial messages being received? If it is every 5 seconds (as in your tests) there is no reason to handle it here. You should do it in your back ground code and not in Tasmota.

HI @arendst , I understand what you say. so I am going to delete anything about D_JSON_SERIALRECEIVED.

About SerialStart and SerialIgnore I think they are good commands. In my case I am receiving one big message every seconds, that is why I want to ignore several (duplicate or not, I only want a message every 5 or 10 seconds or even more seconds)

arendst:
SerialIgnore is now implemented in Serial Bridge as ``SetOption35 0..255``

ortegafernando:
> SetOption35

THanks, do you want me to edit the wiki page?

barbudor:
> > SetOption35
> 
> Thanks a lot, do you want me to edit the wiki page?

That would be very kind
Thanks

ortegafernando:
Hi, PR done for wiki docs (developement branch).

Whay about SerialStart? I hope you can find it also usefull.

ortegafernando:
@arendst Bot has closed this PR. It is only pending the SerialStart stuff. I think it is a good way to improve hardware and software serial (I have read several old diccusions about this topic). DO you think is time to re-opening it again? Thanks.

PR #17184, opened 2022-11-24
BLE eQ3 - allow to set retries at runtime

fightforlife:
## Description:

This PR introduces a new parameter in the eQ3 thermostat driver and allows the user to set the number of connection retries. Curently is is hardcoded to 4.
This is needed if the thermostats are far away and not always respond correctly after 4 tries.

This PR will still set the number of retries to 4 at boot. But at runtime the user can change the number of retries with command "trvretries". (or autoexec.bat or rules)
In my testing I had no problems with retries up to 30.

## Checklist:


fightforlife:
I have this running with 30 ble retries max on Version 12.2.0.5. So far 16 days of uninterrupted uptime.

PR #17427, opened 2022-12-17
remove manual upload port and switch to autodiscovery in platformio

blakadder:
## Description:
Remove manual port entry as default and use Platformio port discovery as default.

Needed setting`common.upload_port` to a MAC default so the CI doesn't break

Option to set upload port is now in `platformio_override.ini` to keep user configurations out of default files. Being empty enables port discovery on my Windows system.

Tested on different boards. Does not work if multiple flashable devices are connected.

## Checklist:


Jason2866:
CI Fails on Mac

platformio_override_sample.ini
```diff
@@ -48,6 +48,10 @@ default_envs            =
 
 
 [env]
+; *** Define serial port used for erasing/flashing/terminal
+upload_port                 = COM10
+monitor_speed               = 115200
```
Jason2866, line 53: Why override `monitor_speed` ?

blakadder:
doesn't matter, mac is not playing along as always

platformio.ini
```diff
@@ -35,10 +35,10 @@ custom_unpack_dir           = unpacked_littlefs
 build_unflags               = ${core.build_unflags}
 build_flags                 = ${core.build_flags}
 monitor_speed               = 115200
-monitor_port                = COM5
+; monitor_port                = COM5
```
Jason2866, line 38: Why not remove?

blakadder: i dunno, cause mac might have some objections

platformio_override_sample.ini
```diff
@@ -113,7 +113,6 @@ lib_extra_dirs          = ${library.lib_extra_dirs}
 ;board_build.partitions  = partitions/esp32_partition_app2944k_fs2M.csv
 monitor_speed           = 115200
 ; *** Serial port used for erasing/flashing the ESP32
-;upload_port             = ${common.upload_port}
 upload_port             = COM4
```
Jason2866, line 116: Not consistent. Should be `upload_port = `

platformio.ini
Jason2866: It is only a comment.

PR #17432, opened 2022-12-18
Fix upload_port is mandatory to compile custom build on Linux

barbudor:
## Description:

Upload port setting is now mandatory to compile (even without upload), at least on Linux
Custom builds fails because they use the upload_port from platformio_override(_sample).ini which are empty
PR makes them inherit from the default env 

## Checklist:


Jason2866:
@barbudor The idea from @blakadder was to make autoport finding of plaltformio possible. This is gone by this PR. Practically it reverts the PR.
Since custom env are already customized, why not adding the `upload_port` there or do your override of the `upload_port` just in your `platformio_override.h`.
It is made for such settings.

I do find this PR wrong.

PR #17447, opened 2022-12-19
Fix on init procedure

stefanbode:
## Description:

#17403

## Checklist:


pawol:
I have tested shuttermotorstop on the rest of controllers and everywhere was the value 65535. After setting onto 500. Each roller controller works

stefanbode:
> 65535. After setting onto 50

Thanks for the information. I checked on my devices and always got a "0" in the case of a missing initialization. @arendst : What is the default for the config parameters if not used? "FFFFFF" or "0000000"

arendst:
On a new install it's 0. On an updated system it may be anything as settings may have been moved around within the settings area over time. I suggest to test for any value outside your valid value range.

PR #17626, opened 2023-01-07
WIP Tuya MCU Bridge driver alternative to the TuyaMCU driver

dgwynne:
I wrote this because I want to operate the PWM LEDs on a fan from the ESP chip rather than leave it on the MCU. Tasmota offers far more control than what the MCU offers. However, with the current TuyaMCU driver I cannot use it and the PWM light driver at the same time. I wanted a much simpler and less integrated TuyaMCU driver to work with.

I am also sending this pull request in mostly as a backup of my work, I'm not going to get upset if you (Tasmota) are not interested in this or the idea of pushing such a huge breaking change on your userbase. You might find some of the code interesting though.

## Description:

The main difference is this driver does not try and wire MCU data points (Dps) into the tasmota power/light/etc controls. Instead, each Dp ends up being relayed directly to MQTT and the rules subsystem. If you want to change the state of something wired up to the MCU, you send tuyamcu specific commands to manipulate the Dp.

Each Dp gets a type and id specific topic that is sent to MQTT. eg, Dp id 1 type bool looks like tele/%topic%/TUYAMCUBOOL1. To change state you send a TuyaMCUBool1 command (the command is actually TuyaMCUBool and the index value is used as the DpId, which is nice and symmetrical) with the new value.

Currently Rules operate on TuyaMCU#TypeDpid things, eg, "rule1 on TuyaMCU#Bool1 do power %value% endon" updates the state of the power of something configured on the tasmota device when the state of the thing on the MCU changes too.

Dp endpoints and their values are remembered by the driver, and only changes to the values are relayed. Only bool, enum, and value types are currently handled.

As a more concrete example of how this works, the fan I was trying to integrate with is a Ventair Skyfan DC with a built in CCT light. The fan comes with a rf433 remote that's wired up the MCU. An "interesting" feature of the remote is that even though it has buttons to toggle the state of the fan and light, and up/down buttons for each of the fan and light, the remote seems to send it's idea of the state rather than the individual button presses for the MCU to update the state with. eg, the fan state gets toggled on the remote, and then the remote sends "the fan is on" or the "fan is off" rather than "toggle the fan". This is interesting because you can just take the MCU reported values and update the state of the light on the tasmota device rather than try and sync the tasmota state with the MCU.

So, here are the Dps on the fan:

```
tele/tasmota_756B78/TUYAMCUBOOL1 0
tele/tasmota_756B78/TUYAMCUENUM2 0
tele/tasmota_756B78/TUYAMCUVALUE3 0
tele/tasmota_756B78/TUYAMCUENUM8 0
tele/tasmota_756B78/TUYAMCUBOOL15 0
tele/tasmota_756B78/TUYAMCUVALUE16 0
tele/tasmota_756B78/TUYAMCUENUM19 0
tele/tasmota_756B78/TUYAMCUENUM22 0
```

DpId 1 is fan power
DpId 3 is fan speed (values 1 to 5 inclusive)
DpId 8 is the fan direction (forward/reverse)
DpId 15 is the light power
DpId 16 is the light dimmer level (1 to 5 inclusive)
DpId 19 is the CCT value (0 to 2 inclusive)
DpId 22 is a timer thing

I can't tell what DpId 2 is. I wish it was a toggle for the onboard buzzer on the fan, but changing it doesn't do anything I can see.

After snipping the MCU connections to the LEDs and patching pins from the ESP chip into their place, I can use rules like this:

```
SetOption 20
Rule1
  on TuyaMCU#Bool15=0 do power off endon
  on TuyaMCU#Value16=1 do backlog dimmer 20; power on endon
  on TuyaMCU#Value16=2 do backlog dimmer 40; power on endon
  on TuyaMCU#Value16=3 do backlog dimmer 60; power on endon
  on TuyaMCU#Value16=4 do backlog dimmer 80; power on endon
  on TuyaMCU#Value16=5 do backlog dimmer 100; power on endon
  on TuyaMCU#Enum19=0 do CT 153 endon
  on TuyaMCU#Enum19=1 do CT 327 endon
  on TuyaMCU#Enum19=2 do CT 500 endon
```

## Todo

- better relaying of the wifi/mqtt status to the MCU
- handling wifi reset requests from the MCU
- low power stuff?
- support for sending status updates and device info queries
- supporting the raw and string Dp types
- restarting the tuya mcu state machine?
- restarting the rx state machine when no bytes are rxed for a period of time
- time sync

## Checklist:


arendst:
Thx. Will merge once I finish some tests. Nice to see different ways of C programming resulting in the same result (meaning programming style).

Regarding esp32c3 serial, it should work just as well. Unless you expect more than one serial interface (over the console, either serial or CDC) as the C3 only has one additional UART and tasmota does not support software serial on esp32.

tasmota/tasmota_xdrv_driver/xdrv_65_tuyamcubr.ino
```diff
@@ -0,0 +1,916 @@
...
+
+	dp->dp_value = XdrvMailbox.payload;
+
+	data.h.dpid = dp->dp_id;
+	data.h.type = dp->dp_type;
+	data.h.len = htons(dt->t_len);
+	dt->t_wr(data.value, dp->dp_value);
+
+	tuyamcubr_send(sc, TUYA_CMD_SET_DP, &data, len);
+
```
arendst, line 502: Pls change TUYA_CMD_SET_DP to TUYAMCUBR_CMD_SET_DP

tasmota/tasmota_xdrv_driver/xdrv_65_tuyamcubr.ino
```diff
@@ -0,0 +1,916 @@
...
+#ifndef nitems
+#define nitems(_a)		(sizeof((_a)) / sizeof((_a)[0]))
+#endif
+
+#ifndef CTASSERT
+#define CTASSERT(x)		extern char  _ctassert[(x) ? 1 : -1 ]	\
+				    __attribute__((__unused__))
+#endif
+
+#define TUYAMCUBR_LOGNAME		"TuyaMCUBr"
```
arendst, line 38: Pls shorten log name to three character in line with Tasmota logging. Suggesting "TYB"

vic42:
Not sure this file is contained intentionally https://github.com/arendst/Tasmota/pull/17626/files#diff-ce9dde7afe3598171690658547081fc8e27885c5903a1c33465e07f0f00318f4

arendst:
Yes, that file shouldn't be there.

dgwynne:
> Thx. Will merge once I finish some tests. Nice to see different ways of C programming resulting in the same result (meaning programming style and punch card width!).
> 
> Regarding esp32c3 serial, it should work just as well. Unless you expect more than one serial interface (over the console, either serial or CDC) as the C3 only has one additional UART and tasmota does not support software serial on esp32.
> 
> UPDATE: Just did a test on an ESP32-C3 without CDC and that works fine. A test on an ESP32-C3 with CDC fails. It fails on ANY configured serial device, not just yours so I have an issue with ESP32-C3 and CDC I need to sort out.
> 
> UPDATE2: Regarding ESP32-C3 with CDC I did some more tests. Appears when selecting GPIO18/19 it doesn't work indeed most probably because that's where my device has connected it's USB connection. If I try other GPIO's (from 6 to 21 excl. 18/19) serial devices work fine including yours. What hardware ESP32-C3 with CDC did you try? I tried the cripple Lolin C3-mini (which seems not to be so crippled in my case).
> 
> UPDATE3: Yep. Checked the schematics and indeed the CDC/USB interface is connected to GPIO18/19 so that explains why it didn't work in the first try.

I was trying on a luatos board (https://templates.blakadder.com/luatos_CORE-ESP32.html) with GPIO8 and 9 as TuyaMCUBr Rx and Tx respectively, GPIO12 is LedLink, and GPIO18 and GPIO19 for PWM channels. I'll move the PWM pins and give it a try.

dgwynne:
> Not sure this file is contained intentionally https://github.com/arendst/Tasmota/pull/17626/files#diff-ce9dde7afe3598171690658547081fc8e27885c5903a1c33465e07f0f00318f4

it's just a bit of defense. the tuyamcubr code and the mielhvac code both use these macros. if the order they're included changes then the preprocessor might get upset.

tasmota/tasmota_xdrv_driver/xdrv_65_tuyamcubr.ino
dgwynne: done

PR #17724, opened 2023-01-16
Tuya - WIFI_STRENGTH + Save check on MCU_CONF

barbudor:
## Description:

**Related issue (if applicable):** fixes https://github.com/arendst/Tasmota/discussions/17625

On a Tuya device for which the non-ESP module has been replaced by an ESP12F
- Adding support for command GET_WIFI_STRENGTH (0x24) that was not yet implemented
- Adding some fail safe check on command MCU_CONF when mode 2 provides GPIO for LED_1 and KEY_1 (Button_1) that are not compatible for ESP8266 (but was ok for the original non-ESP MCU).
Before the PR, the GPIO were blindly enforced, leading to a FRC and configuration reset because of incompatibility with flash pins.

Fail safe adds warning message if either the provided GPIO is illegal or is already mapped to something else than LED1/KEY1
Exemple message for a MCU_CONF message providing GPIO9 for LED1 and GPIO6 for KEY1.
```
23:52:56.791 TYA: In use or illegal gpio9 for LED1, ignored
23:52:56.792 TYA: In use or illegal gpio6 for KEY1, ignored
```
This message will mean that some more hacking is required to route the LED1 and KEY1 signal to other GPIOs that are safe for ESP8266. LED_1 and Button_1 can then be set directly in the template and Tuya conf will be ignored.

## Checklist:


SteWers:
@barbudor I had a lot of trouble while using GPIO9 with a user-template after your commit https://github.com/arendst/Tasmota/pull/17724/commits/fa94a3b0f656d00a1cf0b67d855f95c9c93cc2cb. The device (ESP 8285) is running for years and I only did an update to 12.3.1.5.
The additional check for `USER_MODULE == Settings->module` causes, that GPIO9 cannot be used. It can be solved with SO51, but there is no hint to be aware of that.
When the version will be released, many users will have the same trouble as me. Please rethink the implementation of the additional check. Maybe there will be another solution to avoid the use of GPIO9/10 accidentally.

barbudor:
Hi @SteWers 
Could you please provide me with more details ?
What device ? What template ?

On other devices the lack of this test lead to boot loop with configuration resets which let the user totally in the blind on what was happenning.

Thanks

SteWers:
Hi @barbudor,
of course. It is a quiet simple setup. The device is a Sonoff Dual R2 (39). Because of free GPIOs I use this device to communicate via RS485 with my Solax inverter. GPIO9 is connected to the RTS line. The definition is done via template, so `Settings->module` equals `USER_MODULE`. Because of this GPIO9 (and 10) is not usable. Here is the template:
`{"NAME":"Dual R2","GPIO":[0,3488,0,3520,0,225,7776,32,224,544,0,0,0,0],"FLAG":0,"BASE":39}`
Perhaps it is meanwhile possible to check the ESP Chip ID rather then the configured module.
BR

barbudor:
Oh
It's not related to Tuya at all ?

I have to think about it
Thanks for reporting.

SteWers:
No. It isn't related to Tuya. That's a good point. Maybe the check could be combined with the use of Tuya.

barbudor:
@SteWers PR #17835 should solve your problem.
Please let me know if you can test when it is merged

SteWers:
> @SteWers PR #17835 should solve your problem. Please let me know if you can test when it is merged

The code looks good. I'll test it and let you know.

SteWers:
Hi @barbudor,
works like a charm. 👍
Thanks!

PR #17775, opened 2023-01-23
TuyaMCUBr: support on/true/off/false/toggle in the tuyamcubool command.

dgwynne:
## Description:

I wanted a tasmotized wall switch to be able to blindly send "toggle" to a fan/light and have it do the right thing. The dp value is kept by the driver, so it can easily read, modify, and write it.

I'm hardcoding the "on"/"off"/"toggle" values to compare against, should I be getting them from a macro instead?

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_65_tuyamcubr.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_65_tuyamcubr.ino
@@ -586,18 +586,15 @@ tuyamcubr_cmnd_data_bool(void)
 		return;
 	}
 
-	if (strcasecmp(XdrvMailbox.data, "off") == 0 ||
-	    strcasecmp(XdrvMailbox.data, "false") == 0 ||
-	    strcmp(XdrvMailbox.data, "0") == 0)
-		value = 0;
-	else if (strcasecmp(XdrvMailbox.data, "on") == 0 ||
-	    strcasecmp(XdrvMailbox.data, "true") == 0 ||
-	    strcmp(XdrvMailbox.data, "1") == 0)
-		value = 1;
-	else if (strcasecmp(XdrvMailbox.data, "toggle") == 0 ||
-	    strcmp(XdrvMailbox.data, "2") == 0)
+	switch (XdrvMailbox.payload) {
+	case 0:
+	case 1:
+		value = XdrvMailbox.payload;
+		break;
+	case 2:
 		value = !dp->dp_value;
-	else {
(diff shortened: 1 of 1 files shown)


barbudor:
> I'm hardcoding the "on"/"off"/"toggle" values to compare against, should I be getting them from a macro instead?

The values are overridable by the user using command `StateText` so you could get the actual value from calling `GetStateText(...)` which takes as argument:
- 0 for OFF
- 1 for ON
- 2 for TOGGLE
- 3 for HOLD

Only numerics, there are no macro defined for 0/1/2/3

dgwynne:
> > I'm hardcoding the "on"/"off"/"toggle" values to compare against, should I be getting them from a macro instead?
> 
> The values are overridable by the user using command `StateText` so you could get the actual value from calling `GetStateText(...)` which takes as argument:
> 
> * 0 for OFF
> * 1 for ON
> * 2 for TOGGLE
> * 3 for HOLD
> 
> Only numerics, there are no macro defined for 0/1/2/3

Does that work in addition to on/true, or instead of?

arendst:
As user input use ``GetStateNumber(const char *state_text)`` which returns standard Tasmota numbers. Look for the function name in the code for it's translations:

```
const char kOptions[] PROGMEM = "OFF|" D_OFF "|FALSE|" D_FALSE "|STOP|" D_STOP "|" D_CELSIUS "|DOWN|" D_CLOSE "|"   // 0
                                "ON|" D_ON "|TRUE|" D_TRUE "|START|" D_START "|" D_FAHRENHEIT "|" D_USER "|"  // 1
                                "TOGGLE|" D_TOGGLE "|" D_ADMIN "|"                                            // 2
                                "BLINK|" D_BLINK "|"                                                          // 3
                                "BLINKOFF|" D_BLINKOFF "|"                                                    // 4
                                "UP|" D_OPEN  "|"                                                             // 100
                                "ALL" ;                                                                       // 255

const uint8_t sNumbers[] PROGMEM = { 0,0,0,0,0,0,0,0,0,
                                     1,1,1,1,1,1,1,1,
                                     2,2,2,
                                     3,3,
                                     4,4,
                                     100,100,
                                     255 };
```

dgwynne:
> As user input use `GetStateNumber(const char *state_text)` which returns standard Tasmota numbers. Look for the function name in the code for it's translations:

Looks like that's called as part of the `XdrvMailbox` setup, so I'm just using the payload value to switch between behaviours.

PR #17804, opened 2023-01-28
Add KS0429 TDS driver

Theosakamg:
## Description:

Add TDS (Total Dissolved Solids) KS0429 driver.

## Checklist:


barbudor:
This driver doesn't blend into the Tasmota current architecture of having all analog sensors inside `xsns_02_analog.ino`
As such
- It doesn't make it obvious to the user that the GPIO must be configured on a GPIO that supports Analog
- I believe it doesn't allow to be mixed correctly with other analog sensors

The current analog driver may be a bit messy and would probably appreciate some refactoring but is it safe to start this way ?

PR #17930, opened 2023-02-11
update: Added airm2m esp32c3 core device

xe5700:
## Description:

Added airm2m esp32c3 core device

## Checklist:


s-hadinger:
Thanks for the PR. I would rather have an autoconfig file instead of a module. @arendst 
Thoughts?

arendst:
Agree with @s-hadinger. Modules are passe at the cost of code space and can easily be implemented as autoconfig files.

sfromis:
Looks like this could better simply have been standard templates, does not even need autoconf features

Jason2866:
There is no need to add this C3 Board in code. The device can be configured using a Template.

PR #17990, opened 2023-02-19
Update xdrv_79_esp32_ble.ino

AleXSR700:
Increase maximum length of Alias as the short MaxAlias is not in line with name length and limits the user (e.g. if sensors have room names as appendix like "_master_bathroom").

## Description:

Self encountered issue that the short alias length limits the user in Home Automation environments. Although such a length is always a bit arbitrary, 32 chars is also in line with name length.

## Checklist:
  - [x ] The code change is tested and works with Tasmota core ESP32 V.2.0.6

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
@@ -204,7 +204,7 @@ namespace BLE_ESP32 {
 /////////////////////////////////////////////////////
 
 #define BLE_ESP32_MAXNAMELEN 32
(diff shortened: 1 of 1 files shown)


barbudor:
As this is even not stored in flash, it may be just more simpler to use a String that use dynamic memory and won't have any limit (but RAM)
If you're not feeling confortable with such a change, let's go for that for now and I'll look into it later

AleXSR700:
> As this is even not stored in flash, it may be just more simpler to use a String that use dynamic memory and won't have any limit (but RAM)
That would be even better. I thought that maybe such a change could cause issues elsewhere. But dynamic adjustment via console would be even better. We could align the default value with name length, though, so set it to 32 anyway.

PR #18008, opened 2023-02-20
Set correct OTA URL for all Tasmota build env 

Jason2866:
## Description:

The `OTA_URL` is now set in the PIO env. The attempts to do where not working in the past since the define (which is a Macro) was extended wrong.

By using the syntax `-D DEFINE_NEEDED='"string"'` the Macro is parsed and extended correctly.

All `#define OTA_URL` are removed in source code!
All [env) need to have `-DOTA_URL='"path_to_ota_file"'` or an empty entry `-DOTA_URL='""'`

This Option makes it possible to define more settings without the need of `user_config_override.h`.

## Checklist:


barbudor:
Nice 👍

PR #18035, opened 2023-02-23
Add speed and heading info to GPS driver

pkkrusty:
Added NAV-VELNED messages to driver, allowing speed and heading info on the main web page and JSON messages. Probably could be optimized; mostly another exercise in copy/paste. Also reduced threshold for movement to 100 from 1000. Testing indicated this was more reasonable when horizontal accuracy was 5m or less (which is normal for non-obstructed sky)

## Description:

## Checklist:
  - [ x] The code change is tested and works with Tasmota core ESP32 V.2.0.7


arendst:
Edge case considering how many portable tasmotas are out there. What's the impact on code size and ram usage? If too much use defs to save resources for non-portable users.

pkkrusty:
Good point. I'll put changes into a #define.

pkkrusty:
Adjusted code so requires #define USE_GPS_VELOCITY. 0k4 code space. Maybe 40 bytes RAM (not sure).

DouglasRudd:
G'Day.
I am developing a device that will be used inside a company vehicle while it is being used within our warehouse yard to track and monitor movement and speed, the yard has 100% wifi coverage on our local Lan, I have tried the Unofficial Bin from the web installer but the value for the lon: has the alt: appended to the end of it, I tried to recompile from both the Main and the Dev branch and both of them do not return the lat: value, when I try to compile with #define USE_GPS_VELOCITY. there is an error ( missing the leading " at line 921). I have been trying to debug this my self and have found that for some reason 
on line 901, dtostrfd((double)UBX.rec_buffer.values.lat/10000000.0f,7,lat); is rendering a null.  any help would be appreciated. 
thanks

arendst:
Latest dev (v14.1.0.2) fixes the compile issue. I see this with regards to lat:

image

sfromis:
I noticed that the "GPS Speed" field lacks a unit. Ideally, it could align with what can be configured via the command SpeedUnit. Sure, that is mentioned as being for wind speed, but the units should be perfectly fine for transportation speed too.

PR #18117, opened 2023-03-05
Create WS_epaper27B_display.ini

aoe1:
Config to use a Waveshare WFI0190CZ22 2.7" display with dev firmware and a patch to utilise :f full screen commands (https://github.com/arendst/Tasmota/compare/development...aoe1:Tasmota:patch-1) (don't forget to set the Waveshare HAT driver resistor to 0.47R (A) as per https://www.waveshare.com/wiki/E-Paper_ESP32_Driver_Board#Hardware_Connection)

## Description:

## Checklist:


gemu2015:
while this will work (with the modified udisplay)  it is not the intended mode of operation.
the section :f is used for displays with full and partial update to switch between both modes.
your display is full update mode only.
so you simply can add the :f data to the :I section. then there should be no need to modify udisplay.cpp

also the descriptor has a typo:

:Bb0,0

aoe1:
You are right, it works with just the ini sequence, too. What is tricky is the reset sequence, I've only managed manual reset, not via RST pin - I suspect it might be the timing reset vs init, I will experiment further and submit a more conservative epaper 2.7 .ini in the next few days


-------- Original Message --------

PR #18121, opened 2023-03-05
Refactored Berry rule engine and support for arrays

s-hadinger:
## Description:

Refactored Berry rule engine and support for arrays - with hopefully no regression.

The refactored version moves more processing when creating the rule, less when running the rule matching. Also reduces pressure on the GC.

Now supports index for arrays.

Ex: `ENERGY#Power[1]` matches `100` in `{"Energy":{"Power":[80,100,120]}}` 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 - Shelly Pro 4PM using standard MCP23xxx driver and needs one time Auto-Configuration
 
 ### Changed
+- Refactored Berry rule engine and support for arrays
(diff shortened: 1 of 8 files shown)


sfromis:
While this is a nice improvement, people may get confused about a trigger from a classic rule not working the same in Berry. To get the 2nd value of 100, the classic rule trigger would be `ENERGY#Power[2]`.

s-hadinger:
Ah yes, I missed that. Let me fix it.

s-hadinger:
Fixed in #18123

PR #18148, opened 2023-03-10
remove absolute url

benderl:
## Description:
The "Manage File System" page uses an absolute URI including wifi ip to switch between flash and sd card. If your device uses an ethernet port and has wifi turned off or not configured, the link target resolves to `http://0.0.0.0/ufsd?`.

This PR changes the target to only use a relative URI without ip. So the browser will use the ip or hostname of the current page.

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.7

I have only a ESP32 dev board and cannot test on ESP8266 hardware, but the change is minimal and should also be valid for this platform.


Jason2866:
@bend @s-hadinger imho it would be better to fix the IP `0.0.0.0` to the one which is used and valid
Dont know the side effects (breaking change?) by this PR

s-hadinger:
In principle I agree that not specifying the IP is better. Unless it brakes something (which I don't see why it would), I approve.

Jason2866:
Let us see if issues arrives ;-)

PR #18210, opened 2023-03-17
ESP32 GPS NTP Server

NoroNetwork:
Hello 

https://tasmota.github.io/docs/GPS-NTP-server/

Is it posible to add suport for ESP32 and fast NTP server response? 
And it is posible to do Sensor60 9 Automaticly? Because even if i do rule it is ignored....

This library was build for ESP8266 witch is slow (1s update interval of Time)

ESP32 should have enough power to update every milisecond atleast... 

Thanks :)


shamankayka (approved):

shamankayka (approved):
CHANGELOG.mdCHANGELOG.md

arendst:
Not relevant

PR #18214, opened 2023-03-17
Support for pmsa003i

FunJp:
## Description:

Add support for the PMSA003I I2C air quality sensor

## Checklist:


arendst:
Thx. Looks almost ok.

Pls put the global vars in a struct to solve possible duplications.

FunJp:
Hey,

I assume that this was what you had in mind about all the global variables ? If you also would prefer to also have the followings in the struct just let me know.

`bool pmsa003i_type = false;
bool pmsa003i_ready = false;
uint8_t warmup_counter;  // count for warmup`

Also, while looking thru other sensor, I notice that the xsns_18_pms5003.ino could also benefits from using a similar struct. I won't touch it in this PR, but it might be worth a look too at some point

arendst:
Yeah I like to have them all in a struct but then do not call it ``data`` as being the most obvious variable name....

Call it related to your driver like ``Pmsa003i`` which is most probably unique within the tasmota global variable world.

arendst:
Thx.

PR #18295, opened 2023-03-29
Shutter esp32 support up to 16 shutters

stefanbode:
## Description:
- Enhance shutter to new ESP32 capabilities
- Refactor ShutterButtons
- Increase shutters up to 16
- introduce tilt change with buttons
- maintain 5x 6x 7x... button press with different reconfig options
- setoption74 1 is required to decouple relays and buttons (maybe this needs to be a default
- removed "simultanous hold" event on buttons
- shutterbuttons and MQTT integrations needs testing
- 

## Checklist:

PLEASE DO NOT PUSH TO MAIN NOW. I want to check with others if additional changes are required

### DIFF
--- a/tasmota/tasmota_support/support_command.ino
+++ b/tasmota/tasmota_support/support_command.ino
@@ -917,24 +917,8 @@ void CmndStatus(void)
   }
 
 #ifdef USE_SHUTTER
-  if (Settings->flag3.shutter_mode) {
-    if ((0 == payload) || (13 == payload)) {
-      Response_P(PSTR("{\"" D_CMND_STATUS D_STATUS13_SHUTTER "\":{"));
-      for (uint32_t i = 0; i < MAX_SHUTTERS; i++) {
-        if (0 == Settings->shutter_startrelay[i]) { break; }
-        if (i > 0) { ResponseAppend_P(PSTR(",")); }
-        ResponseAppend_P(PSTR("\"" D_STATUS13_SHUTTER "%d\":{\"Relay1\":%d,\"Relay2\":%d,\"Open\":%d,\"Close\":%d,"
-                                   "\"50perc\":%d,\"Delay\":%d,\"Opt\":\"%s\","
-                                   "\"Calib\":[%d,%d,%d,%d,%d],"
-                                   "\"Mode\":\"%d\"}"),
-                                   i, Settings->shutter_startrelay[i], Settings->shutter_startrelay[i] +1, Settings->shutter_opentime[i], Settings->shutter_closetime[i],
-                                   Settings->shutter_set50percent[i], Settings->shutter_motordelay[i], GetBinary8(Settings->shutter_options[i], 4).c_str(),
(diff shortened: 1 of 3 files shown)


arendst:
Compile for ESP8266 fails because the return from ButtonHandler should be abool.

Pls add the correct false/true at the correct location.

stefanbode:
ok, this was not the plan :-) Interesting method to include and exclude with the IF statement. Never seen before.

arendst:
It allows you and testers to continue developing while users still have the default driver on esp32

JosefRypacek:
Do we have any expectation when USE_SHUTTER_ESP32 will be default / removed from the code and users will be able to use pre-builded binaries?

stefanbode:
I will Check with Theo. We have no more bugs. The only thing I have concerns is that the config save is different to the normal procedure. You have to save the shutter cfg from the file system too

arendst:
I'll enable the new driver for the esp32 today.

The additional config file save, or actually the restore is indeed an issue. I need to dive into this as the same is true for the esp32 energy driver. As far as I know the current solution of restoring the additional config file won't work as it will be overwritten by the current default config.

On my plate.

stefanbode:
Hi Theo, if we TAR all files on esp32 and then using the config backup and restore to handle the TAR file this should be not much work. It will also be future prove if more config files appear.

stefanbode:
Saving the config file and restoring does work for the shutter. There is no accidental overwrite. This is tested.

arendst:
Yes. I'm thinking of a tar like solution too but there are some catches. I want it to be initiated by the current config save and restore process as it needs a restart after restore. I also only want to save .xdrvXX and .xsnsXX files. I again want it to be transparent for filesystems so it needs to work on esp8266 too.

All in all I need time to come up with a decent, small code, solution.

As you said the current shutter file restore works. This will make enabling the new esp32 driver easier. In that case I will enable it ASAP but not before tomorrow as I'm out today.

arendst:
> Do we have any expectation when USE_SHUTTER_ESP32 will be default / removed from the code and users will be able to use pre-builded binaries?

As of v13.0.0.2 it is released and define USE_SHUTTER_ESP32 is removed.

JosefRypacek:
Nice! Yesterday I've flashed my 9 devices which were probably on 12.4.0.5 custom build with `USE_SHUTTER_ESP32` and probably also with defined Wi-Fi credentials.

5 of them booted with default configuration and I had to connect to Wi-Fi AP and setup everything.

The remaining 4 booted successfully, but without shutter configuration.
So I set shutters, tilt, buttons and 2 of them rebooted with default configuration and I had to connect to the AP too.

Now I have 9 functional devices and shutters works as expected.

stefanbode:
Sorry this is definitely not as it should work and was tested. The shutters should copy the actual settings. Also the boot in default is very strange. @arendst theo anything you have an idea?

arendst:
Without a log it's hard to say what went wrong. 

I do see strange code with regards to ``ShutterSettingsDelta()`` (wrong versions).

I'm still rewriting the driver config save stuff so for now do not touch the shutter driver until I'm done.

PR #18298, opened 2023-03-30
Update xnrg_01_hlw8012.ino

Superpfusch:
I am not shure if more changes are needed like changing the datatype to float? but i realised while researching the measurement capabilitys of smartplugs that there is some information loss through first averaging and then dividing. I got it running in my version with 0.1V steps, but there are a lot for the normal user unpleasant changes which would clutter this change proposal.

## Description:

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.7

### DIFF
--- a/tasmota/tasmota_xnrg_energy/xnrg_01_hlw8012.ino
+++ b/tasmota/tasmota_xnrg_energy/xnrg_01_hlw8012.ino
@@ -153,7 +153,7 @@ void HlwEvery200ms(void) {
       DigitalWrite(GPIO_NRG_SEL, 0, Hlw.select_ui_flag);
 
       if (Hlw.cf1_pulse_counter) {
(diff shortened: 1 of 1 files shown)


arendst:
Thx for your observation.

As this will change current calibrated values I won't implement this.

The HLW sensor isn't the best around anyways so let's keep the observed 0.3V step for now.

maurerle:
I think, this would only have a floating point precision impact.

Instead of dividing through `cf1_voltage_pulse_length` (which is the same as `cf1_pulse_length`), it multiplies with the reciprocal value.
Which should be the same, except for floating point calculation errors.

The intent of this PR was to increase the accuracy of voltage measurements - so a matching name of this PR would be very useful.

If this really helps, I would like to look into this, as it probably does not need any kind of recalibration, right?
But first, I think it is needed to double-check the current behavior.

PR #18310, opened 2023-03-31
make IRHVAC send incremental changes.

TMaYaD:
## Description:

This change will allow mqtt messages simpler by allowing something like `publish cmnd/ir-bridge/IRHVAC {"Power": "On", "Incremental": true}`.
The simpler format is required for working with mqtt clients like iotMQTTPanel or similar on android.

Limitations:
- Requires full message be sent at-least once. I work around this with a `system#boot` rule
- Gets messy with multiple devices. I have only one AC per room, and this doesn't doesn't interfere with other IR devices.

TODO:
- [ ] Persist `irac_prev_state` across reboots.
- [ ] Support multiple devices.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.7

### DIFF
--- a/tasmota/include/i18n.h
+++ b/tasmota/include/i18n.h
@@ -561,6 +561,7 @@
     #define D_JSON_IRHVAC_STATE_MODE_SEND_ONLY "SendOnly"
     #define D_JSON_IRHVAC_STATE_MODE_STORE_ONLY "StoreOnly"
     #define D_JSON_IRHVAC_STATE_MODE_SEND_STORE "SendStore"
+  #define D_JSON_IRHVAC_INCREMENTAL "Incremental"
 #define D_JSON_IRRECEIVED "IrReceived"
 
 // Commands xdrv_06_snfbridge.ino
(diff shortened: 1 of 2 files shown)


TMaYaD:
I do not have an ESP32 IR Blaster to test.

s-hadinger:
Thanks. I would rather have this in the standard build rather than a compile option. Can we think of a distinct command syntax for incremental? For ex adding a json field like "incremental":true?

TMaYaD:
My original idea was to leverage `Mem<x>` (or similar) as stores. 

`{..., "Storage": "Mem1"}` would store and/or retrieve from `Mem1`. This will allow incremental operations for multiple devices, if desired.

However, I have a lot of learning to do before I understand the code base enough to do that.

s-hadinger:
Relying on Mem1 sounds overly complicated to me. Why not go back to your original idea, but as a runtime option, not a compile-time option?

Jason2866:
@TMaYaD  Will you do the changes?

TMaYaD:
Yes. I'm away on vacation right now. Will get back to it after 14th.

TMaYaD:
@Jason2866 Updated and tested on my IR blaster. It turned out to be simpler than I expected.

s-hadinger:
bump

pkkrusty:
Nice improvement. Looking forward to testing this. Will make node-red flows simpler too.

PR #18413, opened 2023-04-14
enhanced/fixed SDM72 module

jdelker:
## Description:

SDM72 smart meter support in tasmota is currently somewhat limited, considering that only two values (current power & total energy) are provided.
This change uses additional modbus registers to enhance the available metrics:

grafik

## Checklist:

### DIFF
--- a/tasmota/tasmota_xnrg_energy/xnrg_18_sdm72.ino
+++ b/tasmota/tasmota_xnrg_energy/xnrg_18_sdm72.ino
@@ -40,13 +40,28 @@
 TasmotaModbus *Sdm72Modbus;
 
 const uint16_t sdm72_register[] {
-  0x0034,   // 0 SDM72D_POWER               [W]
-  0x0156,   // 3 SDM72D_TOTAL_ACTIVE        [kWh]
+  0x0000,  // Phase 1 line to neutral volts [V]
+  0x0002,  // Phase 2 line to neutral volts [V]
+  0x0004,  // Phase 3 line to neutral volts [V]
+  0x0006,  // Phase 1 current               [A]
+  0x0008,  // Phase 2 current               [A]
+  0x000A,  // Phase 3 current               [A]
+  0x000C,  // Phase 1 power                 [W]
+  0x000E,  // Phase 2 power                 [W]
+  0x0010,  // Phase 3 power                 [W]
+  0x0018,  // Phase 1 volt amps reactive    [VAr]
+  0x001A,  // Phase 2 volt amps reactive    [VAr]
+  0x001C,  // Phase 3 volt amps reactive    [VAr]
+  0x001E,  // Phase 1 power factor
+  0x0020,  // Phase 2 power factor
+  0x0022,  // Phase 3 power factor
+  0x0046,  // Frequency of supply voltages
+  0x0156,  // Total Energy                  [kWh]
 #ifdef SDM72_IMPEXP
-  0x0500,   // 1 SDM72D_IMPORT_POWER        [W]
-  0x0502,   // 2 SDM72D_EXPORT_POWER        [W]
-  0x0048,   // 4 SDM72D_IMPORT_ACTIVE       [kWh]
-  0x004A    // 5 SDM72D_EXPORT_ACTIVE       [kWh]
+  0x0500,   // Total import active power    [W]
+  0x0502,   // Total export active power    [W]
+  0x0048,   // Import active energy         [kWh]
+  0x004A    // Export active energy         [kWh]
 #endif  //  SDM72_IMPEXP
 };
 
@@ -77,6 +92,8 @@ void Sdm72Every250ms(void)
       AddLog(LOG_LEVEL_DEBUG, PSTR("SDM: SDM72 error %d"), error);
     } else {
       Energy->data_valid[0] = 0;
+      Energy->data_valid[1] = 0;
+      Energy->data_valid[2] = 0;
 
       float value;
       ((uint8_t*)&value)[3] = buffer[3];   // Get float values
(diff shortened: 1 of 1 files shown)


arendst:
What is the advantage of copying the exact functionality of the SDM630 three phase driver into the (original) single phase SDM72 driver.

Why don't you use the SDM630 driver in your case?

Or if you want more registers look at the latest Modbus driver allowing a user to select every register they need using a JSON config file.

jdelker:
> What is the advantage of copying the exact functionality of the SDM630 three phase driver into the (original) single phase SDM72 driver.

Because there seem to be slight differences in functionality between SDM630 and SDM72.
The latter is not capable of tracking import/export energy on each phase, but only provides the sum of all phases.
However, it provides detailed metrics for each phase, as SDM630 does.
So, they are quite close, but not identical in functionality.
 
> Why don't you use the SDM630 driver in your case?

I did, but for some reason the total_energy metric did not work correctly (not updating).
Rather than sorting that out (for a driver not really matching the hardware anyway), I'd thought to simply add what's missing in the SDM72 driver. 

> Or if you want more registers look at the latest Modbus driver allowing a user to select every register they need using a JSON config file.

That might also be an option, which I've not considered yet. I'll give it a try.

Besides all that, is there a particular reason why the SDM72 driver shall be kept single-phase only?
Is that a compatibility issue, I'm not aware of?

arendst:
> Besides all that, is there a particular reason why the SDM72 driver shall be kept single-phase only?
Is that a compatibility issue, I'm not aware of?

At least the last two years SDM72 users never complaint about missed functionality simply because their single phase device didn't have the functionality of your three phase SDM72. Imagine what happens in their case having this updated driver; lots of non-information. If you want your changes to be merged you'll need to add legacy support to the current users preferrable using a define to distinguish between current situation and your suggested changes.

jdelker:
> At least the last two years SDM72 users never complaint about missed functionality simply because their single phase device didn't have the functionality of your three phase SDM72. Imagine what happens in their case having this updated driver; lots of non-information. If you want your changes to be merged you'll need to add legacy support to the current users preferrable using a define to distinguish between current situation and your suggested changes.

Well, from my knowledge the various SDM72 devices were always 3-phase meters. But certainly, there may have been differences in the Modbus registers for the particular models, which could potentially conflict with this modifications.

I tried the generic modbus driver with the [provided script](https://tasmota.github.io/docs/Smart-Meter-Interface/#sdm72d-modbus). This works just fine and provides all the required data.
Side note:
For me, this is good enough. Unless you see any added value by evolving this PR any further, I'm ok to surrender it.

arendst:
> I tried the generic modbus driver with the [provided script](https://tasmota.github.io/docs/Smart-Meter-Interface/#sdm72d-modbus). This works just fine and provides all the required data.

I had hoped you tried the non-script modbus driver (``xnrg_29_modbus.ino``) but this is OK by me. Let's close this as I want to rid of these specific hardware type of drivers anyway.

jdelker:
> I had hoped you tried the non-script modbus driver (`xnrg_29_modbus.ino`) but this is OK by me. 

Uhh!? Now you've got me confused... There's another modbus driver? Could you point me to the particular part in the docs?
Obviously this hasn't catched my eye yet - or I've something completely mixed up in my head.

> Let's close this as I want to rid of these specific hardware type of drivers anyway.

Fine with me.

arendst:
> There's another modbus driver? Could you point me to the particular part in the docs?

Yep. I tend to favour programming over documentation. I suggest you take a look at file ``xnrg_29_modbus.ino``. At the top you'll find some documentation. In folder ``energy_modbus_configs`` you'll find some basic configs and additional documentation too. The driver is default enabled in ``tasmota32.bin`` so all it needs is a JSON config file....

If you manage to get it working nothing stops you from adding to the documentation ;-)

jdelker:
Ok, I'll have a look at it.
Thanks for the constructive chat.
Abandoning this PR now.

jdelker:
For the records:
Using generic modbus driver (`xnrg_29_modbus.ino`) is actually the best way to go here. It's very flexible and does not require any customization in the code if additional metrics are required. 
Unfortunately, it's a hidden gem, that has yet to be documented. Without @arendst's hint and some code studying, I would have never found it.

So here in a nutshell, what I learned from this:
- In the GPIO config, choose the `Nrg_Mbs_RX` and `Nrg_Mbs_TX` connectors for the generic modbus driver.
- The driver requires a configuration for the particular meter device. Have a look in `energy_modbus_configs/configurations.md` on your options and provided default configs for various models.
- For my SDM72D-M, I provided the following config as `modbus.json`, which provides almost all available metrics to Tasmota:
  `{"Name":"SDM72D-M","Baud":9600,"Config":"8N1","Address":1,"Function":4,"Voltage":[0,2,4],"Current":[6,8,10],"Power":[12,14,16],"ApparentPower":[18,20,22],"ReactivePower":[24,26,28],"Factor":[30,32,34],"Frequency":70,"ExportActive":74,"Total":342,"User":[{"R":42,"J":"AverageVoltage","G":"Average Voltage","U":"V","D":21},  {"R":46,"J":"AverageCurrent","G":"Average Current","U":"A","D":22},{"R":48,"J":"TotalCurrents","G":"Total Currents","U":"A","D":22},{"R":52,"J":"TotalPower","G":"Total Power","U":"W","D":23},{"R":56,"J":"TotalApparentPower","G":"Total Apparent Power","U":"W","D":23},{"R":60,"J":"TotalReactivePower","G":"Total Reactive Power","U":"W","D":23},{"R":62,"J":"TotalPowerFactor","G":"Total Power Factor","U":"W","D":2},{"R":72,"J":"ImportActive","G":"Import Active","U":"kWh","D":24},{"R":396,"J":"DiffPower","G":"Diff Power","U":"kWh","D":24},{"R":1280,"J":"TotalImportActive","G":"Total Import Active","U":"W","D":23},{"R":1282,"J":"TotalExportActive","G":"Total Export Active","U":"W","D":23}]}`

PR #18481, opened 2023-04-22
Worked on ESP32 dimmer with Zero cross

stefanbode:
Until now the ESP32 does not support zero-cross dimmer. I take a sneak how they did in in ESPhome and adapted the approach to TASMOTA. At the end it works that smooth that likely i will change ESP8266 either so we have a common code. Currently ESP8266 is not touched. There is a minor issue with savedata == default. When changing the dimmer value the interrupts get stopped during write of the config data to flash.

## Description:
#14978 

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.2.0.8


stefanbode:
Let me fix the issue. Some #ifdef missing. also checking on Dimmer 2..4. Will let you know when ready

tasmota/tasmota_xdrv_driver/xdrv_91_zerocrossDimmer.ino
```diff
@@ -0,0 +1,235 @@
...
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifdef USE_AC_ZERO_CROSS_DIMMER
+/*********************************************************************************************\
+ * Zero-Cross AC Dimmer PMM 1..xx use 
+\*********************************************************************************************/
+
+#define XDRV_91             1
```
arendst, line 25: Take xdrv_68 as it's supposed to work for esp8266 too. Do not forget to set it to 68 too. It currently is set to 1 which is not good.

stefanbode:
> Good choice to start a new driver.

Yes, it started to get a mess. And I do not like mess. All moved to xdrv68 and als corrected the 1 -> 68. Was not aware that this is important. I got one problem on the ESP32. with savedata ON I got a HANG->CRASH after changing the dimmer value. I already know that during flash write ALL interrupts get "skipped". At least on the counter. Any idea to fix?

stefanbode:
I only happens if the high frequency trigger is in. If all dimmers OFF a savedata 1 works. Here the example with dimmer1 ON. It is not a crash. looks like a long hang
```
17:48:27.132 CMD: savedata 1
17:48:27.137 SRC: WebConsole from 192.168.178.156
17:48:27.140 CMD: Grp 0, Cmd 'SAVEDATA', Idx 1, Len 1, Pld 1, Data '1'
ets Jun  8 2016 00:22:57

rst:0x7 (TG0WDT_SYS_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0030,len:184
load:0x40078000,len:12680
load:0x40080400,len:2908
entry 0x400805c4

00:00:00.003 HDW: ESP32-D0WDQ6 
00:00:00.031 UFS: FlashFS mounted with 304 kB free
00:00:00.048 CFG: Loaded from File, Count 91
00:00:00.058 QPC: Count 1
```

stefanbode:
I worked around by changing the interrupt time. Now it is stable also with savedata. I did not find the hook that notify me to disable because there is an savedata comming. From my point it looks on ESP32 that disableAlarm does work quite well. Anyhow we can commit it here. It works for Esp8266 and ESP32 and for ESP32 it checks on the savedata setting and choose the right timing.

Jason2866:
> 
> Stopping interrupts on esp32 is an art! Just calling noInterrupt won't work. You'll need to dive Into idf or take a look at the ds18b20 driver.

This should be solved (a while ago) in Arduino code. The code of disabling interrupts was non functional since the beginning until this PR  https://github.com/espressif/arduino-esp32/pull/7226

PR #17032, opened 2022-11-09
Dingtian relay board driver

barbudor:
## Description:

**Related issue (if applicable):** fixes #16797

[Dingtian relay boards](https://www.aliexpress.com/item/1005004281943758.html) are ESP32 based boards with 8, 16, 24 or 32 relays and inputs.
Relays are driven through x595 shift registers and inputs are read from x165 shift registers
Because the design is using some GPIO for both '595 and '165, those are not independant and could not be managed through the existing Shift595 driver

This driver is not included in any official Tasmota build. You must [compile your own build](https://tasmota.github.io/docs/Compile-your-build/) by adding the following line in your `user_config_override.h`:
```
#define USE_DINGTIAN_RELAY
```

The driver define 5 News GPIOs (from [here](https://github.com/dtlzp/relay_dev_demo/tree/main/gpio_pinout)):
```
|                   |  8-relay | 16-relay | 32-relay | Description                                                |
|-------------------+----------+----------+----------+------------------------------------------------------------|
| GPIO_DINGTIAN_CLK | GPIO 14  | GPIO 14  | GPIO 14  | Serial clokc for both 595 and 165                          |
|    CLK Index      |    1     |    2     |   4      | Number of 74HC165 and 74HC595                              |
| GPIO_DINGTIAN_SDI | GPIO 13  | GPIO 13  | GPIO 13  | Serial input for 595                                       |
| GPIO_DINGTIAN_Q7  | GPIO 16  | GPIO 35  | GPIO 35  | Serial output of 165                                       |
| GPIO_DINGTIAN_PL  | GPIO 32  | GPIO 0   | GPIO 0   | Latch 165 inputs (but also controls OE of the 595)         |
| GPIO_DINGTIAN_RCK | GPIO 15  | GPIO 15  | GPIO 15  | Latch of 595 outputs (but also control CLK inhibit of 165) |
```

The CLK GPIO supports index 1 to 4 to specifiy the number of shift registers:
- 1 : 8 relays/inputs boards
- 2 : 16 relays/inputs boards
- 3 : 24 relays/inputs boards (not really existing)
- 4 : 32 relays/inputs boards

Exemple for the 8 relay board:
image
image

WebGUI:
image

Inputs state are reported through SENSOR message at teleperiod like for PCF8574/MCP230xx port extenders:
```
[...]

PR #18535, opened 2023-04-29
Added support for Dingtian Relay Board v3.6.10

jeroenst:
## Description:
Dingtian has done a hardware revision (v3.6.10) because the relays where toggling at power on and buzzing when relay was activated.

The PL pin from the input register is now disconnected from the ESP32 GPIO32.
The OE pin from the output register is still connected to the ESP32 GPIO32, and has to be pulled low after the output shift register is initialized.

I want to create documentation next week.

**Related issue (if applicable):** https://github.com/arendst/Tasmota/pull/17032

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.2.0.8

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
@@ -41,7 +41,7 @@ struct DINGTIAN_DATA {
   uint8_t     first;            // index of 1st Tasmota relay assigned to 1st Dingtian relays
   int8_t      key_offset;       // index of virtual key
   // pins
-  uint8_t     pin_clk, pin_sdi, pin_q7, pin_pl, pin_rck;
+  uint8_t     pin_clk, pin_sdi, pin_q7, pin_pl, pin_oe, pin_rck;
 } *Dingtian = nullptr;
 
 
@@ -56,7 +56,7 @@ uint32_t DingtianReadWrite(uint32_t outputs)
 
   // setup
   digitalWrite(Dingtian->pin_rck, 0);    // rclk and clkinh to 0
-  digitalWrite(Dingtian->pin_pl, 1);      // load inputs in '165, ready for shift-in (side effect '595 in tri-state)
+  if (PinUsed(GPIO_DINGTIAN_PL)) digitalWrite(Dingtian->pin_pl, 1);      // load inputs in '165, ready for shift-in (side effect '595 in tri-state)
   for ( int i = Dingtian->count ; i > 0 ; i-- ) {
     // relay out to '595
     digitalWrite(Dingtian->pin_sdi, outputs & 1);
@@ -70,7 +70,8 @@ uint32_t DingtianReadWrite(uint32_t outputs)
   }
   // ending
   digitalWrite(Dingtian->pin_rck, 1);    // rclk pulse to load '595 into output registers
-  digitalWrite(Dingtian->pin_pl, 0);      // re-enable '595 ouputs
+  if (PinUsed(GPIO_DINGTIAN_PL)) digitalWrite(Dingtian->pin_pl, 0);      // re-enable '595 ouputs (old board version)
+  if (PinUsed(GPIO_DINGTIAN_OE))digitalWrite(Dingtian->pin_oe, 0);      // enable '595 ouputs (new board version)
 
 #ifdef DINGTIAN_INPUTS_INVERTED
   return ~inputs;
@@ -88,26 +89,28 @@ uint32_t DingtianReadWrite(uint32_t outputs)
 
 void DingtianInit(void) {
   if (PinUsed(GPIO_DINGTIAN_CLK, GPIO_ANY) && PinUsed(GPIO_DINGTIAN_SDI) && PinUsed(GPIO_DINGTIAN_Q7)
-   && PinUsed(GPIO_DINGTIAN_PL) && PinUsed(GPIO_DINGTIAN_RCK)) {
+   && (PinUsed(GPIO_DINGTIAN_PL) || PinUsed(GPIO_DINGTIAN_OE)) && PinUsed(GPIO_DINGTIAN_RCK)) {
     // allocate Dingtian data structure
     Dingtian = (struct DINGTIAN_DATA*)calloc(1, sizeof(struct DINGTIAN_DATA));
     if (Dingtian) {
       // get pins
-      Dingtian->pin_clk = Pin(GPIO_DINGTIAN_CLK, GPIO_ANY);   // shift clock  : 595's SCLK & 165's CLK
-      Dingtian->pin_sdi = Pin(GPIO_DINGTIAN_SDI);             // Serial out   : 595's SER
-      Dingtian->pin_q7  = Pin(GPIO_DINGTIAN_Q7);              // Serial in    : 165's Q7
-      Dingtian->pin_pl  = Pin(GPIO_DINGTIAN_PL);              // Input load   : 595's nOE  & 165's PL (or SH/LD on some datasheet)
-      Dingtian->pin_rck = Pin(GPIO_DINGTIAN_RCK);             // Output load  : 595's RCLK & 165's CLKINH
+      Dingtian->pin_clk = Pin(GPIO_DINGTIAN_CLK, GPIO_ANY);                     // shift clock   : 595's SCLK & 165's CLK
+      Dingtian->pin_sdi = Pin(GPIO_DINGTIAN_SDI);                               // Serial out    : 595's SER
+      Dingtian->pin_q7  = Pin(GPIO_DINGTIAN_Q7);                                // Serial in     : 165's Q7
+      if (PinUsed(GPIO_DINGTIAN_PL)) Dingtian->pin_pl  = Pin(GPIO_DINGTIAN_PL); // Input load    : 595's nOE  & 165's PL (or SH/LD on some datasheet)
+      if (PinUsed(GPIO_DINGTIAN_OE)) Dingtian->pin_oe  = Pin(GPIO_DINGTIAN_OE); // Output enable : 595's nOE (v3.6.10)
+      Dingtian->pin_rck = Pin(GPIO_DINGTIAN_RCK);                               // Output load   : 595's RCLK & 165's CLKINH
       // number of shift registers is the CLK index
       Dingtian->count   = ((GetPin(Dingtian->pin_clk) - AGPIO(GPIO_DINGTIAN_CLK)) + 1) * 8;
 
-      AddLog(LOG_LEVEL_DEBUG, PSTR("DNGT: clk:%d, sdi:%d, q7:%d, pl:%d, rck:%d, count:%d"),
-        Dingtian->pin_clk, Dingtian->pin_sdi, Dingtian->pin_q7, Dingtian->pin_pl, Dingtian->pin_rck, Dingtian->count);
+      AddLog(LOG_LEVEL_DEBUG, PSTR("DNGT: clk:%d, sdi:%d, q7:%d, pl:%d, oe:%d, rck:%d, count:%d"),
+        Dingtian->pin_clk, Dingtian->pin_sdi, Dingtian->pin_q7, Dingtian->pin_pl, Dingtian->pin_oe, Dingtian->pin_rck, Dingtian->count);
 
       DINGTIAN_SET_OUTPUT(Dingtian->pin_clk, 0);
       DINGTIAN_SET_OUTPUT(Dingtian->pin_sdi, 0);
       DINGTIAN_SET_INPUT( Dingtian->pin_q7);
       DINGTIAN_SET_OUTPUT(Dingtian->pin_pl, 0);
+      //Do not initialize Dingtian->pin_oe so the relays will not toggle while restarting
       DINGTIAN_SET_OUTPUT(Dingtian->pin_rck, 0);
 
       Dingtian->first = TasmotaGlobal.devices_present;
--- a/tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
@@ -38,11 +38,12 @@
  */
 
 struct DINGTIAN_DATA {
-  uint32_t    outputs;          // keep ouputs state
-  uint32_t    last_inputs;      // previous inputs state
-  uint8_t     count;            // number of relay and input (8 * numver of shift registers)
-  uint8_t     first;            // index of 1st Tasmota relay assigned to 1st Dingtian relays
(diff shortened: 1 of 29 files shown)


jeroenst:
Tagging @barbudor because he was the initial developer of the dingtian code.

barbudor:
Pleas allow me some time to review but from your description I think no change is needed in the code
You can just document that GPIO32 has to be set to OUTPUT_LOW
What would be the GPIO for PL though?

jeroenst:
No that doesn't work because output of gpio32 will be low before the output shift register is initialized, causing the relays to click at boot (and also maybe on restart).

jeroenst:
One moment, fixing an issue on first boot for this one.

jeroenst:
I don't know why, but the build is failing on the DE version. I tried to fix this but I can't find why the compiler is complaining.

barbudor:
It's not failing on the DE version but on the AD version because you missed this language in your update

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
```diff
@@ -83,31 +91,31 @@ uint32_t DingtianReadWrite(uint32_t outputs)
...
-        Dingtian->pin_clk, Dingtian->pin_sdi, Dingtian->pin_q7, Dingtian->pin_pl, Dingtian->pin_rck, Dingtian->count);
+      AddLog(LOG_LEVEL_DEBUG, PSTR("DNGT: clk:%d, sdi:%d, q7:%d, pl:%d, oe:%d, rck:%d, count:%d"),
+        Dingtian->pin_clk, Dingtian->pin_sdi, Dingtian->pin_q7, Dingtian->pin_pl, Dingtian->pin_oe, Dingtian->pin_rck, Dingtian->count);
 
       DINGTIAN_SET_OUTPUT(Dingtian->pin_clk, 0);
       DINGTIAN_SET_OUTPUT(Dingtian->pin_sdi, 0);
       DINGTIAN_SET_INPUT( Dingtian->pin_q7);
       DINGTIAN_SET_OUTPUT(Dingtian->pin_pl, 0);
+      //Do not initialize Dingtian->pin_oe so the relays will not toggle while restarting
+      DINGTIAN_SET_OUTPUT(Dingtian->pin_pl, 0);
```
barbudor, line 118: redundant lines. You have twice the same command.
Is that intended ?

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
```diff
@@ -83,31 +91,31 @@ uint32_t DingtianReadWrite(uint32_t outputs)
  * Driver initialisation
  */
 
-#define DINGTIAN_SET_OUTPUT(pin,value)  { pinMode((pin), OUTPUT); digitalWrite((pin), (value)); }
-#define DINGTIAN_SET_INPUT(pin)         { pinMode((pin), INPUT); }
-
 void DingtianInit(void) {
   if (PinUsed(GPIO_DINGTIAN_CLK, GPIO_ANY) && PinUsed(GPIO_DINGTIAN_SDI) && PinUsed(GPIO_DINGTIAN_Q7)
-   && PinUsed(GPIO_DINGTIAN_PL) && PinUsed(GPIO_DINGTIAN_RCK)) {
+   && (PinUsed(GPIO_DINGTIAN_PL) || PinUsed(GPIO_DINGTIAN_OE)) && PinUsed(GPIO_DINGTIAN_RCK)) {
```
barbudor, line 96: I don't get that
By default PL should always been set for all models of board
OE should be an optional pin only for new boards

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
```diff
@@ -56,7 +59,7 @@ uint32_t DingtianReadWrite(uint32_t outputs)
 
   // setup
   digitalWrite(Dingtian->pin_rck, 0);    // rclk and clkinh to 0
-  digitalWrite(Dingtian->pin_pl, 1);      // load inputs in '165, ready for shift-in (side effect '595 in tri-state)
+  if (PinUsed(GPIO_DINGTIAN_PL)) digitalWrite(Dingtian->pin_pl, 1);      // load inputs in '165, ready for shift-in (side effect '595 in tri-state)
```
barbudor, line 62: See my other comment below. PL is a mandatory pin so this test is useless.

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
```diff
@@ -70,7 +73,12 @@ uint32_t DingtianReadWrite(uint32_t outputs)
   }
   // ending
   digitalWrite(Dingtian->pin_rck, 1);    // rclk pulse to load '595 into output registers
-  digitalWrite(Dingtian->pin_pl, 0);      // re-enable '595 ouputs
+  if (PinUsed(GPIO_DINGTIAN_PL)) digitalWrite(Dingtian->pin_pl, 0);      // re-enable '595 ouputs (old board version)
```
barbudor, line 76: same comment as above

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
```diff
@@ -70,7 +73,12 @@ uint32_t DingtianReadWrite(uint32_t outputs)
...
+  if (PinUsed(GPIO_DINGTIAN_OE))
+  {
+    digitalWrite(Dingtian->pin_oe, 0);      // enable '595 ouputs (new board version)
+    DINGTIAN_SET_OUTPUT(Dingtian->pin_oe, 0);
+  }
```
barbudor, line 81: 1) better test pin_oe than making a call to PinUsed() which is longer (insure that pin_oe has a default value meaning not used)
2) any way to avoid doing that at each loop but only once ?

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
```diff
@@ -83,31 +91,31 @@ uint32_t DingtianReadWrite(uint32_t outputs)
...
-      Dingtian->pin_clk = Pin(GPIO_DINGTIAN_CLK, GPIO_ANY);   // shift clock  : 595's SCLK & 165's CLK
-      Dingtian->pin_sdi = Pin(GPIO_DINGTIAN_SDI);             // Serial out   : 595's SER
-      Dingtian->pin_q7  = Pin(GPIO_DINGTIAN_Q7);              // Serial in    : 165's Q7
-      Dingtian->pin_pl  = Pin(GPIO_DINGTIAN_PL);              // Input load   : 595's nOE  & 165's PL (or SH/LD on some datasheet)
-      Dingtian->pin_rck = Pin(GPIO_DINGTIAN_RCK);             // Output load  : 595's RCLK & 165's CLKINH
+      Dingtian->pin_clk = Pin(GPIO_DINGTIAN_CLK, GPIO_ANY);                     // shift clock   : 595's SCLK & 165's CLK
+      Dingtian->pin_sdi = Pin(GPIO_DINGTIAN_SDI);                               // Serial out    : 595's SER
+      Dingtian->pin_q7  = Pin(GPIO_DINGTIAN_Q7);                                // Serial in     : 165's Q7
+      if (PinUsed(GPIO_DINGTIAN_PL)) Dingtian->pin_pl  = Pin(GPIO_DINGTIAN_PL); // Input load    : 595's nOE  & 165's PL (or SH/LD on some datasheet)
+      if (PinUsed(GPIO_DINGTIAN_OE)) Dingtian->pin_oe  = Pin(GPIO_DINGTIAN_OE); // Output enable : 595's nOE (v3.6.10)
```
barbudor, line 105: PL should be mandatory, OE optional
If OE not defined, a default value wouldbe in pin_oe to say "not used" to optimize above test

barbudor (commented):
I understand your comment about restart requiring to delay setting OE low

However I think that the PL pin should be mandatory as it is required for both models of boards and it is a functional requirement for inputs to have PL.
OE being available separately only on some boards should be optional.

I don't see a case where you could define OE but not define PL.

In any case, that's 2 different incompatible hardware so template would need to be different.

Thanks a lot for the improvement.

my 2 cents

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: PL is not mandatory for the new board.

OE or PL is mandatory.

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: 1. Ok
2. Yes I can change that

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: PL is only mandatory for old board, for new board OE is mandatory

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: No, it's wrong, I will fix it, thank you!

jeroenst:
> I understand your comment about restart requiring to delay setting OE low
> 
> However I think that the PL pin should be mandatory as it is required for both models of boards and it is a functional requirement for inputs to have PL. OE being available separately only on some boards should be optional.
> 
> I don't see a case where you could define OE but not define PL.
> 
> In any case, that's 2 different incompatible hardware so template would need to be different.
> 
> Thanks a lot for the improvement.
> 
> my 2 cents

PL is only mandatory for the old board.
OE is mandatory for the new board.

I don't know why PL is no longer needed, but it's the instructions that I got from Dingtian and the inputs work fine without PL.

barbudor:
I would like to understand that statement better because I can't see how the '165 would work without PL
How is it cabled on the new board? 
Is it connected to the RCLK of the '595? Or most likely inverted? 
PL must go from low to high just before the shift starts to latch the input values in the shift register and stay high during the shift. 
RCLK must go from low to high at the end of the shift to transfer outputs from shift registers to output registers
So if PL = /RCLK, that would work

If this is the case I suggest to harden the pin testing to insure that at least PL or OE are defined

jeroenst:
> I would like to understand that statement better because I can't see how the '165 would work without PL How is it cabled on the new board? Is it connected to the RCLK of the '595? Or most likely inverted? PL must go from low to high just before the shift starts to latch the input values in the shift register and stay high during the shift. RCLK must go from low to high at the end of the shift to transfer outputs from shift registers to output registers So if PL = /RCLK, that would work
> 
> If this is the case I suggest to harden the pin testing to insure that at least PL or OE are defined

I asked Dingtian how that they latch the inputs now PL is not needed anymore. I don't have an answer to that question yet. I tested it on the new board without PL and it works like our should. I don't have a clue how it does without PL.

barbudor:
Could you please kindly try to follow the track on the PCB ?

jeroenst:
Yes, I will try

jeroenst:
I did change the code according to your comments, only using pin_oe instead of PinUsed(GPIO_DINGTIAN_OE) is a bit tricky I think, because what value can be used for unused state..?

jeroenst:
I did some measurements and reverse engineering of the board, and the RCLK of the 74HC595 is connected to an C04F ic which is probably an logic inverter (although the pins do not match the C04F datasheet I found). The output of this inverter is connected to PL of the 74HC165.

dingitan

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: PL is driven via and inverter connected to RCLK.

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: PL is driven via and inverter connected to RCLK.

tasmota/tasmota_xdrv_driver/xdrv_90_esp32_dingtian_relay.ino
jeroenst: PL is driven via and inverter connected to RCLK.

jeroenst:
Ok now all is set I think. If you want I can create an value (like 255?) when an PL or OE is not configured but imo PinUsed is the best way without using a arbitrary value that's maybe used for other purposes.

The only issue is that it won't compile with the language configurations, can you tell me how to fix that?

jeroenst:
I think this is now ready and can be merged.

jeroenst:
@barbudor Any comments or recommendations on this PR? Is it approved to be merged?

barbudor:
Hi @jeroenst , @arendst 

Sorry I was busy this week-end
Yes, all good for me
Thanks @jeroenst

jeroenst:
@barbudor, no problem, thank you for your support!

DavidNetron:
Hello. I could ask for a ready compiled tasmota for the Dingtian 16 Channel Relay Board (DT-R016)-V.3.6.0. It would help me a lot. I have no knowledge and I came across a board during automation and I need to connect it to the system. Thank you very much

barbudor:
@DavidNetron a PR thread is not the right place for that request. You should ask in the Dingtian discussion https://github.com/arendst/Tasmota/discussions/16797

Beside I strongly recommend you mean how to do custom builds as it is easy with tools such as TasmoCompiler
https://tasmota.github.io/docs/Compile-your-build/

jeroenst:
@DavidNetron here you go:
[tasmota32.factory.zip](https://github.com/arendst/Tasmota/files/14663260/tasmota32.factory.zip)

Beware when ordering the board ask for a **unlocked version** which can be flashed. The default board shipped is locked.

DavidNetron:
It is easy for me to design machinery......Thank you very much for your help. I will be able to move on. Have a nice day

sq4bja:
I have a Dingtian CH8 with working Tasmota. Everything looks fine, but how can I link events from inputs, e.g., IN1, to outputs, e.g., POWER1? What Rule should I write so that a high state on IN1 activates POWER1?

teixeluis:
Hello, I have the 8 ch board v3.6.13, and I find that setting GPIO32 to Dingtian PL allows the Ethernet chip to work properly, but the relays will glitch on power up. If on the other hand I set it to Dingtian OE, relays don't glitch, but I have no Ethernet. Anyone came across this behaviour? Any idea what could help? Thanks

PR #18553, opened 2023-05-01
[ZCDimmer] reboot bugfix

SteWers:
## Description:

- Bugfix of reboots on ESP8266
- Display value of ``ZCDimmerSet`` on Webpage

See also discussion [#14978](https://github.com/arendst/Tasmota/discussions/14978)

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.8

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_68_zerocrossDimmer.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_68_zerocrossDimmer.ino
@@ -24,6 +24,8 @@
 
 #define XDRV_68             68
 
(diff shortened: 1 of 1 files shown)


stefanbode (approved):

stefanbode:
can you cancel the pull request? I copied your code to mine. Regarding the ESP8266 we must do it a bit different. Just commenting out would kill the accuracy for the ESP8266. Then ZCDimmerSet would not make any sense anymore because you do not hit in time the sinus. I used the dynamic trigger capability of the ESP8266 like they do in ESPeasy. Not the accuracy is slightly worse than on ESP32 but still ok. Currently do a "crash test" two dimmers defined and checking if there is a hardware reboot.

PR #18610, opened 2023-05-08
IRremoteESP8266 upgraded to v2.8.5

mbronk:
## Description:
Upgrades IRremoteESP8266 `v2.8.4` ->  [**`v2.8.5`**](https://github.com/crankyoldgit/IRremoteESP8266/releases/tag/v2.8.5). 

**Related issue (if applicable):** `n/a`
## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.9
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #18612, opened 2023-05-08
IR: Extended the A/C protocol with Command/iFeel/SensorTemp

mbronk:
## Description:
This is a follow-up to PR https://github.com/arendst/Tasmota/pull/18610 (_upgrading IRremoteESP8266 `v2.8.4` ->  [**`v2.8.5`**](https://github.com/crankyoldgit/IRremoteESP8266/releases/tag/v2.8.5)_). 

### Changes:
  - Adds support for new IR A/C fields _(introduced in 2.8.5)_: `Command`, `iFeel`, `SensorTemp`
  - Makes the command/response field set vary by `Command` (one of: `Control | Config | iFeel | Timer`)

**Related issue (if applicable):** `n/a`
## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.9


mbronk:
@s-hadinger I personally recommend reviewing commit-by-commit (it _should_ make more sense in these increments). If it helps, I can break it down further/into smaller PRs also

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
```diff
@@ -221,37 +222,75 @@ void IrReceiveInit(void)
   irrecv->enableIRIn();                  // Start the receiver
 }
 
+namespace {
```
s-hadinger, line 225: Just curious, why a namespace?

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
```diff
@@ -221,37 +222,75 @@ void IrReceiveInit(void)
...
+  void addFloatToJson(JsonGeneratorObject& json, const char* key, float value) {
+    if (floorf(value) == value) {
+      json.add(key, (int32_t) floorf(value));       // integer
+    } else {
+      // TODO can do better here
+      json.addStrRaw(key, String(value, 1).c_str());   // non-integer, limit to only 1 sub-digit
+    }
```
s-hadinger, line 232: I suggest you use `ext_snprintf`. It already handles removing trailing digits if they are zeroes.

For example to have at most 2 digits (keep in mind that float is passed by pointer, not by value)
```
void addFloatToJson(JsonGeneratorObject& json, const char* key, float value) {
  char s[16];
  ext_snprintf(s, sizeof(s), PSTR("%-2_f"), &value);
  json.addStrRaw(key, s);
}
```

See https://tasmota.github.io/docs/Sensor-API/#extension-to-vsnprintf

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
```diff
@@ -378,6 +417,16 @@ bool strToBool(class JsonParserToken token, bool def) {
   }
 }
 
+int16_t timeStrToMinutesInt(const JsonParserToken& token, int16_t def) {
+  std::tm t = {};
+  if (token.isStr() && strptime(token.getStr(), "%H:%M", &t)) {
```
s-hadinger, line 422: String literals must be pushed in Flash with `PSTR("%H:%M")` but must be accessed only with 32 bits aligned reads on ESP8266. I'm not sure that `strptime()` is compatible with string literal in Flash.

Can you assess the flash impact of using `strptime()`? I'm concerned that it brings a lot of useless code. For ESP32 it's ok, for ESP8266 it's not.

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
mbronk: Honestly, just a habit.
I mean, they _technically_ are meant to be somewhat better (see [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities),  [here](https://stackoverflow.com/questions/4422507/superiority-of-unnamed-namespace-over-static) or [here](https://abseil.io/tips/186) for rationale), but this doesn't fully apply to here and I have no strong preference either way....
Would you like them removed?

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
mbronk: Thx, will refactor (TBH, I just copied the code from below as-is, w/o any second thoughts)

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
mbronk: Thanks, good catch! I'll see what it does to flash size and report back.
TBH, I started to wonder if it really makes sense keeping it. I added for symmetry (the output is `HH:MM` so I figured it'd be nice to be able to have input in this format...), not really considering device limitations.
If it consumes a lot of flash, I technically could add a "manual" code doing string split by colon and conversion... but it smells of needless complexity as well...

----
tl;dr; Now that I think about it... it might be an overkill to add in the first place (and int input in minutes might be enough). Thoughts?

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
s-hadinger: Ah ok. This part of C++ is obscure to me. Here it's a better version of `static`. Leave it as it is.

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
s-hadinger: Keep it as simple as possible. ESP8266 is **REALLY** limited in resources, every byte counts. If input in minutes is ok, leave it simple. Conversion from more complex formats must be done where you have more resources (like MQTT gateway).

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
mbronk: Changed as suggested (added a separate value for `n/a` though as I found it non-intuitive to see `-100.0` when I was testing it. Please check if you're OK with it

tasmota/tasmota_xdrv_driver/xdrv_05_irremote_full.ino
mbronk: Now removed alltogether.

FYI the flash impact was:

* with **strptime**
```
RAM:   [=====     ]  50.0% (used 40932 bytes from 81920 bytes)
Flash: [======    ]  64.7% (used 662580 bytes from 1023984 bytes)
```

* without
```
RAM:   [=====     ]  48.9% (used 40096 bytes from 81920 bytes)
Flash: [======    ]  64.4% (used 659096 bytes from 1023984 bytes)
```

... so saved ~3kB of Flash and ~800B of RAM (well worth it! thx for catching it)

s-hadinger: Indeed it is not worth the cost

s-hadinger:
All good with me, thanks a lot

PR #18634, opened 2023-05-10
start modus tcp bridge automatically

f-reiling:
## Description:
This change adds the possiility to start the TCP modus bridge directly at system start with the default TCP port (502). No need to send the start command or to define a rule.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.9


Jason2866:
Add a few words in description why this change is usefull or needed.

f-reiling:
> Add a few words in description why this change is usefull or needed. Why not using a simple rule for?

Description has been updated. Instead of using a simple rule, this works directly after flashing the software without any configuration. Especially if multiple devices are used, this solution is easier to maintain.

Jason2866:
Thats not the design goal of Tasmota to "hard code" all possible use cases.
Let us see what the project owner is thinking about.

arendst (approved):
As this is TCP based anyway I can see the advantage optionally starting the server.

PR #18637, opened 2023-05-10
esp32_webcam: faster page load

Staars:
## Description:

Faster web interface by avoiding re-initialisation calls of the hardware and removing a delay from the render section, instead adding error catching to the browser.

I did not see any negative side effects in multiple attempts, but this has to be tested on more devices.  
Works for me on Firefox and Chrome. Safari has issues with MJPEG for years and will probably show no picture after initial success and some page reloads/navigation actions.  
This will not solve any problems with failing hardware initialization, but performance in my setup is way more snappy now and as we had some recent reports about non-working ESP32-cams with newer Tasmota versions I can confirm, that the current driver is NOT fundamentally broken. Basically I had zero issues in extensive tests in the last 2 days since the last PR.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
@@ -974,14 +974,14 @@ uint32_t WcSetStreamserver(uint32_t flag) {
 
 void WcInterruptControl() {
   WcSetStreamserver(Settings->webcam_config.stream);
-  WcSetup(Settings->webcam_config.resolution);
+  if(Wc.up != 0) {WcSetup(Settings->webcam_config.resolution);}
 }
 
 /*********************************************************************************************/
 
 
 void WcLoop(void) {
-  if (4 == Wc.stream_active) { return; }
+  // if (4 == Wc.stream_active) { return; }
 
   if (Wc.CamServer) {
     Wc.CamServer->handleClient();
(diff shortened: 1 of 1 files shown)


btsimonh:
note in my latest PR,
I removed the   Wc.up = 0; here:
```
void CmndWebcamInit(void) {
  Wc.up = 0;
```
and replaced with 
```
void CmndWebcamInit(void) {
  WcSetup(Settings->webcam_config.resolution);
```

the wcsetup will deinit if wc.up before initting.
Doing a camera setup again without deinit first seemed to cause issues for me - maybe because AIThinker has no CAM_RESET gpio?

Staars:
The command `CmndWebcamInit`  does not make so much sense to me anyway, maybe there are useful applications for it with other setups than mine.
By adding  `Wc.up = 0;` I only tried to restore the old behavior in order to force a new initialization, but I did not need it for myself. Thus I have no reservations against changes here and what you wrote about no CAM_RESET sounds plausible to me.

btsimonh:
I was having extreme issues with 0x105 (basically, I2C not found/working), and these occurred randomly.  So for me wcinit command was a good test vehicle to cause camera re-setup :).  It only has practical benefit to start the camera when not looking at the webui - in some blog or other I found someone calling it from outside to 'get the camera running' - maybe for rtsp.

Staars:
Indeed, I only used the web UI so far.
So it is needed for basically everything else … makes sense.

PR #18655, opened 2023-05-14
Add mutex to many camera functions.

btsimonh:
## Description:

Add mutex to many camera functions.
Not a guaranteed fix every time, but does improve chances of AIThinker CAM working.

**Related issue (if applicable):** helps with [discussion](https://github.com/arendst/Tasmota/discussions/18482)

## Checklist:
  - [*] The code change is tested and works with Tasmota core ESP32 V.2.0.9

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
@@ -96,9 +96,17 @@
  * remarks for AI-THINKER
  * GPIO0 zero must be disconnected from any wire after programming because this pin drives the cam clock and does
  * not tolerate any capictive load
+ * the AITHINKER module does not have CAM_RESET - so if you get the camera into a bad state, power off restart is the only way out.
  * flash led = gpio 4
  * red led = gpio 33
  * optional rtsp url: rtsp://xxx.xxx.xxx.xxx:8554/mjpeg/1
+ * 
+ * SH 2023-05-14 - added mutex for many webcam functions - this is to prevent multi-threaded access to the camera functions, which 
+ * can case error 0x105 upon re-init.
+ * Errors 0x103 and 0xffffffff could indicate CAM_PWDN incorrect.
+ * 
+ * I2C use: if USE_I2C is enabled, you can set GPIO26 to I2c_SDA/2 and GPIO27 to I2C_SCL/2, and then use the shared I2C bus 2.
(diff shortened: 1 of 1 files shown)


btsimonh:
I noticed that through a wcinit command, the stream stopped.  Added a commit to allow the stream to continue after the camera is re-configured.

@arendst - 
I have noticed that a software restart (or firmware upgrade..) can (rarely) result in a 0x105 error.
I think this is because the camera is left mid-i2c, and then can no longer be discovered.
As there are devices without a GPIO for CAM_RESET, I'd like to 'close' the camera before restarting.
Would FUNC_SAVE_BEFORE_RESTART be suitable?
hmm..  just tested and using that in xdrv 81 cauases restart to hang...
Advice?

btsimonh:
looks like FUNC_SAVE_BEFORE_RESTART is not a good one - it's called on saveall.  But I can't see an API command to close a driver...

arendst:
Does a close also happens when you stop the stream. Ie I use the GUI. When I close the GUI I see the stream is stopped in the console.

If yes, try what happens if you do this and then do a restart to see if it comes up alive again. If it doesn't comes back up then any pre-closure will fail too.

btsimonh:
will give it a go tonight.  Basically, restart from the web console (so stream is already stopped).
But I'd really like to do a full deinit as well if you think of some way to get a call through at all restart points.

arendst:
I suppose you know this:
```
/**
 * @brief Initialize the camera driver
 *
 * @note call camera_probe before calling this function
 *
 * This function detects and configures camera over I2C interface,
 * allocates framebuffer and DMA buffers,
 * initializes parallel I2S input, and sets up DMA descriptors.
 *
 * Currently this function can only be called once and there is
 * no way to de-initialize this module.
 *
 * @param config  Camera configuration parameters
 *
 * @return ESP_OK on success
 */
esp_err_t esp_camera_init(const camera_config_t* config);
```
where it states no way no to do a de-init.

Take a look at ``esp_camera.h`` for further info (in framework-arduinoespressif32).

btsimonh:
yep, saw that...   but also note it says to call camera_probe before calling the fn, and you can't because it's static....
I tested the wcinit cmnd repeatedly, which deinits then inits, and it's quite reliable with the mutexes in place.
Plus they have a deinit function - so I'm guessing the comments to be old.

Final PR additions above.  Let's not worry about closing on restart.  I tried to reproduce on another module, and could not in 10 or so tries, so unless someone has this specific issue and can't power off restart, we can defer it.

PR #17177, opened 2022-11-24
AC-Dimmer Power lookup table

stefanbode:
## Description:
AC-Dimmer was not linear on power because of the behavior AC-current to calculate the power through the integral over SINUS. Lookup table fix the linearity issue. 
**Related issue (if applicable):** fixes #<Tasmota issue number goes here>

## Checklist:
  - [] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [x] The code change is tested and works with Tasmota core ESP32 V.2.0.5
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #18666, opened 2023-05-15
LED PWM ac_dimmer curve was wrongly applied instead of Gamma

s-hadinger:
## Description:

Fixes regression introduced in #17177 where ac_dimmer curve was applied instead of Gamma curve. Now the ac_dimmer is only applied if `SetOption99 1` is applied.

Fixes #18639

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Partition_Manager.tapp fixed
 - Berry fixed a rare condition when a GC causes a memory corruption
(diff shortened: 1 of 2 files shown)


stefanbode:
Thanks!!

SteWers:
Is `analogWrite` still suitable at this place since PWM is no longer used with the new driver version?

stefanbode:
Let me check at the weekend. Right now I'm pretty sure the complete #ifdef zero cross dimmer can be moved away. Anyhow I want to check with a real device if this also works on the init and if there is a sideeffect if the led value is present at this time

PR #18711, opened 2023-05-22
Add MFRC522 support for customizable SPI pins

AlKi:
## Description:
(MF)RC522 reader modules cannot be used with e.g. ESP32-cam modules, since the driver uses the default SPI interface.
This code change enables custom builds to use the SPI pins configured via a template (e.g. via web UI)

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.9 (Note: tested on ESP32-cam)


arendst:
Agree with @Jason2866. I'll need to dive in why this needs to be fixed the way it is proposed here.

arendst:
I see the issue. Will fix in a generic way. Thx for your input.

arendst:
Thx. Merged in some way. See above.

AlKi:
Oh, nice!
I added the #ifdef since I was not sure if this would break things.
I'm happy this lead to your nice (and way cleaner) fix in #18711 :)

PR #18826, opened 2023-06-08
Add I2S Audio defines and humanise some messages

blakadder:
## Description:

Added most of the I2S Audio driver defines to my_user_config.h and fixed up some log messages to be more human friendly

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -1116,6 +1116,7 @@
 
 #define USE_NETWORK_LIGHT_SCHEMES                 // Add support for light schemes via DDP (via UDP on port 4048)
 
+//#define USE_I2S                                 // Add I2S support (used with Berry Audio and TensorFlow lite)
 //#define USE_SPI                                 // Add support for hardware SPI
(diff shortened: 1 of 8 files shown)


Staars:
I think this is a difficult topic.
The define `USE_INMP441` can be misleading, because every INM441 board I am aware of can be configured for using the right channel by connecting 2 pins.
I would prefer some kind of descriptor to setup the i2s-driver in order to pass the neede values at runtime, that get later get passed to:
```static const i2s_config_t i2s_config = {
    .mode = I2S_MODE_MASTER | I2S_MODE_TX,
    .sample_rate = 44100,
    .bits_per_sample = 16,
    .channel_format = I2S_CHANNEL_FMT_RIGHT_LEFT,
    .communication_format = I2S_COMM_FORMAT_STAND_I2S,
    .intr_alloc_flags = 0, // default interrupt priority
    .dma_buf_count = 8,
    .dma_buf_len = 64,
    .use_apll = false
}; ```

This could be done in an ini file like .drvset_i2s, which could be used from other drivers too.
Thus we could avoid a large chunk of defines and special builds.
/

PR #18829, opened 2023-06-09
Add wcclient.get_bytes() to Berry

btsimonh:
(cherry picked from commit 5903b21448a3f088b12bd47a04996ff84e467a3a)

## Description:

Add webclient.get_bytes() to Berry to allow binary files to be read from the web.

My first time modifying the Berry native code, so pls confirm I did it right?

## Checklist:

### DIFF
--- a/lib/libesp32/berry_tasmota/src/be_webclient_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_webclient_lib.c
@@ -30,6 +30,7 @@ extern int wc_getstring(bvm *vm);
 extern int wc_writefile(bvm *vm);
 extern int wc_writeflash(bvm *vm);
 extern int wc_getsize(bvm *vm);
+extern int wc_getbytes(bvm *vm);
 
 #include "be_fixed_be_class_webclient.h"
 
@@ -68,6 +69,7 @@ class be_class_webclient (scope: global, name: webclient) {
     write_file, func(wc_writefile)
     write_flash, func(wc_writeflash)
     get_size, func(wc_getsize)
+    get_bytes, func(wc_getbytes)
 }
 @const_object_info_end */
 
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_webclient.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_webclient.ino
@@ -654,5 +654,57 @@ extern "C" {
   }
 }
 
+class StreamBytes: public Stream
+{
+public:
+  StreamBytes(uint8_t * buf_in, int len_in) : buf(buf_in), offset(0), len(len_in) {};
+
+  size_t write(const uint8_t *buffer, size_t size) override {
+    // AddLog(LOG_LEVEL_INFO, "FLASH: addr=%p  hex=%*_H  size=%i", addr_start + offset, 32, buffer, size);
+    if (size > 0) {
+      if (offset + size > len){
+        AddLog(LOG_LEVEL_ERROR, "BERRYWC: bufer overrun");
+        return size;
+      }
+      memcpy(buf+offset, buffer, size);
+      offset += size;
+    }
+    return size;
+  }
+  size_t write(uint8_t data) override {
+    write(&data, 1);
+    return 1;
+  }
+
+  int available() override { return 0; }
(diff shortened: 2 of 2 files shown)


btsimonh:
Hold this for a day...  I'm looking to add a generic webclient.write(bytes|string) function with the intent of being able to send multipart, but need to test a bit....

s-hadinger:
Very nice, thanks.

s-hadinger:
This is an excellent start. It could be possible to optimize further and to have the `bytes()` buffer allocated first and the stream writing into it. This would avoid a double allocation of memory.

btsimonh:
Is that not what be_pushbuffer does?  I'd assumes this allocated the bytes, and that be_pushbytes just confirmed it on the stack.
But maybe we don't even want the be_pushbytes at all?  Can you check what the two functions do?
I have a nasty feeling that we're adding two things to the stack...

s-hadinger:
We're indeed pushing and allocating twice the buffer on the heap.

To make it only once is a little trickier:
- instanciate a new `bytes()` object and pass the size to force allocation of the buffer
- call `resize()` to force the internal buffer to be the right size
- call `_buffer()` to get the pointer of the buffer in memory
- fill the buffer with the data
- (opt) call `resize()` again if the data is actually shorter than expected

One example is here:
```C

static int i_readbytes(bvm *vm)
{
    int argc = be_top(vm);
    be_getmember(vm, 1, ".p");
    if (be_iscomptr(vm, -1)) {
        void *fh = be_tocomptr(vm, -1);
        size_t size = readsize(vm, argc, fh);
        if (size) {
            /* avoid double allocation, using directly the internal buffer of bytes() */
            be_getbuiltin(vm, "bytes");
            be_pushint(vm, size);
            be_call(vm, 1);  /* call bytes() constructor with pre-sized buffer */
            be_pop(vm, 1);  /* bytes() instance is at top */

            be_getmember(vm, -1, "resize");
            be_pushvalue(vm, -2);
            be_pushint(vm, size);
            be_call(vm, 2); /* call b.resize(size) */
            be_pop(vm, 3);  /* bytes() instance is at top */

            char *buffer = (char*) be_tobytes(vm, -1, NULL); /* we get the address of the internam buffer of size 'size' */
            size = be_fread(fh, buffer, size);

            /* resize if something went wrong */
            be_getmember(vm, -1, "resize");
            be_pushvalue(vm, -2);
            be_pushint(vm, size);
            be_call(vm, 2); /* call b.resize(size) */
            be_pop(vm, 3);  /* bytes() instance is at top */
        } else {
            be_pushbytes(vm, NULL, 0);
        }
        be_return(vm);
    }
    be_return_nil(vm);
}
```

btsimonh:
Theoretically, we could cope with content-length -1 by expanding the byte buffer - not efficient, but maybe better than not allowing?

s-hadinger:
I don't understand. Expanding a buffer always results in a new allocation. So it's better to allocate the right size in the first place

btsimonh:
yep, but if there is no content-length header, we don't know how much we're going to get :).  Obviously it's better to have a content-length.... but we should cope without, even if it's more expensive.

s-hadinger:
I wouldn't support binary read if Content-Length is not present. Unless you have a specific use case for it?

sfromis:
Except for images, most binary reads could be well-served by having a fixed initial buffer of a "reasonable" size, as an argument to `get_bytes`, and then throw an error if exceeded.

Of course, well-behaved servers "should" provide a length, but I did observe that webcam `snapshot.jpg` (which i some time ago contemplated using from another Tasmota device) is not among those, instead having `ContentLength=-1`.

btsimonh:
haha :).  I will fix that!  The new driver is coming on nicely... I'm now considering whether to make a true berry class for it, or continue to work with evil Tas commands that return native addresses and lengths in json.  But I don't really know where to start on adding a Berry class to the webcam driver.  (I DO call functions on a berry class if it exists, and that works nicely if you create the class in a berry script)..

btsimonh:
see new PR based on your advice and some research :).
The StreamBeBytesWriter(vm, incr) may be useful for other things as well?

s-hadinger:
Always handy to have it. Thanks for this.

PR #18830, opened 2023-06-09
Matter fix Illuminance to uint

s-hadinger:
## Description:

Fix Illuminance to unsigned int, to conform with spec.

## Checklist:


sfromis:
Fixing illuminance to an integer does make it work "less well" for sensors having useful decimals on the lux value, but if the specs stipulate this.....

s-hadinger:
The specs says "unsigned int 16 bits". Before the fix it was "signed int 16 bits"

s-hadinger:
The formula is 'log10(lux)*10000+1'

I thin they screwed up the formula. You can have very small increments above 1 but nothing below 1. 

It should have been 'log10(lux+1)*10000'. But now the standard is bad

s-hadinger:
See #18847, I'm actually fixing the formula so it's close enough to the standard, but also usable for low lux values. Just substract 1 to the value reported by the controller.

PR #18837, opened 2023-06-10
Berry Webclient: wc_getbytes improvment

btsimonh:
Change StreamBytes to StreamBeBytesWriter, and improve wc_getbytes

This reduces allocation to a single allocaiton, but also theroetically allows the Bytes object to grow if content-length was unset.

(cherry picked from commit 936477f5977dd1a4f899498cc7cc0baf05b34a89)

## Description:

based on guidance from @s-hadinger, this re-works wc_getbytes in the Berry webclient to do a single memory allocation when content-length is present.

It will refuse lengths over 32kbytes (same as wc_getstring).

In theory, it will cope with content-length not present.  It would allocate 1K, and then expand the bytes object by (size received + 2K) each time the new data will not fit.
This feature is not tested and is not protected against massive incoming data, but in theory would throw on allocation failure (e.g. Bytes size limit reached

It is tested with small files (success) and with a 33k file (correctly throws a BE exception).

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_webclient.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_webclient.ino
@@ -654,21 +654,59 @@ extern "C" {
   }
 }
 
-class StreamBytes: public Stream
+
+// a stream which writes to the Bytes object on the top of the stack
+class StreamBeBytesWriter: public Stream
 {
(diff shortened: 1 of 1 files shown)


s-hadinger:
Perfect. I even forgot that you could send NULL to be_pushbytes and have a fresh new buffer.

Nice job!

PR #18853, opened 2023-06-11
Berry: add path.rmdir(path), path.mkdir(path)

btsimonh:
## Description:
This is a very minimial change which allows the existing FS module mkdir and rmdir to be used via the tasmota path module in berry.
Use case is managing files on SD card, or organising berry modules in folders.

It's only applicable to Berry and hence ESP32 devices.

Berry test code:
```
import path

print(path.mkdir('/testdir'))
print(path.mkdir('/testdir/test'))
var f = open('/testdir/test/test.txt', 'w')
print(f)
f.write('test test')
f.close()

print(path.remove('/testdir/test/test.txt'))
print(path.rmdir('/testdir/test'))
print(path.rmdir('/testdir'))

print(path.listdir('/testdir/test'))

#will return false
print(path.rmdir('/testdir'))

```

## Checklist:

### DIFF
--- a/lib/libesp32/berry_tasmota/src/be_path_tasmota_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_path_tasmota_lib.c
@@ -18,7 +18,10 @@
 #include "be_sys.h"
 #include <time.h>
 
+// from be_port.c
 extern int m_path_listdir(bvm *vm);
+extern int m_path_mkdir(bvm *vm);
+extern int m_path_rmdir(bvm *vm);
 
 static int m_path_exists(bvm *vm)
 {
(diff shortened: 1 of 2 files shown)


s-hadinger:
Looks interesting. How does it work integrated FS module and web interface?

btsimonh:
You have always been able to mkdir in Scripts.... so it's not any different.
The web interface currently displays folders.  And you can click into them, but not delete them.
But it also displays ALL folder content recursively - I don't think that's ok for my 32 Gbyte SD card FULL of pictures.
I'm looking at an update for that too, and adding folder delete.  But it's completely separate.  I may even be able to reduce the code size there (since that affects ESP8266 as well).  Just about to do some compiles and see how the ram and flash occupation changes.

The affect on flash and RAM of this PR could be interesting.
e.g. unless USE_SCRIPTS is defined, mkdir and rmdir from LittleFS may have been excluded from the link.
If that was the case, then the PR will have a larger impact than the 'minor' changes above indicate...
For me, for ESP32, it should not be a problem.
Your opinion?

@Jason2866 's comment here https://github.com/tasmota/Berry_playground/discussions/19 led me to think harder about how a PR could impact.... (thankyou jason).

PR #18859, opened 2023-06-12
UFILESYS UI changes, static file serving

btsimonh:
Allow recursive folder display to be disabled by 
`#define UFILESYS_NO_RECURSE_GUI`
Enable static file serving on /fs/ by 
`#define UFILESYS_STATIC_SERVING`

### `#define UFILESYS_NO_RECURSE_GUI`
If a filesystem has folders containing large quantities of files, it seems better not to recurse into the folder when displaying the filesystem content.
This define removes the recursion code - also saving 80 bytes of flash
Whether this should be a default is above my consideration, it saves flash and (in my opinion) improves the UI.  But who knows if someone is scraping the UI for file listings, which this would disturb?

### `#define UFILESYS_STATIC_SERVING`
_Note: costs 1844 bytes of flash and 40 bytes of RAM on esp8266 builds. Probably an equivalent load on esp32_
Static file serving from `/fs/` allows the UFS to host webpages/apps, or serve saved images from webcam, etc.
We do this to good effect in OpenBeken.
I would use this to develop a simple webcam timelapse viewer, so you don't even need to retrieve the images to view the timelapse sequence.
It only allows access to existing files, and cannot list folders.  A simple Berry function could provide that.
I'm sure this should not be enabled on esp8266.
It may be useful on esp32, assuming we have spare flash.

(I did have a much more comprehensive update, with GUI folder delete, and other 'nice' stuff, but could not produce a PR with that which reduced the flash size... so decided to stick to simple.)

Apologies for presenting two features in one PR.

## Checklist:
  - [?] Only one feature/fix was added per PR and the code change compiles without warnings


btsimonh:
pause for a moment ref static serving.  I missed something...
testing with SD card should make
/sd/folder/file
available.
but I find it at /folder/file.

btsimonh:
merge dev back in, just to kick the checks...  not sure why some were 'cancelled' last time....

s-hadinger:
Thanks. Just to be sure, it serves only what's in `/fs/` folder ? If yes, why name it `fs`? Why not `/www/` which would be more explicit?

btsimonh:
no specific reason...  just because it serves the whole filesystem.
It may be nice to have config, or only turn it on by command, and to specify both root url and served folder - not everyone will want to serve the whole FS the whole time.

p.s. jason said this was already possible, but after an hour of searching I find no reference to it.  Are you aware of anything for static file serving in TAS?

s-hadinger:
It was not possible natively, but somehow probably with some Berry scripting. If we believe it's a rare need, having compile option is a fair choice (not enabled by default).

However, I would strongly recommend to choose a more explicit directory than `fs`. Having a `fs` directory within a file-system is not explicit that it serves as static content. I suggested `www` because it's the default in Apache server.

btsimonh:
ahh...  no, you misunderstand.
e.g. the content I'm accessing is via a url which includes /fs/:
http://192.168.1.190/fs/sd/timelapse/index.html
so..  e.g. you can also access the autoexec.be at:
http://192.168.1.190/fs/autoexec.be
/fs/ maps to the ROOT of the filesystem, not the other way around.

This is not great because the complete filesystem is exposed.  But then it is already via the web ui....

s-hadinger:
Oh, this is even worse than what I thought. You should NOT expose the entire file-system, only files that are in a specific directory. Filesystem can contain moderately sensitive data (TLS private key, Zigbee encryption key...).

btsimonh:
tend to agree...
Should I remove that from this PR, leaving only the non-recursion, and think on a new PR where we can issue commands to expose parts people want?
Personally, I see a need to be able to serve the SD card (it's been running for days with >5000 images on the card so far, and no need to remove it...); but also could see a need to serve some files off flash.  These requirements conflict unless we share root or have it configurable; so i guess it must be made configurable.

s-hadinger:
I would just change the PR to server the content of `/www/` to prefix `/www/`. Can you store the images in the `www` directory?

btsimonh:
That last commit fixes it.  Sorry, one more commit to come - to remove the comments about it notw working..  one moment.

btsimonh:
fully working.... and may solve some other issues.  (why was is going through lib\libesp32\Zip-readonly-FS\src\ZipReadFS.cpp anyway?)

btsimonh:
auth on it's way.

btsimonh:
As per the comments in the source:
* like "UFSServe <fs path>,<url>[,<noauth>]"
* e.g. "UFSServe /sd/,/mysdcard/,1" - will serve the /sd/ fs folder as https://<ip>/mysdcard/ with no auth required
* e.g. "UFSServe /www/,/" - will serve the /www/ fs folder as https://<ip>/ with auth required if TAS has a password setup
* <noauth> defaults to 0 - i.e. the default is to require auth if configured
* it WILL serve on / - so conflicting urls could occur.  I beleive native TAS urls will have priority purely becasue this handler was added after the TAS handlers.
* you can serve multiple folders, and they can each be auth or noauth
*
* by default, it also enables cors on the webserver - this allows you to have 
* a website external to TAS which who's pages can access the files, else the browser refuses.

It's a lot more competent than the original one line PR, but is now a complete solution.
The only 'missing' part could be that it does not serve folder lists, but this is generally an optional feature in any webserver, and mostly turned off for good reason.  (although for me it's a nice-to-have, I'm not going to add it).

btsimonh:
I'm closing this because the PR does too much.

I will split into separate PRs.  

1/ to remove UI directory recursion unless a define is set to enable it.

2/ Add optional static serve, depending on a define.

br,  simon

chrifey:
Hi @btsimonh,
thanks for your work. I'm just trying the motion/timelapse features out, but do not fully understand/find how to implement this part, that seems to enable hosting the html file and the javascript code.
I tried to apply this PR to the latest development branch, but I'm not able to access files on the filesystem via the http server.
Do you have any hint or is this progressing at another place?

Thanks in advance for your feedback!
Christian

PR #18880, opened 2023-06-14
Add support for SGP41 TVOC/NOx Sensor

precurse:
## Description:
This adds support for the SGP41 TVOC/NOx sensor and uses the official driver from Sensirion. I was hoping to merge this with the existing SGP40 driver, but it seems that this is non-trivial at this time. 

Since this device uses the same I2C ID as the SGP40, it must be enabled with `i2cdriver 82` after disabling `i2cdriver98 0`

## Checklist:


lib/lib_i2c/arduino-i2c-sgp41/src/SensirionI2CSgp41.cpp
```diff
@@ -0,0 +1,184 @@
...
+        return error;
+    }
+
+    error = SensirionI2CCommunication::sendFrame(SGP41_I2C_ADDRESS, txFrame,
+                                                 *_i2cBus);
+    if (error) {
+        return error;
+    }
+
+    delay(50);
```
s-hadinger, line 107: I understand that it's supposed to be the official library. But such `delay()` are very problematic with Tasmota because they freeze evevery thing and makes the device unresponsive. It's considered a very bad practice in Tasmota to have any `delay()` for more than 5 ms.

Could it be refactored to make the read 50ms later in Tasmota event loop?

lib/lib_i2c/arduino-i2c-sgp41/src/SensirionI2CSgp41.cpp
```diff
@@ -0,0 +1,184 @@
...
+    SensirionI2CTxFrame txFrame =
+        SensirionI2CTxFrame::createWithUInt16Command(0x280E, buffer, 3);
+
+    error = SensirionI2CCommunication::sendFrame(SGP41_I2C_ADDRESS, txFrame,
+                                                 *_i2cBus);
+    if (error) {
+        return error;
+    }
+
+    delay(320);
```
s-hadinger, line 133: Same but worse problem here.

tasmota/tasmota_xsns_sensor/xsns_109_sgp41.ino
```diff
@@ -0,0 +1,201 @@
...
+int32_t voc_index_sgp41;
+int32_t nox_index_sgp41;
+
+uint16_t conditioning_s = 1;
+
+GasIndexAlgorithmParams voc_algorithm_params;
+GasIndexAlgorithmParams nox_algorithm_params;
+
+/********************************************************************************************/
+void sgp41_print_serial(uint16_t serialNumber[], uint8_t serialNumberSize) {
```
s-hadinger, line 57: Please remove any `Serial.print()`. We must not pollute the serial bus with such logs.

Use `AddLog()` to emit proper logs.

s-hadinger (commented):
Thanks for this PR. It's unfortunate that the official lib does `delay()` in the code which has very bad effect on the rest of Tasmota. If each sensor lib did the same, devices with multiple sensors could become unusable.

lib/lib_i2c/arduino-i2c-sgp41/src/SensirionI2CSgp41.cpp
precurse: Ahh, that makes sense! Thanks for the explanation. I'll see if I can get these fixed up

precurse:
Hey @s-hadinger I've managed to split out most of the functions so far. I used the 50ms function to check if there needs to be a reading, which will initiate a reading. Then every 1 second, it does the reading (if there's one pending). This works, for a bit of time, and then finally the whole I2C bus seems to crap out with a `{"I2CScan":"Error 4 at 0x01"}` error message when I run an `i2cscan`. 

I ran a logic analyzer to trace what was happening, and it seems at some random point, the SGP41 seems to respond with a `NACK` when the ESP8266 attempts to perform a read command, about 921ms (well after the 50ms minimum requirement) after the write command was issued.

Screenshot here: 
image

I'm still learning I2C, but wanted to check if this was behavior you've seen before. I'll keep troubleshooting, but it's very strange behavior. Thanks!

s-hadinger:
Thanks, I'm not familiar enough with I2C but this doesn't seem familiar

precurse:
> Thanks, I'm not familiar enough with I2C but this doesn't seem familiar

The SGP41 board I bought had a pull-up resistor for SDA and SCL, which I'm sure was causing the issue. I cut the traces to it on my board and so far it's been stable. Once I know it's stable, I'll update my PR with the new code :)

precurse:
Alright, I got it @s-hadinger . I've had this driver running for a while and it's running nicely. I removed the `delay()` and `Serial.print()` use. Let me know if there's anything else that you'd like to see fixed. 

I've also found a clever way of testing for an SGP40 vs SGP41, so I'm going to work on that next. The drivers are nearly identical, except for one function that SGP40 uses differently. This should allow for a universal SGP4x driver, without the need to manually handle drivers.

precurse:
Hey @s-hadinger  , the SGP41 sensor has been running for quite a few days and working reliably.

I just cleaned up some of the other Tasmota markdown files, JSON/web output to use i18n values, and confirmed the driver works on the ESP32 platform. I started to use SGP4X consistently throughout, which is the basis for the universal driver once I receive my SGP40 to confirm.

Let me know if anything else needs to be fixed up!

Jason2866:
> The SGP41 board I bought had a pull-up resistor for SDA and SCL, which I'm sure was causing the issue. The ESP8266 already has an internal I2C pull-up, so having 2 in parallel is not good. I cut the traces to it on my SGP41 board and so far it's been stable. Once I know it's stable, I'll update my PR with the new code :)

Thats weird. Since i2c is a bus it needs terminating. The internal pull up resistor of the esp8266 is so high that it is not enough to terminate the i2c bus correctly. AND it is no problem to have both active.

Jason2866:
@TD-er Do you have experience with the Sensirion SGP41?

precurse:
> Thats weird. Since i2c is a bus it needs terminating. The internal pull up resistor of the esp8266 is so high that it is not enough to terminate the i2c bus correctly. AND it is no problem to have both active.

I thought the D1 Mini had an internal I2C pull-up, but may have been mistaken then. It's possible the SHT3X board I have soldered has an I2C pull-up as well, since I'm reading 10K ohms across VCC and SDA/SCL.

Jason2866:
Thats explains it.

TD-er:
> @TD-er Do you have experience with the Sensirion SGP41?

I don't, but @tonhuisman did adapt the code for the SGP4x/SGP41 sensor for ESPEasy quite recently.
See: 

- https://github.com/letscontrolit/ESPEasy/blob/mega/src/_P147_SGP4x.ino
- https://github.com/letscontrolit/ESPEasy/blob/mega/src/src/PluginStructs/P147_data_struct.h
- https://github.com/letscontrolit/ESPEasy/blob/mega/src/src/PluginStructs/P147_data_struct.cpp

hoanglongutc:
> I thought the D1 Mini had an internal I2C pull-up, but may have been mistaken then. It's possible the SHT3X board I have soldered has an I2C pull-up as well, since I'm reading 10K ohms across VCC and SDA/SCL.
Hi,
Have you tested on nodemcu esp8266?

hoanglongutc:
Hi @precurse,
can you share a screenshot of the configure module
image

sfromis:
Should simply be the standard `I2C SCL` and `I2C SDA` pins, nothing special for each type of I2C sensor.

sfromis:
I2C works fine on my ESP8266 devices, including NodeMCU dev boards.

hoanglongutc:
> I2C works fine on my ESP8266 devices, including NodeMCU dev boards.

HI, I am using version "Program Version 13.1.0(sensors)" does it support sgp41 sensors?
I have connected and configured i2c, however I don't see the sensor appear
when running the i2cscan command:
MQT: stat/xxx/RESULT = {"I2CScan":"No devices found"}
Looking forward to your support. Thanks

barbudor:
"No device found" start very bad
Whether the right driver is there or not, whether it is enabled or not, i2cscan should return a result
Start by fixing your hardware

hoanglongutc:
> "Không tìm thấy thiết bị" khởi động rất tệ Cho dù có trình điều khiển phù hợp hay không, dù có bật hay không, i2cscan sẽ trả về kết quả Bắt đầu bằng cách sửa phần cứng của bạn

I also suspect the hard drive is faulty, I'm ordering another piece of hardware. Thank you

TD-er:
Some tips for I2C:

- Not too long wires (typically upto few decimeters)
- Use wires of same length for SDA and SCL
- Use pull-up resistors between SDA/VCC and SCL/VCC. Typically 4k7 or 10k.
- When using longer cables, you may consider twisted cable, where you twist SDA with GND in a pair and SCL with GND in another twisted pair. Something like UTP cable is fine for this.
- Lower I2C clock to 100 kHz when seeing lots of errors and/or using long cable.

hoanglongutc:
hi,
I changed a new sensor, when I ran i2scan the results came back, but tasmota still hasn't recognized the sensor. Look forward to the help
image

image
image

sfromis:
As visible on the I2CDriver output, the driver number 82 is not included in the binary you're running, You need a custom build including
`#define USE_SGP4X`
(The driver is also default-included for ESP32 builds)

hoanglongutc:
I tried using ESPHOME, everything worked normally (this proves that the hardware and connection are not the problem causing the error).
image

Jason2866:
> > https://tasmota.github.io/docs/Compile-your-build/
> 
> really confusing, when I rebuild the custom version i2cscan can't find the device. image

Disable all other i2c drivers. See https://tasmota.github.io/docs/I2CDEVICES/#management

hoanglongutc:
Based on your suggestion I solved the problem, thank you very much

deepbass909:
I believe the driver is sending out the wrong value for Domoticz. I've configured the value of Sensor idx 9 AirQuality for Domoticz, but receive the raw sensor-value, which isn't of any use in Domoticz. Instead the "SGP40 Air quality" is needed.

I believe the issue lines in  the file xsns_109_sgp4x.ino lines 222 - 224, that state the wrong parameter for Domoticz. If I'm reading the source code correctly, the value needs to be the calculated voc_index_sgp4x.


#ifdef USE_DOMOTICZ
      if (0 == TasmotaGlobal.tele_period) DomoticzSensor(DZ_AIRQUALITY, srawVoc);
#endif  // USE_DOMOTICZ

PR #18895, opened 2023-06-17
add path.isdir for berry, plus save ~87 bytes of flash in the process.

btsimonh:
## Description:

Adds path.isdir( ) for Berry.  It's the one feature which prevents invention of new html/javascript/berry file browsing techniques.  Of course, you can path.listdir, but you get the same result for files and folders (empty set? - maybe it should return nil for files?).

This was done by combining in a number of path functions into a single function, so saving code repetition, and saving a whole 87 bytes after the addition.

Tested AFTER the ZipFile fixes in https://github.com/arendst/Tasmota/pull/18894 which I found when trying to use only the zip filesystem, so may depend on that PR.

tested with the berry script found in https://github.com/arendst/Tasmota/pull/18894 with the path.isdir tests uncommented.

If it seems overly complex, then ditch it.

br, Simon

## Checklist:

### DIFF
--- a/lib/libesp32/berry_tasmota/include/be_port.h
+++ b/lib/libesp32/berry_tasmota/include/be_port.h
@@ -1,0 +1,10 @@
+/********************************************************************
+ * Tasmota - constants used in be_port and be_path_tasmota.lib.c
+ *******************************************************************/
+#define MPATH_MKDIR 0
+#define MPATH_RMDIR 1
+#define MPATH_LISTDIR 2
+#define MPATH_ISDIR 3
+#define MPATH_EXISTS 4
+#define MPATH_MODIFIED 5
+#define MPATH_REMOVE 6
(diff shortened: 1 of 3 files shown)


s-hadinger:
Thanks a lot

btsimonh:
(apologies for the splurge of PRs.  Having 10 days with the wife away, so have some time)

btsimonh:
@s-hadinger - Just a note that in testing I crashed TAS (no reboot, and 150m away :( ) by path.listdir() of a folder with probably >1000 files.  Difficult to test on my dev unit with serial, as it has no SD.
We should probably limit it to some small (25?) number and do something like add a '_more' filename at the end, just to prevent the crash/endless loop.
Then, plan to later implement an iterating folder lister, so people have the choice.

(I implemented a folder listing web endpoint in Berry - so if it finds the marker, it could easily swap to an iterating mode, and 'stream' the results to the caller).

I will PR the 'quick fix' for consideration after testing.
Looking at it, we're not be closing the files we're listing either.

PR #18906, opened 2023-06-19
Bugfix - prevent Berry path.listfile() crash by limiting to 25 files

btsimonh:
## Description:

During testing, listing a folder with >1000 files on SDCard crashed TAS, and it did not reboot (had to visit the unit).

Unfortunately I'm not able to get a serial log to see what was actually going on;   but path.listdir() is simply not designed to handle large quantities, and had a bug whereby it was not closing files.

So this PR limits it to 25 files, and adds a filename `_more` to the end if the limit is reached, as well as closing the files.

A longer term solution to listing larger folders would consist of an iteration over the folder from Berry.  More thought and discussion needs to go into this before implementation.

The PR is low risk, but does change the behaviour of path.listdir() by limiting the number of files returned - this could affect exisitng users, so please review and change the number 25 as required.

## Checklist:

### DIFF
--- a/lib/libesp32/berry_tasmota/src/be_port.cpp
+++ b/lib/libesp32/berry_tasmota/src/be_port.cpp
@@ -101,6 +101,7 @@ BERRY_API void be_writebuffer(const char *buffer, size_t length)
 // provides MPATH_ constants
 #include "be_port.h"
 extern "C" {
+#define BE_MAX_FILELIST_COUNT 25
     // this combined action is called from be_path_tasmota_lib.c
     // by using a single function, we save >200 bytes of flash
     // by reducing code repetition.
@@ -147,10 +148,11 @@ extern "C" {
                         File *dir = (File *)be_fopen(path, "r");
                         if (dir) {
                             switch (action){
-                                case MPATH_LISTDIR:
+                                case MPATH_LISTDIR:{
+                                    int maxcount = BE_MAX_FILELIST_COUNT;
                                     // fill out the list object
                                     dir->rewindDirectory();
-                                    while (1) {
+                                    while (maxcount--) {
                                         File entry = dir->openNextFile();
                                         if (!entry) {
                                             break;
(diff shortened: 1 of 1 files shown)


btsimonh:
the first casualty of this mod is my 'zap(folder)' function, which now can't complete as it only deletes the first 25 files in folder/subfolders!...  but it's better than it killing tas.

s-hadinger:
Thanks, although 25 seems a little low. Would it be safe to return at least 50 or 100 ?

btsimonh:
we could estimate, using the max filename len in the TAS file driver of 48, that 100 would take ~5kbytes.
That does not sound unreasonable.
How much memory do we consider 'normal' in a running Berry?

Also, what should we use as the 'terminator'?  Could we add 'nil' rather than a string - possibly more unique?

p.s. I'm guessing it's possible to add a 2nd optional parameter to listdir which could indicate 'start index', then we would not need a new API to get more...

Staars:
I tried to recreate your crash, but was unsuccessful.
After polluting the filesystem with 200 empty files, I could only see the truncation of the list like so:
```
15:06:57.230 ['/.settings', '/bspl0001.c', '/bspl0002.c', '/bspl0003.c', '/bspl0004.c', '/bspl0005.c', '/bspl0006.c', '/bspl0007.c', '/bspl0008.c', '/bspl0009.c', '/bspl0010.c', '/bspl0011.c', '/bspl0012.c', '/bspl0013.c', '/bspl0014.c', '/bspl0015.c', '/bspl0016.c', '/bspl0017.c', '/bspl0018.c', '/bspl0019.c', '/bspl0020.c', '/bspl0021.c', '/bspl0022.c', '/bspl0023.c', '/bspl0024.c', '/bspl0025.c', '/bspl0026.c', '/bspl0027.c', '/bspl0028.c', '/bspl0029.c', '/bspl0030.c', '/bspl0031.c', '/bspl0032.c', '/bspl0033.c', '/bspl0034.c', '/bspl0035.c', '/bspl0036.c', '/bspl0037.c', '/bspl0038.c', '/bspl0039.c', '/bspl0040.c', '/bspl0041.c', '/bspl0042.c', '/bspl0043.c', '/bspl0044.c', '/bspl0045.c', '/bspl004
```

Not necessarily related, I have rewritten the the MPATH_LISTDIR part to use the newer Arduino API and the speed of the above operation increased dramatically (from >5 seconds to about 700 milliseconds). 
Did you encounter a watchdog related crash?

btsimonh:
The TAS did not come back in a few minutes, and I did not have serial on it.  The particual cam unit is incredibly reliable normally, but having thought about it, it is entirely possible that I was doing as lot in Berry with each file, and it was just taking a VERY long time.

Staars:
Yes, got confused by the heat 😀 

The list is intact and has the correct size of 200.

btsimonh:
your PR also avoids the non closure issue, so I'll close this one in favour of yours.... :)
Also, it ended 'nicely'. (i.e. an exception).  Thanks for gripping this!

PR #18926, opened 2023-06-21
allow folders to be added to the filesystem

Staars:
## Description:

Adding files to the filesystem in the build process did require adding single file paths line by line in the env: section.
Now it is possible to just add a directory like so:
```
custom_files_upload         = ${env:tasmota32_base.custom_files_upload}
                              /Users/user/Developer/_test/fs/
```  

There is no error catching and the idea is to prepare one folder with the contents of the filesystem, that will be flashed together with the firmware. This could be added later, but should not be necessary.

@Jason2866 Plz take a look.

## Checklist:

### DIFF
--- a/pio-tools/post_esp32.py
+++ b/pio-tools/post_esp32.py
@@ -77,7 +77,10 @@ def esp32_build_filesystem(fs_size):
             else:
                 print("Failed to download: ",file)
             continue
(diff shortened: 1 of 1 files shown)


Jason2866 (approved):
LGTM

PR #18937, opened 2023-06-22
Berry added f-strings as an alternative to string formatting

s-hadinger:
## Description:

Berry add support for f-strings (similar to python). f-strings are converted internally by lexer to their format equivalent and then parsed as normally. It works as a transpiler, syntactically rewriting code before parsing.

Principle:
`f"begin  { <expr> : <format>}  end"` becomes `format("begin  %<format>  end", expr`

Example:
`f"2 + 2 equals {2+2:i}"` becomes `format("2 + 2 equals %i", 2+2)`

If the format is unspecified, `%s` is inserted:
`f'"foo" + "bar" equals {"foo"+"bar"}` becomes `format("\"foo\" + \"bar\" equals %s", "foo"+bar")`

If you accidentally insert `%` in the format, it is ignored:
`f"1/3 equals {1.0 / 3.0 :%.3f}"` becomes `format("1/3 equals %.3f", 1.0 / 3.0)`

To output `{` use double brackets: (note that `%` are automatically doubled)
`f"{{ 1% }}"` becomes `format("{ 1%% }")`

Use the `=` (equal sign) to also output the expression:
`f"{id=:i} {name=}"` becomes `format("id=%i name=%s", id, name)`

Note: f-strings takes less than 1KB of flash memory in the lexer

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 - Berry `string.format()` now automatically converts type according to format
 - Matter add friendly-name (NodeLabel) to each endpoint
 - Berry add global function `format` as a simpler syntax to `string.format`
+- Berry added f-strings as an alternative to string formatting
 
 ### Breaking Changed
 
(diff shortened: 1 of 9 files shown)


barbudor:
Berry is more and more pythonic 🤣 
Next step, remove `end` 😵‍💫

s-hadinger:
Nah...

PR #18975, opened 2023-06-26
quote path

ggottwald:
## Description:
building an environment from a path that contains whitespaces an error will be thrown.
Path: `/Volumes/My Code/Tasmota`
Error:
```
esp_idf_size: error: argument map_file: can't open '/Volumes/My': [Errno 2] No such file or directory: '/Volumes/My'
```

## Checklist:

### DIFF
--- a/pio-tools/metrics-firmware.py
+++ b/pio-tools/metrics-firmware.py
@@ -7,7 +7,7 @@ from os.path import join
 def firm_metrics(source, target, env):
     if env["PIOPLATFORM"] == "espressif32":
         import tasmota_metrics
-        env.Execute("$PYTHONEXE -m tasmota_metrics " + str(tasmotapiolib.get_source_map_path(env).resolve()))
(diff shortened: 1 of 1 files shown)


sfromis:
You should use path names without spaces. While you may fix one single place, that does not mean that spaces are supported.

ggottwald:
Didn't have any issue in a earlier version with those directories and also not after fixing this. In my opinion directories with spaces should be supported, since we are not in the 90s

Jason2866:
Platformio is not designed for whitespaces too. So as sfromis wrote, fix an obvious one and you will run someday in another one. Maybe not so easy to find.
The rule don't use whitespaces and special characters is still the way to go.
I will not merge. It is a reminder no to use whitespaces. Thx for the PR which is reminding me to add a check for whitespaces and special chars in Path and stop the compile process with a clear understandable error message.

Jason2866:
Just search for issues in Platformio github. You will find not just one.

Jason2866:
Thx, will merge added a warning with #18985 not to have whitespaces in path.

PR #18997, opened 2023-06-29
Berry `import strict` now detects useless expr without side effects

s-hadinger:
## Description:

Berry: new test added when compiling in strict mode `import strict`. Now the parser detects when statements or expression don't have any side effect, which means that they do nothing. It's most certainly a mistake.

This detects for example `self.attr 0` which should have been `self.attr = 0`, we was silently accepted as two useless expression.

```berry
> def f() 1 end
syntax_error: stdin:1: strict: expression without side effect detected

> def f(x) x 0 end
syntax_error: stdin:1: strict: expression without side effect detected

> def f(x) return x end     # this is ok
```

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking Changed
 - Berry `bool( [] )` and `bool( {} )` now evaluate as `false` (#18986)
+- Berry `import strict` now detects useless expr without side effects
 
 ### Changed
 - Matter support for temperature in Fahrenheit (`SetOption8 1`) (#18987)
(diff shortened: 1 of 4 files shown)


sfromis:
Good improvement :-) 

One funny edge case:
```
var a,b
def f() b:=a end
BRY: Exception> 'syntax_error' - input:1: strict: expression without side effect detected
```
Not a very intuitive message, when the walrus purpose is to actually have a side effect. Of course, not very useful to use walrus when a simple assignment would be the thing to do, hence I do not really see this as a "bug".

s-hadinger:
Nah this is a bug. I thought I marked a walrus assignment as a side effect.

s-hadinger:
Fixed in #18999

PR #19018, opened 2023-07-03
Berry `var` allowed in with walrus operator `:=`

s-hadinger:
## Description:

Berry, now `var` is allowed with `:=` walrus operator to create a local variable and assign it within an expression. The scope is the same as the expression.

```berry
def f()
  if (var a := 1)
    var b = 2
    print(a, b)
  end
  print(a)
end
```

Unit tests not done yet.

## Checklist:


s-hadinger:
There is still an issue, the following would wrongly compile:
```berry
def f()
  if (var a := a)
    print(a)
  end
end
```

PR #19024, opened 2023-07-03
Restart 4 option to do a ESP.deepSleep(0)

tjareson:
## Description:
Implementing a "restart 4" option in order to put tasmota into deep sleep without wake up time for usage in battery dependent use cases where deep sleep wake up should only happen by hardware triggers. 

https://github.com/arendst/Tasmota/discussions/19016

## Checklist:
  - [X ] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9

### DIFF
--- a/tasmota/tasmota_support/support_command.ino
+++ b/tasmota/tasmota_support/support_command.ino
@@ -1151,6 +1151,9 @@ void CmndRestart(void)
     }
     break;
 #endif  // ESP32
+  case 4:
+    ESP.deepSleep(0);  // deep sleep mode with only hardware triggered wake up
+    break;
   case -1:
     CmndCrash();    // force a crash
     break;


arendst:
I see your goal. I also see why you won't get feedback because it deepleeps to fast. 

What about a few seconds delay before executing deepSleep(0)? The correct location for the deepSleep would be in support_tasmota.ino like EspRestart().

I'll implement a more robust way than this.

tjareson:
Thanks, I was also thinking about delay, but wasn't sure if a simple delay() is a good practice for that?
Also I didn't really know how many seconds would make sure that it reliable works and not end up in a race condition where sometimes messaging would be fast enough, sometimes not.

arendst:
Another issue (I think) is the fact that the hardware wakes up anyway after 2 hours. Not sure if that's the case but I seem to remember that's why we have a seperate deepsleep driver to support longer periods of sleep. @stefanbode knows all about this..

arendst:
Just implemnted in latest dev branch. Use command ``restart 9`` to go in deepsleep. If all's well a hardware reset will wakeup and act as wakeup from deepsleep (just tested and did) I still don't know if it stays sleeping for hours.

stefanbode:
The esp8266 still was not able to do more than 74 min deepsleep. Even if documentation say it could. This was about 1 year ago. Did not test recently. Anyhow the wake-up is only 2sec. Nothing that will drain the battery. Should not worry you. Esp32 deepsleep was never really part of the development. It just uses esp8266 methods, even if there are more options.

stefanbode:
A hardware deepsleep should also work for days without wake-up.

PR #19028, opened 2023-07-04
MI32 legacy / Berry: add wifi_probe for BLE commissioning

Staars:
## Description:

In order to use Bluetooth for sending Wifi credentials in initial device setup using the open standard **Improv**, it is useful to test, if an AP is really accessible at runtime - without reboot in the running Bluetooth session.
This PR adds the command `wifi_probe(string ssid, string password)`to the BLE module of Berry, that will try to connect (typically when in Tasmotas AP mode). The result can later be checked with something like `tasmota.wifi()['ip']`, that should not be "0.0.0.0" or throw an exception in the case of success.

This function is not Bluetooth specific, but at the moment there is no other known use case. That's why it gets added to an optional non-core module. If needed it should move to another module in the future.

## Checklist:


sfromis:
Sounds like the same use case as the existing [WifiTest](https://tasmota.github.io/docs/Commands/#wifitest)<x> command, only available in AP mode, where it tests credentials and saves in a SSID slot.

Staars:
😀 .. that would of course be better!
I can check later and would gladly close this PR.

Staars:
Thanks to @sfromis !!

I did search on the Berry side but not in the commands.

Will close it due to unnecessary function duplication,

PR #19030, opened 2023-07-04
Added Windvane ADC Support

computergeek1507:
## Description:
Added support for Windvane direction based on ADC value.
Tested with the [Sparkfun Weather Kit ](https://www.sparkfun.com/products/15901) and a Wemos D1 mini.

## Checklist:


tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
```diff
@@ -707,6 +710,48 @@ void AdcShow(bool json) {
...
+        }
+        if(dir_val > 2.00 && dir_val < 2.20 ) {
+          wind_chr = D_TX20_SOUTH D_TX20_WEST;
+        }
+        if(dir_val > 3.00 && dir_val < 3.25 ) {
+          wind_chr = D_TX20_WEST;
+        }
+        if(dir_val > 2.8 && dir_val < 2.99 ) {
+          wind_chr = D_TX20_NORTH D_TX20_WEST;
+        }
```
barbudor, line 743: Number of tests can be halved in 2 by ordering properly and using else if
Also this leaves empty holes like 2.995 which falls in no cases

Just an opinion: I know you didn't invented those direction macros but wouldn't an angle easier to process downstream in a JSON data ?
I could make sense to use human readable directions in hte Web UI but angle in the JSON

sfromis: I agree that an angle would be preferable for JSON, and the web UI is not meant to be "the useful place" anyway.

And it does look odd that values 
0.75..1.20 is south
1.30..1.70 is northeast
2.00..2.20 is southwest
2.50..2.70 is north
I was expecting the analog value to have a circular dependence on the angle, but your values are jumping around. Is that really correct?

computergeek1507: It uses a resistor network with reed switches according to the data sheet, https://cdn.sparkfun.com/assets/d/1/e/0/6/DS-15901-Weather_Meter.pdf

tasmota/include/tasmota_template.h
```diff
@@ -181,6 +181,7 @@ enum UserSelectablePins {
   GPIO_OPTION_E,                       // Emulated module
   GPIO_SDM230_TX, GPIO_SDM230_RX,      // SDM230 Serial interface
   GPIO_ADC_MQ,                         // Analog MQ Sensor
+  GPIO_ADC_WIND_DIR,                         // Analog Wind Direction
```
arendst, line 184: This needs to move to the end of the list

tasmota/include/tasmota_template.h
```diff
@@ -434,6 +435,7 @@ const char kSensorNames[] PROGMEM =
   D_SENSOR_OPTION " E|"
   D_SENSOR_SDM230_TX "|" D_SENSOR_SDM230_RX "|"
   D_SENSOR_ADC_MQ "|"
+  D_SENSOR_ADC_WIND_DIR  "|"
```
arendst, line 438: Ditto.

tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
arendst: I suggest to convert the received values to degrees and use that to index in the winddirection text string like TX20 does but then without the NNE options. Then output both string and degrees again like TX20 does.

arendst: In addition move from ADC float ASAP to degree integer lowering code size considerably.

Jason2866:
@computergeek1507 Are you going to change the code?

PR #19059, opened 2023-07-06
do not auto-resize FS with embedded files in the firmware

Staars:
## Description:

At first boot after flash or resetting the boot count to 0 with `reset 99` Tasmota will try to automatically resize the filesystem partition to the maximum size.
This is problematic when files are embedded in the firmware binary, because they are erased in this process which renders the whole file embedding useless.
Until now it was necessary to flash with fixed partition values to prevent this deletion.

This PR checks, if there are more files than `.settings` in path `/` and will not call the resize function.

This is not perfect, as we have to live with the smaller FS, but if a larger FS is needed, then another firmware (without embedded files) should be chosen anyway.

For the future it would be super nice to let the FS grow non-destructive in Tasmota or add some firmware patching magic to esptool.

@s-hadinger Please check, if there is a better solution to accomplish this.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_7_berry_embedded.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_7_berry_embedded.ino
@@ -43,12 +43,18 @@ const char berry_prog[] =
   "def log(m,l) tasmota.log(m,l) end "
   "def load(f) return tasmota.load(f) end "
 
-  // try to resize FS to max at first boot
+  // try to resize FS to max at first boot, if not already populated with embedded files
   // "tasmota.log('>>> bootcount=' + str(tasmota.settings.bootcount), 2) "
   "if tasmota.settings.bootcount == 0 "
-    "import partition_core "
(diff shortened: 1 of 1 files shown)


s-hadinger:
@Jason2866 Do you know if there is a way to increase the partition size of LittleFS without erasing the partition?

Staars:
Yes, that sounds like a potential solution.
I assume, that for our use case the situation should be somewhat easier as we have a fresh FS with only one added file. (.settings) so that the layout at this point in time should be linear.
Could it be enough to just increase the block count in the super block according to the new partition scheme?

Jason2866:
Well, looks nice. But this would mean we need to fork [IDF LittleFS](https://github.com/joltwallet/esp_littlefs.git), since it is a wrapper which uses the orig. LittleFS.
Will have a look how much effort it is to do and to maintain.

Staars:
I am almost sure, that we can expand the partition and the FS purely in Berry while keeping the files.
Will try to prove my theory soon.

Let‘s close this one.

PR #19111, opened 2023-07-14
fix: remove minimal password length in UI

okainov:
For some reason there is a limit on minimal length for MQTT password in UI. However it makes no sense as there is no such restriction in backend and MQTT user can have shorter passwords on the broker side.

Before this change it could be workarounded with Chrome Dev Console, but the right solution would be to remove this limit at all

## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
@@ -1950,7 +1950,7 @@ const char HTTP_FORM_MQTT1[] PROGMEM =
   "<p><b>" D_CLIENT "</b> (%s)<br><input id='mc' placeholder=\"%s\" value=\"%s\"></p>";
 const char HTTP_FORM_MQTT2[] PROGMEM =
   "<p><b>" D_USER "</b> (" MQTT_USER ")<br><input id='mu' placeholder=\"" MQTT_USER "\" value=\"%s\"></p>"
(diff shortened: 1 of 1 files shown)


arendst:
I would agree if not for the many conversations went on over this in the past.

As you noticed, the backend works just fine. Let's keep it this way.

okainov:
>I would agree if not for the many conversations went on over this in the past.

Could you please point me which conversations do you mean? I'm not aware of history of the project and rather looking from end-user point of view.


>As you noticed, the backend works just fine. Let's keep it this way.

I don't get this. Keeping it like it is keeps the inconsistency and makes it really inconvenient to use for those who have short passwords without any real reason...

okainov:
>Short passwords are a bad practice. The Tasmota WebUI should not promote nor make it easy to implement bad practices.

Sure, but it has nothing to do with this issue.

Two comments:
- If you say so, then Tasmota backend should also report a corresponding error and don't behave inconsistently with the frontend.
- In context of MQTT, Tasmota is not "promoting" anything but serves as a **client** for *existing* MQTT servers. If the server was configured in this or another way, this shouldn't really matter for Tasmota, it should be able to work with whatever valid servers and credentials as long as they're valid as per MQTT spec. And, BTW, [per MQTT spec](https://www.hivemq.com/blog/mqtt-security-fundamentals-authentication-username-password/#:~:text=you%20can%20send%20a%20username%20without%20password) there can be users without password at all which should be also accepted.

s-hadinger:
Theo already responded. `MqttPassword` command works as you expect.

PR #19134, opened 2023-07-17
New DEEPSLEEP topic for HA + Battery Level % support

stefanbode:
To allow better integration into HA LWT topic will report details of sleep status: 16:54:04.388 MQT: hm/tele/ESP_3284D1/LWT =
DeepSleep

Additionally introduce batterylevel from 0..100 % to be able to send to HA and similar integrations. batterypercentage must be set in a rule or something similar because ways to calculate can be very different. value = 101 does meant powerplug and value is not shown. This is the default

## Description:

## Checklist:

### DIFF
--- a/tasmota/include/tasmota_types.h
+++ b/tasmota/include/tasmota_types.h
@@ -733,8 +733,8 @@ typedef struct {
   uint16_t      artnet_universe;           // 734
   uint16_t      modbus_sbaudrate;          // 736
   uint16_t      shutter_motorstop;         // 738
-
-  uint8_t       free_73A[3];               // 73A
+  uint8_t       battery_level_percent;     // 73A
+  uint8_t       free_73B[2];               // 73B
 
   uint8_t       novasds_startingoffset;    // 73D
   uint8_t       web_color[18][3];          // 73E
--- a/tasmota/tasmota_support/settings.ino
+++ b/tasmota/tasmota_support/settings.ino
@@ -980,6 +980,7 @@ void SettingsDefaultSet2(void) {
   if (Settings->sleep < 50) {
     Settings->sleep = 50;                // Default to 50 for sleep, for now
   }
+  Settings->battery_level_percent = 100;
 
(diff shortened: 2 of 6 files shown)


stefanbode:
Good objection. The intend is separate between an offline device and a device just sleeping. After the deepspleeptime the device is offline if it does not connect intime. 

Yes, it is a breaking change because you now only get offline on a crashing device and not on a battery powered deepsleep. What about setting the LWT to sleep and report the deepsleep parameters to another topic. Still a breaking change but then not suddenly a JSON.

barbudor:
AFAIK you can't change the payload for the last will message: it is set on connect and then if the connection timeout triggers on the broker, it will be published as expected, unless you have closed properly the connection which voids the last will message. 
Of course you can publish whatever you want to the topic before closing the connection.
To be tested...

stefanbode:
@barbudor : This is what I was thinking, too. But the LWT is normally only for a unforeseen disconnects. On a graceful disconnect of the client the LWT "Offline" should not raise. In TASMOTA there is a hack that "Offline" is send in any case, but you can disable this hack and ensure the "last send topic" stay if you gracefully disconnect the client. http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718090  .

stefanbode:
LWT is back to "Offline" until further discussion and decision.

barbudor (commented):
Hello @stefanbode
This line breaks build where Tas Discovery is not enabled
Could you please wrap it in a #ifdef? 
I currently don't have access to a computer to make a PR
Thanks

Just in case this comment is not pointing to the right line, it's about the call to 'TasDiscovery()' in the CmndDeepSleepTime()

Thanks

arendst:
I'll fix it now. Thx.

arendst:
Fixed!

barbudor:
Thanks a lot Theo

PR #19138, opened 2023-07-17
More user-friendly defaults for DisplayMode and DisplayDimmer

sfromis:
## Description:

Changing default for DisplayMode from 1 to 0. Users are getting confused by the display doing something they did not ask for (and not being aware of DisplayMode). This got worse with LVGL/HASPmota displays becoming common, with users having much less reason to dive into the old DisplayXxxxx commands. And it may even be hard to see that it is even a display of time/date causing the display to flicker.

Changing default for DisplayDimmer from 10% to 50%. The low brightness of 10% is not always easy to see, especially in daylight. 50% is generally better, while not going "full blast" with 100%.

**Related issue (if applicable):** fixes #[18996](https://github.com/arendst/Tasmota/discussions/18996),

## Checklist:

### DIFF
--- a/tasmota/tasmota_support/settings.ino
+++ b/tasmota/tasmota_support/settings.ino
@@ -1289,12 +1289,12 @@ void SettingsDefaultSet2(void) {
 
   // Display
 //  Settings->display_model = 0;
-  Settings->display_mode = 1;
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks

PR #19160, opened 2023-07-21
Bugfix battery reporting

stefanbode:
fix issue with init value
value 255 reported on new devices
## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_support/support_tasmota.ino
+++ b/tasmota/tasmota_support/support_tasmota.ino
@@ -831,7 +831,7 @@ void MqttShowState(void)
 
 
 // Battery Level expliciet for deepsleep devices
-if (Settings->battery_level_percent != 101) {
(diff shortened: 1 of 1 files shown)


arendst:
Are you sure this was a new device? It should have been 0 in that case.

Anyway. The only solution for new settings values initialization is adding to the settings.ino file based on version number. I'll add it now making this PR obsolete.

arendst:
Implemented in v13.0.0.3

stefanbode:
Thanks Theo. When I'm searching for I never find the section to init the values. I was surprised it was 255. I just updated one of my devices and saw "batterypercent:255". Anyhow the change is also fine because valid values are 0...100 and 101 for a non battery device. anything above fly in by magic.

stefanbode:
The init value must be 101 in this case!!!!

arendst:
It is.

PR #19206, opened 2023-07-28
Add influxdb_period variable

pkkrusty:
influxdb_period could be set in userconfig_override, but would not populate.

## Description:

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.11


arendst:
SET_INFLUXDB_PERIOD is never used in Tasmota. Why this change?

arendst:
Nope. It doesn't work that way. A setting variable to be used in userconfig_override needs a code construct like user definable TELE_PERIOD at the cost of code size increase only used during initial configuration.

sfromis:
The way to have IFXPeriod included in `user_config_override.h` would be to have it as a command in `USER_BACKLOG`, which is executed once when initializing the configuration for the first time. Not tested if that would work.

pkkrusty:
Will try that, thanks

Issue #16985, opened 2022-11-03
Power switches and remains on after temperature sensor is not alive


blacknell:
### PROBLEM DESCRIPTION
After the reported temperature is detected as not alive the power switch is turned on, resulting in huge energy costs.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): Sonoff Elite POW320D
- [X] Tasmota binary firmware version number used: 12.2.0
  - [ ] Pre-compiled
  - [X] Self-compiled
- [ ] Flashing tools used: _____
- [X] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
07:31:36.841 SRC: Backlog
07:31:36.845 CMD: Grp 0, Cmd 'MODULE', Idx 1, Len 0, Pld -99, Data ''
07:31:36.856 MQT: stat/tasmota_CC7FF4/RESULT = {"Module":{"0":"Sonoff POWR320D"}}
07:31:37.092 SRC: Backlog
07:31:37.097 CMD: Grp 0, Cmd 'GPIO', Idx 1, Len 3, Pld 255, Data '255'
[...]

PR #19279, opened 2023-08-08
Improvements to thermostat debug output

blacknell:
## Description:

It's very fairly difficult to figure out why the controller is behaving the way it does, so these changes to log lines only make this a lot easier. Also, debug output can now be grep'd for THE: to speed up logging analysis.

add: log prefix 'THE' added to debug output
add: error log message when sensor is detected as not alive
add: debug output of main controller parameters when thermostat enabled
add: update DEBUG_THERMOSTAT to only control the virtual switch (debug output can now be used when this flag has not been compiled in)
add: human readable output on debug lines eg 100ths of degrees also displayed as fractional degrees
fix: comment typos and small grammatical changes for clarity

**Related issue (if applicable):** fixes # N/A relates #16985 

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.11

### DIFF
--- a/tasmota/include/i18n.h
+++ b/tasmota/include/i18n.h
@@ -824,6 +824,7 @@
 #define D_LOG_TCP "TCP: "          // TCP bridge
 #define D_LOG_BERRY "BRY: "        // Berry scripting language
 #define D_LOG_LVGL "LVG: "         // LVGL graphics engine
+#define D_LOG_THERMOSTAT "THE: "   // Thermostat driver
 
 /********************************************************************************************/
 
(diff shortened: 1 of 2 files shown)


blacknell:
Suggest maintainers do a squash merge if this isn't already standard practice.

blacknell:
In my view `DEBUG_THERMOSTAT` should be changed.

It served 2 purposes in the current code which were to

- enable debug output lines (which were in themselves controlled by runtime `LOG_LEVEL_DEBUG` by the end user)
- compile in a virtual switch (which was only a Domoticz topic/payload output anyway).

I've changed this so that `DEBUG_THERMOSTAT` only compiles in the virtual switch but in my opinion it should be renamed to `ENABLE_THERMOSTAT_DEBUG_VIRTUAL`. Some people might consider this breaking (even though it's debug only) so I didn't make this change. I'm happy to add a commit for this if the maintainers agree.

arendst:
Thx.

PR #19015, opened 2023-07-03
Add new webcam driver.  

btsimonh:
## Description:

To use the old one, define USE_WEBCAM_LEGACY

This is a re-submisison of the webcam PR, with the implementation of it changed to by default replace the old one, making this PR generate a directly testable and usable binary.

Runs all camera stuff from a task.

Never fails at startup on my (very bad) esp32cam board.

Enables motion detect, Timelapse, direct image saving etc. from Berry scripts and as TAS commands.

Increased reliability. Increased flexibility.

Documented here: https://github.com/btsimonh/TasmotaBerryPLay/blob/master/tasdocs/esp32cam.md

Example Timelapse and Motion [here](https://github.com/btsimonh/TasmotaBerryPLay/tree/master/timelpase) (needs more description, but works very well. Can use staticfileserving from the other PR to provide direct viewing and download of complete image sets.

Should work with Scripts, but not tested for a while (@gemu2015 ?).

Checklist:

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [x] The code change is tested and works with Tasmota core ESP32 V.2.0.10
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #19280, opened 2023-08-08
Add new webcam driver over latest dev.

btsimonh:
## Description:

To use the old one, define USE_WEBCAM_LEGACY

This is a re-submisison of the webcam PR, with the implementation of it changed to by default replace the old one, making this PR generate a directly testable and usable binary.

Runs all camera stuff from a task.

Never fails at startup on my (very bad) esp32cam board.

Enables motion detect, Timelapse, direct image saving etc. from Berry scripts and as TAS commands.

Increased reliability. Increased flexibility.

Documented here: https://github.com/btsimonh/TasmotaBerryPLay/blob/master/tasdocs/esp32cam.md

Example Timelapse and Motion [here](https://github.com/btsimonh/TasmotaBerryPLay/tree/master/timelpase) (needs more description, but works very well. Can use staticfileserving from the other PR to provide direct viewing and download of complete image sets.

Should work with Scripts, but not tested for a while (@gemu2015 ?).

Checklist:

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
@@ -19,7 +19,8 @@
 
 #ifdef ESP32
 #ifdef USE_WEBCAM
-#ifdef USE_WEBCAM_LEGACY
+// defining USE_WEBCAM_V2 will use xdrv_81_esp32_webcam_task.ino instead.
+#ifndef USE_WEBCAM_V2
 
 /*********************************************************************************************\
  * ESP32 webcam based on example in Arduino-ESP32 library
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
@@ -19,6 +19,8 @@
 
 #ifdef ESP32
 #ifdef USE_WEBCAM
+#ifdef USE_WEBCAM_LEGACY
+
 /*********************************************************************************************\
  * ESP32 webcam based on example in Arduino-ESP32 library
  *
@@ -1531,5 +1533,6 @@ bool Xdrv81(uint32_t function) {
   return result;
 }
 
+#endif  // USE_WEBCAM_LEGACY
 #endif  // USE_WEBCAM
 #endif  // ESP32
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam_task.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam_task.ino
@@ -22,6 +22,8 @@
 // defining USE_WEBCAM_V2 will this file rather than xdrv_81_esp32_webcam.ino
 #ifdef USE_WEBCAM_V2
 
+#undef WEBCAM_DEV_DEBUG
+
 /*********************************************************************************************\
  * ESP32 webcam based on example in Arduino-ESP32 library
  *
@@ -360,7 +362,9 @@ public:
 
 localOV2640Streamer::localOV2640Streamer(SOCKET aClient, int width, int height) : CStreamer(aClient, width, height) {
   clearframe();
-  AddLog(LOG_LEVEL_DEBUG,PSTR("CAM: Created RTSP streamer width=%d, height=%d"), width, height);
+#ifdef WEBCAM_DEV_DEBUG  
+  AddLog(LOG_LEVEL_DEBUG,PSTR("CAM:RTSP w%d h%d"), width, height);
+#endif
 }
 void localOV2640Streamer::setframe(BufPtr ptr, int len) {
   f_ptr = ptr;
@@ -492,7 +496,9 @@ bool pic_alloc(struct PICSTORE *ps, int width, int height, int jpegsize, int for
   int orglen = 0;
 
   if (!ps){
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_ERROR, PSTR("CAM: pic_alloc ps null"));
+#endif
     return false;
   }
 
@@ -505,7 +511,9 @@ bool pic_alloc(struct PICSTORE *ps, int width, int height, int jpegsize, int for
   }
 
   if (!len){
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_ERROR, PSTR("CAM: pic_alloc invalid format or len 0"));
+#endif    
     return false;
   }
 
@@ -696,13 +704,17 @@ void WcInterrupt(uint32_t state) {
   if (state) {
     // Re-enable interrupts
     cam_start();
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: cam_start()"));
+#endif    
     Wc.disable_cam = 0;
   } else {
     // Stop interrupts
     Wc.disable_cam = 1;
     cam_stop();
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: cam_stop()"));
+#endif
   }
 }
 
@@ -745,7 +757,9 @@ bool WcPinUsed(void) {
 //    }
   }
 
+#ifdef WEBCAM_DEV_DEBUG  
   AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: i2c_enabled_2: %d"), TasmotaGlobal.i2c_enabled_2);
+#endif
 
   if (!PinUsed(GPIO_WEBCAM_XCLK) || !PinUsed(GPIO_WEBCAM_PCLK) ||
       !PinUsed(GPIO_WEBCAM_VSYNC) || !PinUsed(GPIO_WEBCAM_HREF) ||
@@ -787,7 +801,9 @@ void WcFeature(int32_t value) {
       wc_s->set_reg(wc_s, 0x103, 0xff, 0xcf);   // COM1: Allow 7 dummy frames
       break;
   }
+#ifdef WEBCAM_DEV_DEBUG  
   AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: Feature: %d"), value);
+#endif  
 }
 
 void WcApplySettings() {
@@ -866,8 +882,9 @@ void WcSetDefaults(uint32_t upgrade) {
     WcSetMotionDefaults();
   #endif
 
+#ifdef WEBCAM_DEV_DEBUG  
   AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: Defaults set"));
-
+#endif
   if (Wc.up) { WcApplySettings(); }
 }
 
@@ -908,7 +925,9 @@ uint32_t WcSetup(int32_t fsiz) {
 
   TasAutoMutex localmutex(&WebcamMutex, "WcSetup", 200);
 
+#ifdef WEBCAM_DEV_DEBUG  
   AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: WcSetup"));
+#endif  
   // if 15, make it -1, so disableing
   if (fsiz >= FRAMESIZE_FHD) { fsiz = -1; }
 
@@ -924,7 +943,9 @@ uint32_t WcSetup(int32_t fsiz) {
 
   if (Wc.up) {
     esp_camera_deinit();
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: Deinit"));
+#endif    
     //return Wc.up;
   }
   Wc.up = 0;
@@ -953,12 +974,16 @@ uint32_t WcSetup(int32_t fsiz) {
     if(TasmotaGlobal.i2c_enabled_2){              // configure SIOD and SIOC as SDA,2 and SCL,2
       config.sccb_i2c_port = 1;                   // reuse initialized bus 2, can be shared now
       if(config.pin_sccb_sda < 0){                // GPIO_WEBCAM_SIOD must not be set to really make it happen
+#ifdef WEBCAM_DEV_DEBUG  
         AddLog(LOG_LEVEL_INFO, PSTR("CAM: Use I2C bus2"));
+#endif        
       }
     }
     config.pin_pwdn = Pin(GPIO_WEBCAM_PWDN);       // PWDN_GPIO_NUM;
     config.pin_reset = Pin(GPIO_WEBCAM_RESET);    // RESET_GPIO_NUM;
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: Template pin config"));
+#endif    
   } else if (Y2_GPIO_NUM != -1) {
     // Modell is set in camera_pins.h
     config.pin_d0 = Y2_GPIO_NUM;
@@ -977,7 +1002,9 @@ uint32_t WcSetup(int32_t fsiz) {
     config.pin_sscb_scl = SIOC_GPIO_NUM;
     config.pin_pwdn = PWDN_GPIO_NUM;
     config.pin_reset = RESET_GPIO_NUM;
+#ifdef WEBCAM_DEV_DEBUG  
     AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: Compile flag pin config"));
+#endif    
   } else {
     // no valid config found -> abort
     AddLog(LOG_LEVEL_DEBUG, PSTR("CAM: No pin config"));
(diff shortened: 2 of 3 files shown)


btsimonh:
see https://github.com/arendst/Tasmota/pull/19015 for info on the previous PR.

The cam ran on that PR for 28 days (and is still running on that).

This PR is the exact same driver, pasted on top of current dev.  I have not done extensive testiing on latest dev....

@gemu2015 - I would appreciate your thoughts and feedback, as the originator of the original driver, and a great inspiraiton.  I beleive I've solved some of the underlying issues with some boards (lack of reset line), but also made signifiant improvement in reliability by accessing stuff only from one place.  Also, I hope it still fully supports your Scripts based stuff - but now does support Berry, with a few extra features for reading poictures, etc. directly which will enable use with possible current/future AI technologies, as well as sending images to remote locations (e.g. google drive, or ftp).  Happy to discuss any time.

gemu2015:
@btsimonh 

i am busy with other things, but made a short check with my s3cam.

1. got compile error when ENABLE_RTSPSERVER is not defined.
2. with task driver i didn't get the cam initialized 
3. legacy driver got picture without script, but none with script

have to check in more detail what is wrong, but will take some time

btsimonh:
thankyou @gemu2015 - I will fix the compilation thing with RTSP disabled tonight, and re-check and post my user config.
What was the user config you used?
Ref the legacy driver, there have been no changes for some time...  there were some mutexes added some time back - maybe these affected the script video display?  I'll try to review the commits...

gemu2015:
now also compiled for esp32 cam and it works with and without script.
on my esp32s3 legacy also works now.
however task version crashes and i later have to open the case for serial reflash.
my s3 cam is however a very special build since it also uses bidirectional audio.
on power on i hear strange cracklings from the speaker as if i2s is failing.
probably your version is interfering with i2s ?
if i find the time i will build a version without audio and try again

btsimonh:
hmm..  it could be that by putting the cam in a separate core/thread, it's starving the i2s audio processing.  Certainly, the cam does run 'as fast as it can' - e.g. at max framerate if not transferring the video anywhere.
Check the shared I2C enabling too - not sure if this has any impact, but Theo enabled it by default a few months ago.
"task version crashes" - if you have serial on it, maybe we get a crashdump?

btsimonh:
@gemu2015 - I just ordered the required audio modules (I hope they are right).
https://www.amazon.co.uk/dp/B07TSC1QCH?psc=1&smid=AWLU8WJU8S0VS&ref_=chk_typ_imgToDp
https://www.amazon.co.uk/dp/B08BCHHZPN?psc=1&smid=AWLU8WJU8S0VS&ref_=chk_typ_imgToDp
You can never have too many toys!

gemu2015:
does not seem to be related to i2s however. disabled all audio code and got the result below
Seems to be an S3 implementation incompatibility.
S3 cam driver is very different from standard ESP32. (see CAM driver source code)

```
00:00:00.000 HDW: ESP32-S3 (PSRAM)
00:00:00.006 UFS: FlashFS mounted with 10244 kB free
00:00:00.010 CFG: Loaded from File, Anzahl 18
00:00:00.016 QPC: Count 1
00:00:00.017 SCR: ufilesystem found
00:00:00.024 Projekt tasmota - webcam Version 13.0.0.2(tasmota)-2_0_10(2023.08.12 12:38:01)
00:00:00.607 WIF: verbinden mit AP1 WLAN-7490 Channel 1 BSSId CC:CE:1E:61:99:91 in Modus 11n als sonoff-6016...
00:00:02.477 WIF: verbunden
00:00:02.728 HTP: Web-Server aktiv bei sonoff-6016 mit IP-Adresse 192.168.188.88
12:46:13.710 MQT: Verbindungsversuch...
12:46:13.858 MQT: verbunden
12:46:13.859 MQT: tele/sonoff/LWT = Online (beibehalten)
12:46:13.861 MQT: cmnd/sonoff/POWER = 
12:46:13.863 MQT: tele/sonoff/INFO1 = {"Info1":{"Module":"ESP32S3","Version":"13.0.0.2(tasmota)","FallbackTopic":"cmnd/webcam_fb/","GroupTopic":"cmnd/tasmotas/"}}
12:46:13.874 MQT: tele/sonoff/INFO2 = {"Info2":{"WebServerMode":"Admin","Hostname":"sonoff-6016","IPAddress":"192.168.188.88","IP6Global":"","IP6Local":"fe80::6ab6:b3ff:fe21:3780"}}
12:46:13.896 MQT: tele/sonoff/INFO3 = {"Info3":{"RestartReason":"Software reset CPU","BootCount":6}}
12:46:17.079 QPC: Reset
12:46:17.934 MQT: tele/sonoff/STATE = {"Time":"2023-08-12T12:46:17","Uptime":"0T00:00:08","UptimeSec":8,"Heap":246,"SleepMode":"Dynamic","Sleep":50,"LoadAvg":19,"MqttCount":1,"Wifi":{"AP":1,"SSId":"WLAN-7490","BSSId":"CC:CE:1E:61:99:91","Channel":1,"Mode":"11n","RSSI":56,"Signal":-72,"LinkCount":1,"Downtime":"0T00:00:03"}}
12:46:53.549 UPL: Datei Config_s3cam_12.0.2.3-2.dmp
12:46:53.586 UPL: erfolgreich 4096 bytes
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x40056ea0  PS      : 0x00060e30  A0      : 0x8002f1fd  A1      : 0x3fceb9d0  
A2      : 0x00000000  A3      : 0x00000000  A4      : 0xffffffff  A5      : 0x00000000  
A6      : 0x00000000  A7      : 0x00000002  A8      : 0x00000000  A9      : 0x3fcebaf0  
A10     : 0x3fce9710  A11     : 0x00000000  A12     : 0x3fc9dc38  A13     : 0x00000000  
A14     : 0x00000001  A15     : 0x00000000  SAR     : 0x0000001c  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000000  LBEG    : 0x40055601  LEND    : 0x4005560e  LCOUNT  : 0xfffffff4  


Backtrace: 0x40056e9d:0x3fceb9d0 |<-CORRUPTED
```

btsimonh:
Yep, it uses rom based jpg routines - so some of my modified conversion routines may not 'fit' (but should not be run in the above log).
Also uses a different clock scenario, and a different gpio setup routine.

1/ would you be happy with a `#define` which enforces legacy for S3 until it can be resolved (probably when you have more time...)?
2/ the most likely parts are the assertion of pwdn pin in wcSetup.  Try commenting those?
Do the addresses in the core dump take us anywhere useful in the map file?

br,  Simon

btsimonh:
fixed without RTSP build, and a small change around us of pin_pwdn (like, don't use it if it's not configured...).

btsimonh:
@gemu2015 - do I need some special config for it to find layer3.h when USE_SHINE is enabled for audio?
Do I need USE_SHINE?

gemu2015:
@btsimonh 
USE_SHINE is an mp3 encoder needed for mp3 coding of a microphone if you want mp3 (unidirectional) streaming or storing of mp3 audio files.

Audio Bridge does not require MP3, it uses binary streaming over UDP (bidirectional, half duplex).

btsimonh:
I'm thinking my first usecase for audio would be just bringing it to the web interface - seems this is supported, but uses the mp3 encoder :(.  I'll try again with a full clean & build, and if that fails, examine the build files in more detail (complete rebuild really stresses my 10 yr old 8gb i7).  My second use case MAY be to look at streaming the audio with the video over RTSP, but that's much more involved, coordinating the video and audio, capturing the timestamps, and modifying the rtsp library for the muxing.  New toys should arrive later today.

gemu2015:
Video with audio would be very cool.  
ESP32 MP3 audio transfer slows down video to 1/3 frame rate
while on ESP32-S3 there is no remarkable reduction in frame rate. (s3-cam should be your next toy  :-))

i was working on a door station with bidirectional audio which works very well with audio.
i adapted the fastest display i could get with 20 ms full display update rate, but could not get
a fluent video display from the cam to the display, only about 1-3 frames per second.

i postponed this now and leave it to other folks.




i currently postponed this

gemu2015:
#define USE_WEBCAM
#define USE_WEBCAM_LEGACY
#define MP3_MIC_STREAM
#define USE_I2S_MIC
#define USE_SHINE
#define USE_I2S_AUDIO
#define MP3_STREAM_PORT 83
#define MIC_CHANNELS 2
#define I2S_BRIDGE

remark:
SHINE encoder only works with 32khz and up, so we either must define 2x16 or 1x32 kHz channels

btsimonh:
@gemu2015 - what board do you build for S3?  Seems if I just build S3, I get no PSRAM, but if I build tasmota32s3-qio_opi-all I get lgvl.h errors from scirpter.  Now trying a modified tasmota32s3-qio_opi-all....  I have it 'working' but seems wifi and cam interfere, but maybe because no PSRAM.

btsimonh:
@gemu2015 - your opinion on S3 webserving timings appreciated.
I killed my task, and it's still really bad - see discussion in general.  My 'normal' esp32 is still not as good as a ESP8266....  what kind of HTTP response time did you get on your original firmware?  e.g. just going from main menu to consoles?

btsimonh:
@gemu2015 - pls try again.
For me, my wifi on S3 is terrible - If I attenuate it a bit (finger on antenna), it becomes near perfect.
But cam now works quite well when wifi is ok.

gemu2015:
@btsimonh 
compiled with your last build and s3cam now works with and without script. i am having no wifi issues at all.
however also the old legacy driver never had any problems with my s3 cam

according to wifi issues keep in mind that the cam clock line is a severe source of interference with the wifi signal.  all my cam problems i had also with esp32 cams came from that line (provided a stable power supply)

i also have had wifi issues with my rgb display (sunton) also from interference with high speed display clock.  i had to adjust display clock to minimize this effect.

btsimonh:
note.  wcclock 11 or 12 gives better results than wcclock 10.  So I guess 10mhz is bad...  and detuning it is better. (1-200 ms web response at 10, 30ms resonse at 11.)  For my hardware ... and wifi interaction :(

btsimonh:
@arendst - I'm pretty happy with this now.  It's fine on S3 - with the caveat that my board's clock line is very noisy as the observations above indicate.
All my stability and wifi concerns have been resolved with the understanding that the XCLK line frequency caused them all.

Over the coming weeks, I'll look at an autoconf for my specific board; I beleive that autoconf runs some berry at startup, and so could pre-configure the clock.

I have concerns over the webcam driver as a whole - but they apply to both new and old.  e.g. if there is no pin config, it will default to something close to the AIThinker board, and this could in theory prevent full boot.  I think it would be better in the absence of pin config to not try to start the camera hardware.  It may also be better to have a default wcclock which is detuned from 20mhz, since 20mhz seems to be a bad one for wifi interferance.  But ALL sample code has 20mhz.... so maybe that's the recommended for the cameras.

The new driver does allow `wcresolution -1` to prevent camera startup.

One thing of general note for ESP32 (maybe only S3?) was that one issue I discovered was that an addLog() very close to a task ending caused a guru mediation (invalid access).  I added an (excessive) delay after the addLog(), and never saw the issue again.  This may have implications for other use of tasks on TAS.  It appears that addLog within a running task is fine - it's just if that class terminates very shortly after the addLog.  I looked at the logging source, and could not see anything obvious, but it's as though we're keeping a pointer to task memory, and using it after addLog returns (i.e. in this case task termination).
This could have implications beyond tasks.

br, Simon

btsimonh:
p.s. I did not experiment with audio as well yet.  It will be a few months before I can get to play with those new toys...

barbudor:
@btsimonh are you calling addLog in your camera thread?
If yes, I am not surprised of bad side effects. Tasmota and it's functions are meant to be ran from a single thread. The is no lock mechanism on resource access so if tasmota core thread is doing an addLog at the same time your camera thread does, bad things are highly expected.
A delay shouldn't change anything

btsimonh:
@barbudor - addLog is protected by a mutex in ESP32, so should be safe.  It's been heavily tested in BLE some years ago, but maybe not this specific circumstance of a thread ending.

Jason2866:
My 2 cents. The standard Webcam driver setup should only include the Webcam feature.
Tasmota way. Keep it small and simple by default. Enhanced features (motion detection....) consuming flash and RAM should only be enabled via compile time settings, like it is done fore the rtsp stuff.

btsimonh:
@Jason2866 - I could re-work again to slim down, disable motion, but I'd like to leave the TAS commands for picture saving/reading jpeg in for Berry use.

## What I'd have in mind is:
1/ Add `#define` for motion
1b/ Add `#define' for picture manipulation (encode/decode - only really required for motion and advanced AI related possible future features).  Check and remove crossover with @arendst 's existing JPEG decode routines, which I think I reproduced independently :(.
2/ disable the driver unless pins are configured (I feel danger in the defaults as many modules have different pins)
3/ add a pin type for CAM_PWR - I've only just realised that this is different to CAM_PWDN.  CAM_PWR on AIThinker boards controls the actual 2 voltage regs - nothing to do with the camera as such, but does enable the board to go into a low(er) power draw.  CAM_PWDN is a camera control line, often not connected.  The esp cam driver is not designed to toggle the CAM_PWR line, hence the 0x105 errors.  (I'm looking at if we can deepsleep, and so run timelapse off battery).
4/ Allocate the main control structure only if enabled - so saving RAM if the driver is not enabled.
5/ _maybe_ move motion into ANOTHER task, so that it's use does not affect framerate.  i.e. minimise related cam thread processing to picture capture to the buffer, but process the buffer analysis seperately.
6/ move all motion code to a different file so the basic driver is easier to maintain?
7/ detailed code-size reduction.

I do not anticipate any major feature changes, except for the motion being optional.
The only 'extra' feature which would be REALLY nice would be to test and tune audio with video, but I don't see this as achievable within this step (e.g. I'd like to investigate audio in the RTSP, and real-time audio stream to the main menu or a dedicated browser-style interface, so enabling a 'door bell' style device.  In theory it will work to browser now, but in practice I beleive it will need some tuning).

But... I'll be out of action for about a month from wed (holiday then business), so it's not going to be quick, unless I had help.  I would guess that the revision could be available in 8-12 weeks, or 16 weeks to include some further long-period tests.

## So we have three options ref this PR:
1/ put on hold and anticipate a future update.
2/ publish as is to get feedback, and anticipate a future update.
3/ change such that inclusion of the 'new' driver is controlled by an additional `USE_WEBCAM_2" rather than be the default, publish and hope for feedback.

I'm happy with any of the above, although getting feedback on performance and features would be nice.
A list of real cam use cases would also be really good, so we know where the focus should be. (maybe run a poll?)

### Size comparison:
(expand to understand @Jason2866 's valid concerns).
<details>
<summary>Build Size comparison</summary>

Bases on building WITHOUT scripts (since with scripts includes motion code linking...).

DRAM increases by 172 bytes.
Flash increases by 13.7 KBytes.
Total image size increase by 13668 bytes.

'new' driver:
```
Total sizes:
Used static DRAM:   64820 bytes (  59760 remain, 52.0% used)
      .data size:   21076 bytes
      .bss  size:   43744 bytes
Used static IRAM:   89894 bytes (  41178 remain, 68.6% used)
      .text size:   88867 bytes
   .vectors size:    1027 bytes
Used Flash size : 1189556 bytes
           .text:  948971 bytes
         .rodata:  240329 bytes
Total image size: 1300526 bytes (.bin may be padded larger)
```

'old' driver:
```
Total sizes:
Used static DRAM:   64692 bytes (  59888 remain, 51.9% used)
      .data size:   21108 bytes
      .bss  size:   43584 bytes
Used static IRAM:   89894 bytes (  41178 remain, 68.6% used)
      .text size:   88867 bytes
   .vectors size:    1027 bytes
Used Flash size : 1175856 bytes
           .text:  938855 bytes
         .rodata:  236745 bytes
Total image size: 1286858 bytes (.bin may be padded larger)
```
</details>

arendst:
As it's esp32 it's not so much the size increase but the possible lack of complete compatibility with the current driver. For me I feel you tend to go a long way to add features you need while losing the compatible and possibly legacy functionality.

I suggest to keep your driver as a V2 version for users feeling comfortable to use it while keeping the current driver for all current users feeling comfortable with it's current performance.

btsimonh:
@arendst - I agree with you.  In this case I strived to retain every existing feature, but it's not possible to test them all without help, and for those from whom I need the help most, time is a precious comodity.
Tonight or early tomorrow, I will update to swap the `USE_WEBCAM_LEGACY` to `USE_WEBCAM_V2` and reverse the meaning, such that the default will be the original driver.

btsimonh:
I did an hour of work to separate the motion features, and (see sizes added to above) there is a significant saving (8k reduction of flash compared to release driver.  11k reduction of DRAM compared to release driver), even over the original driver.  I did not try to run it yet, but it builds fine with and without motion.  I suggest this is a worthwhile effort regardless of which is default.  Let me test a little, and if it's the same as the PR for normal and S3, I'll update as above, but also with the separation (separate .ino file for motion, so making it simpler as well).

btsimonh:
All done.
Logging minimised too.

Jason2866:
Fine for me now 👍

btsimonh:
@Jason2866 - thankyou for running it up.  I believe that my S3 board is hardware compromised (Freenove - I'll report to them at some point, they seem quite responsive.  I got s/n 85, so it's a fairly new board for them).  They ran XCLK all around the houses.  At some point I'll remove the pin on that GPIO to reduce the stub.
Ref the new arduino core, I did not try it.

btsimonh:
thankyou @arendst :).
please make sure @btsimonh is mentioned against any issues or questions.
I'll be away for a couple of weeks now, but mentions go into my main work email, so I will see them upon return.

belveder79:
I installed the new development snapshot and it works really decently... I did not test any of the motion features, but from what I saw so far it is soooooooo more stable and faster compared to the other version - amazing...

btsimonh:
@belveder79 - using Berry, you should be able to serve a simple html webpage from the SD card or flash.  This would then allow you to create a page dedicated to the PTZ control, rather than use the webui as a basis.
It's also easy enough to add a button in the webui which redirects to the PTZ page.  Since most of these devices have plenty of filesystem flash and probably an SD card, the html filesize is not a problem...
( if you dig through [https://github.com/btsimonh/TasmotaBerryPLay](https://github.com/btsimonh/TasmotaBerryPLay/blob/master/tasdocs/esp32cam.md) you will find my webcam berry timelapse code, which includes adding such a link button (`web_add_main_button()`), and it has an HTML - not accessing the live image, and served using static file serving, not via Berry reading and serving the file - but it's a simple example. )

belveder79:
@btsimonh very cool, will check out... thx

PR #19285, opened 2023-08-09
Improvements to PID controller driver

blacknell:
## Description:

These improvements aid the usability of the PID controller primarily by adding an explicit local sensor compile time option and adding missing sensor information to the JSON output.

Note. Although the PID algorithm is inherently designed to work with any driven set point type, when using with a local sensor the assumption was that sensor would be a temperature one (as it most often will be). Furthermore, the code asks for the global default temperature sensor which is end user configurable using command `GlobalTemp2` to select an index selector. However, if this sensor is missing then the index will point to the next available sensor which will result in the algorithm attempting to attain the wrong outcome. The approach chosen here is to allow a sensor name to be chosen eg `DS18B20` and a parameter of that sensor eg `Temperature`. This is much more robust since if that sensor disappears then the device will be flagged as missing and the PID takes action. Furthermore, it is now possible to have the algorithm try and drive a different sensor parameter type eg. `Humidity` or `Pressure` to the set point.
 

- add: display key controller data in the Tasmota web front end
- add: allow a local sensor name to be chosen
- add: define the local sensor measurement parameter
- add: limit sensor not seen error log message to every 60 seconds to avoid flooding the logs
- add: include interval in seconds since sensor data was last updated in json output
- fix: properly update the maximum interval time so missing sensor data is properly alerted is the command `PIDMaxInterval` is used to change this

When this PR is merged I'll update documentation accordingly.

**Related issue (if applicable):** fixes N/A

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.2.0.11


arendst:
Thx again

PR #19299, opened 2023-08-11
ESP32: pass flashmode at build time to macro definition

Staars:
## Description:

This removes all run time computations as this value is known at build time.
The flash mode is taken from the build environment, passed as a build flag and converted to a string literal by the preprocessor.
Saves 200 bytes of flash on the ESP32, is faster at run time (not perceptible) and should be pretty future proof aka extensible.

## Checklist:

### DIFF
--- a/pio-tools/add_c_flags.py
+++ b/pio-tools/add_c_flags.py
@@ -5,7 +5,3 @@ env.Append(CFLAGS=["-Wno-discarded-qualifiers", "-Wno-implicit-function-declarat
 
 # General options that are passed to the C++ compiler
 env.Append(CXXFLAGS=["-Wno-volatile"])
-
-# Pass flashmode at build time to macro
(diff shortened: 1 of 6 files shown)


Staars:
Okay, the forgotten 8266 ....
But even better - the old platform should benefit in the same manner, although I did not measure the size difference.
I wanted to avoid to create another Python script in /pio-tools, that's why I added the code to pre_source_dir.py as this script is executed for all platforms.

arendst:
Once upon a time (set)flashmode on ESP8266 was used to fix flash chip lane count. It appeared four lane compiled binaries would fail on two lane hardware. During OTA upload the flashmode would be changed to two lane just before executing the new binary.

In the debug code you'll still find the command ``SetFlashmode`` to fix (and test) current report implementation.

Over time the OTA flashmode test has vanished and so this implemntation of only reporting the compiled binary flashmode might well do.

And pls do not forget ESP8266. 80% of my tasmota is ESP8266 based hardware. The ESP32 berry,matter,bluetooth thingy is still niche with regards to ready sold "usable" (not being dev boards with fancy displays etc) overpriced hardware.

PR #19302, opened 2023-08-12
Improve no load detection for ADE7953 energy sensor

paulusbrand:
Removes software based no load threshold.
Lowers ADE7953 builtin no load detection threshold to be able to measure 5 watt power levels.

## Description:

Since the ADE7953 has builtin no load detection and it's enabled by default. We do not need to check it in software.
The shelly pro 1 pm with stock software is able to measure loads around 5W. Tasmota software can currently not measure below 10W even when software based no load detection is disabled.
The ADE7953 builtin no load detection threshold is a bit high by default. That's why the thresholds for active power and reactive power are lowered during initialization. No load detection for apparent power is disabled by default. With these settings a load of 5W can be measured.

## Checklist:


arendst:
Thx! Currently evaluating ...

PR #19303, opened 2023-08-13
fix OneWire for IDF5.1 and C2/C6

Staars:
## Description:

Besides a new `include` the common difference is XTENSA vs RISCV.

## Checklist:


arendst:
Is __riscv currently defined?

Staars:
Seems so, got the hint from here:
https://github.com/arendst/Tasmota/blob/development/lib/default/Ext-printf/src/ext_printf.cpp

arendst:
While scanning idf 5.0 I still see specific defines for C3, C2, C6 etc.

I prefer to keep these defines over __riscv which is hardly used in idf AND may be used by other hardware too (like highfive1 in lib OneWire.h)

Staars:
Okay, I will refactor it. (I was so proud about this short solution 😄)

PR #19388, opened 2023-08-25
Added passive mode for Sen5x sensor (required for Ike@ Vindstyrka)

evilJazz:
## Description:

Added a passive mode in sen5x sensor for parasitic installations. This skips reset & initialization of the sen5x sensor on startup and reduces the polling to every 10 seconds to not interfere with and confuse the other I2C master on the bus, e.g. Ike@ Vindstyrka.
Use SetOption156 1 to enable passive mode and restart the device.

## Checklist:


evilJazz:
This is a hack for the Ike@ Vindstyrka where an ESP is installed parasitically by connecting to the SCA, SCL and 5V + GND pads on the main board. The MCU on the main board already set up the SEN54 sensor. 

The default handling in Tasmota's Sen5x sensor adaptation is to reset the sensor and start measuring. When this is done, the MCU on the main board gets confused and displays wrong results and apparently also resets everything, which leads to all weird and bouncing values in both systems.

To fix the issue, we need to prevent Tasmota from resetting the sensor. Also sending the I2C frame "start measuring" confuses the Ike@ MCU, so this has to be disabled as well. Additionally we need to reduce the frequency of polling (current fix is every 10 seconds), since the Ike@ MCU does not seem to handle another I2C master on the same bus gracefully.

evilJazz:
If SetOption156 is not the way to go for persisting the passive mode, please advise how to handle this more gracefully and I'll rewrite. I already read through the sensor docs + sources and it looks like another sensor (Sensor13) is relying on a user-defined boot rule to set up the sensor. What is the best practice here?

Issue #17751, opened 2023-01-20
Power update via MQTT is not working as expected


Civlo85:
### PROBLEM DESCRIPTION
I want to send the actual Power consumption measured by my tasmota device via MQTT to my local broker. But it does not work as expected. I set up my "powerdelta" to 10 in order to get updates whenever my power consumption changes by 10%. That was the idea. When I look to the console output, I can see, that the trigger seems to work, but the value in the MQTT publish is wrong! It seems to be a little bit behind. How to solve this problem?

Please have a look at this console output:

I have attached my coffee machine to this device and want to measure the energy consumption of the boiler within the machine. Now the machine is pre-heated and just turns the boiler on and off to hold the temperature. We are in idle mode with an power soncumption of 6 Watts. Now lets look at the logs:

```lua
09:38:32.459 MQT: tele/MSD05/MARGINS = {"MARGINS":{"PowerDelta":1164}}
09:38:32.468 MQT: tele/MSD05/SENSOR = {"Time":"2023-01-20T09:38:32","ENERGY":{"TotalStartTime":"2022-12-21T21:21:11","Total":2.002,"Yesterday":1.097,"Today":0.302,"Power":42,"ApparentPower":42,"ReactivePower":0,"Factor":1.00,"Voltage":230,"Current":0.184}}
```
We see, that the "PowerDelta" seems to trigger. It detects an increase of 1164 Watts. So in total we should habe 1170 Watts. But when we look to the MQTT message, we see that a power consumption of only 42 Watts is transmitted! Why?

Ok, now lets wait until the boiler turns off again.
```lua
09:38:45.430 MQT: tele/MSD05/MARGINS = {"MARGINS":{"PowerDelta":-435}}
09:38:45.439 MQT: tele/MSD05/SENSOR = {"Time":"2023-01-20T09:38:45","ENERGY":{"TotalStartTime":"2022-12-21T21:21:11","Total":2.006,"Yesterday":1.097,"Today":0.307,"Power":727,"ApparentPower":1165,"ReactivePower":910,"Factor":0.62,"Voltage":228,"Current":5.104}}

09:38:46.418 MQT: tele/MSD05/MARGINS = {"MARGINS":{"PowerDelta":-721}}
[...]

PR #19397, opened 2023-08-27
EnergyMargins - always send MQTT telemetry message

SteWers:
## Description:

The issue https://github.com/arendst/Tasmota/issues/17751 was solved with the commit https://github.com/arendst/Tasmota/commit/86979646fff377a0acfdd88b1c7e1d202f97a5a6. This causes the behavior, which is discussed here https://github.com/arendst/Tasmota/discussions/19386.

This PR solves both problems:
- A MQTT message is send immediately when a margin fires.
- Additionally a message is send after 2 seconds when the energy reading is stable.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.11


arendst:
IRMC

PR #19412, opened 2023-08-29
Don't reset the MAX17043 battery fuel gauge after waking from Deep Sleep

blacknell:
## Description:

The Analog Devices MAX17043 is designed to continually monitor battery voltage over time and using dynamic algorithms present a much more accurate battery percentage remaining value to the system. In reality, almost all such systems would use deep sleep to conserve battery power (waking for, say, 10 seconds every 10 minutes).

However, the current driver resets the chip on every boot resulting in its internal algorithm having to start again from scratch. It will eventually converge on a decent number but this is made much more difficult (or even impossible) if it's reset on every reboot. I have noted inconsistent spikes in percentage after a reboot.

This PR detects if waking from deep sleep and skips the MAX17043 `Power on Reset` and `Quick Start` device commands.

Note that this assumes the hardware has been designed so the MAX17043 remains powered up when the ESP deep sleeps. There is little point in using this device in another way if you want accurate convergence on remaining charge.

**NOTE:** @arendst I did consider adding code to issue the Tasmota `BatteryPercentage` command but all the existing code seems to be defined in Zigbee specific sources files eg `D_CMND_ZIGBEE_BATTPERCENT`. I'm happy to do this if maintainers are ok this is the right approach. Personally, this feels right as it would be automatically added to the state json payload.

- fix: don't reset or quick start the device when coming out of deep sleep
- add: update global battery percentage when max17043 reports new value
- chore: refactor command and config values as define directives
- chore: refactor percentage symbol to use Tasmota define directive

**Related issue (if applicable):** fixes #N/A

## Checklist:

### DIFF
--- a/tasmota/include/tasmota.h
+++ b/tasmota/include/tasmota.h
@@ -515,10 +515,10 @@ enum DevGroupShareItem { DGR_SHARE_POWER = 1, DGR_SHARE_LIGHT_BRI = 2, DGR_SHARE
 
 enum CommandSource { SRC_IGNORE, SRC_MQTT, SRC_RESTART, SRC_BUTTON, SRC_SWITCH, SRC_BACKLOG, SRC_SERIAL, SRC_WEBGUI, SRC_WEBCOMMAND, SRC_WEBCONSOLE, SRC_PULSETIMER,
                      SRC_TIMER, SRC_RULE, SRC_MAXPOWER, SRC_MAXENERGY, SRC_OVERTEMP, SRC_LIGHT, SRC_KNX, SRC_DISPLAY, SRC_WEMO, SRC_HUE, SRC_RETRY, SRC_REMOTE, SRC_SHUTTER,
-                     SRC_THERMOSTAT, SRC_CHAT, SRC_TCL, SRC_BERRY, SRC_FILE, SRC_SSERIAL, SRC_USBCONSOLE, SRC_SO47, SRC_MAX };
+                     SRC_THERMOSTAT, SRC_CHAT, SRC_TCL, SRC_BERRY, SRC_FILE, SRC_SSERIAL, SRC_USBCONSOLE, SRC_SO47, SRC_SENSOR, SRC_MAX };
 const char kCommandSource[] PROGMEM = "I|MQTT|Restart|Button|Switch|Backlog|Serial|WebGui|WebCommand|WebConsole|PulseTimer|"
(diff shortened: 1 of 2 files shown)


stefanbode:
The BatteryPercentage is NOT zigbee specific. You can set this with a command or rule easily just like you set power.

stefanbode:
I checked the code and the approach is quite interesting. What I do miss is the option to change the behavior to the old model. Can we assume that ALL users will use the AD device with POWER ON when using deepsleep. I do agree that this might be ok if you wake up every 10min. But what about every 4h. Maybe we should make it a little bit more configurable. Mayby something comming from rules var?

blacknell:
This did occur to me and mostly I prefer to make code backwards compatible so it works like it used to. I thought a new `SetOption` would be overkill. I settled for a DEBUG_MORE message to warn someone what's going on.

However, the impact of not issuing the reset and quick start during a wake up but when the sensor had been powered down is not that significant. The power reset command performs the same as physically powering on so that happens anyway. The quick start is only needed in cases when power up is very noisy (per the data sheet).

IMO this is moot because any design that powers down the sensor for hours, then powers it on for a few seconds (or even minutes) in the expectation it will give a meaningful gauge level is fundamentally flawed. Granted these devices converge quicker than old techniques but they are only useful if continually monitoring battery draw and charge. They are low power devices too (75µA down to 3µA with I2C held low on sleep).

My vote is to leave it as is

stefanbode:
ok agree. at the end the argument that you get messy data anyhow if you freshly restart the device make it. Then the only question I have (I support the PR), does it really benefit to use deepsleep in this case?

blacknell:
> ok agree. at the end the argument that you get messy data anyhow if you freshly restart the device make it. Then the only question I have (I support the PR), does it really benefit to use deepsleep in this case?

Most of the power is likely to be taken by the ESP, so I'd say yes. Deepsleep makes a massive difference therefore. But the purpose of the sensor is to give a realistic percent remaining indicator. Done well I'd say there work very well together. It can help decide how much deepsleep to set. The solar powered weather station I'm building runs down to 90% during the night and is back to 100% by 10am. It'll be interesting to see how that works in the winter.

blacknell:
> The BatteryPercentage is NOT zigbee specific. You can set this with a command or rule easily just like you set power.

I'll do some research into the history but I propose to refactor so the names don't include zigbee since they really aren't.

blacknell:
@arendst This is all done now and ready for merge. I have some other PR's pending so would appreciate closing so I can set up my fork repo properly and get them done [using best practice](https://github.com/arendst/Tasmota/discussions/19481)

PR #19433, opened 2023-08-31
Signal exit sleep

s-hadinger:
## Description:

While working on HDMI CEC, I need to publish and handle messages quickly as they arrive, and not necessarily wait for the next tick which can be 50ms in the future.

I added a simple flag, marked as `volatile` to be usable from ISR, that sends a virtual `signal` to exit the (dynamic) sleep and trigger a tick event. Therefore the maximum delay is 1ms. The signal is reset when exiting sleep.

A driver may request to run in fast loop mode (equivalent of `Sleep 0`) by continuously setting the signal at each `FUNC_LOOP` for as long as needed.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ## [13.1.0.2]
 ### Added
+- internal signal to exit sleep mode usable from ISR
 
 ### Breaking Changed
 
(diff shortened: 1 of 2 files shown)


arendst:
Did you look for FUNC_SLEEP_LOOP?

It does exactly that.

s-hadinger:
Yes indeed...

PR #19458, opened 2023-09-04
Prep Arduino 3.0: fix shine for gcc12

Staars:
## Description:

GCC 12 is very strict about types and does not allow to cast `int` to `int32_t` and vice versa, so this PR just makes everything `int`.

I can confirm that the I2S microphone and MP3 encoding is working with Arduino 3.0 and Tasmota, but the needed changes will come later.

## Checklist:


arendst:
How come this fails while tasmota is using int32_t too instead of int???

s-hadinger:
If the native lib uses `int`, then `int` it must be

Staars:
AFAIK the problem is casting from `int` to `int32_t` and back, not using `int32_t` per se.
I do not agree with the strictness of GCC (unless I simply do not understand what's going on here).

PR #19542, opened 2023-09-18
Windmeter: add measure interval setting

### LINKED ISSUE #9091: Increase measurement duration for windmeter (Sensor68)

_A clear and concise description of what the problem is._
Right now the measurement duration is fixed to 1 second. This means when there are no counts within 1 second the wind speed will be 0 even when the anemometer cups are rotating slowly.
I use a Davis anemometer which has a measurement duration of 2.5 seconds (according to datasheet). The minimum resolution for this gauge is 0.444 rotations per second (1.6 km/h).
With the current measurement duration of windmeter my minimum wind speed is 3.6 km/h.
_A clear and concise description of what you want to happen._
Be able to set the measurement duration with sensor settings.
I thought about a solution but unfortunately I am not an experienced programmer (I could try anyway with some help).
One way to solve it would be to increase a counter in the main loop which will be executed every second (FUNC_EVERY_SECOND -> WindMeterEverySecond).
When a set amount of counts is reached the main loop to calculate wind speed is executed. The set amount of counts is the measurement duration in 1 second steps which can be set by a sensor setting.
**Additional context**  

malbinola:
## Description:
Original implementation was designed for a small cup anemometer (approx 6cm radius) where measuring speed every second was sufficient. As reported by @Ingenieur89 several months ago, when using an anemometer with a higher radius, is useful to delay speed computation, otherwise the speed may not get computed below a certain value.

New setting `windmeter_measure_intvl`, by default set to 1s (like before), allows to increase the resolution of the speed measurement. Using command `Sensor68 6, X` (where is X from 1 to 255), you can specify to compute the speed every `X` seconds.

Finally, now the computation is based on the real time-delta between measurements and not on the name of the timer function (***EverySecond) as happened before.

Compilation for ESP32 succeeded but unfortunately i do not have an ESP32 device to test it on the field.

**Related issue (if applicable):** fixes #9091 

## Checklist:

### DIFF
--- a/tasmota/include/tasmota_types.h
+++ b/tasmota/include/tasmota_types.h
@@ -848,8 +848,9 @@ typedef struct {
   uint8_t       hdmi_cec_device_type;      // F61  - v13.1.0.1 (was ex_modbus_sbaudrate v12.2.0.5)
 
   uint8_t       modbus_sconfig;            // F62
+  uint8_t       windmeter_measure_intvl;   // F63
 
-  uint8_t       free_f63[13];              // F63 - Decrement if adding new Setting variables just above and below
+  uint8_t       free_f64[12];              // F64 - Decrement if adding new Setting variables just above and below
 
   // Only 32 bit boundary variables below
   uint32_t      touch_threshold;           // F70
--- a/tasmota/tasmota_xsns_sensor/xsns_68_windmeter.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_68_windmeter.ino
@@ -31,6 +31,7 @@
 #define WINDMETER_DEF_PULSES_X_ROT    1     // Number of pulses for a complete rotation
 #define WINDMETER_DEF_PULSE_DEBOUNCE  10    // Pulse counter debounce time (milliseconds)
 #define WINDMETER_DEF_COMP_FACTOR     1.18  // Compensation factor
+#define WINDMETER_DEF_MEASURE_INTVL   1     // Speed measurement interval: speed value will be computed every X (seconds)
 #define WINDMETER_DEF_TELE_PCHANGE    255   // Minimum percentage change between current and last reported speed in order to trigger a new tele message (0...100, 255 means off)
 #define WINDMETER_WEIGHT_AVG_SAMPLE   150   // No of samples to take
 
@@ -60,7 +61,8 @@ float const windmeter_2pi = windmeter_pi * 2;
 struct WINDMETER {
   volatile uint32_t counter_time;
   volatile unsigned long counter = 0;
-  //uint32_t speed_time;
+  uint8_t measure_counter = 0;
+  uint32_t measure_time = 0;
   float speed = 0;
   float last_tele_speed = 0;
 #ifndef USE_WINDMETER_NOSTATISTICS
@@ -79,7 +81,7 @@ void IRAM_ATTR WindMeterUpdateSpeed(void)
   if (time_diff > Settings->windmeter_pulse_debounce * 1000) {
     WindMeter.counter_time = time;
     WindMeter.counter++;
-//    AddLog(LOG_LEVEL_DEBUG, PSTR("WMET: Counter %d"), WindMeter.counter);
+    //AddLog(LOG_LEVEL_DEBUG, PSTR("WMET: Counter %d"), WindMeter.counter);
   }
 }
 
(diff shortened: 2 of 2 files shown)


Ingenieur89:
Thanks @malbinola for coming back and bringing this improvement!
I will look into it again.

Is it possible to compute by fractions of a second like 2.5s? Not just 1, 2, 3, ... seconds.

You have 3 parameters here: 
windmeter_pulses_x_rot, windmeter_radius, windmeter_speed_factor 
which are only used in this formula if I see it right:
 // speed = ( (pulses / pulses_per_rotation) * (2 * pi * radius) * anemometer_factor ) / delta_time
    WindMeter.speed = (((WindMeter.counter / Settings->windmeter_pulses_x_rot) * (windmeter_2pi * ((float)Settings->windmeter_radius / 1000)) * ((float)Settings->windmeter_speed_factor / 1000)) / ((float)(time - last_time) / 1000000));
Maybe you could save on memory if you merge them together to one factor?
Example:
speed = pulses * factor / delta_time
factor = 2 * pi * radius * anemometer_factor / pulses_per_rotation
The user has to solve this equation one time and hand it over to the settings.
Also this could help the user to customize their calculation.

Best regards!

malbinola:
Ok, let me know.

> Is it possible to compute by fractions of a second like 2.5s? Not just 1, 2, 3, ... seconds.

Yes but it's not an immediate modification. Speed computation is based on FUNC_EVERY_SECOND callback, so logic needs to be migrated for instance to FUNC_EVERY_250_MSECOND and handle time fractions.
I think that seconds here are enough. Why do you need fractions?

> Maybe you could save on memory if you merge them together to one factor?

With this PR, i've tried to keep changes to a minimum. But of course it can be pre-computed into a single factor in order to improve internal calculations. Do yo mean internally? Right?
From a user perspective is easier to deal with three named parameters from `Sensor68 #,X` command instead of having a single compressed factor to set. Then we need to take into account backward compatibility.
Anyway, this could be the 2.0 improvement! :)

Ingenieur89:
> Yes but it's not an immediate modification. Speed computation is based on FUNC_EVERY_SECOND callback, so logic needs to be migrated for instance to FUNC_EVERY_250_MSECOND and handle time fractions. I think that seconds here are enough. Why do you need fractions?

I checked again, don't necessarily need fractions. I use a Davis windmeter and in the calculation they go with a measurement period of 2.5 seconds but I can convert the equation to 3 seconds, no issue. This is what i use currently anyway.

>With this PR, i've tried to keep changes to a minimum. But of course it can be pre-computed into a single factor in order to improve internal calculations. Do yo mean internally? Right?

Yes, I meant a single factor for internal calculation.

>From a user perspective is easier to deal with three named parameters from Sensor68 #,X command instead of having a single compressed factor to set. Then we need to take into account backward compatibility.

I didn't think about backward compatibility, this would be a breaking change...

>Anyway, this could be the 2.0 improvement! :)

This was just what I had on my mind. Just leave it as it is. Perfect!
I saw some commented code of a wind direction calculation. This could be interesting to implement in the future, too.

malbinola:
> I saw some commented code of a wind direction calculation. This could be interesting to implement in the future, too.

I agree, it would be cool! This implementation was inspired by tx20/tx23 sensor (xsns_35_tx20) that have support for wind direction information; some code was ported as starting point for future implementations.
Unfortunately, i have only a cup anemometer with no direction measurement.

@Ingenieur89 what kind of ESP device do you have? You can confirm that modification also work on ESP32?
Thank you

Jason2866:
PR needs to be tested with an ESP32. Without no merge.

Ingenieur89:
> > I saw some commented code of a wind direction calculation. This could be interesting to implement in the future, too.
> 
> I agree, it would be cool! This implementation was inspired by tx20/tx23 sensor (xsns_35_tx20) that have support for wind direction information; some code was ported as starting point for future implementations. Unfortunately, i have only a cup anemometer with no direction measurement.
> 
> @Ingenieur89 what kind of ESP device do you have? Can you confirm that modification also work on ESP32? Thank you

I use an ESP32 devkit c (not sure which model).
I will try to find time on the weekend and test it on a separate device. 
Do you know how I can download all the files with your PR included? Im not used to Github

Ingenieur89:
> > I use an ESP32 devkit c (not sure which model).
> 
> I think that is enough. What do you think @Jason2866 ? If he asserts that works successfully on that, can i mark the checklist as done?
> 
> > Do you know how I can download all the files with your PR included? Im not used to Github
> 
> @Ingenieur89 This is the [source branch](https://github.com/malbinola/Tasmota/tree/windmeter-measure-intvl) of my PR: you can click on "Code" (green button) then "Download ZIP".

I flashed your branch to the ESP32 devkit c and it is working fine for a few days. Thank you for the update!

malbinola:
Great news @Ingenieur89 !! Thank you for the feedback!
Checklist updated ;)

PR #19558, opened 2023-09-20
Berry counter

s-hadinger:
## Description:

Berry add ability to read and change Counter values:
- if `USE_COUNTER` is not defined, or GPIO are not configured, all functions return `nil`
- `gpio.counter_read(counter:int) -> int or nil` read counter value, or return nil if counter is not used
- `gpio.counter_set(counter:int, value:int) -> int or nil` set the counter value, return the actual value, or return nil if counter is not used
- `gpio.counter_add(counter:int, value:int) -> int or nil` add to the counter value, return the actual value, or return nil if counter is not used

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Support for Shelly PlusPMMini, Plus1Mini and Plus1PMMini
 - Matter support for Virtual Devices controllable via Rules or Berry (#19520)
+- Berry read and write Counters
 
 ### Breaking Changed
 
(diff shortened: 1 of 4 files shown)


sfromis:
Seems to work, also for `CounterType 2` giving millis.

Wondering about whether indexing should start at 0 or 1. Sure it can be "less confusing" to have same index numbers as on the Counter command etc, instead of 1 lower. OTOH `set_power` and `get_power` uses zero-based indexes. Thus there is a "consistency issue".

s-hadinger:
Ah. Good question. For consistency it should start at 0, right?

sfromis:
Well, one-based would be consistent with base Tasmota, but zero-based would be more internally consistent within Berry, including with `gpio.pin_used(gpio.CNTR1, 2)` which returns false with two counters, but true for 2nd counter when using index 1 instead.

While using the lower level `gpio` module functions with Counter pins is probably not too useful, except for checking if pins are assigned as expected, it does look like the `gpio` module already leans toward zero-based.

I suppose that's a "yes" :wink:

PR #19580, opened 2023-09-23
Allow webbutton up to 32

barbudor:
## Description:

**Related issue (if applicable):** fixes #19324

My proposition to support WebButton up to 32
No data stored in Settings => must use `autoexec.bat` or `on system#init` rule to set on each reboot
Usage:
- adds 176 B of flash
- adds 64 B of RAM

(Note: I have tested an alternative with +8 B of RAM instead of +64 but +224 B of flash, I believe it's better to optimize flash usage)

There's no #define. Not easy to add due to MAX_BUTTON_TEXT being defined as a const in tasmota.h (not overridable)

image

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.2.0.13

### DIFF
--- a/tasmota/include/tasmota.h
+++ b/tasmota/include/tasmota.h
@@ -122,7 +122,7 @@ const uint8_t MAX_STATE_TEXT = 4;           // Max number of State names (OFF, O
 const uint8_t MAX_NTP_SERVERS = 3;          // Max number of NTP servers
 const uint8_t MAX_RULE_MEMS = 16;           // Max number of saved vars
 const uint8_t MAX_FRIENDLYNAMES = 8;        // Max number of Friendly names
-const uint8_t MAX_BUTTON_TEXT = 16;         // Max number of GUI button labels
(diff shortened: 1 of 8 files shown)


arendst:
Thx for the idea. Give me time to evaluate.

barbudor:
Hi @arendst ,
Except for core features, I'm much in favor of putting things that are not often used elsewhere than Settings.
More drastically, settings for a feature that is not used (or even not in the build) should not take up room.

You may have suggested something based on JSON but that means adding JSON parsing to every driver and we would likely loose in code-bytes what we would have saved in settings-bytes.
So why not using something that is already there : the command parser and autoexec.bat
Generalizing a small file system, even on 1MB build could be very helpfull with that
And using 1 single file (autoexec.bat) vs 1 file per driver, is using more efficiently the flash estate.

One point is that in that case, settings changed in the console or through MQTT are not automatically saved (vs automatic save of Settings). This can be seen as a disadvantage as user need to remind to update his autoexec.bat, but that could also be an advantage when tinkering as a simple restart reverts to the last-known-good state.

PR #19658, opened 2023-10-03
Rename VERSION to TASMOTA_VERSION as a guard against external macros

Staars:
## Description:

Related to:
https://github.com/Jason2866/Tasmota-build/actions/runs/6394601232/job/17356385986

If we include an external library with not-so-smart naming conventions regarding macros we can run into compile errors with Tasmotas global variable `VERSION`.

This error could be fixed in an external library too, but maybe we give our variable a more specific name in this case. For my eyes this make our code a tiny bit more readable, but in the end it is a matter of personal preference.

@arendst Please take a look.

## Checklist:

### DIFF
--- a/tasmota/include/tasmota_version.h
+++ b/tasmota/include/tasmota_version.h
@@ -20,6 +20,6 @@
 #ifndef _TASMOTA_VERSION_H_
 #define _TASMOTA_VERSION_H_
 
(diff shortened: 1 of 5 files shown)


Staars:
Upps, a PR race condition.

I close this one, maybe we pick up the idea later.

s-hadinger:
Ha, race condition indeed. I came out with a less impactful solution, just using `Settings->version` instead

PR #19694, opened 2023-10-07
use partiton table from flash for "Download Filesystem"

Staars:
## Description:

No changes to Tasmota code.

This PR shall fix the custom target "Download Filesystem", which was broken for devices with more flash than the usual 4 MB and with an expanded filesystem. This produces a mismatch between partition table in flash and partition table in the partition.csv-file.

Now the script reads the real partition table from the flash with the help of esptool.py and adjusts the settings on the fly.

Side note: downloading from the flash seems to be more fragile than uploading, so the "upload_speed" property in the build environment (there is no extra download_speed in Platformio) may need to be reduced.

The script does still have only very basic error handling.

## Checklist:

### DIFF
--- a/pio-tools/custom_target.py
+++ b/pio-tools/custom_target.py
@@ -91,62 +91,10 @@ def _parse_size(value):
         return int(value[:-1]) * base
     return value
 
-def _parse_partitions(env):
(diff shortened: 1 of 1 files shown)


Jason2866:
Thx!

PR #19732, opened 2023-10-13
Update xsns_02_analog.ino for calibrated values on ESP32

lalo-uy:
ESP32  ADC has variables errors, but calibrations values are put in eFuses at factory test. analogReadMilliVolts() from api correct the ADC reading with tata values. Since now the value are mV and not ADC numbers, the fast solution is to revert it back to ADC numbers: divide by Vcc  => ANALOG_V33*1000  and multiply by ANALOG_RANGE Other way is to modify all sensor cases to account for the different unit.

## Description:

## Checklist:
  - [X ] Only one feature/fix was added per PR and the code change compiles without warnings
  - [X ] The code change is tested and works with Tasmota core ESP32 V.2.0.14
  - [X ] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_02_analog.ino
@@ -335,10 +335,17 @@ uint16_t AdcRead(uint32_t pin, uint32_t factor) {
   uint32_t samples = 1 << factor;
   uint32_t analog = 0;
   for (uint32_t i = 0; i < samples; i++) {
+#ifdef ESP32 
+    analog += analogReadMilliVolts(pin);  // get the value corrected by calibrated values from the eFuses
+#else
     analog += analogRead(pin);
+#endif
     delay(1);
   }
   analog >>= factor;
+#ifdef ESP32
+  analog = analog/(ANALOG_V33*1000) * ANALOG_RANGE; // go back from mV to ADC
+#endif
   return analog;
 }
 


sfromis:
This was discussed on Discord, starting here:
https://discord.com/channels/479389167382691863/490244847568158751/1162097276609626153

lalo-uy:
This begin with a esp32s with 10k 1% pullup on a 10k ntc.
With Tasmota32 13.1 got 3 to 4 C plus readings at ambien temp.
Try to tune adcparam, to no avail.
Same hw on an Arduino UNO was perfect to +- 0.2C

Then I came across this
https://www.makerfabs.cc/article/cautions-in-using-esp32-adc.html

Dig deep in Tasmota code and arduino esp32 libs to see if that was used in
hal.
Finaly patch as the PR states and got the correct readings.

Going from mV to ADC number is a quick hack, the wrigt solution is to edit
all ADC use cases to account for the different units.

arendst:
Thx.

I can see what you aim for but am not sure if ANALOG_RANGE is ALWAYS the fixed value or, in some cases a user entered value (AdcParam 2 for ADC_RANGE)

Jason2866:
@arendst The idea to use calibrated values is nice. Do you think there is a way to implement it?

arendst:
Let's merge and see what happens.

PR #19756, opened 2023-10-15
Teleinfo use Apparent Power as Active Power approximation

s-hadinger:
## Description:

#19381 removed Active Power in favor of Apparent Power (which is the only information provided by Teleinfo). However this was a breaking change and Active Power is always 0.

I propose to copy Apparent Power to Active Power as an approximation (which should not be too bad).

Note: Teleinfo provides instant current as IINST, but it's an integer which is not precise enough.

@hallard Are you ok with this? Do you have a better way to get Active Power?

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 - MAX31855/MAX6675 sensors driver support up to 6 (#19329)
 - ESP32 analog from `analogRead()` to calibrated `analogReadMilliVolts()` (#19732)
 - I2S refactoring in preparation for core 3 (#19749)
+- Teleinfo use Apparent Power as Active Power approximation
 
 ### Fixed
 - ESP32 shutter frequency (#19717)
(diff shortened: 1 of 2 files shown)


hallard:
@s-hadinger it was also in my thought doing something like that during #19381 but was not sure of side effects so I let only one, but this one got my preference, so yes I'm ok and that's the best we can do with what we have. Some meters have `TENSION` label that I may use to try to have IINST more precise (divide `PAPP` by `TENSION`) but once agin it's just empiric.

ouinouin:
hi, @hallard  ,
first thanks for your great job around teleinfo.
this weird behavior of the telinfo thing in all modes to not show the active power was really a stupid thing from Enedis , they even didnt manage to go for a floating point number on amperes, which gives us a 230w step if we multiply power by voltage.

Despite that, would disagree to make the assumption that : active power = apparent power.
I know as customers, we dont pay for reactive power, but i think its great to show up those figures, since the assumption of apparent power = active power is very wrong is all houses i go to do some measures, reason being there is a lot of small transformers that completely disturb the current quality, add to this some "line assist" UPS or other oddities, and the power factor is often near 0,6 , 0,7.
as our times are challenging energy wise, and as some people are having more small solar at home , i think more and more people will be sensitive to the concept of apparent power, (for example to have better autonomy from a UPS, you get more bang for the bucks if your power factor is high).

as a remedial , to guess the power in watts, i do the derivation on a short time range of EAST, and manage to get some useful values, even if the derivation means the value is lagging a bit behind.
Wh/time = W :-)

otherwise on mode standard, you have the 30 min averaged : Puissance Active Soutirée and Linky Puissance Active Soutirée N-1 that could be another solution (maybe in fact choosing the method should be an option.)

thanks again for all the job around opensource and teleinfo if only @ENEDIS had a look at the job done here and could push an upgrade on their protocol...
 tell us what you think if you have time , i m curious to have your opinion.

PR #19814, opened 2023-10-21
Add command ``GpioRead``

s-hadinger:
## Description:

Add command `GpioRead` which is used to quickly debug digital inputs giving their current values. `digitalRead()` is called on all configured GPIO although the static value may not be of any sense for many protocols (I2C...).

Examples:
```
GpioRead

14:45:08.240 RSL: RESULT = {"GPIORead":{"GPIO32":0,"GPIO33":0}}
```
 
**Related issue (if applicable):** fixes #19810

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - I2C bus2 support to SHTxX temperature and humidity sensor
 - I2C bus2 support to HYTxxx temperature and humidity sensor
 - I2C bus2 support to SI1145/6/7 Ultra violet index and light sensor
+- Add command ``GpioRead``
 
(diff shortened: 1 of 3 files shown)


arendst:
This only works for configured gpios in templates. It shows no user configured gpios. Let me chk how to fix this.

s-hadinger:
Ah Yes, I'm not familiar with that part

sfromis:
If would make sense to also get a peek at pins not (yet) configured, like when hunting for where inputs are, maybe as an option. Including analog too of course :wink:

s-hadinger:
I'm concerned about potential side effect of reading "forbidden" GPOIs. Also the output would be more or less unreadable. Just configure every gpio as INPUT 😉

PR #19856, opened 2023-10-28
Update be_lexer.c

aziemer:
Added check for unterminated block comments

## Description:

Comment blocks started with '#-' but not closed caused a read past EOS and threw a message concerning this erroneous token - but gave no hint, that an unterminated comment was the culprit in the first place.

## Checklist:

### DIFF
--- a/lib/libesp32/berry/src/be_lexer.c
+++ b/lib/libesp32/berry/src/be_lexer.c
@@ -260,6 +260,7 @@ static void skip_comment(blexer *lexer)
 {
     next(lexer); /* skip '#' */
     if (lgetc(lexer) == '-') { /* mult-line comment */
+        int lno = lexer->linenumber;
         int mark, c = 'x'; /* skip first '-' (#- ... -#) */
         do {
             mark = c == '-';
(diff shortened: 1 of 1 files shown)


s-hadinger:
Good catch. Could you please submit the same PR upstream? Or do you want me to do it?

aziemer:
Would you do it please?

aziemer:
Just to be pedantic: there is another thing in the same function that needs attention:
Any '-#' appearing within a string would terminate a comment block...

s-hadinger:
Thanks. I think we can skip this additional check

lib/libesp32/berry/src/be_lexer.c
```diff
@@ -269,6 +270,11 @@ static void skip_comment(blexer *lexer)
             }
             c = next(lexer);
         } while (!(mark && c == '#') && c != EOS);
+        if (c == EOS) {
+        	char tmp[64];
```
s-hadinger, line 274: Ouch sorry. I didn't check the details. You shouldn't be using sprintf as it is considered as a dangerous function. 

You should be doing `be_lexererror(lexer, be_pushfstring(lexer->vm,...))`

Check in be_parser.c for examples (push_error)

aziemer:
Oh, sorry, didn't know that. Have changed it accordingly and sent a PR. I'm still new to Berry, I have used it with a NSPanel and recently ported Berry to an Application of my own running on an ARM-m7 (ATSAME70). It works like a charm there, I'm really impressed! I Had a self-written BASIC like interpreter running there before which was WAY slower than Berry.

aziemer:
PS: The added #include <stdio.h> should be removed too, then

s-hadinger:
There is no such include, that's the point

aziemer:
Oh man, so I even forgot that - it's not my day or I'm growing old, but I'm just 63, so that can not be it...

s-hadinger:
It's done.

PR #19866, opened 2023-10-29
Fix for CmndHDMIAddr function.

anishsane:
1. Typical values for XdrvMailbox.payload are 0x1000...0x4000. Hence the check should be (value > 0).
2. Don't overwrite the user supplied value with value read from the hardware.

## Description:

https://github.com/arendst/Tasmota/discussions/19128#discussioncomment-6892010

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_70_0_hdmi_cec.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_70_0_hdmi_cec.ino
@@ -1116,7 +1116,7 @@ bool IRAM_ATTR CEC_Device::setLineState(bool state, bool check)
 // manage callbacks
 void CEC_Device::OnReady(int logical_address)
 {
-  if (_on_rx_cb) { _on_ready_cb(this, logical_address); }
+  if (_on_ready_cb) { _on_ready_cb(this, logical_address); }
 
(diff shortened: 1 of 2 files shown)


s-hadinger:
Hmm. This is embarrassing, the test is missing a `!` and is hence reversed. 

Good catch

s-hadinger:
The test in `CmndHDMIType` is also reversed

s-hadinger:
Please note that hardware address always takes priority over user supplied address

anishsane:
> Please note that hardware address always takes priority over user-supplied address

This is a bit debatable.
On one side, you are right, that the hardware address should be used whenever possible, because that is the correct value.
One the other side, we have to respect user's choice as well.

Maybe we can return failure from this set function when I2C pins are connected and hence the hardware address can be read.
Thus, there will not be any surprises for the user later.
Some error message could be shown like, "Cannot override the hardware address with the user-supplied address."
Then the user would be aware that the set addr command was unsuccessful and tasmota is using the address queried from the EDID.

s-hadinger:
The idea was also to save to settings the last known hardware address, in case reading from EDID fails for whatever reason. It was designed as a cache, not a way to spoof the address

PR #19877, opened 2023-10-30
Feature/tuya mcu update

benjaminnestler:
## Description:

Okay, this is the initial try to support the OTA-Upgrade of the TUYA MCU with tasmota. You can activate this feature by defining the Macro `USE_TUYA_MCU_UPGRADE`. 

There is a new TuyaCommand added `TuyaUpgrade` which needs 3 arguments (strings) to start an OTA of the tuya mcu:
- version of OTA firmware
- md5 checksum of OTA firmware file
- http-server and path to the OTA firmware file

You can start the OTA procedure via WebComand (Console) or MQTT.

Example:
`TuyaUpgrade 1.1.3,85c449b5cb2502279399dfec57ad30ac,http://192.168.178.23/16025109522a96f16f7ae.bin`

The first step is to load the OTA binary to a non-volatile memory:
- For ESP8266 the internal flash in the range of the tasmota-OTA partition is used 
- For ESP32 devices the UFS is used to store the OTA binary

In the Tuya documentation you'll find this simple (and not complete) sequence diagram
tuya-update-sequence

If OTA binary transfer succeeds, tasmota request the tuya mcu for supported update package size. MCU should response with a dedicated transfer buffer size (typically 256 bytes).

Tasmota starts to transmit the OTA binary from the source nvm to the tuya mcu. After receiving a response to an update package the next OTA file chunk will be transferred with a delay of one second.

After the full transfer of OTA file the tuya mcu should restart und tasmota can query the product ID to verify the tuya mcu version.

I don't know if there are devices out there, for which we can request a tuya mcu firmware file. Maybe a simple request in direction to the device producer leads to a pleasing response ... or not. 

Some companies using a pre-installed Tasmota version on their commercial devices. And maybe these are the companies, which are providing firmware updates for the tuya MCU.
- https://wesmartify.de/search?q=tasmota&options%5Bprefix%5D=last
- https://nous.technology/smart/tasmota/

I've two different devices (two thermostats) with tuya mcu test-firmwares to verify the OTA process on ESP32  and ESP8266 device.

## Checklist:


tasmota/tasmota_xdrv_driver/xdrv_16_tuyamcu_upgrade.ino
```diff
@@ -0,0 +1,390 @@
...
+    _current_packet = nullptr;
+  }
+}
+
+bool TuyaUpgBuffer::init (uint32_t len, char* checksum, char* filename) {
+  reset();
+#ifdef ESP8266
+  // prepare to use OTA-partition
+  _start = (ESP.getSketchSize() + _flash_sector_size - 1) & (~(_flash_sector_size - 1));
+  _end = (uint32_t)&_FS_start - 0x40200000;
```
arendst, line 70: Dangerous as it can overwrite ESP8266 settings located in eight banks below _FS_start. See Tasmota OTA function which makes sure the latests settings are stored safely before the Tasmota OTA is allowed to start,

barbudor:
Tasmota OTA already includes support for upgrading other MCUs such as an Arduino (with Tasmota Client) or zigbee processors, etc...
It allows to use the standard file upload web page without the need to setup a local http server
May be path to investigate?

tasmota/tasmota_xdrv_driver/xdrv_16_tuyamcu_upgrade.ino
benjaminnestler: Fixed this by commit [c2e284b](https://github.com/arendst/Tasmota/pull/19877/commits/c2e284b565a9f56fd39f8b90a5f4d1bcb53415f8)

benjaminnestler: Maybe we should take a look at FLOG::init(void) in [support_flash_log.ino](https://github.com/arendst/Tasmota/blob/development/tasmota/tasmota_support/support_flash_log.ino). That was my original source of code snippet copy to prepare the OTA part.

benjaminnestler:
Hi @barbudor,
I think it's a good idea! You are the pros of tasmota ;-) Thanks for your inspiration. I will check this path.

benjaminnestler:
Hi @barbudor,
I've added the path to upload the Tuya mcu firmware file via Webserver upload page.

barbudor:
> Hi @barbudor, I've added the path to upload the Tuya mcu firmware file via Webserver upload page.

Cool 👍

tasmota/tasmota_xdrv_driver/xdrv_16_tuyamcu_upgrade.ino
Jason2866: Yeah, this code part is rather old. At this time many API / functions in Tasmota just did not exists.

Jason2866:
@benjaminnestler Still draft? If you want a review remove the Draft status

benjaminnestler:
Hi @barbudor, @Jason2866 or @arendst 

what is your best practice to refresh a pull request feature branch in sync with the current development branch?
Should I rebase the feature branch from time to time until the PR is merged?

barbudor:
Unless the PR and the others gets in conflict because they are touching similar parts of the code, it's not necessary.

benjaminnestler:
Found a bug ... after fixing it, I will reset the draft state of the PR.

benjaminnestler:
Hi,

Is there a good solution to debug/log Tasmota firmware in the case of an unknown source of firmware reset?

Currently, I am working on the Tuya MCU firmware upgrade process. Generally, the process runs and flashes the Tuya MCU via Tasmota. I have implemented a new TuyaMCU command to upload the Tuya MCU file and transfer this file via the Tuya protocol to the Tuya MCU.

@barbudor gave me a hint that other co-processors can be updated through the WebUI upload via the file dialog. Therefore, I want to implement this approach as well. However, following this path, I ran into a Tasmota reset before the firmware could be transferred to the Tuya MCU. I need a solution to detect the source of the reset.
If I disconnect the hardware connection (RX and TX) to the Tuya MCU, the reset doesn't occur, and the Tasmota firmware does what I want in the case of no responses from the Tuya MCU.

arendst:
Do you see an exception or a watchdog? The latter is likely caused by your tuya upload code taking too long in between returning to the schedular to reset the watchdog timer and more importantly keeping network active.

benjaminnestler:
Okay, with the help of the map file, I can pinpoint where the error might be.

benjaminnestler:
What else needs to be done:

I'm considering adding a progress indicator that can be shown in the WebUI during the upload and transfer to the Tuya MCU. The overall process duration depends on the OTA file size. Currently, one chunk is transferred to the Tuya MCU every second. So, the entire process takes a few minutes.

arendst:
>  I added a dropdown box to the WebUI to select the desired OTA upload file. 

The fact that this is currently not present is keeping the code as small as possible. Until tuya testing for differences in the first never seemed to be an issue. I know that ":" to distinguish file type isn't a secure solution but then, it was only supposed to be used for TasmoClient.

Having a drop down list still is not a secure way of uploading files; users can still select the wrong type of file.

Any way, I suggest to make the drop down list optional using a define to keep the code as small as possible.

benjaminnestler:
So, I've checked the source code in xdrv_01_9_webserver.ino. I can't think of a clever way to reliably distinguish between Tasmota firmware and Tuya MCU firmware. The issue is that the Tuya firmware can contain anything as the first byte. Both 0xE9 and 0x1F can occur. This makes it impossible to differentiate between the firmware to be uploaded. I need another decision criterion. It's not sufficient to only query the presence/configuration of Tuya RX and TX pins because, in this case, a Tasmota firmware update would always end up in the Tuya MCU update branch. That's why I added manual selection through the dropdown box.

Does anybody have a good idea on how to make the distinction without manual selection? @Jason2866 @barbudor @arendst

barbudor:
Basic question : where do you get tuya MCU firmware from? Are some vendors provide updates for download? Or are you scrapping them spying the communication on a stock device?

benjaminnestler:
> Looking further in the first block, is there no common tuya diferentiatorin there?

No, sorry, there is no specific marker byte used.

I've contacted two vendors and obtained their Tuya MCU firmwares. Both of them respond in the same way. There are no unique identification bytes that can be utilized.

benjaminnestler:
> Basic question : where do you get tuya MCU firmware from? Are some vendors provide updates for download? Or are you scrapping them spying the communication on a stock device?

Yes, indeed, I am in contact with two manufacturers and have some firmware files for OTA tests.

benjaminnestler:
What can I do to ensure that this branch doesn't fade into oblivion?

PR #19922, opened 2023-11-04
Add Support for AHT30 Temperature and Humidity Sensor

legchenkov:
## Description:

Implemented sensor detection, initialization, and data reading based on both the AHT20 and the AHT30 datasheet specifications.
Configured the sensor polling interval to 10 seconds to prevent sensor heating.

## Checklist:
  - [+] The pull request is done against the latest development branch
  - [+] Only relevant files were touched
  - [+] Only one feature/fix was added per PR and the code change compiles without warnings
  - [+] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [+] The code change is tested and works with Tasmota core ESP32 V.2.0.14
  - [+] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).


arendst:
In addition to the datasheet, did you TEST if your changes work? I see you also change AHT1x reset command. Did you test AHT1x functionality?

legchenkov:
> In addition to the datasheet, did you TEST if your changes work? I see you also change AHT1x reset command. Did you test AHT1x functionality?

I tested the code on two ESP32 boards:
1)UICPAL Type-C ESP32-C3 16Pin Development Board Mini
2)WeAct ESP32 Development Board TYPE-C CH340K WiFi+Bluetooth Dual Core ESP32-DOWD-V3
In both cases, it works without any problems.
But the reset command is really different from the one in the datasheet. And if it is wrong, it means it has no noticeable effect.
Initially, the AHT30 sensor didn't work because of an incorrect initialization command in the code for AHT20. After fixing the initialization command, it worked, but it was getting 3 degrees hotter due to frequent measurements. So I had to reduce their frequency. Now the measurements are accurate.
Anyway, I will do some more testing.

legchenkov:
The reset command in my code did not match the AHT10/20/30 datasheets. However, the reset function is not used in the driver, so the error did not affect the driver. Apparently, the initialization function is sufficient for the sensor to work.

legchenkov:
> This PR also fixes work with AHT20+BMP280 module from aliexpress. The AHT20 part of module does not work with #define AHTX_CMD 0xB1 and works with #define AHTX_CMD 0xBE

Very good! I'm just waiting for the AHT20+BMP280 to be delivered.

legchenkov:
Added Russian translation of untranslated element "Dew point" for AHTxx sensors.

dadaweb77:
I tested the code to ESP32 boards:
Sonoff Extreme mini R4.
It works without any problems.
Good work

Noschvie:
@dadaweb77 how have you connected the sensor to the "Sonoff Extreme mini R4"? Thanks!

dadaweb77:
> @dadaweb77 how have you connected the sensor to the "Sonoff Extreme mini R4"? Thanks!

Simple:
**R4mini**-------**AHT30**
u2 (+3.3V) -------     Vin  (red wire)
GND             -------     GND (withe wire)
TX                 -------     SCL (yellow wire)
RX                 -------     SDA (green wire)

Use this template
{"NAME":"Sonoff MINIR4","GPIO":[32,608,0,640,0,0,0,0,0,0,0,0,0,0,0,576,0,0,0,0,0,0,224,160,0,0,0,0,0,0,0,0,0,0,0,0],"FLAG":0,"BASE":1}

P_20231127_091330

tasmota/tasmota_xsns_sensor/xsns_63_aht1x.ino
```diff
@@ -56,12 +59,15 @@
...
 
-#define AHT1X_CMD_DELAY      40
-#define AHT1X_RST_DELAY      30
+#define AHT1X_CMD_DELAY      20
+#define AHT1X_RST_DELAY      10
 #define AHT1X_MEAS_DELAY     80 // over 75ms in datasheet
 
-#ifdef USE_AHT2x
-  #define AHTX_CMD     0xB1 // Cmd for AHT2x
+#ifdef USE_AHT3x
```
s-hadinger, line 66: I'm not a big fan of having a compile type option to make it AHT20 or AHT30 compatible but never both. Can you detect the type of device by probing an internal value, and adapt accordingly?

tasmota/language/ru_RU.h
```diff
@@ -84,7 +84,7 @@
 #define D_DATA "Данные"
 #define D_DARKLIGHT "Темный"
 #define D_DEBUG "Отладка"
-#define D_DEWPOINT "Dew point"
+#define D_DEWPOINT "Точка росы"
```
s-hadinger, line 87: Please make a separate PR for language support. Let's not mix topics.

tasmota/tasmota_xsns_sensor/xsns_63_aht1x.ino
```diff
@@ -56,12 +59,15 @@
 #define AHT_TEMPERATURE_OFFSET  50
 #define KILOBYTE_CONST          1048576.0f
 
-#define AHT1X_CMD_DELAY      40
-#define AHT1X_RST_DELAY      30
+#define AHT1X_CMD_DELAY      20
+#define AHT1X_RST_DELAY      10
```
s-hadinger, line 63: Any view on the previous values? Did you check with AHT10/AHT20? Did you check data sheets?

s-hadinger (changes requested):
Thanks for your PR. Can you please check the questions?

tasmota/tasmota_xsns_sensor/xsns_63_aht1x.ino
Jason2866: Maybe probing with the different CMD hex values?

legchenkov: Technically, there's no need to differentiate between AHT20 and AHT30 during compilation, as test results suggest they are fully compatible. If the command code is corrected and a delay is implemented after sending the start measurement command, both AHT20 and AHT30 will operate correctly. I couldn’t find a way to programmatically distinguish between them. The addition of the AHT30 driver selection was only made to ensure the corresponding name is displayed in the Tasmota interface.

tasmota/language/ru_RU.h
legchenkov: Okay, I will.

tasmota/tasmota_xsns_sensor/xsns_63_aht1x.ino
legchenkov: Personally, I have only tested the performance of the AHT20 and AHT30. They work. I do not have other versions of these sensors. The delays are taken from the datasheet for AHT20. In AHT30 datasheet initialization requires 5ms delay, and there is nothing about softset. The measurement time is the same for all three versions - 75ms. AHT10 datasheet requires less than 20ms for initialization after power on. But it happens at power-on and it is definitely faster than the ESP itself is initialized.

tasmota/tasmota_xsns_sensor/xsns_63_aht1x.ino
s-hadinger: Ok I understand and I think we can merge as-is. However I would largely prefer a common code for ATH10/20/30 and dynamic selection at run-time. This is in the philosophy of Tasmota.
Could you propose a future evolution in that direction?

s-hadinger (approved):
Ok to merge as-is, but I would largely prefer a future version that unifies ATH10/20/30 into a single code base, and dynamic selection

legchenkov:
> Ok to merge as-is, but I would largely prefer a future version that unifies ATH10/20/30 into a single code base, and dynamic selection

To make a universal driver that will work for all versions of this sensor seems realistic to me. But it will be possible to distinguish only AHT10, because it has a different code for sending a command. The other versions of AHT I do not know how to distinguish from each other programmatically. Anyway, I'll give it a try.

tobiloeb:
Is this feature already merged and available? I would like to use the AHT30 as well. :)

divadiow:
is there a test bin somewhere? i can't seem to find a workflow creation from PR

PR #20011, opened 2023-11-13
[Thermostat] Fix JSON index

SteWers:
## Description:

The index count of thermostat instances begins with `1`, but in the tele/SENSOR message it starts with `0`. This fixes the index to start always with `1`.

**Please pay attention:** Because of changing the index, the parsing of the JSON message has to be adjusted!
| old | new |
|--------|--------|
|`"Thermostat0":{"ThermostatModeSet":1,...`|`"Thermostat1":{"ThermostatModeSet":1,...`|

Also the index will be shown in the WebUI.

## Checklist:


SteWers:
@arendst Do you think, it's better to make a hint in the changelog / release notes, because of the breaking change?

PR #20013, opened 2023-11-14
Add the possibility to set global compile time settings

Jason2866:
## Description:

used in first step for the streaming Partition Wizard feature. The feature is great in most cases but for env trimmed to use as less as possible resources it is contra productive. 
Example: "accidently" hitting the Button on a C2 build can lead to a crash since 20k of heap needed for Partition Wizard can be to much.

@s-hadinger Comments, Opinions? Fine for you? 

It can now easily en/disabled in `platformio_override.ini`
in section [tasmota] with 

```
[tasmota]
; *** Global build / unbuild compile time flags for ALL Tasmota / Tasmota32 [env]
;build_unflags               =
build_flags                 =
```

## Checklist:

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -65,14 +65,9 @@ lib_extra_dirs              =
 ;enable_esp32_gz = 1
 ; Uncomment and specify a folder where to place the firmware file(s) (default set to folder build_output)
 ;bin_dir = /tmp/bin_files/
+; Global build flags (used for all env) can be overriden in "platformio_override.ini"
 build_unflags               =
-build_flags                 = -DTASMOTA  ; flag indicating that we are compiling Tasmota
-                              -DUSE_BERRY_PARTITION_WIZARD
-; *********************************************************************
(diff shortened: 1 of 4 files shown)


s-hadinger:
Sure. 

We could also add a test about the available memory and hide the button if memory is too low. This would be maybe more future proof

Jason2866:
Okay, so let's us merge this and a option to hide the button automatically and as an option for the user maybe a "Set Option"?

s-hadinger:
Do you have a map of env dependencies? I'm a bit lost.

It seems that `-DUSE_BERRY_PARTITION_WIZARD` is not applied to Core3 builds, at least not ESP32C3

Jason2866:
@s-hadinger Oops, overlooked the core3 part.

Jason2866:
Having two esp32 core setups to handle is tricky. If this works i will add to the `platformio_tasmota_core3_env_samle.ini`

PR #20036, opened 2023-11-16
Label of upload button

SteWers:
## Description:

As the label of the upload button is concatenated at runtime, it does not match in all languages. To solve this, there is now a full label text for `Start upgrade` and `Start restore`.

## Checklist:


barbudor:
The reason was to save flash bytes

SteWers:
I'm aware of saving memory. This PR costs 16 bytes, but I think that worth it.

PR #20059, opened 2023-11-19
Fix Berry parser error in specific cases

s-hadinger:
## Description:

Fix error in Berry parser that would cause registers not to be freed after an access via GETINDEX or GETMETHOD.

## Checklist:


s-hadinger:
Let's see for now. Haspmota was also recompiled with no change in the bytecode

PR #20088, opened 2023-11-23
Add Support for Avago Tech Bluetooth Buttons

kabongsteve:
## Description:

Support Bluetooth Buttons directly (without need for hub) and potentially other Bluetooth buttons

https://www.aliexpress.com/item/1005005791145037.html?spm=a2g0o.order_detail.order_detail_item.3.1478f19chQFmBA

Avago Tech Buttons come in 1 to 3 button versions and simply broadcast a Manufacturer packet with the following details without encryption

/*
  Offset Button Manufacturer Data
  0 Company ID = 2 bytes 0x0043
  2 Unknown = 4 bytes 7fffffff
  6 Counter = 1 byte
  7 Unknown = 1 byte 05
  8 Switch = 2 bytes  - unique per switch unit
  10 Unknown = 1 byte 00
  11 Button = 1 bytes 0x (x = 1 to 3)   single button 1 only dual button 1 and 3
  12 Unknown = 4 bytes 50000100
*/

NOTE: To use these buttons

- BLEAddrFilter must be set to 1 or above
- MI32Option1 1  - otherwise teleperiod will act like a button press
- Due to all buttons using same static random MAC address C1:A2:A3:A4:A5:A6 I dynamically update the MAC to include the Switch as the last two MAC values.  Thus slots into the MI32 sensors functionality

- If BLEAlias is used to remember buttons, both the static MAC address and the pseudo MAC addresses must be Aliased
eg
BLEAlias C1A2A3A4A5A6=AnyButton
BLEAlias C1A2A3A40058=Button3   (where 0058 is the Switch ID)

## Checklist:
  - [X ] The pull request is done against the latest development branch
  - [NA] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_62_esp32_mi_ble.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_62_esp32_mi_ble.ino
@@ -24,6 +24,8 @@
   --------------------------------------------------------------------------------------------
   Version yyyymmdd  Action    Description
   --------------------------------------------------------------------------------------------
+  0.9.2.2 20231123  changed - added support for Avago Tech Bluetooth buttons                              
+  -------
   0.9.2.1 20210217  changed - make features alos depend on received data - i.e. 'unknown' devices will show what they send.
                               Add MI32Option6 1 to switch to tele/tasmota_ble/<somename> style MQTT independent of HASS discovery.
   -------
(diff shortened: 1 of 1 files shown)


barbudor:
Interresting, other Buttons seemed to be a pain to integrate due to encryption.
Too late for me as I decided to go zigbee for such use cases but great to see working solution in BLE.

kabongsteve:
@barbudor   Similar issues, but I wanted units to directly interact with Tasmota and not require a separate service like MQTT or HomeAssistant.

IE my door/window sensors talk directly to each shutter controller so opening the window Locks the shutter to not operate.

These buttons are for my home grown Lift and (some of the shutters) and again want it running almost standalone without needing wifi or internet.

I was very pleased to find these and at <$5 each a bargain, especially with the fact there is no encryption .

PR #20117, opened 2023-11-28
Integrate TIMERS into Deepsleep

stefanbode:
Added new functionality to control the Deepsleep Wakeup process with TIMERS.
Timers must be type of "RULE"
"rule1 Wakeup" and ARMED and REPEAT will be used
Timer can use: TIME and SUNRISE and SUNSET with offset
Current +- variation is not supported

If above is NOT set the old behavior with "deepsleeptime xx" is still supported to wakeup on a regular base.

## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_29_deepsleep.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_29_deepsleep.ino
@@ -38,7 +38,7 @@
 
 const uint32_t DEEPSLEEP_MAX = 10 * 366 * 24 * 60 * 60;  // Allow max 10 years sleep
 const uint32_t DEEPSLEEP_MAX_CYCLE = 60 * 60;            // Maximum time for a deepsleep as defined by chip hardware
-const uint32_t DEEPSLEEP_MIN_TIME = 5;                   // Allow 5 seconds skew
+const uint32_t DEEPSLEEP_MIN_TIME = 15;                  // Allow 15 seconds skew
 const uint32_t DEEPSLEEP_START_COUNTDOWN = 4;            // Allow 4 seconds to update web console before deepsleep
 
 const char kDeepsleepCommands[] PROGMEM = D_PRFX_DEEPSLEEP "|"
@@ -53,7 +53,7 @@ uint8_t deepsleep_flag = 0;
 bool DeepSleepEnabled(void)
 {
   if ((Settings->deepsleep < 10) || (Settings->deepsleep > DEEPSLEEP_MAX)) {
-    Settings->deepsleep = 0;     // Issue #6961
+    Settings->deepsleep = 0;    // Issue #6961
     return false;               // Disabled
   }
 
@@ -144,7 +144,7 @@ void DeepSleepCalculate()
         // day_bitarray>0 otherwhise no weekday selected
         // rule keyword "Wakeup"
         // Timer action: rule
-        if (xtimer.arm && day_bitarray && GetRule(0) == "Wakeup" && POWER_BLINK == xtimer.power) {
+        if (xtimer.arm && day_bitarray && GetRule(0) == "Wakeup" && bitRead(Settings->rule_enabled, 0) && POWER_BLINK == xtimer.power) {
 #ifdef USE_SUNRISE
           if ((1 == xtimer.mode) || (2 == xtimer.mode)) {      // Sunrise or Sunset
             ApplyTimerOffsets(&xtimer);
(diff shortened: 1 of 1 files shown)


Jason2866:
@stefanbode Compile errors in special builds variants https://github.com/Jason2866/Tasmota-build/actions/runs/7023562479/job/19110370399
Could it be that when rules are disabled the compile fails?

stefanbode:
Thanks will check tomorrow morning. Should skip if rules and script in not enabled. Thanks for pointing

stefanbode:
1. You have to set currently deepsleeptime to 60 or something valid. Then you have 60 seconds.
2. If the device start not with a deepsleep wakeup you have 60 seconds. Countdown in the logs. Disable timers will make it
3. Yes and only captured by deepsleep. Documentation will need a change

smhc:
1. That is perhaps a bit confusing as "DeepSleepTime" will instruct when to sleep in this mode, but when to wakeup in the other mode.
2. I believe a common use case would be to sleep until a certain time each day then remain awake until ready to sleep again. e.g a lamp that goes into deep sleep during the day but is awake all night or something. (Or in my case a cat flap that wakes twice a day to perform an operation then wants to sleep immediately after the action is done).

How would that work?
1. Set a rule timer for sunset
2. On wake - how to keep awake? Do both a DeepSleepTime 0 and turn off the timer?
3. At sunset - how to sleep immediately? Re-enable the timer and do a "DeepSleepTime 1" ? (hopefully the "1" doesn't impact the next wake). 

Do you think a mandatory "Restart 9" or "DeepSleepResume" command that actually initiates the sleep might be a reasonable solution to all of the above.?

smhc:
Although I think a wake from "Restart 9" seems most intuitive, another option is using a timer to sleep. i.e "Rule1 Sleep" and "Rule2 Wakeup". This would allow you to schedule when to be asleep vs wake. I'm not sure about the implementation difficulty of this however.

stefanbode:
Regarding 1 I agree. This is not optimal right now. The main problem is there is no change in Timers and therefore a change here I cannot catch. There must be a restart or the deepsleeptime to get notification this is changed. I will check today how to optimize. 

My answer on 2 was wrong. You want to keep the device awake. Easiest way is to create a rule on system#start rule1 0. Maybe it is an idea to use the ONCE flag on rules. This is currently unused. In this case the rule Wakeup got disabled once go to deepsleep. To go to sleep on sunset currently a rule1 1;restart 1 will do the job. Please be aware minimum code addition is required. This is not a pc. Introducing all this Wakeup, Sleep, depsleepresume is an overkill. We have to be smarter.

smhc:
Yeah I agree a 'sleep timer' UI is probably overkill. It was just one possible solution to the problem of when to sleep. I don't think an automatic sleep immediately after wake is ideal when using a wake timer. So either an explicit sleep or scheduled sleep could solve that.

Writing a bit of code is fine - if anything I was proposing mandating writing some code to explicitly sleep instead of it being automatic.  I think it's more intuitive to have the wake timer only be concerned with wakes, rather than it also implying an automatic sleep that you need to disable on boot.

stefanbode:
@smhc : There is a new push on the way. As the initial build eas to integrate the timers in general the new push is to optimize the work and automation with deepsleep.

What changed:
as soon as there is a working configuration for deepsleep the 60 second timer starts. No need anymore for a reboot or a deepsleep whatever. To prevent a deepsleep best is to disable all timers in the UI until ready and then switch it on. 
the rule1 now also support the status 5 (ONCE). This does mean if you set `rule1 5` the deepsleep will take the next possible event and go to deepsleep. After wakeup it will stay alive. E.g. with this you are easily able to wakeup the device daily at sunrise.

What do you need to send the device back to deepsleep?
I would go with a rule2 and an event. 
`ON Time#Minute=%sunset% DO rule1 5 ENDON`

this will trigger the 60 second countdown before the device go to sleep.

smhc:
Hmm ok, so essentially turning on the special rule is how to turn on the sleep schedule and initiate first sleep.

Iterating the rules on "Restart 9" for one that is marked "Wakeup" still seems a bit more intuitive for the stay awake after wake scenario. But if "rule1 5" does actually go to sleep immediately (or close to it) then that is still workable for the "stay awake" scenario.

I'll give it a try over the next few days and see how it goes.

pkkrusty:
Looks like this new feature is actively being tweaked, so documentation probably doesn't exist yet. Is there anywhere laid out exactly how you envision it working? Is the goal to set a (for example) 12 hour timer, then, with deepsleeptime set to 60, the unit will deepsleep for 60 second intervals (like current implementation) until the timer runs out, then it will remain awake?

stefanbode:
Documentation already updated. The timers are a replacement for the Intervall to get more control. E.g. wake-up every morning on sunrise and goto sleep again. 

Intervall is still working, but you cannot mix.

stefanbode:
@smhc : the idea is nice but there is no option to select the rule that gets executed on a timer. I thought it is different but it isn't. Always all rules get called on the event. The 1,2,3,4... is only valid for the relays

smhc:
@stefanbode 

The point was to use the restart9 command to initiate the sleep.
It'd be the same code as this:

https://github.com/arendst/Tasmota/blob/28c8f8cf4707800b86fc066e22100f6fcf84ad54/tasmota/tasmota_xdrv_driver/xdrv_29_deepsleep.ino#L146

(which just uses the minimum across all timers 
https://github.com/arendst/Tasmota/blob/28c8f8cf4707800b86fc066e22100f6fcf84ad54/tasmota/tasmota_xdrv_driver/xdrv_29_deepsleep.ino#L161 )

but executed immediately by the Restart 9 command callback instead of using a countdown initiated in the deep sleep drivers TelePeriod callback. 

i.e 

```
CmdRestart9() {
   if Rule1=wakeup {
     DeepSleepCalculate()
   }
   if RtcSettings.nextwakeup > 0 {
     DeepSleepStart()
   } else {
     deepsleep()
   }
}
```

PR #20133, opened 2023-11-30
fixed possible char array overflow

ghost:
the temporary char arrays size was fixed to 4 bytes before. snprintf was set for the second char, so maximum size must be limited to 4-1, instead of 4 bytes. to avoid further mistakes usage of a #define
(assuming that flaw made never problems as the number of attached sensors was usually <99)

## Description:

## Checklist:
  - [ X] The pull request is done against the latest development branch
  - [ X] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_39_thermostat.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_39_thermostat.ino
@@ -1388,6 +1388,8 @@ uint8_t ThermostatGetDutyCycle(uint8_t ctr_output)
   return value;
 }
 
+#define TEMPLEN (4)
+
 void ThermostatGetLocalSensor(uint8_t ctr_output) {
   String buf = ResponseData();   // copy the string into a new buffer that will be modified
   JsonParser parser((char*)buf.c_str());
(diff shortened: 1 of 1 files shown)


stegerfa:
Yes, but "good catch" on my own mistake... Sorry, it seems I wasn't thinking carefully enough then... Today: second look, found.

stegerfa:
sfromis, thanks for that hint, I changed the name of the preprocessor-constant

tasmota/tasmota_xdrv_driver/xdrv_39_thermostat.ino
```diff
@@ -1398,9 +1400,9 @@ void ThermostatGetLocalSensor(uint8_t ctr_output) {
     if (  (THERMOSTAT_SENSOR_NUMBER > 1)
         &&(THERMOSTAT_CONTROLLER_OUTPUTS > 1)
         &&(ctr_output < THERMOSTAT_SENSOR_NUMBER)) {
-      char temp[4];
+      char temp[TASMOTA_XDRV_39_THERMOSTAT_TEMPLEN];
       temp[0] = IndexSeparator();
-      snprintf(&temp[1], 4, "%u", (ctr_output + 1));
+      snprintf(&temp[1], TASMOTA_XDRV_39_THERMOSTAT_TEMPLEN-1, "%u", (ctr_output + 1));
```
eku, line 1405:
```suggestion
      snprintf(&temp[1], sizeof(temp)-1, "%u", (ctr_output + 1));
```

PR #20143, opened 2023-12-02
Change HX711 default calibration percentage to 10 from 1

pkkrusty:
Current calibration resolution of 1 gives a resolution of +/-75g on a 20kg load cell, whereas the cheap aliexpress HX711 modules are generally able to output 1g precision on those load cells (https://forum.arduino.cc/t/hx711-noise-free-resolution-tests/351959). Setting calibration precision to 10 or higher allows more accurate calibration. @jeroenst did the PR for this, he may have insight on why the current default is 1.

## Description:

## Checklist:
  - [x ] The code change is tested and works with Tasmota core ESP8266 V.2.7.4.9
  - [x ] The code change is tested and works with Tasmota core ESP32 V.2.0.14

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_34_hx711.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_34_hx711.ino
@@ -44,7 +44,7 @@
 #define HX_SCALE             120     // Default result of measured weight / reference weight when scale is 1
 #endif
 #ifndef HX711_CAL_PRECISION
-#define HX711_CAL_PRECISION  1     // When calibration is to course, raise this value.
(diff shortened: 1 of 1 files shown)


jeroenst:
This pull request can not be accepted because it would change the reading values on current devices out in the field. This define is created so a developer can change the precision upon build, not to be changed in the default builds.

jeroenst:
@arendst can we change the define into a user configurable variable in the console?

jeroenst:
Even better would be writing a manual page for the hx711 here: https://github.com/arendst/Tasmota/pull/20143# . But comments are also an idea.

PR #20257, opened 2023-12-17
Unify de_DE

SteWers:
## Description:
Make de_DE more unify

## Checklist:

### DIFF
--- a/tasmota/language/de_DE.h
+++ b/tasmota/language/de_DE.h
@@ -28,7 +28,7 @@
  * Use online command StateText to translate ON, OFF, HOLD and TOGGLE.
  * Use online command Prefix to translate cmnd, stat and tele.
  *
- * Updated until v13.2.0.2 - Last update 19.11.2023
+ * Updated until v13.3.0.1 - Last update 17.12.2023
 \*********************************************************************/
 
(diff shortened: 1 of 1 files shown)


Noschvie (approved):

tasmota/language/de_DE.h
```diff
@@ -348,7 +348,7 @@
 #define D_MQTT_TLS_ENABLE "MQTT TLS"
 #define D_HTTP_API "HTTP API"
 #define D_HTTP_API_ENABLE "HTTP API aktivieren"
-#define D_FRIENDLY_NAME "Name [Friendly Name]"
+#define D_FRIENDLY_NAME "Friendly Name"
```
eku, line 351: Why not translate this one?

SteWers: Not everything should be translated as the are no better words in german. The previous expression breaks the two column design and doesn't say more than the new one.

Jason2866:
Whats better with Einstellungen as Konfiguration? Later on Konfiguration is still used.
I see no unify.  Anyways idk really whether the one or the other word is used...

SteWers:
That's philosophical. I don't know, which word is better, but for me Einstellungen feels better. At the underlying settings pages Einstellungen is also used. So it is unify. Only on save, restore and reset Konfiguration is used. I think, that it is the better word at that place.
But you're right. Out of the tasmota build, there are a few points not unify. I'll think about them.

Issue #5512, opened 2019-03-22
syslog log not complete where MQTT off


apneef:
### ISSUE DESCRIPTION - TROUBLESHOOTING
I am using a SHP6.
I have enable syslog to remote host.

When via console getting status of all, then in the console the status lines are displayed until status11.
But on the syslog host, there is only until status3

If i turn MQTT on, the status is being displayed correct in the syslog server.
  

### REQUESTED INFORMATION
_Make sure these boxes are checked before submitting your issue. Thank you_

**FAILURE TO COMPLETE THE REQUESTED INFORMATION WILL RESULT IN YOUR ISSUE BEING CLOSED**

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Sonoff-Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Sonoff-Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in issues (https://github.com/arendst/Sonoff-Tasmota/issues)
- [x] Searched the problem in the wiki (https://github.com/arendst/Sonoff-Tasmota/wiki/Troubleshooting)
- [x] Searched the problem in the forum (https://groups.google.com/d/forum/sonoffusers)
- [x] Searched the problem in the chat (https://discord.gg/Ks2Kzd4)
- [x] Device used (i.e. Sonoff Basic) : Blitwolz SHP6 - sonoff basic
- [x] Tasmota binary firmware version number used : 6.5.0 / pre-compiled
- [ ] Development IDE - Compiler / Upload tools used : ____ / ____
- [x] Provide the output of command ``status 0`` :

STATUS 0 OUTPUT HERE:

15:15:21 CMD: status 0
15:15:21 SRC: WebConsole from 192.168.2.249
15:15:21 RSL: Received Topic /status, Data Size 1, Data 0
15:15:21 RSL: Group 0, Index 1, Command STATUS, Data 0
15:15:21 RSL: STATUS = {"Status":{"Module":45,"FriendlyName":["SONOFF_01"],"Topic":"SONOFF_01","ButtonTopic":"0","Power":1,"PowerOnState":3,"LedState":1,"SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":0}}
[...]

PR #20260, opened 2023-12-18
Realtime syslog

budulinek:
## Description:

## Checklist:

At the moment, MQTT logging through `MqttPublishLoggingAsync()` function and syslog logging through `SyslogAsync()` are provided at a second interval. The reason for the delay is a fear by developers that “immediate logging would delay time critical code execution severely” because Tasmota is running on a single-threaded setup without asynchronous operations. 
This PR removes the artificial delay. `SyslogAsync()`  is called directly in `Scheduler()` which means that syslog is provided immediately. 

# IMPACT ASSESSMENT

Syslog messages (UDP) are provided immediately. Since the main concern of the devs were possible (blocking) delays during code execution, I will do a brief assessment of this PR on the speed of code execution under various scenarios:

# 1. You do not use syslog

This PR will have no impact on your code . The [SyslogAsync()](https://github.com/arendst/Tasmota/blob/2204ecdc516b885e88f02caff5c83688e7d0b1af/tasmota/tasmota_support/support.ino#L2289)  function returns immediately if TasmotaGlobal.syslog_level equals 0 (syslog is disabled). There is no blocking delay.

# 2. You do not care about syslog
This PR will have no impact on your code execution. TasmotaGlobal.syslog_level is set to 0 by default.

# 3. You use syslog (you have enabled syslog)
There are two possible sources of blocking delay:

### a) The provided syslog server host name can not be resolved
The `WifiHostByName()` function is blocking. If the syslog server (provided by you) does not exist on the network, the code execution will be blocked for the duration of DNS_TIMEOUT (1000 millis). But if the syslog host name can not be resolved ( `WifiHostByName()` returns with 0 ), TasmotaGlobal.syslog_level is temporarily set to 0 for the period of SYSLOG_TIMER (600 seconds). After this SYSLOG_TIMER period is over, syslog_level returns back to the value set by the user (you). Which means that another blocking delay (1s long) will occur after 600 seconds). 
Impact assessment: this PR will not lead to additional (or longer) blocking delays. Both BEFORE and AFTER this PR, you have a 1s blocking delay every 600s. Regardless of your syslog level. Even if you set syslog to 4, you will have a 1s blocking delay every 600s if your server name can not be resolved.

**BEFORE:**
-	log event is recorded (stored in buffer)
-	non-blocking delay for up to 1s
-	the 1s interval is over, SyslogAsync() is called
-	WifiHostByName() blocks the code execution for 1s
-	syslog_level is temporarily set to 0 for 600s (within these 600s you will not experience any additional blocking delays)

**AFTER:**
-	log event is recorded (stored in buffer)
-	SyslogAsync() is called immediately
-	WifiHostByName() blocks the code execution for 1s
-	syslog_level is temporarily set to 0 for 600s (within these 600s you will not experience any additional blocking delays)

### b) Possible delay caused by sending individual UDP messages.
Individual log messages are send within a [while loop](https://github.com/arendst/Tasmota/blob/2204ecdc516b885e88f02caff5c83688e7d0b1af/tasmota/tasmota_support/support.ino#L2299)  If there are blocking functions within this loop, code execution is blocked. There already is a blocking `delay()` ( 1 millisecond as a solution to #5512 issue). In my opinion, other functions ( `PortUdp.beginPacket()` , `PortUdp.write()` , `PortUdp.endPacket()` ) are non-blocking. But even if they were blocking, this PR actually improves (not worsens) code execution!
Let’s assume that all functions within the while loop ( `delay()` , `PortUdp.beginPacket()` , `PortUdp.write()` , `PortUdp.endPacket()` ) cause a blocking delay T for each syslog UDP message. Let’s assume high syslog_level which causes 3 log messages within 1s:

**BEFORE**
-	3 log event is recorded (stored in buffer)
-	non-blocking delay for up to 1s
-	the 1s interval is over, `SyslogAsync()` is called
-	there are 3 log messages in a buffer
-	the while loop blocks code execution for the period 3 x T
-	`SyslogAsync()` returns

**AFTER**
-	log event is recorded (stored in buffer)
-	`SyslogAsync()` is called immediately
-	the while loop blocks code execution for the period T
-	`SyslogAsync()` returns
-	log event is recorded (stored in buffer)
-	`SyslogAsync()` is called immediately
-	the while loop blocks code execution for the period T
-	`SyslogAsync()` returns
-	log event is recorded (stored in buffer)
-	`SyslogAsync()` is called immediately
-	the while loop blocks code execution for the period T
-	`SyslogAsync()` returns

The whole idea of delaying syslog messages (current situation) is wrong. It is always better to have multiple short blocking delays than one long blocking delay.


barbudor:
Sending logs should remains outside of the critical loop.
The need for "real-time" logs has not been demonstrated

sfromis:
It could be feasible to have your borderline case for using syslog as transactions instead of logging served by having your change depend on `#ifdef`, which can be set (or not) at build time, with the default being the current implementation.

barbudor:
Moving a logging system to "real-time" in order to get a UDP based publication mechanism as an alternative to MQTT is failed-by-design

If what you need is an alternative UDP publication mechanism propose an optional driver just like KNX  or Domoticx

budulinek:
Hi folks,
thanks for your quick replies!

@sfromis 
> Pushing the delays till after the current activity generating messages means faster execution of the primary function, and then any blocking network operations can be executed after Tasmota has done the processing which possibly created multiple messages. Hence it can be clearly worse (in terms of slowing down actions) to always take multiple shorter delays immediately before completing the function.

The primary function is called by `Scheduler()` . In my PR, the `SyslogAsync()` function is also called by the `Scheduler()`, not by the primary function. The primary function generates multiple log messages, stores them in the log buffer (I am not eliminating the buffer!). The primary function does its job and returns to `Scheduler()` . Only then the `SyslogAsync()` is called. My PR does not "slow down actions" of the primary function.

Also, please note that within the `SyslogAsync()` function, all (potentially) blocking code is within [this while loop](https://github.com/arendst/Tasmota/blob/2204ecdc516b885e88f02caff5c83688e7d0b1af/tasmota/tasmota_support/support.ino#L2299) (and 1 log message = 1 run of this loop). If the blocking code was outside of this while loop (some kind of network initiation?), the economics of the code execution would be different and I would support delaying `SyslogAsync()`. But this is not the case and I stand by my assertion that it is always better to call `SyslogAsync()` immediatelly by the `Scheduler()` (i.e. after the primary function returns to `Scheduler()` ).

@barbudor 
> Sending logs should remains outside of the critical loop.

This is a single-threaded setup. At the end of the day, everything is executed within this "critical loop" (i.e. `Scheduler()` ). The only question is when to execute (send logs) and for how long your "critical loop" ( `Scheduler()` ) will be blocked. You face 2 options:

1. Send syslogs immediatelly after the primary function does its job (returns to `Scheduler()` ). Your "critical loop" is blocked for multiple short delays. Inbetween these short delays, your "critical loop" is ready to execute the primary function.
2. Delay sending syslog, store more messages in the buffer and face longer delay down the road. Your "critical loop" will be blocked later, but by longer delay. 

> Moving a logging system to "real-time" in order to get a UDP based publication mechanism as an alternative to MQTT is failed-by-design

1. Being "real time" is the expected behaviour of syslog (regardless of how you want to use it).
2. With all respect, this comment is irrelevant for this PR. Sure, people can use (or misuse) syslog in a way it was not meant to be used. But that should not be the (sole) reason for delaying syslog.

> If what you need is an alternative UDP publication mechanism propose an optional driver just like KNX or Domoticx

Actually.... No. I do not need "an alternative UDP publication mechanism". I want syslog which behaves in a way people expect it to behave (= real time), unless there is a reason for a delay. And so far I do not see any reason for the delay.

arendst:
The OP's solution still allows for completing the main task before updating syslog; only at the end of the scheduler the syslog is updated (so not at every log addition).

With sleep set to 50 syslog is also updates as soon as after 50mSec.

Let's see if shutter users like to change (as it sprawls the log when enabled).

PR #20278, opened 2023-12-21
Replace ellipsis

SteWers:
## Description:
Replace ellipsis in most common files
## Checklist:


arendst:
Without looking at the code I'm quite sure your character is not an utf-8 (ascii) character and hence will use more than one byte.

SteWers:
Yes, you're right. In UTF-16 it is 0x2026 and in UTF-8 0xE2 0x80 0xA6. So the code isn't smaller, but it looks better in the output…

arendst:
Code size over cosmetics ;-)

SteWers:
OK, but the code size is the same. From 0x2E 0x2E 0x2E to 0xE2 0x80 0xA6.

arendst:
But I don't like the cosmetics.

arendst:
And my legacy tools (like Termite) hate it too:
```
14:20:47.049-029 MQT: Attempting connectionâ€¦
```

SteWers:
Uhh, that's bad. In VS and Teraterm the serial output is OK.

sfromis:
I'd say that anything which replaces standard 7-bit (or maybe 8-bit) ASCII by use of more specialized Unicode characters should be avoided, unless there is a **very** good reason. Which there is not in this case, and personally I do not even think that it looks better.

And while I agree that Unicode should be used widely when i18n is a possible issue, replacing any character codes (except for 7-bit US-ASCII giving the same UTF-8 "encoded" data), there still are lots of tools or encoding specifications needing to be upgraded to painlessly support Unicode in all weird tools relevant to a development toolchain.

PR #20352, opened 2023-12-30
MiElHVAC: don't send commands to the AC while reading data from the AC.

dgwynne:
on some units it can take around 250ms to reply to a request, by which time we're shoving another command to the unit. if this happens, the unit gives up and starts replying to the new command, which can again take 250ms. in this situation effectively nothing gets through.

avoid this by checking if we're in the parser state machine. this also gives us timeout handling.

tested on 4 different AC units. one which was unusable before is now functioning as expected, and the other 3 appear just as functional as they were before.

## Description:

## Checklist:


arendst:
thx

PR #20365, opened 2024-01-01
Add a second recovery path if BLE stack stops being able to receive adverts

lis0r:
Add a second recovery path if BLE stack stops being able to receive adverts

## Description:

I have twice seen the receiving of passive adverts stall, requiring a device restart to resume their reception. The issue is highly intermittent, and I have not found a means of reproducing it at will. I suspect the issue is related to https://github.com/espressif/esp-idf/issues/4001 .

The logs contain a lines stating "BLE: scan stall? no adverts > 120s, restart BLE". The restart of NimBLE is successful, meaning that Tasmota does not restart. However, the restart of NimBLE is not sufficient to rectify the underlying problem, so the message repeats every 2 minutes.

I have adjusted the check so that it will try to restart NimBLE after 120s of no advertisements. After this, if there is still no advertisements after 240s, it will restart Tasmota.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
@@ -421,6 +421,7 @@ uint64_t BLELastLoopTime = 0;
 int BLEScanTimeS = 20; // scan duraiton in S
 int BLEMaxTimeBetweenAdverts = 120; // we expect an advert at least this frequently, else restart BLE (in S)
 uint64_t BLEScanLastAdvertismentAt = 0;
+uint32_t BLEScanLastAdvertisementAtResets = 0;
 uint32_t lastopid = 0; // incrementing uinique opid
 uint32_t BLEResets = 0;
 // controls request of details about one device
@@ -1333,6 +1334,7 @@ class BLEAdvCallbacks: public NimBLEAdvertisedDeviceCallbacks {
     TasAutoMutex localmutex(&BLEOperationsRecursiveMutex, "BLEAddCB");
     uint64_t now = esp_timer_get_time();
     BLEScanLastAdvertismentAt = now; // note the time of the last advertisment
+    BLEScanLastAdvertisementAtResets = BLEResets; // note the number of resets when the last advertisement occurred
 
     uint32_t totalCount = BLEAdvertisment.totalCount;
     memset(&BLEAdvertisment, 0, sizeof(BLEAdvertisment));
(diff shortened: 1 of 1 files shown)


Jason2866:
There will be no merge. No driver is allowed to restart Tasmota just when the specific driver fails forever reason.

s-hadinger:
There is one exception though with zigbee. If the MCU goes into an invalid state, there is no choice other than a full restart of both Tasmota and the MCU. 

I'm not familiar with BLE. What does not adverts mean? Can it happen if there are no BLE devices in range?

barbudor:
I assume this also occurs only when using active scan?
I'm all passive with PVVX'ed Mija and I haven't seen this blocking

Probably can also be done with a rule

lis0r:
It is definitely not a case of no devices in range - my house is a BLE heavy environment. The node in question normally reports 13 devices, and there are more devices just slightly out of range. From what I can see, the reason it finds zero devices is because the BLE controller has taken an error so severe that even restarting the BLE stack is insufficient to fix it. Manually restarting Tasmota quickly and definitively fixes the BLE stack.

I have active scans disabled to save power, and I operate purely on receiving passive adverts from a number of MJ_HT_V1,  MCCGQ02HL, and SJWS01LM devices running their stock firmware. As mentioned, it's highly intermittent - I've been running otherwise OK for months, but it's happened twice within the last few days, with no recent changes to set up or radio environment.

It should be noted that the BLE driver can already restart Tasmota: if restarting NimBLE actively fails or times out, Tasmota will be restarted after 10 seconds. This change is effectively building upon that model to cope with a false positive.

I will investigate rules.

s-hadinger:
Let me rephrase my question. What happens if a user enables the driver but has zero BLE device in range? could Tasmota also reboot every 240 seconds?

Jason2866:
> It should be noted that the BLE driver can already restart Tasmota: if restarting NimBLE actively fails or times out, Tasmota will be restarted after 10 seconds. This change is effectively building upon that model to cope with a false positive.

This part of the code should not be there. Seems it was overlooked in the review.

Jason2866:
Iirc this driver is not ready for Arduino core 3.0.0 since it uses h2zero `NimBLE-Arduino` library. For Arduino core 3.0.0 to support BLE with C2 and C6 too it is needed to switch to h2zero `esp-nimble-cpp`. To use this library some changes are needed.

lis0r:
Berry bodge around, for others looking for a similar work around:

```
last_ads = 0
def ble_trigger(value, trigger, msg)
  if value['adverts'] == last_ads
    tasmota.cmd("restart 1")
  end
  last_ads = value['adverts']
end
tasmota.add_rule("BLE", ble_trigger)
```

PR #20392, opened 2024-01-04
Support LWT messages when using TASMESH

### LINKED ISSUE #20369: Fire trigger on Tasmesh broker when a node disconnects
### PROBLEM DESCRIPTION
When using [Tasmesh](https://github.com/arendst/Tasmota/blob/516621af5250d3ec39dfb349db1f07ba1730e7c0/info/xdrv_57_tasmesh.md), nodes won't send any MQTT Online or Offline messages, meaning that tools like Home Assistant won't be able to use these nodes as they will permanently appear offline. Sending the Online message is easy enough through the use of a rule on the node, e.g. `ON system#init DO Backlog MeshNode 00:11:22:33:44:55; Publish tele/tasmota_XXXXX/LWT Online ENDON`, however sending the Offline message is currently impossible: the node won't be able to send it as it's now powered off, and the broker never realizes that the node is gone and doesn't fire any triggers.

The best solution would be to somehow implement LWT messages on the broker, but in the meantime a nice workaround would be for the broker to detect when a node has gone offline and print a message which can then be used as a trigger to run a `Publish` command with the LWT Offline message.

### REQUESTED INFORMATION

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions). Related discussions: #
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
  - [ ] Pre-compiled
  - [ ] Self-compiled
```lua
  Configuration output here:


```
```lua
  Rules output here:


```
```lua
  STATUS 0 output here:


```
```lua
  Console output here:


```

### TO REPRODUCE

### EXPECTED BEHAVIOUR
_A clear and concise description of what you expected to happen._

### SCREENSHOTS

### ADDITIONAL CONTEXT

pipe01:
## Description:

This PR enables the TASMESH broker to detect when nodes come online and offline through the use of heartbeat packets. When a change is detected, the corresponding LWT Online or Offline message is sent through MQTT.

**Related issue (if applicable):** fixes #20369 and #20375

## Checklist:

### DIFF
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -395,7 +395,9 @@
                                                  // Enabled by default on ESP32 and variants
 
 // -- ESP-NOW -------------------------------------
-//#define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
+#define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
+#define TASMESH_HEARTBEAT      1                 // If enabled, the broker will detect when nodes come online and offline and send Birth and LWT messages over MQTT correspondingly (0 = off, 1 = on)
+#define TASMESH_OFFLINE_DELAY  3                 // Maximum number of seconds since the last heartbeat before the broker considers a node to be offline
 
 // -- OTA -----------------------------------------
 //#define USE_ARDUINO_OTA                          // Add optional support for Arduino OTA with ESP8266 (+13k code)
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -395,9 +395,9 @@
                                                  // Enabled by default on ESP32 and variants
 
 // -- ESP-NOW -------------------------------------
-#define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
-#define USE_TASMESH_HEARTBEAT                    // If enabled, the broker will detect when nodes come online and offline and send Birth and LWT messages over MQTT correspondingly
-#define TASMESH_OFFLINE_DELAY  3                 // Maximum number of seconds since the last heartbeat before the broker considers a node to be offline
+//#define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
+//#define USE_TASMESH_HEARTBEAT                    // If enabled, the broker will detect when nodes come online and offline and send Birth and LWT messages over MQTT correspondingly
+//#define TASMESH_OFFLINE_DELAY  3                 // Maximum number of seconds since the last heartbeat before the broker considers a node to be offline
 
 // -- OTA -----------------------------------------
 //#define USE_ARDUINO_OTA                          // Add optional support for Arduino OTA with ESP8266 (+13k code)
--- a/tasmota/my_user_config.h
+++ b/tasmota/my_user_config.h
@@ -396,7 +396,7 @@
 
 // -- ESP-NOW -------------------------------------
 #define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
-#define TASMESH_HEARTBEAT      1                 // If enabled, the broker will detect when nodes come online and offline and send Birth and LWT messages over MQTT correspondingly (0 = off, 1 = on)
+#define USE_TASMESH_HEARTBEAT                    // If enabled, the broker will detect when nodes come online and offline and send Birth and LWT messages over MQTT correspondingly
 #define TASMESH_OFFLINE_DELAY  3                 // Maximum number of seconds since the last heartbeat before the broker considers a node to be offline
 
 // -- OTA -----------------------------------------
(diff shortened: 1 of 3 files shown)


tasmota/my_user_config.h
```diff
@@ -395,7 +395,9 @@
                                                  // Enabled by default on ESP32 and variants
 
 // -- ESP-NOW -------------------------------------
-//#define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
+#define USE_TASMESH                              // Enable Tasmota Mesh using ESP-NOW (+11k code)
```
barbudor, line 398: Please leave all the below lines commented in your PR
Users wanting to enable the feature will add them in their user_config_override.h

tasmota/tasmota_xdrv_driver/xdrv_57_9_tasmesh.ino
```diff
@@ -649,6 +679,16 @@ void MESHEverySecond(void) {
       MESHsetWifi(1);
       WifiBegin(3, MESH.channel);
     }
+
```
barbudor, line 682: Suggestion: send the heartbeat request only if no other packets came in within the last second

tasmota/my_user_config.h
```diff
@@ -395,7 +395,9 @@
...
+#define TASMESH_HEARTBEAT      1                 // If enabled, the broker will detect when nodes come online and offline and send Birth and LWT messages over MQTT correspondingly (0 = off, 1 = on)
```
s-hadinger, line 399: We usually take the convention of the feature to be defined instead of having value 1. Also prefix with `USE_`

Should be `#define USE_TASMESH_HEARTBEAT`

tasmota/my_user_config.h
```diff
@@ -395,7 +395,9 @@
...
+#define TASMESH_OFFLINE_DELAY  3                 // Maximum number of seconds since the last heartbeat before the broker considers a node to be offline
```
s-hadinger, line 400: Please change to `#define USE_TASMESH_OFFLINE_DELAY  3`

tasmota/tasmota_xdrv_driver/xdrv_57_9_tasmesh.ino
```diff
@@ -546,6 +546,25 @@ void MESHevery50MSecond(void) {
...
+                _peer.isAlive = true;
+                char stopic[TOPSZ];
+                GetTopic_P(stopic, TELE, _peer.topic, S_LWT);
+                MqttPublishPayload(stopic, PSTR(MQTT_LWT_ONLINE));
+              }
+              break;
+            }
+          }
+#else
+        break;
```
s-hadinger, line 565: Why don't you leave the `break;` with heartbeat enabled?

s-hadinger (changes requested):
Please change the defines

tasmota/my_user_config.h
barbudor: Configuration values generally do not have the `USE_` prefix

s-hadinger: Ah ok. This are my own conventions then. Still I prefer to have the same prefix for all, it makes searches easier

tasmota/tasmota_xdrv_driver/xdrv_57_9_tasmesh.ino
pipe01: Missed that, thanks

tasmota/tasmota_xdrv_driver/xdrv_57_9_tasmesh.ino
pipe01: I think it would be a good idea to send the heartbeat regardless because maybe the packet we sent before didn't make it to the broker (e.g. maybe the TTL was too short)

Jason2866: @barbudor Okay, or should be changed?

barbudor: I don't think it's critical, we can leave it like this

Jason2866:
@Staars fine for you?

Staars:
No complaints from my side.
We had the topic of LWT on our radar relatively early on, but have not yet started any concrete implementations. These are always best created based on concrete problems and solutions. I assume the PR works as intended and has been tested for some time.

I remember thinking relatively early on about the concept of different nodes, which could be "always on" or "uncertainly accessible". The former should send a guaranteed heartbeat, where the absence would indicate an error.
The latter would only report when they wake up from a deep sleep, for example, and would not take part in forwarding data packets. In order to control this remotely, a functioning LWT would be necessary. However, due to a lack of concrete applications, I didn't pursue the concept any further.

PR #20408, opened 2024-01-06
Add support for pipsolar inverter

chefpro:
## Description:

This PR adds support for solar inverter from "EASUN Power" and simular. A lot of inverters from china use this protocol. Baudrate and serial protocol parameter are settable. Polling of data can be switched off.
The advantage over implementations from ESPEasy and others are, that QET,QEY,QEM,QED,QEH and DAT is also supported and the webinterface in tasmota which makes it easy to look at values without broker, mqttclient and such.

Berry scripts to sync date and time in inverters with local time are possible through DAT call.
Also through the QEx-Api it is possible to paint graphs of historical data in webinterface with berry without the need of heavyweight database and graphics framework.

I would not recommed to enable it by default, because of memory footprint.

The implemention is asyncron and shoud not slow down tasmota too much.

## Checklist:
  - [X ] The pull request is done against the latest development branch
  - [ X] The code change is tested and works with Tasmota core ESP32 V.2.0.14


arendst:
Thx.

As you noticed I won't enable it by default.  Too much code duplication and non-standard mqtt messages probably flooding the rules engine. It's your call.

LittleCypher:
Hi,
Thank you very much for your great work, before I was reading every minute from serial and saving the data to mySQL, so I'm really  looking forward to implement this as a simpler way.  The only problem is that I don't really know how to enable it and implement it to tasmota. Can you please write a few lines on how I can do that?
Thanks in advance!

ouinouin:
@chefpro thanks for this contribution , i gor a voltronic clone : superwatt VMII, ill try your code , maybe i ll ping you  in tasmota comments if i get some questions on how to use it ( i have plenty of friends having those inverters, and the esphome soluce is complicated to integrate.

ouinouin:
@chefpro , i compiled successfully on an esp32S3 but its not working, i dont have enough coding knowledge to understand from the code how to set the baudrate, sbaudrate command is not available despite having selected pipsolartx and pipsolarrx , but i see in your code that it can claim the hardware serial, but i dont know what is the condition to go hardware or software.
at the end , i d like to be able to send some commands to change parameters of the inverter, will it be possible or you implemented just a simple read of parameters ?

PR #20494, opened 2024-01-14
Berry example - FTP server

Staars:
## Description:

Feature rich FTP server in Berry, that is a bit exaggerated considering the supported commands for demonstration purposes. Written from scratch in a week, so obviously not very polished nor long term tested.
Supports active and passive mode - passive seems to work a bit more stable.

Download should work pretty well with over 400 kB/s at times on my S3 dev-board, basically I never had a problem here and this probably is a valid use case.
Upload at about 60 kB/s with mediocre stability. At least ( tested with FileZilla) resumed upload works most of the time. By design of FTP it is impossible to check for a valid upload for the server. Only the client can check the size afterwards.

In my tests, the upload of the Arduino version was more stable, but I have seen silent upload errors here too.

Compatibility to various FTP clients on macOS is higher than the Arduino version. Especially  the pure GUI applications do a lot of stuff under the hood (double uploads, renaming or copying on the server) - I definitely do not recommend these for the ESP32 (Arduino or Berry).
I could never get it to work with the builtin FTP client from macOS (with cmd + K from Finder), but every other tested client did work.

I did see occasional crashes on the TTGO watch, maybe there are still bugs on my side.
Changing from passive to active mode at runtime did seem to bring problems too.

For daily use I would recommend to slim this down to the needed commands from the chosen FTP client which i.e. can be seen in the detailed log of FileZilla.

Usage:
Save file as `ftp.be`

Start with `br load("ftp")` from the serial/web console or with `load("ftp")` from Berry.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.5

### DIFF
--- a/tasmota/berry/drivers/ftp.be
+++ b/tasmota/berry/drivers/ftp.be
@@ -1,0 +1,564 @@
+#- Simple FTP server in Berry by Christian Baars
+#  supports active and passive mode - but passive is preferred!
+#  only light error handling
+-#
+
+class PATH    # helper class to hold the current directory 
+    var p  #  path components in a list
+
+    def init()
+        import string
+        self.p = []
+    end
+
+    def set(p)
+        import string
+        import path
+
+        if path.isdir(p) != true
+            return false
+        end
+
+        var new = string.split(p,"/")
+        self.p = []
+        for c:new
+            if c != ""
+                self.p.push(c)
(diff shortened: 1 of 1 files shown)


s-hadinger:
Awesome job, thanks!

PR #20523, opened 2024-01-16
Rename "Konsolen" to "Werkzeuge"

SteWers:
I think it is also the better word in german. Refering to https://github.com/arendst/Tasmota/commit/d51340961fe2487c673b157335cdf85e6d73106c.

### DIFF
--- a/tasmota/language/de_DE.h
+++ b/tasmota/language/de_DE.h
@@ -275,7 +275,7 @@
 #define D_CONFIGURATION "Einstellungen"
 #define D_INFORMATION "Informationen"
 #define D_FIRMWARE_UPGRADE "Firmware Update"
-#define D_MANAGEMENT "Konsolen"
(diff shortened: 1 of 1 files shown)


eku:
> I think it is also the better word in german.

I'm sorry, but `Tools` is not German. How about `Verwaltung`?.

Jason2866:
Do find both suggestions not perfect.

Staars:
Besides I always use the stock english version, I think the most correct term would be:
_Dienstprogramme_ - which is quite long
a bit shorter:
_Werkzeuge_

Tools is definitely no translation and if you want to read _tools_ at that place, then just flash the english version.

SteWers:
> Do find both suggestions not perfect.

I also think it's not perfect.
For me, `Werkzeuge` is the best suggestion.

Issue #20521, opened 2024-01-16
Exception 6 IntegerDivideByZero When set ShutterMode 4


rt400:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._
When set ShutterMode 4 i got Exception 6 IntegerDivideByZero

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [X] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [X] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [X] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [X] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [X] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [X] Device used (e.g., Sonoff Basic): ESP32 Lolin
- [X] Tasmota binary firmware version number used: 13.3.0.3
  - [X] Pre-compiled
  - [ ] Self-compiled
- [X] Flashing tools used: Esptool_____
- [X] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:
{"NAME":"ESP32-DevKit","GPIO":[1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,0,1,1,1,0,0,0,0,1,1,1,1,1,0,0,1],"FLAG":0,"BASE":1}
{"Module":{"1":"ESP32-DevKit"}}
[...]

PR #20524, opened 2024-01-17
Fixed DIV/0

stefanbode:
## Description:
#20521

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.5

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_27_esp32_shutter.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_27_esp32_shutter.ino
@@ -687,7 +687,7 @@ void ShutterInit(void)
           Shutter[i].min_realPositionChange = 0;
         break;
         case SHT_COUNTER:
-          Shutter[i].min_realPositionChange = SHT_DIV_ROUND(Shutter[i].min_realPositionChange, Shutter[i].motordelay);
+          Shutter[i].min_realPositionChange = SHT_DIV_ROUND(Shutter[i].min_realPositionChange, Shutter[i].motordelay > 0?Shutter[i].motordelay : 1);
         break;
       }
 


tasmota/tasmota_xdrv_driver/xdrv_27_esp32_shutter.ino
```diff
@@ -687,7 +687,7 @@ void ShutterInit(void)
           Shutter[i].min_realPositionChange = 0;
         break;
         case SHT_COUNTER:
-          Shutter[i].min_realPositionChange = SHT_DIV_ROUND(Shutter[i].min_realPositionChange, Shutter[i].motordelay);
+          Shutter[i].min_realPositionChange = SHT_DIV_ROUND(Shutter[i].min_realPositionChange, Shutter[i].motordelay > 0?Shutter[i].motordelay : 1);
```
arendst, line 690: Is there no better way to make sure ``Shutter[i].motordelay`` well never become 0? User entry checks?

stefanbode: Good point. This is a special requirement only for shuttermode 4 and the stepper motors. On all other shuttermodes the correct and default value is 0. If checked on userinput I would need to change it when you change the shuttermode and additionally check on the cmdShutterMotordelay what is the current shuttermode and reject. For sure possible, but maybe not less code

arendst: OK. Let's do it this way then if see if it at least solves the exception 6. Thx.

barbudor: Something I often use when I want to avoid a zero value is an OR with 1
```
result = value | 1;
```
Of course it may modified the value but depending on use case, it might not be a problem
And it's very light on code

PR #20525, opened 2024-01-17
add support for ams5915 and ams6915 temerature and pressure sensors

baschdello:
## Description:
Adds support for another sensor: AMS5915 and AMS6915 from Analog Microelectronics. There are many types of these sensor series available for various pressure ranges. Pressure range can be set with sensor command _Sensor114_. Pmin and Pmax of sensor is saved at sensor configfile on the filesystem (as discussed in #15791). If filesystem is not available then pmin and pmax defaults to 0.
Command example for sensor type "AMS 5915-0005-D-B" (differential and bidirectional): _Sensor114 -5 5_

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP8266 V.2.7.5

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
@@ -59,7 +59,7 @@ struct {
   int16_t   pmax;
 } amsx915Settings;
 
-void AmsDetect(void) {
+void Amsx915Detect(void) {
   // sensor datasheets does not provide any information/commands on sensor
   // specific register so it is not possible to properly detect sensor type
   if(Wire.available() != 4) {
@@ -77,7 +77,7 @@ void AmsDetect(void) {
     AddLog(LOG_LEVEL_INFO, PSTR("CFG: AMS detected"));
 }
 
-bool AmsCommand() {
+bool Amsx915Command() {
   int32_t vals[2];
   ParseParameters(2,(uint32_t *)vals);
   if(XdrvMailbox.data_len >= 3 && XdrvMailbox.data_len < 13) {
@@ -85,7 +85,7 @@ bool AmsCommand() {
       if (vals[1] >= -32768 && vals[1] < 32768) {
         amsx915Settings.pmin = (int16_t)vals[0]; // save pmin of sensor
         amsx915Settings.pmax = (int16_t)vals[1];  // same with pmax
-        AmsSettingsSave();
+        Amsx915SettingsSave();
         Response_P(S_JSON_SENSOR_INDEX_SVALUE, XSNS_114, "pressure range set");
         return true;
       }
@@ -96,7 +96,7 @@ bool AmsCommand() {
   return false;
 }
 
-void AmsReadData(void) {
+void Amsx915ReadData(void) {
   uint8_t buffer[4];
   Wire.requestFrom(AMSX915_ADDR, 4);
   if(Wire.available() != 4) {
@@ -114,7 +114,7 @@ void AmsReadData(void) {
   amsx915data.present = true;
 }
 
-void AmsShow(bool json) {
+void Amsx915Show(bool json) {
   if(amsx915data.present) {
     if (json) {
       ResponseAppend_P(PSTR(",\"AMSx915\":{\"" D_JSON_TEMPERATURE "\":%1_f,\"" D_JSON_PRESSURE "\":%2_f}"), &amsx915data.temperature, &amsx915data.pressure);
@@ -133,7 +133,7 @@ void AmsShow(bool json) {
  * Driver Settings load and save
 \*********************************************************************************************/
 
-void AmsSettingsLoad(bool erase) {
+void Amsx915SettingsLoad(bool erase) {
   // Called from FUNC_PRE_INIT (erase = 0) once at restart
   // Called from FUNC_RESET_SETTINGS (erase = 1) after command reset 4, 5, or 6
 
@@ -175,7 +175,7 @@ void AmsSettingsLoad(bool erase) {
 
       // Set current version and save settings
       amsx915Settings.version = AMSX915_VERSION;
-      AmsSettingsSave();
+      Amsx915SettingsSave();
     }
     AddLog(LOG_LEVEL_INFO, PSTR("CFG: AMS config loaded from file"));
   }
@@ -186,7 +186,7 @@ void AmsSettingsLoad(bool erase) {
 #endif  // USE_UFILESYS
 }
 
-void AmsSettingsSave(void) {
+void Amsx915SettingsSave(void) {
   // Called from FUNC_SAVE_SETTINGS every SaveData second and at restart
 #ifdef USE_UFILESYS
   uint32_t crc32 = GetCfgCrc32((uint8_t*)&amsx915Settings +4, sizeof(amsx915Settings) -4);  // Skip crc32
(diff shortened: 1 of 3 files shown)


tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,251 @@
...
+      }
+    }
+    Response_P(S_JSON_SENSOR_INDEX_SVALUE, XSNS_114, "invalid pressure range [-32768..32767]");
+    return false;
+  }
+  return false;
+}
+
+void AmsReadData(void) {
+  uint8_t buffer[4];
```
s-hadinger, line 100: Why don't you test for `amsx915data.present`? It is not good to send a request on the I2C bus if there is no device connected. Did I miss something?

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,251 @@
...
+    Serial.println("AMS error read!");
+    amsx915data.present = false;
+    return;
+  }
+  buffer[0] = Wire.read();
+  buffer[1] = Wire.read();
+  buffer[2] = Wire.read();
+  buffer[3] = Wire.read();
+
+  amsx915data.pressure = ((256*(buffer[0]&0x3F)+buffer[1])-1638.0)*(amsx915Settings.pmax-amsx915Settings.pmin)/13107+amsx915Settings.pmin;
```
s-hadinger, line 112: Do you really need floating point here?

If you really need them, please force to single precision float with `1638.0f`

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,251 @@
...
+  amsx915data.temperature = (((256.0*buffer[2]+buffer[3])*200.0)/65536)-50;
```
s-hadinger, line 113: Same, force every literal to `float`, otherwise the computation is done with `double` and the result is converted to `float`. Which increases A LOT the code size

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,251 @@
...
+#else
+  // Try to load file /.drvset122
+  char filename[20];
+  snprintf_P(filename, sizeof(filename), PSTR(TASM_FILE_SENSOR), XSNS_114);
+
+  if (erase) {
+    TfsDeleteFile(filename);  // Use defaults
+  }
+  else if (TfsLoadFile(filename, (uint8_t*)&amsx915Settings, sizeof(amsx915Settings))) {
+  Serial.println("TfsLoadFile");
```
s-hadinger, line 161: You should not have `Serial.print` anywhere. Please remove

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,251 @@
...
+  int16_t   pmax;
+} amsx915Settings;
+
+void AmsDetect(void) {
+  // sensor datasheets does not provide any information/commands on sensor
+  // specific register so it is not possible to properly detect sensor type
+  if(Wire.available() != 4) {
+    // Error
+    amsx915data.present = false;
+    AddLog(LOG_LEVEL_INFO, PSTR("CFG: AMS not detected!"));
```
s-hadinger, line 68: You shouldn't log at `INFO` level any undetected device, otherwise the logs would be crowded by all the potential devices you may connect. Please remove, or at least demote to `LOG_LEVEL_DEBUG`

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,251 @@
...
+
+    return;
+  }
+  while(Wire.available()) {
+    Wire.read();
+  }
+  amsx915data.present = true;
+  I2cSetActiveFound(AMSX915_ADDR, "AMSx915");
+    AddLog(LOG_LEVEL_INFO, PSTR("CFG: AMS detected"));
```
s-hadinger, line 77: Is `AMS detected` detailed enough. Why not `AMSx915 detected` ?

baschdello:
Hi s-hadinger,
good points. I will check and edit asap.

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,217 @@
...
+  if (!I2cEnabled(XI2C_86)) { return false; }
+
+  bool result = false;
+
+  switch(function) {
+    case FUNC_PRE_INIT:
+      Amsx915SettingsLoad(0);
+      break;
+    case FUNC_INIT:
+      Amsx915ReadData();
```
s-hadinger, line 191: I'm still not ok with this. You never check if the device is actually plugged into the bus, and blindly send requests to the bus every second. We can't accept this kind of behavior from every driver, it would overwelm the bus and take a long time for each I2C request to time out when the device is not present.

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
```diff
@@ -0,0 +1,217 @@
...
+
+const uint16_t AMSX915_VERSION = 0x0100;       // Latest sensor version (See settings deltas below)
+
+// Global structure containing driver saved variables
+struct {
+  uint32_t  crc32;    // To detect file changes
+  uint16_t  version;  // To detect driver function changes
+  int16_t   pmin;
+  int16_t   pmax;
+} amsx915Settings;
```
barbudor, line 60: Allocate data dynamically IF the sensor is detected, and use that pointer as a test point for presence in other calls such as the call back
See existing drivers such as https://github.com/arendst/Tasmota/blob/development/tasmota/tasmota_xsns_sensor/xsns_112_ens210.ino

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
barbudor: See other drivers such as the one I pointed above for rules of good design

barbudor (changes requested):

tasmota/tasmota_xsns_sensor/xsns_114_amsx915.ino
baschdello: Think about the following case: it is possible that sensor is detached from bus (by user/contact problem/emi). If I check for `amsx915data.present` and sensor is temporary fallen of the bus, there will be no redetection of sensor. Because i2c is a master-slave-bus, it is necessary to poll the specific device address as sensors are slaves. If we won't do this, we can not redetect sensor and we need to reset the complete software which is not good at standalone measuring devices. For example the BME sensors are redetected after I remove and reattach them. I wanted to have this exact behavior at AMSx915 implementation.
I will add a counter like @barbudor showed in ens210 implementation to send requests only n seconds.

barbudor: I understand your point but 
1) Tasmota is not meant to detect sensors that are added dynamically
2) One good reason for that is you can destroy sensor or ESP by doing so

So if you want to allow redetecting a sensor that was there at the beginning but temporarily disappeared, you have to 
- Detect it once at start and have an indicator of detection (I suggest using the pointer for the dynamically allocated memory as such, no need for a `xxxx.present` boolean)
- As recommend, use that indicator to ignore totally the driver is the sensor was not detected at boot

That will prevent to poll the I2C bus for nothing it the sensor was not there at the beginning and still allow to redetected if it temporarilly disapear.

People tends to build generic binaries that include all kind of sensors they may use in order to have 1 binary they can use on different boards.
Acting as you do means that this driver is using RAM and CPU even when on a board where this sensor is not present.

Just to make it clear, I think you understand from @s-hadinger comments and mine this is a hard requirement to have the sensor merged in Tasmota

baschdello:
Hello,
I had time to test the code on the ESP32 and it works. I've noticed and fixed a missing initialisation of variable 'i'.
Do I have to make a fresh pull request to get the sensor code merged into tasmota?
Best regards
Bastian

PR #20561, opened 2024-01-22
Utouch update

gemu2015:
## Description:

Add XPT2046

now all discrete touch driver chips are supported

i updated display.docs and will provide samples for each chip

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_10_scripter.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_10_scripter.ino
@@ -5875,7 +5875,7 @@ extern char *SML_GetSVal(uint32_t index);
           goto exit;
         }
 #endif // USE_TTGO_WATCH
-#if defined(USE_FT5206) || defined(USE_XPT2046) || defined(USE_LILYGO47) || defined(USE_UNIVERSAL_TOUCH) || defined(USE_GT911)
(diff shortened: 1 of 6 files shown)


tasmota/tasmota_xdrv_driver/xdrv_10_scripter.ino
Jason2866, line 1: You noticed all touch drivers are now  possible via Utouch driver. Why still using the specific drivers here? Is a further PR planned to remove?
Mid / long term goal (discussed with Theo) is to remove specific display drivers and use only your great piece of UDisplay driver. Same should be done for UTouch driver.
So no more specific drivers for display and touch.

gemu2015: i used this for comparing and testing.  will be removed later.
there is only 1 display driver not yet ported to Universal mode: RA8876. i am working on this but this is a very complex driver and will take a while.

i have a lot of displays and testet the touch driver with several but perhaps we should wait removing drivers until more people also have tested it.

Jason2866: Thx for the answer. So everything goes into direction "Universal" driver. Great.
As said no hurry. It is a mid/long term goal. It is done, when it is done ;-)

PR #8021, opened 2020-03-29
Enhanced counter debouncing

to-scho:
New commands CounterDebounceLow and CounterDebounceHigh to allow individual debounce times for low and high pulse widths to discard non valid falling edges. These are checked before legacy CounterDebounce checks distance between to valid falling edges.
Useful to have robust counter results when using e.g. an TCRT5000 optical sensor to count electromechanical "Ferraris" electricity meters.

## Description:

**Related issue (if applicable):** fixes #7488, fixes #516

## Checklist:
  - [x] The pull request is done against the latest dev branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR.
  - [x] The code change is tested and works on core Tasmota_core_stage
  - [ ] The code change pass travis tests. **Your PR cannot be merged unless tests pass**
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

PR #20712, opened 2024-02-12
Fix counter rising edges when CounterDebounceLow/High is set

curzon01:
## Description:

Counters were only set on falling edges, regardless of whether [CounterDebounceLow](https://tasmota.github.io/docs/Commands/#counterdebouncelow) or [CounterDebounceHigh](https://tasmota.github.io/docs/Commands/#counterdebouncehigh) was set or not.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_01_counter.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_01_counter.ino
@@ -71,7 +71,7 @@ void CounterIsrArg(void *arg) {
     // passed debounce check, save pin state and timing
     Counter.timer_low_high[index] = time;
     Counter.pin_state ^= (1<<index);
-    // do not count on rising edge
+    // for rising edge
     if bitRead(Counter.pin_state, index) {
       // PWMfrequency 100
       // restart PWM each second (german 50Hz has to up to 0.01% deviation)
@@ -79,7 +79,10 @@ void CounterIsrArg(void *arg) {
 #ifdef USE_AC_ZERO_CROSS_DIMMER
       if (index == 3) ACDimmerZeroCross(time);
 #endif //USE_AC_ZERO_CROSS_DIMMER
-      return;
+      // do not count on rising edge if CounterDebounceLow and CounterDebounceHigh is unset
+      if (0 == Settings->pulse_counter_debounce_low && 0 == Settings->pulse_counter_debounce_high) {
+        return;
+      }
     }
   }
 


arendst:
DO you have a use-case for this? I ask because there never was a reason to change this code part as far as I know and the impact is not clear to me.

curzon01:
> DO you have a use-case for this? 

Tipping bucket rain gauge:
Each tilting movement closes or opens a reed contact, i.e. each tilting movement corresponds to a fixed amount of rain, but you have to count both the rising and the falling edge

> I ask because there never was a reason to change this code part as far as I know and the impact is not clear to me.

I wonder why the counter does not worked as described for CounterDebounceLow/High, maybe I just misunderstand the descritiopn for [CounterDebounceLow](https://tasmota.github.io/docs/Commands/#counterdebouncelow)/[CounterDebounceHigh](https://tasmota.github.io/docs/Commands/#counterdebouncehigh):

"_When unequal zero tasmota will check falling and rising edges on the counter's GPIO._"

I just wonder what CounterDebounceLow/High exists for if you can't use it to recognize both edges separately.

curzon01:
Theo, as I read the origin PR (#8021) this PR will not be backward compatible, I think the misleading part is the description for CounterDebounceLow/High.

Anyway, I find the separate detection of falling and rising edges useful in cases like this. Should I solve this via an SO or an additional CounterType?

arendst:
Looking at the important part of the code:
```
      Counter.any_counter = true;
      pinMode(Pin(GPIO_CNTR1, i), bitRead(Counter.no_pullup, i) ? INPUT : INPUT_PULLUP);
      if ((0 == Settings->pulse_counter_debounce_low) && (0 == Settings->pulse_counter_debounce_high) && !Settings->flag4.zerocross_dimmer) {
        Counter.pin_state = 0;
        attachInterruptArg(Pin(GPIO_CNTR1, i), CounterIsrArg, &ctr_index[i], FALLING);
      } else {
        Counter.pin_state = 0x8f;
        attachInterruptArg(Pin(GPIO_CNTR1, i), CounterIsrArg, &ctr_index[i], CHANGE);
      }
```
wouldn't it make sense in your case not to use PCD_low and PCD_high but just use the original basic debounce which uses the CHANGE interrupt?

arendst:
Would this solve your issue:
```
    // do not count on rising edge
    if bitRead(Counter.pin_state, index) {
      // PWMfrequency 100
      // restart PWM each second (german 50Hz has to up to 0.01% deviation)
      // restart initiated by setting Counter.startReSync = true;
#ifdef USE_AC_ZERO_CROSS_DIMMER
      if (index == 3) { ACDimmerZeroCross(time); }
      return;
#else
      if (!Settings->flag6.counter_both_edges) {
        return;
      }
#endif //USE_AC_ZERO_CROSS_DIMMER
    }
```
making sure the ZeroCross dimmer keeps working.

curzon01:
try to summarize
- PCD_low/PCD_high already activates IRQ on CHANGE
- PCD_low/PCD_high was originally intended by the author only to add additonal pre-debounces for falling and rising edges, not to count both separately.
- PCD_low/PCD_high cmnd description is misleading in that both edges are counted

I want to be backward compatible not touching the current behaviour but allow additonal count of falling & rising edges by an optional setting e.g CounterType 2 or some SO

curzon01:
> Would this solve your issue:
> 
> ```
>     // do not count on rising edge
>     if bitRead(Counter.pin_state, index) {
>       // PWMfrequency 100
>       // restart PWM each second (german 50Hz has to up to 0.01% deviation)
>       // restart initiated by setting Counter.startReSync = true;
> #ifdef USE_AC_ZERO_CROSS_DIMMER
>       if (index == 3) { ACDimmerZeroCross(time); }
>       return;
> #else
>       if (!Settings->flag6.counter_both_edges) {
>         return;
>       }
> #endif //USE_AC_ZERO_CROSS_DIMMER
>     }
> ```
> 
> making sure the ZeroCross dimmer keeps working.

Yes, that's in my mind

arendst:
I'll change it now.

arendst:
Done.

PR #20760, opened 2024-02-18
Allow mapping of one PWM channel to multiple GPIOs

sudden6:
## Description:

Allow same PWM channel on multiple GPIO pins on ESP32.

This implementation seems to work, but it has one drawback. It will use a a new PWM hardware channel per configured PWM output, so you can never have more than 16 PWM outputs even if all of them are configured as PWM1. Also, the PWM signal is not guaranteed to be exactly the same only that it has the same duty cycle. The hardware PWM channels are reconfigured at different times and phasing is applied on all outputs. This is no problem for light strips though.

A proper implementation would use the ESP32 GPIO_MUX to route the same LEDC PWM channels to multiple GPIO pins, but this would be more complicated to implement with the current framework of Tasmota.

**Related issue (if applicable):** https://github.com/arendst/Tasmota/discussions/10884

## Checklist:
  - [] Only one feature/fix was added per PR and the code change compiles without warnings

I don't have any ESP8266 devices for testing, but the changes are only in regions that won't affect it. I also took the liberty of adding some documentation in doxygen format, let me know if that's ok.

### DIFF
--- a/tasmota/tasmota_support/support.ino
+++ b/tasmota/tasmota_support/support.ino
@@ -1411,14 +1411,30 @@ void ConvertGpios(void) {
 }
 #endif  // ESP8266
 
-int IRAM_ATTR Pin(uint32_t gpio, uint32_t index = 0) {
+/**
+ * @brief Find the pin number with a given GPIO function
+ * 
+ * @param gpio The GPIO function to look for
+ * @param index Special function to look for, custom encoding per GPIO function
+ * @param start_search Start index in the global GPIO function list, use to search
+ *        for multiple occurences of the same GPIO function on different pins. Pass the value
+ *        returned by previous calls to this function.
+ * @returns The pin number of the first pin matching the GPIO function,
+ *          -1 if none found or -2 on invalid start_search parameter.
+ */
+int IRAM_ATTR Pin(uint32_t gpio, uint32_t index = 0, int start_search = 0) {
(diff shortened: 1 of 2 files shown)


Jason2866:
Without testing if and how it affects the esp8266, there is no merge possible.

s-hadinger:
I will review it

s-hadinger:
I'm favorable to have this special semantics: if two GPIO have the same `PWM<x>` then they are replicating the same behavior. Actually it would make sense to have the same with `RELAY<x>`

I looked at esp-idf api, it should be possible to use the Mux to have only one timer driving 2 GPIOs, but I'm afraid this would be too specific, and fragile whenever esp-idf evolves again. The need is still a niche, so having this limitation of consuming one timer per actual GPIO should be ok.

s-hadinger:
I just need to test on ESP8266 before merging

s-hadinger:
I tested on ESP8266, but it does not expose the same behavior. I.e. the duplication of PWM does not work on ESP8266.

I'm annoyed to not have a consistent behavior on ESP8266 and ESP32. If you want this feature integrated, you need to backport it to ESP8266 as well.
Thanks

sudden6:
@s-hadinger I don't have any ESP8266 device and I don't think the PWM generators are similar enough that I could do a blind port.

PR #20814, opened 2024-02-26
add support for ams5915 and ams6915 temerature and pressure sensors

baschdello:
## Description:
Adds support for another sensor: AMS5915 and AMS6915 from Analog Microelectronics. There are many types of these sensor series available for various pressure ranges. Pressure range can be set with sensor command Sensor114. Pmin and Pmax of sensor is saved at sensor configfile on the filesystem (as discussed in https://github.com/arendst/Tasmota/discussions/15791). If filesystem is not available then pmin and pmax defaults to 0.
Command example for sensor type "AMS 5915-0005-D-B" (differential and bidirectional): Sensor114 -5 5

## Checklist:


tasmota/my_user_config.h
```diff
@@ -738,6 +738,7 @@
 //    #define DS3231_ENABLE_TEMP                   //   In DS3231 driver, enable the internal temperature sensor
 //    #define USE_BM8563                           // [I2cDriver59] Enable BM8563 RTC - found in M5Stack - support both I2C buses on ESP32 (I2C address 0x51) (+2.5k code)
 //    #define USE_PCF85363                         // [I2cDriver66] Enable PCF85363 RTC - found Shelly 3EM (I2C address 0x51) (+0k7 code)
+    #define USE_AMSX915                          // [I2CDriver86] Enable AMS5915/AMS6915 pressure/temperature sensor (+1k2 code)
```
s-hadinger, line 741: Not sure it should be enabled by default. @Jason2866 @arendst

arendst: You're right but in the end it needs to be disabled before merge so pls comment this line and I'll merge.

Jason2866 (approved):

PR #20868, opened 2024-03-04
[tuyamcu_v2] Send updates unconditionally

marcusb:
## Description:

Remove the logic that inhibits the sending of updates to the MCU until the DP's state has been observed at least once in a status report from the MCU. This logic was intended to ensure that a DP is not updated with its current value, which reportedly crashes some very broken devices.

However, other devices like the Feit DIM/WIFI dimmers do not reliably report DPs that haven't been changed. So the required status report would never arrive, resulting in the inability to control these devices remotely, at least until their state had been changed by a manual button press, something that is not always practical as it requires physical device access and needs to be done after every restart.

Removing this logic allows Tasmota to control the device state.

## Checklist:


marcusb:
cc: @btsimonh

PR #20906, opened 2024-03-10
Fix warning caused from `ESPFtpServer.cpp`

Jason2866:
## Description:

undef the already defined `#define F()` to suppress redefine warning.
Imho there should be in general no redefine done. A different define is already in Arduino core

## Checklist:

### DIFF
--- a/lib/lib_div/ESPFtpServer/ESPFtpServer.cpp
+++ b/lib/lib_div/ESPFtpServer/ESPFtpServer.cpp
@@ -25,6 +25,7 @@
 #include <time.h>
 
 #ifdef ESP32
+#undef F
 #define F(A) A
 #endif
 


arendst:
Not sure if the redefine breaks Arduino define of F().....

Jason2866:
Yep, agree. It is in general a dirty hack to redefine. It will break anyways, or do I miss something?

gemu2015:
this is the Arduino 32 define, actually a NOP
#define F(string_literal) (string_literal)

but for strange reasons F() causes an error on ESP32  without the redefine ???

gemu2015:
a little more research

 pgmspace.h
 #define PSTR(s)       (s)
 
 WString.h
 #define FPSTR(str_pointer) (reinterpret_cast<const __FlashStringHelper *>(str_pointer))
 #define F(string_literal) (FPSTR(PSTR(string_literal)))

causes this error
/Users/gerhardmutz1/.platformio/packages/framework-arduinoespressif32/cores/esp32/WString.h:37:29: error: invalid conversion from 'const __FlashStringHelper*' to 'long long unsigned int' [-fpermissive]

i could not evaluate why this happens.

since #define in .cpp files does not affect other files it is no harm to redefine this macro and is not worth the effort for more research

PR #21026, opened 2024-03-24
Fix crash in Universal Touch

s-hadinger:
## Description:

@gemu2015 there is currently a crash if you compile with both `#define USE_XPT2046` and `#define USE_UNIVERSAL_TOUCH` and `:TS,*` is present in `display.ini`.

This is due to initialized values of `ut_*` variables.

With the configuration above, the TS would not work, but at least it wouldn't crash.

If you're ok, I would like to consider `#define USE_UNIVERSAL_TOUCH` as the new default in LVGL build, and remove legacy drivers.

## Checklist:

### DIFF
--- a/lib/lib_display/UDisplay/uDisplay.h
+++ b/lib/lib_display/UDisplay/uDisplay.h
@@ -456,18 +456,18 @@ class uDisplay : public Renderer {
 
   uint8_t ut_array[16];
   uint8_t ut_i2caddr;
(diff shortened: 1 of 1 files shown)


Jason2866:
Yes, let us enable the Utouch driver in variant "LVGL" and disable the currently dedicated touch drivers.

s-hadinger:
Let me merge. Anyways this patch is harmless

PR #21029, opened 2024-03-25
display ini to utouch

gemu2015:
## Description:

replace touch entries with universal touch code

## Checklist:


Jason2866:
@arendst Can you increase patch version number when merging this (and the other related PR)?

s-hadinger:
Awesome.
My only concern is that we don't gracefully manage the transition. If we include Universal Touch in the pre-compiled binaries, then all the previous configuration fail until the user updates the display.ini - and they can't update them unless they migrated to Universal Touch screen.

Is there a way that both changes don't need to be done at the same time?

Or maybe we need to leaves `:TS` for now until user upgrade to latest firmware?

Jason2866:
Maybe a transition version, which has dedicated and Utouch driver enabled?
When this version is launched it modifies the Utouch part in display.ini and restarts with changed settings and does some touch tests (Berry script)?

s-hadinger:
Now that uTouch is default, we can merge

PR #21081, opened 2024-04-01
Added RelayLock to Lock specific relays from beeing changed

stefanbode:
## Description:
Sometimes it might be useful and required to LOCK a relay e.g. to temporary disconnect ALEXA or other smart home from changing a relay. The new command relaylock&lt;x&gt; 1|0 can disable or enable a lock on a specific relay

- interlock still works - higher priority
- running pulsetimers will NOT change after timeout if relaylock set during timer
- shutters gets confused if relay does not switch. Anyhow Shutterlock exist. Therefore acceptable issue

## Checklist:

### DIFF
--- a/tasmota/include/i18n.h
+++ b/tasmota/include/i18n.h
@@ -285,6 +285,7 @@
   #define D_STATUS13_SHUTTER "SHT"
 #define D_CMND_STATE "State"
 #define D_CMND_POWER "Power"
+#define D_CMND_RELAYLOCK "RelayLock"
 #define D_CMND_TIMEDPOWER "TimedPower"
 #define D_CMND_FANSPEED "FanSpeed"
 #define D_CMND_POWERONSTATE "PowerOnState"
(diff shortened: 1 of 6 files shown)


stefanbode:
> Thx for the code. I won't merge this PR as it is. I see a better location in the Settings area and I want to rename the command from `RelayLock` to `PowerLock` as it also works on display power and lights.
> 
> I will implement this my way.
> 
> BTW did you test what happens with bi-stable relays (using two bits within the power uint32_t)?

Hi Theo, "RelayLock" was a hallucination of ChatGPT how to do it. Therefore I assumed this might be most straight forward. Anyhow PowerLock is also fine. In this case ChatGPT has to learn it :-)

Bi-Stable relays is new to me. Never seen this. I will read the doc and test it with the current implementation. Thx.

arendst:
Functionality implemented as command ``PowerLock``.

arendst:
Regarding the bi-stable relay it seems to work fine too. 

So let's see what users come up with.

Thx for the idea.

PR #21102, opened 2024-04-05
Bugfix: power0 if power_lock is used

stefanbode:
## Description:
power0 changed the power regardless if power_lock was set.

## Checklist:

### DIFF
--- a/tasmota/tasmota_support/support_tasmota.ino
+++ b/tasmota/tasmota_support/support_tasmota.ino
@@ -397,11 +397,11 @@ void SetAllPower(uint32_t state, uint32_t source)
         break;
     case POWER_ON:
         // Keep locked bits and set all other to 1
-        TasmotaGlobal.power = (TasmotaGlobal.power & Settings->power_lock) | ~Settings->power_lock;
+        TasmotaGlobal.power = (TasmotaGlobal.power & Settings->power_lock) | (all_on & ~Settings->power_lock);
         break;
     case POWER_TOGGLE:
         // Keep locked bits and toggle all other
-        TasmotaGlobal.power ^= ~Settings->power_lock;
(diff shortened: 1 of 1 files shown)


stefanbode:
@arendst : Was there a reason to set only on the devices_present the power to 1 and not to the complete 32bit array?. 
So is it required to do:
power_t all_on = POWER_MASK >> (POWER_SIZE - TasmotaGlobal.devices_present);

arendst:
> Was there a reason to set only on the devices_present the power to 1 and not to the complete 32bit array?.

Yes. As the print output will show all bits set even if there are just only 1 or two relays/lights present.

tasmota/tasmota_support/support_tasmota.ino
```diff
@@ -392,13 +392,17 @@ void SetAllPower(uint32_t state, uint32_t source)
     power_t all_on = POWER_MASK >> (POWER_SIZE - TasmotaGlobal.devices_present);
     switch (state) {
     case POWER_OFF:
-      TasmotaGlobal.power = 0;
-      break;
+        // keep loocked bits and set all other to 0
+        TasmotaGlobal.power &= Settings->power_lock;
```
arendst, line 396: This will turn on relays which have locked bit set but were OFF. Not good.

Ditto for the other changes.

stefanbode: ok readded the all_on filter and do some testing. Looks reasonable now to me

stefanbode: Regarding: TasmotaGlobal.power &= Settings->power_lock;  Your finding is wrong. it is a &= and therefore keep only the locked relays ON. All other go OFF as intended.

arendst:
Thx. Tested. OK

PR #21173, opened 2024-04-14
Revert "Core 2015" -> back to Core 2.014

Jason2866:
@arendst


arendst:
Let's wait a few days and see if others also have issues with HWCDC. If not it must be me.

PR #21180, opened 2024-04-14
Breaking: ESP32 Core 3.0.0 only support

Jason2866:
## Description:

remove Platformio build support for Core 2.0.x and build all ESP32 variants only with Core 3.0.0 

CI is expected to fail. Changes in Github Actions.

## Checklist:


.github/workflows/Tasmota_build_devel.yml
```diff
@@ -96,14 +96,13 @@ jobs:
...
           - tasmota32solo1-safeboot
           - tasmota32-safeboot
-          - tasmota32c3-safeboot
+          - tasmota32c2-safeboot
           - tasmota32c3cdc-safeboot
           - tasmota32s2-safeboot
           - tasmota32s2cdc-safeboot
           - tasmota32s3-safeboot
           - tasmota32s3cdc-safeboot
           - tasmota32c2-safeboot
```
arendst, line 105: Duplicate tasmota32c2-safeboot ???

.github/workflows/Tasmota_build_master.yml
```diff
@@ -23,14 +23,13 @@ jobs:
...
           - tasmota32solo1-safeboot
           - tasmota32-safeboot
-          - tasmota32c3-safeboot
+          - tasmota32c2-safeboot
           - tasmota32c3cdc-safeboot
           - tasmota32s2-safeboot
           - tasmota32s2cdc-safeboot
           - tasmota32s3-safeboot
           - tasmota32s3cdc-safeboot
           - tasmota32c2-safeboot
```
arendst, line 32: Ditto

arendst:
What is your opinion on changing ``..cdc.bin`` names to "normal" names without the cdc addition? As we now can detect HWCDC over serial, one binary is enough and I see no reason to call it ..cdc.bin

Jason2866:
@arendst I had the same idea in mind. It is technically not 100% correct since the build settings from CDC is different to the non CDC builds. So I choose CDC for the name. And use the CDC build as one and only variant

eku:
Since the update from 13.4 (tasmota32-display) to 14.0, my board no longer starts. With git-bisect I found out that this commit is the cause. It is a Heltic Wifi-Kit-32 V1 with ESP32-D0WDQ6. 

Should I open a ticket?

arendst:
Did you read the changelog for v14.0? Did you notice the major change in display usage?

eku:
> Did you read the changelog for v14.0? Did you notice the major change in display usage?

Sure, I did. #21183 does not break it.

With "no longer starts" I mean it does not start anymore, no output at serial, no WiFi connect. Nothing.

arendst:
With no serial output it's time to call coroner or investigate if the correct binary is loaded....

eku:
> investigate if the correct binary is loaded

At least I flashed the correct factory.bin file, the one from the release page.

And this is where I run out of options for further analysis.

arendst:
Try to flash an older binary to see if you get serial back. If that works just perform an OTA of a newer release instead of serial factory.

eku:
That worked, but it updates the app0 partition only. So the cause is in one of the other partitions, right?

arendst:
If you did a serial flash of a factory image then all is rewritten not just app0. So I cannot understand your question. Using the factory image there is only one app0 partition and one safeboot partition.

arendst:
Yes. Correct. If you want to update the safeboot partition execute command ``upload 2`` in the console. Using command ``restart 3`` switches between the two partitions.

PR #21207, opened 2024-04-17
Add Publish3 command to send binary MQTT data

mbezuidenhout:
## Description:

Add ability to the Publish command to send a hexstring as binary MQTT data.

Publish command now has the following options

`Publish3 | <topic> <payload> = MQTT publish any topic and optional binary payload. For example the payload 7461736D6F7461 will output tasmota`

## Checklist:


sfromis:
With ESP32 and Berry, Tasmota already supports publishing binary data.

tasmota/tasmota_support/support.ino
```diff
@@ -590,6 +590,33 @@ char* Trim(char* p) {
   return p;
 }
 
+size_t HexStringToBinaryString(char** source) {
```
s-hadinger, line 593: Please use `ToHex_P()` instead of adding a new implementation

tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
```diff
@@ -737,6 +737,21 @@ void MqttPublishPayload(const char* topic, const char* payload) {
...
+  if (binary) {
+    if (strlen(response_data) > 3 && strncmp(response_data, "0x", 2) == 0) {
+      binary_length = HexStringToBinaryString(&response_data);
+    }
+  }
+
+  MqttPublishPayload(topic, response_data, binary_length, retained);
+}
+
 void MqttPublish(const char* topic, bool retained) {
```
s-hadinger, line 755: Is this one still needed?

tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
```diff
@@ -737,6 +737,21 @@ void MqttPublishPayload(const char* topic, const char* payload) {
   MqttPublishPayload(topic, payload, 0, false);
 }
 
+void MqttPublish(const char* topic, bool retained, bool binary) {
+  // Publish <topic> default ResponseData string with optional retained
+  char* response_data;
+  uint32_t binary_length = 0;
+  response_data = ResponseData();
```
s-hadinger, line 744: I'm not sure this is valid, since ResponseData() is `(char*)TasmotaGlobal.mqtt_data.c_str();`

I suppose this conversion shouldn't be used unless binary conversion is explicitly needed

s-hadinger (changes requested):
Some changes requestes.

Overall, can you explain the use case? Do we really need to add this option?

barbudor:
Personally I would prefer a prefix like you originally proposed. Whether 0x or $
Nevertheless the smallest solution should prevail

tasmota/tasmota_support/support.ino
mbezuidenhout: The ToHex_P function takes an array and converts it into a hex string. I'm doing the inverse of that.
It takes the binary value and maps it to one of the characters in `"static const char * hex PROGMEM = "0123456789ABCDEF";"`

mbezuidenhout:
Use case for this would be a Sonoff RC to WiFi bridge receiving a signal and then controlling the relays on a WaveShare ethernet 8ch relay. [https://www.waveshare.com/wiki/Modbus_POE_ETH_Relay](https://www.waveshare.com/wiki/Modbus_POE_ETH_Relay)

tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
mbezuidenhout: The function `void MqttPublish(const char* topic, bool retained)` is used in xdrv_12_home_assistant.ino, xdrv_27_shutter.ino and a few others.
The function could just look for strings in the format `0x([0-9A-Za-z]{2})+` but that might cause a breaking change where someone is trying to send a string and not binary data.

Noschvie:
> Use case for this would be a Sonoff RC to WiFi bridge receiving a signal and then controlling the relays on a WaveShare ethernet 8ch relay. https://www.waveshare.com/wiki/Modbus_POE_ETH_Relay

This device supports MQTT? How do you convert MQTT to be accepted by this device?
Why not using Modbus TCP?

mbezuidenhout:
This device supports MQTT. Details are located at the bottom of the wiki. Sending data with Modbus TCP is not supported by tasmota commands. https://tasmota.github.io/docs/Commands/

Noschvie:
> This device supports MQTT. Details are located at the bottom of the wiki.

ah.... using the Waveshare Cloud or with a local MQTT broker too?

Noschvie:
> Sending data with Modbus TCP is not supported by tasmota commands.

maybe a feature request is needed :-)

tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
mbezuidenhout: I'm not sure I understand what you mean. When creating a rule a binary payload cannot be specified so the character string will just contain printable characters.

mbezuidenhout:
> > This device supports MQTT. Details are located at the bottom of the wiki.
> 
> ah.... using the Waveshare Cloud or with a local MQTT broker too?

Yes, on this device you can specify your own local MQTT server.

tasmota/tasmota_support/support.ino
```diff
@@ -590,6 +590,33 @@ char* Trim(char* p) {
...
+  size_t str_len = strlen(*source+2);
+  char *s = *source+2;
```
s-hadinger, line 595: I'm not very comfortable with just skipping blindly the first 2 characters. Could you at least check if they are "0x" or "0X" and skip them in such case, or keep them if it's just HEX data without '0x'?

tasmota/tasmota_support/support.ino
s-hadinger, line 593: If it's binary content, I would prefer `uint8_t **` instead of `char **`. The problem is that signness of `char` is architecture-dependant. So fixing to non-signed avoids some errors.

tasmota/tasmota_support/support.ino
```diff
@@ -590,6 +590,33 @@ char* Trim(char* p) {
...
+  char *new_str = (char *)malloc(str_len/2);
+  for (int i = 0; i < str_len; i++) {
+        unsigned char val;
+        char cur = s[i];
+        if (cur >= 97) {
+            val = cur - 97 + 10;
```
s-hadinger, line 601: `val` could have a much larger value than expected. Can you add a guarding condition, and decide what value you put for illegal charachters? `0` I guess?

tasmota/tasmota_support/support.ino
```diff
@@ -590,6 +590,33 @@ char* Trim(char* p) {
   return p;
 }
 
+size_t HexStringToBinaryString(char** source) {
+  size_t str_len = strlen(*source+2);
+  char *s = *source+2;
+  char *new_str = (char *)malloc(str_len/2);
+  for (int i = 0; i < str_len; i++) {
+        unsigned char val;
```
s-hadinger, line 598: Same, prefer `uint8_t`

tasmota/tasmota_support/support.ino
```diff
@@ -590,6 +590,33 @@ char* Trim(char* p) {
...
+        if (cur >= 97) {
+            val = cur - 97 + 10;
+        } else if (cur >= 65) {
+            val = cur - 65 + 10;
+        } else {
+            val = cur - 48;
+        }
+        /* even characters are the first half, odd characters the second half
+         * of the current output byte */
+        if (i%2 == 0) {
```
s-hadinger, line 609: Maybe a bit picky, but I prefer `i & 1` which is just a bit mask, instead of a full-blown division. (Maybe the compiler replaces it already)

tasmota/tasmota_support/support.ino
```diff
@@ -590,6 +590,33 @@ char* Trim(char* p) {
...
+            new_str[i/2] = val << 4;
+        } else {
+            new_str[i/2] |= val;
+        }
+    }
+    memcpy(*source, new_str, str_len/2);
```
s-hadinger, line 615: There is a big asumption here, that `source` is large enough to handle the data. This can be a major security problem due to buffer overflow. Maybe add a `size_t len` parameter so the caller specifies the buffer size.

Second remark, why allocate a buffer, then copy to `source`. Why not store directly in `source`? This looks like an unnecessary double allocation. Am I missing something?

s-hadinger (changes requested):
Could you please change `HexStringToBinaryString`, there are some security concernes about managing the size of the target buffer, and a possible unnecessary double allocation.

tasmota/tasmota_support/support.ino
mbezuidenhout: Changed the format to no longer require 0x prefix.

tasmota/tasmota_support/support.ino
mbezuidenhout: According to the C and C++ standards, a byte has at least enough bits to store a character from the basic execution character set, which typically means at least 8 bits. Hence, in the context of standard conformant C and C++ compilers, char is always 1 byte.

tasmota/tasmota_support/support.ino
mbezuidenhout: Added check to ensure that all values in string contain 0-9, a-z or A-Z

tasmota/tasmota_support/support.ino
mbezuidenhout: Change now no longer does an unnecessary `malloc`.

tasmota/tasmota_support/support.ino
s-hadinger: Oh, I didn't realize that you are changing the values in place. I'm not a big fan of a function that scraps the source string and converts to binary in place. Please have an input string (`const char *`) and an output buffer (`uint8_t *`) and a length of output buffer (`size_t`)

s-hadinger: Btw, if the caller wants to rewrite the input string, then it's their explicit choice when calling the function.

mbezuidenhout:
Upon reviewing the support.ino file, I discovered a function: `int32_t HexToBytes(const char* hex, uint8_t* out, size_t* outLen)`, designed for converting hexadecimal to binary. However, this function exhibits some issues: firstly, outLen is initialised to zero and not subsequently updated. Secondly, there are no validations to confirm whether the input string is exclusively comprised of hexadecimal characters.

What course of action should I take in this situation?
1. Request a review of the function to solicit feedback or improvements from other team members.
2. Modify the function and include these changes in my existing pull request, which already contains the Publish3 command.
3. Make the necessary modifications to the function and initiate a separate pull request for these changes. Once this pull request is merged, proceed with another pull request for the Publish3 command.

tasmota/tasmota_support/support.ino
mbezuidenhout: Upon reviewing the support.ino file, I discovered a function: `int32_t HexToBytes(const char* hex, uint8_t* out, size_t* outLen)`, designed for converting hexadecimal to binary. However, this function exhibits some issues: firstly, outLen is initialised to zero and not subsequently updated. Secondly, there are no validations to confirm whether the input string is exclusively comprised of hexadecimal characters.

What course of action should I take in this situation?
1. Request a review of the function to solicit feedback or improvements from other team members.
2. Modify the function and include these changes in my existing pull request, which already contains the Publish3 command.
3. Make the necessary modifications to the function and initiate a separate pull request for these changes. Once this pull request is merged, proceed with another pull request for the Publish3 command.

s-hadinger: Thanks for your patience. You can proceed with 2. and include the changes in this PR

PR #21236, opened 2024-04-21
Allow changing WiFi Hostname at compile time from user_config_override.h

ascillato:
## Description:

This small PR adds the option to allow changing the WiFi Hostname at compile time from **user_config_override.h**.

_Usage:_
- Add in **user_config_override.h** file the following so the hostname can be customized at compile time:

```cpp
#undef DEFAULT_WIFI_HOSTNAME
#define DEFAULT_WIFI_HOSTNAME  "MyDevice"
```

_Notes:_
- The default value is `"%s-%04d"` which expands to `<MQTT_TOPIC>-<last 4 decimal chars of MAC address>`
- This PR does not change the FLASH or RAM usage.

## Checklist:

### DIFF
--- a/tasmota/tasmota.ino
+++ b/tasmota/tasmota.ino
@@ -23,8 +23,8 @@
 #endif  // ESP32_STAGE
 #include "include/tasmota_compat.h"
 #include "include/tasmota_version.h"        // Tasmota version information
-#include "include/tasmota.h"                // Enumeration used in my_user_config.h
 #include "my_user_config.h"                 // Fixed user configurable options
+#include "include/tasmota.h"                // Enumeration used in my_user_config.h
 #ifdef USE_TLS
 #include <t_bearssl.h>                      // We need to include before "tasmota_globals.h" to take precedence over the BearSSL version in Arduino
 #endif  // USE_TLS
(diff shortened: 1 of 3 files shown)


tasmota/tasmota.ino
```diff
@@ -23,8 +23,8 @@
 #endif  // ESP32_STAGE
 #include "include/tasmota_compat.h"
 #include "include/tasmota_version.h"        // Tasmota version information
-#include "include/tasmota.h"                // Enumeration used in my_user_config.h
 #include "my_user_config.h"                 // Fixed user configurable options
+#include "include/tasmota.h"                // Enumeration used in my_user_config.h
```
s-hadinger, line 27: I'm concerned that changing the include order has much more impact and may break many stuff. Can you please preserve the include order?

ascillato: Hi, the new order of the includes is needed for this change to work. I already tested and there is no issue with that since **my_user_config.h** is just a lot of macro definitions which don't change anything on **tasmota.h** (except for the default value of WIFI HOSTNAME that this PR changes.

Jason2866:
LGTM

sfromis:
You can have hostname included in the build without a new `#define`, by using the regular `Hostname` command in `USER_BACKLOG`. Commands given there will be executed first boot with new settings.

barbudor:
There's a lot of consts for which I would have loved to do that in the past but I was so afraid of changing that order... 🥶
Thank you for your bravery @ascillato 👍

ascillato:
@sfromis Hi, you are right with that. In fact, I was using that in my devices at home, but there is the issue with the user backlog approach (for this case in particular) that takes time and require a restart making other issues like the WiFi AP name (which uses the Hostname) not taking the change right away showing the previous naming. With this simple change, which adds no extra code to the Firmware, we can allow the user to set the HOSTNAME (or a custom template of the hostname like "%s-%06d") at compile time simplifing the process.

Jason2866:
Let us merge and see what happens ;-)

arendst:
Implemented differently but same result.

```
#define WIFI_DEFAULT_HOSTNAME  "%s-%04d"         // [Hostname] Expands to <MQTT_TOPIC>-<last 4 decimal chars of MAC address>
```

ascillato:
Awesome. Thanks!

PR #21272, opened 2024-04-25
Berry `web_add_handler` called before `Webserver` is initialized

s-hadinger:
## Description:

Bug introduced with `BrRestart`. There is a mechanism to detect if `web_add_handler` was missed because of a BrRestart. It checked if network was up, but failed to check if `Webserver` was also created.

In some cases, the event would be fired too soon, before `Webserver` is created, hence lost.

**Related issue (if applicable):** fixes #21158

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
 - LVGL restore `lv_palette` functions
 - IPv6 support in safeboot
 - LVGL fix memory allocation of flush buffers
+- Berry `web_add_handler` called before `Webserver` is initialized
 
 ### Removed
 - LVGL disabled vector graphics
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_9_berry.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_9_berry.ino
@@ -899,7 +899,7 @@ bool Xdrv52(uint32_t function)
 #ifdef USE_ETHERNET
           network_up = network_up || EthernetHasIP();
 #endif
-          if (network_up) {       // if network is already up, send a synthetic event to trigger web handlers
+          if (network_up && (Webserver != NULL)) {       // if network is already up, send a synthetic event to trigger web handlers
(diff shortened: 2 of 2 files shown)


storchalberto:
I have used benzino77 (developer) to build my version. The version i see is 13.4.1.2. The web_add_handler seems to function well but now the when the browser interrogate the page, it stay waiting the end of page transfer. It seems to have problem the webserver.content_flush(): the browser receive the page but do not terminate the connection. Also, during this, the main page do not respond. If i force the browser to close my page, the main page return to function. This not happens using the standart 13.4.0 version.

s-hadinger:
You should call `webserver.content_stop()` when the page is finished.

Unless you have specific needs, `webserver.content_flush()` is generally not needed

storchalberto:
The webserver.content_stop() after the webserver.content_send of the page add at the end of the page a new <body> tags, with your copirights and so the page is not correct. in the version 14.4.0 the webserver.content_flush() do not do it, end the page  and function correctly

s-hadinger:
Good point. Do you have a test case?

Maybe we should add `webserver.content_close()` to force the end of the page

storchalberto:
the webserver.content_close() do the same. the connection to the browser is not ended

storchalberto:
Here the code for test:
import webserver
class Web_page_demo : Driver
  def page_say_test1()
	var fr=open("test1.htm", 'r')
	var mybuff=fr.read()
	fr.close()
	webserver.content_send(mybuff)
	webserver.content_flush()
  end
  def init()
	print("init")
	tasmota.add_driver(self)
  end
  def web_add_handler()
	webserver.on("/test1.htm", self.page_say_test1)
	print("page added")
  end 
end
web_page_demo_instance = Web_page_demo()
Here the test page (test1.htm):
<html>
<body>
Test page
</body>
</html>
If you replace the webserver.content_flush() with the webserver.content_stop() the page in the browser is completed, vbut not correct due for the addition of your copyright at the end. in the 13.4.0 version it runs correctly

s-hadinger:
I could reproduce, thanks. I suspect it is due to new Arduino framework Core3 that has a slower reuse of non-closed TCP connections.

I'm adding `webserver.content_close()` to solve this

PR #21299, opened 2024-04-28
Update esp32 platform to 2024.04.14

Jason2866:
## Description:

Updated HWCDC code in Arduino. Should fix boot stuck issues with ESP32-S3

@fyi @s-hadinger 

## Checklist:

### DIFF
--- a/tasmota/tasmota.ino
+++ b/tasmota/tasmota.ino
@@ -493,7 +493,9 @@ void setup(void) {
     TasConsole.println();
     AddLog(LOG_LEVEL_INFO, PSTR("CMD: Using USB CDC"));
   } else {
+    #if SOC_USB_SERIAL_JTAG_SUPPORTED  // Not S2
     HWCDCSerial.~HWCDC();       // not needed, deinit CDC
+    #endif  // SOC_USB_SERIAL_JTAG_SUPPORTED
     // Init command serial console preparing for AddLog use
     Serial.begin(TasmotaGlobal.baudrate);
     Serial.println();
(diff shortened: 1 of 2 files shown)


tasmota/tasmota.ino
```diff
@@ -474,10 +474,10 @@ void setup(void) {
 
   bool is_connected_to_USB = false;
 #if SOC_USB_SERIAL_JTAG_SUPPORTED  // Not S2
-  for (uint32_t i = 0; i < 5; i++) {  // wait up to 250 ms - maybe a shorter time is enough
-      is_connected_to_USB = usb_serial_jtag_is_connected();
+  for (uint32_t i = 0; i < 5; i++) {  // wait up to 50 ms - maybe a shorter time is enough
+      is_connected_to_USB = HWCDC::isPlugged();
```
Staars, line 478: This change resolves to:
  inline static bool isPlugged(void) {
    return usb_serial_jtag_is_connected();
  }

So what is the purpose?

tasmota/tasmota.ino
```diff
@@ -474,10 +474,10 @@ void setup(void) {
...
       if (is_connected_to_USB) { break; }
-      delay(50);
+      delay(10);
```
Staars, line 480: Are you sure, that the timeout is now magically shorter? I would expect undefined behavior now.

Jason2866:
Does not fix at all. Still issues with S3.

Staars:
The most recent code changes in Arduino's HWCDC class require some refactoring in Tasmotas USB connection check, if we want to use Arduino's API.

New way is:
1. Always start USB CDC console.
2. Check for plugged USB cable using the Arduino API
3. a) if USB connection is detected proceed with already initialised HWCDC console
    b) if NO connection is detected - deinit HWCDC and fall back to serial.

PR #21329, opened 2024-05-03
Fixed HexToBytes function. Added Publish3 command to send binary data via MQTT.

mbezuidenhout:
## Description:

Add ability to the Publish command to send a hexstring as binary MQTT data.

Publish command now has the following options

Publish3 | <topic> <payload> = MQTT publish any topic and optional binary payload. For example the payload 7461736D6F7461 will output tasmota or the value ff will output the integer value 255

## Checklist:


Jason2866:
Some clear words, for the niche function which was recently added which needs resources in every build, now there is another code increase with this PR.

s-hadinger:
We could consider excluding this feature in SAFEBOOT as it shouldn't be useful there

Jason2866:
> We could consider excluding this feature in SAFEBOOT as it shouldn't be useful there

Fine when doing this way.

PR #21351, opened 2024-05-05
let berry drivers provide "after_teleperiod" handlers

dgwynne:
This is useful if you want to periodically report the state of something handled separately to the sensors output.

## Description:

I'm working on Yet Another(tm) driver for micradar devices, the R60ASM1 in particular, which seems to work in a similar fashion to a TuyaMCU device in that it has a concept of independent datapoints that report at different times.  Because there are so many different datapoints, the driver relays their updates as separate mqtt topics, (eg, tele/%topic%/R60ASM1/BreathingRate or tele/%topic%/R60ASM1/MonitorSwitch) rather than shove several dozen bits of info into the SENSORS or STATE payloads. Some datapoints are updated very infrequently, so it's handy to report them every teleperiod so stupid downstream systems (eg, HA) don't forget what's going on.

C drivers can hook `FUNC_AFTER_TELEPERIOD` for similar reasons. This extends it to Berry drivers too.

## Checklist:

### DIFF
--- a/lib/libesp32/berry_tasmota/src/embedded/driver_class.be
+++ b/lib/libesp32/berry_tasmota/src/embedded/driver_class.be
@@ -14,6 +14,7 @@ class Driver
   var save_before_restart
   var web_sensor
   var json_append
+  var after_teleperiod
   var button_pressed
   var display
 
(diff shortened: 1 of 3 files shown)


dgwynne:
https://github.com/dgwynne/berry is where the berry driver is atm.

s-hadinger:
Awesome, thanks

PR #21352, opened 2024-05-05
Use letsencrypt  ISRG_Root_X1 certificate instead of R3

gladk:
According to the blog post [1] new intermediate certificates will be deployed on 6th of June, 2024.

[1] https://letsencrypt.org/2024/04/12/changes-to-issuance-chains

Thus existing R3 will stop to work for newly issued certificates. This PR replaces R3 by ISRG root one, which should work till 2035.

This change adds about **288** bytes to the final build.

## Description:

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.0.0


s-hadinger:
Why do we need 5 certificates. This is a big size increase

s-hadinger:
I wonder if we can use 'ISRG Root X1' instead which is RSA 4096

gladk:
r3 can be removed about 6th of September 2024. Ass all certificates, issued by letsencrypt have a max expiration length 3 months.

according to [this](https://letsencrypt.org/2024/04/12/changes-to-issuance-chains) picture, only r10 and r11 will be used, but it is not clear, whether we really need r>11.

gladk:
> I wonder if we can use 'ISRG Root X1' instead which is RSA 4096

good catch. Let me check.

gladk:
@s-hadinger thank you for the hint! It really works. I am updating the merge request.

s-hadinger:
Hold on, I had second thoughts and I'm surprised it's working at all since we don't have the full cert chaon

gladk:
The main problem is here that the devices flashed with R3 will probably stop working soon,  because certificate check will fail.

s-hadinger:
I understand. My initial thought was to validate againg `ISRG Root X1`, but I had second thoughts and I understand that a server certificate should only contain the signature from the intermediate CA (R3 or R10-R15), but does not contain the entire signature chain of to the root CA.

Did you actually try replacing CA with `ISRG Root X1`? Does it work?

gladk:
Yes! I checked with ISRG Root X1 on esp8266 and it works! I even changed one random byte to make a negative test, and it failed to connect to the server. So, it works! 

What I did not check is the tasmota32. But I hope it should work. Also, this changes gives +about 290 bytes in comparison to initial build.

s-hadinger:
Awesome, well done. ESP32 uses exactly the same code so I don't expect any difference.

PR #21497, opened 2024-05-25
SML: FastExit for binary SML parsing

rPraml:
## Description:

This improves performance for the SML parser for type 's' (SML binary smart message coding)
We assume, that an SML message always starts with 0x77 / SML_SYNC and start decoding only, if the buffer starts with the sync byte.
If there is a rare case, that an SML message does not start with 0x77, this improvement can be disabled with flag 32

## Details:
I noticed a high load (~170) and slow FPS on the connected display (240x240 ST7789)

With this PR I tried to improve the SML parsing for binary SML data, which seems to have some room for improvement.

As far as I understand the binary parsing, there is a buffer, where each byte is added at the end of the buffer and then the buffer is shifted one byte to the left. 
After each byte, SML_Decode is called, which parses the the descriptor each time and checks if one line matches with the buffer. 
(@gemu2015 is this correct?)

In my case I have 3 Logarex LK13BE606739 meters, where each transmits ~660 bytes each second.
My descriptor has ~ 60 lines, which means that SML_Decode is called ~2000 times/sec. and it will parse and check each line in the descriptor.

As far as I see, all SML messages in binary mode will start with 0x77, so I thought, it could be a good idea to implement a fast exit, and start decoding only, when the buffer starts with the sync byte, which makes the decoding a lot of faster.
(0x77 occurs 23 times in my test sample of 660 bytes)

To test the improvement, I bound this to flag 32, but I want to discuss, if this could be the default. (is 0x77 really a sync byte?)

This change reduces the load from 170 to 100 and gives some extra frames on the TFT

Roland
 

## Checklist:


gemu2015:
that seems to be a good improvement.
i would suggest to make this the default und invert flag & 32 as i also assume that 0x77 is always present.

rPraml:
I've modified the PR a bit. I still have to test this and provide a PR for documentation update...

tasmota/tasmota_xsns_sensor/xsns_53_sml.ino
```diff
@@ -1480,6 +1480,11 @@ void sml_shift_in(uint32_t meters, uint32_t shard) {
     case 's':
       // binary obis = sml
       mp->sbuff[mp->sbsiz - 1] = iob;
+      if (mp->sbuff[0] != SML_SYNC && ((mp->flag & 32) == 0)) {
```
eku, line 1483: Wouldn't the code be easier to understand if a constant with a descriptive name were used for the numerical value 32? And since a bit is obviously being tested here, the use of 0x20 is a good idea.

eku (changes requested):

rPraml:
This PR is tested on an ESP32-S3 and works IMHO

tasmota/tasmota_xsns_sensor/xsns_53_sml.ino
```diff
@@ -1768,6 +1768,15 @@ void SML_Decode(uint8_t index) {
     return;
   }
 
+
+  struct METER_DESC *md = &meter_desc[index];
+  // start of serial source buffer
+  cp = md->sbuff;
+  if (md->flag & 32 && md->type == 's' && *cp != SML_SYNC) {
```
eku: Wouldn't the code be easier to understand if a constant with a descriptive name were used for the numerical value 32? And since a bit is obviously being tested here, the use of 0x20 is a good idea.

eku (changes requested):

tasmota/tasmota_xsns_sensor/xsns_53_sml.ino
rPraml: Is this the same comment as https://github.com/arendst/Tasmota/pull/21497#pullrequestreview-2082718853 ?
This is already fixed.

PR #21502, opened 2024-05-26
Support Levoit Core200s (WIP, Feedback requested)

blenk92:
## Description:
Hi there, I bought one of these devices https://levoit.com/collections/air-purifiers/products/core-200s-smart-true-hepa-air-purifier and figured that is uses a esp32 (the solo1 variant) to talk to another MCU via a serial protocol. I've reversed the protocol (well at least the relevant parts I would say)  to control the air purifier using Tasmota. 

This PR contains my initial implementation, to proof that its generally possible to that (so its still kind of in a POC state in some places). However, it should already give a rough idea on how the protocol works... However,  before I spend additional efforts in making it better, I would like to request some indication if  the project would generally be interested in support these devices. I say 'these' because Levoit offers a bunch of similar devices, I only own a Core200S but I could imagine that other models are similar, maybe even using an extended version of the protocol (and there are still some fields in the status messages that seem to be unused on my device).

Since this is quite a large patch already, I would also like to ask what else I would need provide in order to get the levoit devices (ore at least the core200S) officially supported.

Screenshot to show you how the web interface currently looks like:
Screenshot_2024-05-26_23-59-41

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP32 V.3.0.0

### DIFF
--- a/tasmota/berry/drivers/Core200S.be
+++ b/tasmota/berry/drivers/Core200S.be
@@ -1,23 +1,31 @@
 import string
 import webserver
 
+####
+# Feature state classes
+# In the section below there Classes that represent state of specific features of the air purifier (e.g.
+# if the fan is on or off)
+####
+
+# Generic parent class that implements the shared functionality of all state classes
 class StateEnum
     var value_map
     var value
 
+    # Init a state
+    # param value_map map of valid states in readable and binary form (e.g. {"on":1, "off":0}
+    # param value value that this object reprents
     def init(value_map, value)
         self.value_map = value_map
         if type(value) == 'int'
-            for k : self.value_map.keys()
-                if self.value_map[k] == value
+            for v : self.value_map
+                if v == value
                     self.value = value
                     break
                 end
-            end
+             end
         elif type(value) == 'string'
-            if self.value_map.contains(value)
-                self.value = self.value_map[value]
-            end
+            self.value = value_map.find(value)
         end
 
         if self.value == nil
@@ -39,11 +47,13 @@ class StateEnum
     end
 
     def ==(other)
-        return self.value == other.value
+        return self.value == other.value && type(self) == type(other)
     end
 
 end
 
+# Generic parent class that implementes On-Off only features (so features that only
+# know these two states
 class OnOffState : StateEnum
     static ON = OnOffState("on")
     static OFF = OnOffState("off")
@@ -52,24 +62,28 @@ class OnOffState : StateEnum
     end
 end
 
+# Class that represent if the fan is running or nor
 class FanMode : OnOffState
     def init(value)
         super(self).init(value)
     end
 end
 
+# Class that represents if the child lock is enabled or not
 class ChildLockMode : OnOffState
     def init(value)
         super(self).init(value)
     end
 end
 
+# Class that represents if the sleep mode is enabled or not
 class SleepMode : OnOffState
     def init(value)
         super(self).init(value)
     end
 end
 
+# Class that represent if the the display auto off mode is enabled or not
 class DisplayAutoOffMode : StateEnum
     static OFF = DisplayAutoOffMode("off")
     static ON = DisplayAutoOffMode("on")
@@ -78,12 +92,17 @@ class DisplayAutoOffMode : StateEnum
     end
 end
 
+# Class that represents if the display is currently turned on or not
+# (Note this just represents the current state and can't only be controlled via
+#  sleep mode or display auto off mode)
 class DisplayMode : StateEnum
     def init(value)
         super(self).init({"off": 0, "on": 100}, value)
     end
 end
 
+# Class that represents the current speed level of the fan
+# (Note even if fan is turned off, the last speed state is stored)
 class FanSpeedMode : StateEnum
     static LOW = FanSpeedMode("low")
     static MED = FanSpeedMode("med")
@@ -94,6 +113,7 @@ class FanSpeedMode : StateEnum
     end
 end
 
+# Class that reprensts the state of the night light
 class NightLightMode : StateEnum
     static OFF = NightLightMode("off")
     static LOW = NightLightMode("low")
@@ -103,28 +123,38 @@ class NightLightMode : StateEnum
     end
 end
 
-def modulo(num, mod)
-    return ((num % mod) + mod) % mod
-end
+
+
+####
+# Utility classes
+####
 
 def info_c2s(msg)
     log("C2S: " + msg, 2)
 end
 
 def log_c2s(msg)
-    log("C2S: " + msg)
+    log("C2S: " + msg, 3)
 end
 
+# Class that implements messaging functionality that is need and not provided by the
+# "normal" serial class, such as
+# * read byte by byte
+# * read ahead (reading without actually consuming the content)
+# and some device specifc functionality such a
+# * checksum calculation and verification of each message
+# * tracking of message count
+# * reading of single messages instead of whole serial buffer
 class Core200SSerial : serial
     var msg_count
     var read_buffer
     var error_count
     var expected_replies
 
-    def init()
-        super(self).init(16, 17, 115200)
+    def init(RX, TX)
+        super(self).init(RX, TX, 115200)
         self.reset_counters()
-        self.read_buffer = bytes("")
+        self.read_buffer = bytes()
     end
 
     def deinit()
@@ -137,19 +167,24 @@ class Core200SSerial : serial
         while i < msg.size()
             if (i != 5)
                 checksum -= msg.get(i)
-                checksum = modulo(checksum, 256)
+                if checksum < 0
+                    checksum += 256
+                end
             end
             i += 1
         end
         return checksum
     end
 
+    # send message and wait for reply
+    # message counter is set and message checksum are calculated automatically
     def send(msg)
         self.send_nowait(msg)
-        tasmota.delay(100)
+        tasmota.delay(50)
         return self.read_msg()
     end
 
+    # send message and don't wait for reply here. Instead the reply needs to be handled asynchronously
     def send_nowait(msg)
         if msg.size() >= 2 && msg[1] == 0x22
             self.expected_replies += 1
@@ -169,15 +204,17 @@ class Core200SSerial : serial
             i += 1
         end
 
-        log_c2s("Send " + msg.tohex())
+        log_c2s(f"Send {msg.tohex()}")
     end
 
+    # Return number of available bytes
+    # (is specified because the available() method from serial would return incorrect values)
     def available()
         self.read_buffer += self.read()
-        return self.read_buffer.siez()
+        return self.read_buffer.size()
     end
 
-    # read but don't remove from buffer
+    # read num bytes but don't actually remove them from the buffer
     def read_ahead(num)
         var content = self.read()
         self.read_buffer += content
@@ -189,7 +226,7 @@ class Core200SSerial : serial
         return nil
     end
 
-    # read an remove from buffer
+    # read num bytes and remove them from buffer
     def consume(num)
         var content = self.read_ahead(num)
         if content != nil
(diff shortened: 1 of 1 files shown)


blenk92:
> Hi @blenk92 As this device is based on ESP32, have you considered a Berry driver instead of a native one ? I think it could make more sense

Hi @barbudor,

took me a bit to work myself into berry, but I think you have a point here. I re-implemented the current state in berry (the original change with the native driver is still available at https://github.com/blenk92/Tasmota/tree/core200s_native for reference)

Please let me know how we can proceed to get this merged ;-)

barbudor:
Great @blenk92 

Merge is not under my control.
May be it could be worth adding along a `Core2005.md` doc along to explain how to use the driver, any configuration etc...

arendst:
@s-hadinger is this a good implementation of a compiled-in berry driver? I had hoped this functionality could be added by users during runtime....

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
+import string
+import webserver
+
+class StateEnum
+    var value_map
+    var value
+
+    def init(value_map, value)
```
s-hadinger, line 8: Please add comments to give a hint about parameters, and what the class is suppose to do

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+        self.value_map = value_map
+        if type(value) == 'int'
+            for k : self.value_map.keys()
+                if self.value_map[k] == value
+                    self.value = value
+                    break
+                end
+            end
+        elif type(value) == 'string'
+            if self.value_map.contains(value)
```
s-hadinger, line 18: you can simply do:
`self.value = value_map.find(value)`

If the key does not exist, it returns `nil`

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+import webserver
+
+class StateEnum
+    var value_map
+    var value
+
+    def init(value_map, value)
+        self.value_map = value_map
+        if type(value) == 'int'
+            for k : self.value_map.keys()
```
s-hadinger, line 11: `for v: value_map` allows to iterate already on map values. So you can simplify code

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+class NightLightMode : StateEnum
+    static OFF = NightLightMode("off")
+    static LOW = NightLightMode("low")
+    static FULL = NightLightMode("full")
+    def init(value)
+        super(self).init({"off": 0, "low": 50, "full": 100}, value)
+    end
+end
+
+def modulo(num, mod)
```
s-hadinger, line 106: please avoid generic names, this pollutes the global namespace. Or consider putting all this in a module

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+    return ((num % mod) + mod) % mod
```
s-hadinger, line 107: What is this supposed to do?

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+end
+
+def info_c2s(msg)
+    log("C2S: " + msg, 2)
+end
+
+def log_c2s(msg)
```
s-hadinger, line 114: This does the same as `info_c2s`

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+end
+
+class Core200SSerial : serial
+    var msg_count
+    var read_buffer
+    var error_count
+    var expected_replies
+
+    def init()
+        super(self).init(16, 17, 115200)
```
s-hadinger, line 125: Please avoid hard-coding GPIO numbers. It should at least be optional parameters to the constructor

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+        self.reset_counters()
+        self.read_buffer = bytes("")
```
s-hadinger, line 127: You can just do `bytes()` instead of `bytes("")`

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+        self.close()
+    end
+
+    static def calc_checksum(msg)
+        var checksum = 0xff
+        var i = 0
+        while i < msg.size()
+            if (i != 5)
+                checksum -= msg.get(i)
+                checksum = modulo(checksum, 256)
```
s-hadinger, line 140: for perfomance, it would be better to add the modulo code here, rather than an additional function call

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+            end
+            i += 1
+        end
+        return checksum
+    end
+
+    def send(msg)
+        self.send_nowait(msg)
+        tasmota.delay(100)
```
s-hadinger, line 149: This is not the right way to do it. `tasmota.delay(100)` stops the entire Tasmota for 100ms which is much higher than the standard loop.

Prefer `tasmota.set_timer(100, <your_function>)` for asynchronous callback

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+        msg[5] = self.calc_checksum(msg)
+
+        # send message
+        var i = 0
+        while i < msg.size()
+            self.write(msg[i])
+            i += 1
+        end
+
+        log_c2s("Send " + msg.tohex())
```
s-hadinger, line 172: You can consider also:
`log(f"Send {msg.tohex()}")`

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+    end
+
+    def available()
+        self.read_buffer += self.read()
+        return self.read_buffer.siez()
```
s-hadinger, line 177: typo: `size` instead of `siez`

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+            else
+                max_read_attempts -= 1
+                if max_read_attempts <= 0
+                    if !allow_no_msg
+                        info_c2s("Too many read attempts (couldn't read message header)")
+                        self.error_count += 1
+                    end
+                    return bytes("")
+                end
+                tasmota.delay(100)
```
s-hadinger, line 219: Remove delay

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+        self.timer_remaining = 0
+        self.timer_total = 0
+        self.init_cmds()
+    end
+
+    def init_cmds()
+        ser.flush()
+        ser.reset_counters()
+
+        tasmota.delay(1000)
```
s-hadinger, line 303: Use async function or state machine. You can't stop Tasmota for 1 full second.

tasmota/berry/drivers/Core200S.be
```diff
@@ -0,0 +1,712 @@
...
+            end
+            msg = ser.read_msg(true)
+        end
+    end
+
+    def every_second()
+        self.second_counter += 1
+        if self.second_counter == 60
+            self.second_counter = 0
+            self.set_wifi_led(tasmota.wifi() != {})
```
s-hadinger, line 514: `tasmota.wifi()` is never equal to `{}`. What are you trying to do ? If you want to know if wifi is connected, use `tasmota.wifi('up')`

s-hadinger (changes requested):
Thanks, there are some changes needed, primarily remove the long `delay()` (above 50ms) and do async with `tasmota.set_timer()` instead.

You can also consider packaging all these in a single module, and avoid polluting the global namespace.

tasmota/berry/drivers/Core200S.be
s-hadinger: Oh I understand now. Simply do:
`if (checksum < 0) checksum += 256 end`

It's way faster and simpler

arendst:
OK. I leave it to @s-hadinger then 👍

tasmota/berry/drivers/Core200S.be
blenk92: ah, then i guess I misread the documentation (https://tasmota.github.io/docs/Berry/#tasmota-object) . It says 

>  Retrieves Wi-Fi connection info or empty map.

But i guess this means then that it only returns empty map if wifi is not supported (or maybe disabled) ? Maybe the documentation could be a bit more specific here...

tasmota/berry/drivers/Core200S.be
blenk92: function removed as by your other suggestion

tasmota/berry/drivers/Core200S.be
blenk92: function removed as by your other suggestion

tasmota/berry/drivers/Core200S.be
blenk92: ah yeah, actually i wanted to set a different log level here :-D

tasmota/berry/drivers/Core200S.be
blenk92: done

tasmota/berry/drivers/Core200S.be
blenk92: done

tasmota/berry/drivers/Core200S.be
blenk92: done

tasmota/berry/drivers/Core200S.be
blenk92: done

tasmota/berry/drivers/Core200S.be
blenk92: okay added as parameters to the constructor (i hope this is what you meant)

tasmota/berry/drivers/Core200S.be
blenk92: hmm I'm not entirely sure how this is supposed to work. `send()` will eventually be called by all of the commands. If we switch to asynchronous handling here, how am I supposed to report if the command was successful or not?

tasmota/berry/drivers/Core200S.be
blenk92: hmm, not sure if async behavior is easy to implement here (because the messages kind of need to be guaranteed to be send in this order, or at least first ones). If've seen weird behavior if the order of messages changes. So maybe some simple state machine needs to do the trick.

tasmota/berry/drivers/Core200S.be
blenk92: we are in the middle of reading a message here. Removing this delay can lead to message not being available at the point in time we are in the loop. We can maybe reduce a bit, but right now there is synchronous communication happening, which means missing a message here can lead to incorrect results of commands. Or am I missing something here?

blenk92:
> Thanks, there are some changes needed, primarily remove the long `delay()` (above 50ms) and do async with `tasmota.set_timer()` instead.
> 
> You can also consider packaging all these in a single module, and avoid polluting the global namespace.

Thanks for you review. I tried to adapt according to the comments that you left. There is maybe still something open in regards of the async behavior. Maybe you want to have another look ?

Just FYI: I won't have access to my Core200S for ~ 1.5 weeks. So no hurry with the re-review, because I can't test potentially requested changes on my device until then...

tasmota/berry/drivers/Core200S.be
s-hadinger: Good point. "empty map" is outdated, the object was changed later. I will update the documentation

tasmota/berry/drivers/Core200S.be
s-hadinger: Generally speaking, you first acknowledge the command, and then you post a message once the message is accepted.

tasmota/berry/drivers/Core200S.be
s-hadinger: Let's stick to this, but generally speaking try to avoid this and have some sort of minimalistic state machine instead. `tasmota.set_timer()` is an easy way to launch code in the future, very similar to javascript

tasmota/berry/drivers/Core200S.be
s-hadinger: See above

PR #21528, opened 2024-05-29
xdrv: tcp_bridge: Increase baudrate

shantur:
On ESP32 TCP Serial bridge can support baudrate of more than 115200.

After this change the baudrate is still limited to 306000 due to uint_8 is used for settings->tcp_baudrate.

A 1.5Mbauds have been tested by changing uint_16 for settings->tcp_baudrate on ESP32

## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_41_tcp_bridge.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_41_tcp_bridge.ino
@@ -212,7 +212,7 @@ void CmndTCPStart(void) {
 }
 
 void CmndTCPBaudrate(void) {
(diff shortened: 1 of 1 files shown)


s-hadinger:
I'm afraid it's not that simple:

```
uint8_t       tcp_baudrate;              // F41
```

This means that the max allowed value for now is `306000`. We can think of changing slightly the encoding above `96` (which represents `115200`) to allow for higher rates. But changing the size of `tcp_baudrate` simply does not work, because is would change the alignment of all other settings.

Please come back with a better solution to encode differently above `96` and keep the settings in 8 bits.

arendst:
I'll refactor the tcp_baudrate from uint8_t to uint16_t as I did long time ago for the other baudrates.

Hold your horses.

arendst:
Implemented in v14.0.0.4

PR #21535, opened 2024-05-31
SML: Allow larger offset when using x to ignore bytes

dops-at:
## Description:
STW Klagenfurt sends 355 bytes, and the interesting values starts somewhere at position 304. Therefor we need to set a larger uint range to allow more than 255 bytes to ignore.

Discussion open -> https://github.com/arendst/Tasmota/discussions/21359

## Checklist:


s-hadinger:
If there is no reason to force 16 bits, you can use `uint32_t` instead. It makes the code smaller

PR #21544, opened 2024-06-02
Bugfix: crash when using tjpegd and LVGL

Staars:
## Description:

Tasmotas Arduino framework does include tjpegd in ROM and it clashes with the externally added library.
The external library would be compiled in with LVGL activated in the build environment and lead to repeatable crashes when accessing functions of tjpegd in other code parts.
This will also shrink the firmware binary in these cases.

## Checklist:

### DIFF
--- a/lib/libesp32_lvgl/lvgl/src/libs/tjpgd/tjpgd.c
+++ b/lib/libesp32_lvgl/lvgl/src/libs/tjpgd/tjpgd.c
@@ -24,6 +24,7 @@
 /                     Some performance improvement.
 /----------------------------------------------------------------------------*/
 
+#ifndef TASMOTA // has tjpegd in ROM
 #include "tjpgd.h"
 
 
@@ -1135,3 +1136,5 @@ JRESULT jd_decomp(
 
     return rc;
 }
+#endif // TASMOTA
+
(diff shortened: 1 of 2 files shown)


s-hadinger:
Thanks. We need to keep track of this change otherwise it will be overwritten by the next LVGL update. Do you think of any other option in librariy.json to achieve the same result without changing source code ?

Staars:
Maybe …
I am not sure, if there is guaranteed flag in the Arduino framework.

ESP_ROM_HAS_JPEG_DECODE Could be a candidate.

if this works, the best way would be to use this instead and upstream the change to LVGL.

s-hadinger:
I was thinking about adding in `library.json` of LVGL:

```json
[...]
	"build": {
		"includeDir": ".",
		"flags": [ "-I$PROJECT_DIR/tasmota/lvgl_berry" ],
                 "srcFilter": ["-<libs/tjpgd/**>"]
	},
[...]
```

I haven't tested at all, but that was the spirit of removing a directory from the build chain.

Jason2866:
Yep, suggested solution should work fine. https://docs.platformio.org/en/latest/projectconf/sections/env/options/build/build_src_filter.html#projectconf-build-src-filter

PR #21565, opened 2024-06-04
Berry `input()` returns empty string and does not crash

s-hadinger:
## Description:

Berry `input()` crashed. This function has no use in Tasmota, it now returns an empty string instead of crashing.

## Checklist:


sfromis:
Now, people will think that it is a bug that it returns an empty string :grin:

s-hadinger:
Yes, it's always better than a crash. But it's difficult to return an exception without changing the core of Berry

PR #21568, opened 2024-06-05
Fixed incorrect output size clamping

mbezuidenhout:
## Description:

Fixed incorrect output size clamping. Code was reading past the end of `const char* hex`

`out_len` is the maximum size of the output and thus the value of `bytes_out` cannot be larger than `out_len`.

## Checklist:


s-hadinger:
@arendst This looks ok to me.

PR #21605, opened 2024-06-10
MQTT is not reporting the correct LastChange

stefanbode:
I assume this is just missing. I got some side effects on shutters where on MQTT commands the source is still the LAST one before the MQTT

## Description:

## Checklist:


tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
stefanbode, line 1: Maybe the command must be some lines up. It's on you.

arendst:
I see your point.

I'll merge this PR as you seem to be the only user of ``TasmotaGlobal.last_source``.

The issue is that ONLY the shutter code makes extensive use of calling command functions by their normal functions. The command code was never planned to be used that way. Instead you should have called the command's functionality, so not using the ``XdrvMailbox`` struct. Your approach also provides tons of MQTT messages, which I think, are needless.

PR #21657, opened 2024-06-19
Berry: fix file.readbytes() for large files

Staars:
## Description:

Reading binary files into bytes in Berry leads to data corruption for large files (I did not figure out the threshold) by creating a bytes buffer of the correct size, that has correct data at the front but only holds zeros at the tail end. This fails silently and does not throw any error.

This seems to be related to the resize operation after the actual data read from the file system.

To reproduce we need a test file on the ESP, that is not too small. I used this one: 
https://github.com/Staars/MockUp/blob/main/watermeter.jpg

Then load in the Berry web console:
```
f = open("watermeter.jpg","r")
b =f.readbytes()
f.close()
```

Check the terminating bytes, which should be "FFD9" for a JPG file with:
```
b[-2..]
```

The current version shows "0000". Tested on ESP32 and ESP32-S3.

Unrelated to this PR the C6 crashes with heap corruption:
`CORRUPT HEAP: Bad head at 0x408407e8. Expected 0xabba1234 got 0x52844590`

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.0.1

### DIFF
--- a/lib/libesp32/berry/src/be_filelib.c
+++ b/lib/libesp32/berry/src/be_filelib.c
@@ -79,15 +79,11 @@ static int i_readbytes(bvm *vm)
             be_call(vm, 2); /* call b.resize(size) */
             be_pop(vm, 3);  /* bytes() instance is at top */
 
-            char *buffer = (char*) be_tobytes(vm, -1, NULL); /* we get the address of the internam buffer of size 'size' */
+            char *buffer = (char*) be_tobytes(vm, -1, NULL); /* we get the address of the internal buffer of size 'size' */
             size = be_fread(fh, buffer, size);
 
-            /* resize if something went wrong */
-            be_getmember(vm, -1, "resize");
-            be_pushvalue(vm, -2);
-            be_pushint(vm, size);
-            be_call(vm, 2); /* call b.resize(size) */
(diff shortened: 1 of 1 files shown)


sfromis:
Looks like some much wider issue with non-small files. I tried to transfer a moderately sized file (`ftp.be`) with both Manage File system upload (web or via curl), or the UrlFetch command, on a board with PSRAM available, and this failed. Closest to a symptom was "file too large". I did manage to transfer a somewhat smaller `.be` file, 11 KB instead of failing with 17 KB.

s-hadinger:
I'm afraid there is a deeper bug in the bytes.resize function. Let's put on hold until I see more clearly what is happening

s-hadinger:
Closing this one, the actual fix is https://github.com/arendst/Tasmota/pull/21716

There was a bug in `bytes.resize()`

Staars:
Nice!
(But I still do not really see the value of the second resize procedure.)

s-hadinger:
The last resize is only here in case the number of bytes read is different from what was expected. It should not happen with Tasmota, but you never know. There are many reason why a read could return less bytes than the size of the file.

Jason2866:
@s-hadinger The crash with the C6 is still there

PR #21717, opened 2024-07-02
add support for 4.2 inch epaper display from WeAct Studio (partial up…

timehlers:
02.07.2024

## Description:

Add support for newer epaper display WeAct Studio 4.2" (partial update).

Hi,

I tried to figure out the display.ini requirements for the newer epaper display from AliExpress (WeAct Studio), which does not need any lookup tables anymore for refresh. I did a lot of trial and error but finally got a display.ini working.

Unfortunately this needs some patches also to the uDisplay.cpp code, because:

1. We don't have a ":l", ":L" or ":L\<number\>" section (I set a new ep_mode=3, in case we don't have these sections, but ":f" or ":p").
2. The memory area needs a different setup (uDisplay::SetMemoryArea). For this display we need to swap y-start and y-end, means instead of for example "45,4,00,00,2b,01" we need "45,4,2b,01,00,00".
3. By default the display will omit the first pixel line and place it at the bottom, so we need to reduce the y-position in "uDisplay::SetMemoryPointer" by 1 to fix this.

## Checklist:


Jason2866:
Are the changes fully compatible to the current implementation?

timehlers:
Also, nobody until now should have an epaper section ":f" or ":p", while not having any of these: ":l", ":L" or ":L<number>" sections, because until now no epaper display is supported without lookup tables for refresh -- so nobody should accidentally run now into ep_mode=3 in my opinion.

arendst:
> Also, nobody until now should have an epaper section ":f" or ":p", 

While looking at WS_epaper29_v1_display.ini I see both :f and :p sections so forcing mode 3 as you do doesn't seem the appropriate solution.

timehlers:
I made it a bit more clear now and better programmed, hopefully in the second commit.

timehlers:
Btw, it is this display:

https://a.aliexpress.com/_EwkB0VN

:)

PR #21723, opened 2024-07-03
make Tasmota compatible for Arduino espressif32 based on IDF 5.2.2

Jason2866:
## Description:

with the changes Tasmota can use the Arduino espressif32 framework based on Tasmota IDF 5.2.2

The alternative framework can be enabled in `platformio override.ini` by uncommenting the entry `;platform = https://github.com/Jason2866/platform-espressif32/releases/download/2024.07.20/platform-espressif32.zip`

The framework uses the newer GCC 13.2 toolchain. The benefit is smaller flash usage.
The S3 safeboot (as example) reduces 7kbyte in size.

Thx @Staars for fixing the last issues left ;-)

@arendst @s-hadinger and everyone who likes testing new stuff (greetings to @TD-er ;-))

## Checklist:

### DIFF
--- a/lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
+++ b/lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
@@ -152,7 +152,7 @@ void TasmotaSerial::setTransmitEnablePin(int tx_enable_pin) {
 bool TasmotaSerial::freeUart(void) {
   for (uint32_t i = SOC_UART_NUM -1; i >= 0; i--) {
     if (0 == bitRead(tasmota_serial_uart_bitmap, i)) {
-      m_uart = i;
+      m_uart = uart_port_t(i);
       bitSet(tasmota_serial_uart_bitmap, m_uart);
       return true;
     }
--- a/lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
+++ b/lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
@@ -183,5 +183,19 @@ typedef int SerConfu8;
 
 #define STATION_IF 0
 
+
+// ESP32 needed define to support IDF 5.2 based Arduino builds
+
+// Name has changed
+#ifdef LCD_CAM_LCD_UPDATE_M
+#define LCD_CAM_LCD_UPDATE_REG LCD_CAM_LCD_UPDATE_M
+#endif // LCD_CAM_LCD_UPDATE_M
+
+// IDF 5.2 has changed counting UART channels, SOC_UART_NUM includes now LP UARTS too
+#ifdef SOC_UART_HP_NUM
+#undef SOC_UART_NUM
+#define SOC_UART_NUM SOC_UART_HP_NUM
+#endif
+
 #endif // ESP32
 #endif // __ESP8266TOESP32_H__
--- a/lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.h
+++ b/lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.h
@@ -109,7 +109,7 @@ class TasmotaSerial : public Stream {
     uint32_t m_speed;
     uint32_t m_config;
     HardwareSerial *TSerial;
-    int m_uart = 0;
+    uart_port_t m_uart = uart_port_t(0);
 #endif
 
 };
--- a/lib/lib_div/ArduinoNTPd/NTPServer.cpp
+++ b/lib/lib_div/ArduinoNTPd/NTPServer.cpp
@@ -23,7 +23,7 @@ bool NtpServer::beginListening()
   return false;
 }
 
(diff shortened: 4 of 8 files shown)


lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
```diff
@@ -152,7 +152,7 @@ void TasmotaSerial::setTransmitEnablePin(int tx_enable_pin) {
 bool TasmotaSerial::freeUart(void) {
   for (uint32_t i = SOC_UART_NUM -1; i >= 0; i--) {
     if (0 == bitRead(tasmota_serial_uart_bitmap, i)) {
-      m_uart = i;
+      m_uart = uart_port_t(i);
```
arendst, line 155: Oh my. Not sure if this should be changed like shown. Will need investigation as the m_uart is just supposed to be a bitmap for Tasmota internally.... I'll dive in.

lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
```diff
@@ -183,5 +183,19 @@ typedef int SerConfu8;
...
+// ESP32 needed define to support IDF 5.2 based Arduino builds
+
+// Name has changed
+#ifdef LCD_CAM_LCD_UPDATE_M
+#define LCD_CAM_LCD_UPDATE_REG LCD_CAM_LCD_UPDATE_M
+#endif // LCD_CAM_LCD_UPDATE_M
+
+// IDF 5.2 has changed counting UART channels, SOC_UART_NUM includes now LP UARTS too
+#ifdef SOC_UART_HP_NUM
+#undef SOC_UART_NUM
```
arendst, line 196: Another candidate for deeper investigation. SOC_UART_NUM is used by TasmotaSerial to find number of useable uarts. Tasmota cannot use LP uarts with tasmota code (I think)....

lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
Staars: No, in the code we set  the bitmap `tasmota_serial_uart_bitmap` with the value of `m_uart`. So the latter is not a bitmap and will be used in other parts directly to call `uart_set_rx_full_threshold(m_uart, <value>);`. 

`uart_port_t `was a typedef'd integer before and will become an enum with IDF 5.2.
`uart_set_rx_full_threshold` only will accept `uart_port_t` type after 5.2.

lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
Staars: Agree, this could fail in edge cases.

Jason2866: It behaves like before. With IDF the naming changed. Before IDF 5.2 there was only SOC_UART_NUM (no reference to LP UART). With IDF 5.2 SOC_UART_NUM are ALL UART (including LP UART) So it is redefined to the numbers of "normal" UARTs.

lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
arendst: Thx. Correct

lib/libesp32/ESP32-to-ESP8266-compat/src/esp8266toEsp32.h
arendst: The main problem I see here is that SOC_UART_NUM is used in both TasmotaSerial and xsns_53_sml.ino. Changing it in esp8266toesp32.h is no guarantee that it is changed before te previous mentionend files are compiled.

Why not abandon SOC_UART_NUM and use SOC_UART_HP_NUM instead?

arendst: I'll change it now.

arendst: Just did a test by using current platform and
```
#define SOC_UART_HP_NUM 2
#ifdef SOC_UART_HP_NUM
#undef SOC_UART_NUM
#define SOC_UART_NUM SOC_UART_HP_NUM
#endif
```
Doesn't work. SOC_UART_NUM at the time TasmotaSerial.cpp is accessing it is still 3 on esp32. 

I'll find a better soultion....

Staars: SOC_UART_HP_NUM Is the new choise from Espressif. The defines are basically a backpot from 5.2.
One option would be a separate backport header file, that must be included for dependent drivers and could be removed in the future.
But there are many other ways to solve it for sure.

PR #21743, opened 2024-07-07
Berry: add cam module, img class

Staars:
## Description:

The main intention of this PR is to free up as much heap memory as possible for machine learning applications while using a connected camera module. I measured about 20 kB heap use reduction. This is not meant as replacement for any of the other webcam drivers. Uses JPEGDEC library with SIMD support on the ESP32-S3.

Compile with:
```
-DUSE_BERRY_CAM ; needs -DUSE_BERRY_IMAGE too
-DUSE_BERRY_IMAGE ; could be used without -DUSE_BERRY_CAM
```

- this will strip down the main webcam driver to the setup functions only, but will not influence  the normal webcam firmware in any way
- add `webcam` module to Berry with simple picture acquisition 
- add `img` class to Berry as a thin container for image data in order to read images from the camera or from a file, while providing conversions and extraction of regions of interest

Tested with a self trained model to read 7 segment digits of a display (which could be an "unsmart" device of any kind) with still about 80 kB free heap on a cheap ESP32 cam board with 4 MB PS-RAM.

More examples:
- Show camera data on a connected display as a video (tested on ESP32-S3-Eye)
- Show images from an external camera (another ESP32 in LAN with the normal webcam driver) on a connected display
- Use image data in LVGL as image source

I would update Tasmotas docs with the content of my preliminary documentation: https://staars.github.io/docs/Berry_import/#img-class

## Checklist:


Staars:
C family builds fail because of missing dsp library in our framework.

@Jason2866 Can we add this back for these chips? It will not increase firmware size if not used.

Both additions are 99,9% pointless for PSRAM-less boards anyway.

Jason2866:
@Staars Good timing. Planning a Platform release for tomorrow. The CI should be green with the updated Platform build.

Staars:
A new planned framework update is planned anyway, which should solve the compile issue for the C family of chips.
I would suggest to keep this PR open and we'll see, if the build failures will disappear.

PR #21755, opened 2024-07-09
Bugfix: multiple SHT4x sensors at the same bus, support for SHT40-CD1B

Grumium:
## Description:

- Before this fix, multiple SHT4x sensors on the same bus were not detected because the 'type' variable was set in the outer for-loop. The bug was introduced in https://github.com/arendst/Tasmota/commit/7295bdb5490c723bae082455210df0643e72f7c5. Once a sensor was found, the 'type' variable was reset to 0, preventing the detection of any additional sensors with types greater than 0, such as SHTC3 and SHT4X. Since the SHTC3 has only a single address, this issue only affected SHT4X sensors.

BEFORE:
https://github.com/arendst/Tasmota/blob/5df51a9365c72e0fd0d50b096ec0e2b28f2c23fd/tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino#L122-L132

AFTER:
https://github.com/arendst/Tasmota/blob/3cf3f997250ef7fadcedcc1d6a554533c03fe78a/tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino#L123-L136

- Sensors of type SHT40-CD1B with address 0x46 were also not detected before this fix. Refer to the first page of the [datasheet](https://sensirion.com/media/documents/1D662E57/661CD1C1/HT_DS_Datasheet_SHT4xI-Digital.pdf) for more details.
- Additionally, the bus number was not displayed, making it impossible to distinguish sensors on different buses. To make it less intrusive, the bus number will only be displayed if necessary.

BEFORE:
image
Please note: Only one SHT4x sensor is detected (0x44) although two are present at the same bus

AFTER:
image

## Checklist:

I don't have an ESP8266 at hand. Could somebody like @arendst @stibus please check the functionality?

### DIFF
--- a/I2CDEVICES.md
+++ b/I2CDEVICES.md
@@ -32,7 +32,7 @@ Index | Define              | Driver   | Device   | Address(es) | Bus2 | Descrip
   13  | USE_ADS1115         | xsns_12  | ADS1115  | 0x48 - 0x4B | Yes  | 4-channel 16-bit A/D converter
   14  | USE_INA219          | xsns_13  | INA219   | 0x40 - 0x41, 0x44 - 0x45 |      | Low voltage current sensor
   15  | USE_SHT3X           | xsns_14  | SHT3X    | 0x44 - 0x45 | Yes  | Temperature and Humidity sensor
-  15  | USE_SHT3X           | xsns_14  | SHT4X    | 0x44 - 0x45 | Yes  | Temperature and Humidity sensor
+  15  | USE_SHT3X           | xsns_14  | SHT4X    | 0x44 - 0x46 | Yes  | Temperature and Humidity sensor
   15  | USE_SHT3X           | xsns_14  | SHTCX    | 0x70        | Yes  | Temperature and Humidity sensor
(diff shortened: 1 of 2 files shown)


arendst:
Thx.

Your sensor naming is OK but ... I like to suggest you have a look at the the xsns_09_bmp.ino driver with regards to the sensor names. In case of an ESP8266, which only has one bus, the support for the second bus adds extra code size. Pls consider to add #define ESP32 for the second bus naming like the BMP driver does. This saves precious code space for the SHT3X driver default enabled in ``tasmota-sensors.bin``

Grumium:
Good call! I tried to make it consistent. However, if 'BMP280-77-1`is the template, the `xsns_09_bmp.ino` needs an update as well, since this is not what it currently does:
https://github.com/arendst/Tasmota/blob/5df51a9365c72e0fd0d50b096ec0e2b28f2c23fd/tasmota/tasmota_xsns_sensor/xsns_09_bmp.ino#L566-L567

arendst:
No. That's not what it does! It's an incremental string addition.

Line 560 provides "BMP280-77".
Line 567 provides the above string added with "-1"

Grumium:
I see. Lesson learned. This should do the trick.

PR #21805, opened 2024-07-20
fix pushcolors for displaytext

gemu2015:
## Description:

fixes push colors for display text with RGB displays

## Checklist:

### DIFF
--- a/lib/lib_display/UDisplay/uDisplay.cpp
+++ b/lib/lib_display/UDisplay/uDisplay.cpp
@@ -2213,7 +2213,7 @@ void uDisplay::pushColors(uint16_t *data, uint16_t len, boolean not_swapped) {
 
   //Serial.printf("push %x - %d - %d - %d\n", (uint32_t)data, len, not_swapped, lvgl_param.data);
 
-  // Isolating _UDPS_RGB to increase code sharing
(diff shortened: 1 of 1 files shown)


lib/lib_display/UDisplay/uDisplay.cpp
```diff
@@ -2229,42 +2229,50 @@ void uDisplay::pushColors(uint16_t *data, uint16_t len, boolean not_swapped) {
     // check that bytes count matches the size of area, and remove from inner loop
     if ((seta_yp2 - seta_yp1) * (seta_xp2 - seta_xp2) > len) { return; }
```
s-hadinger, line 2231: I would be more comfortable with `uint32_t` if we ever have bigger displays. Actually forcing `uint16_t` most certainly makes the code bigger

s-hadinger: Never mind, it's 16 bits all over the place

gemu2015: this indeed means that we may not push more than 65k pixels and was inherited from very old drivers
the :B parameter in the display descriptor thus has a limit.
we may change this to 32 bit.

PR #21816, opened 2024-07-22
Update SHT3X driver to bring it into line with other sensors, e.g., HYT..., and avoid unnecessary I²C reads

Grumium:
## Description:
Until now, each sensor was queried at undefined intervals (~800 ms in my case) and additionally within the scope of TELEPERIOD. I think it makes sense to align this with other drivers, such as with the HYT (xsns_97_hyt.ino) and avoid unnecessary reads. Below, I am showcasing the effect with just three I²C devices on the same bus.

BEFORE:
three_dev_before

AFTER:
three_dev_new

## Checklist:
  - [] The code change is tested and works with Tasmota core ESP8266 V.2.7.7

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino
@@ -1,7 +1,7 @@
 /*
   xsns_14_sht3x.ino - SHT3X, SHT4X and SHTCX temperature and humidity sensor support for Tasmota
 
-  Copyright (C) 2022  Theo Arends, Stefan Tibus
+  Copyright (C) 2024  Theo Arends, Stefan Tibus, Jan-David Förster
 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
--- a/tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino
@@ -1,7 +1,7 @@
 /*
   xsns_14_sht3x.ino - SHT3X, SHT4X and SHTCX temperature and humidity sensor support for Tasmota
 
-  Copyright (C) 2024  Theo Arends, Stefan Tibus, Jan-David Förster
+  Copyright (C) 2022  Theo Arends, Stefan Tibus
 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License as published by
@@ -41,6 +41,9 @@ uint8_t sht3x_addresses[] = { 0x44, 0x45, 0x46, 0x70 };
 
 uint8_t sht3x_count = 0;
 struct SHT3XSTRUCT {
+  float   humi = NAN;
+  float   temp = NAN;
+  uint8_t valid = 0;
   uint8_t type;        // Sensor type
   uint8_t address;     // I2C bus address
   uint8_t bus;
@@ -64,9 +67,8 @@ uint8_t Sht3xComputeCrc(uint8_t data[], uint8_t len) {
   return crc;
 }
 
-bool Sht3xRead(uint32_t sensor, float &t, float &h) {
-  t = NAN;
-  h = NAN;
+bool Sht3xRead(uint32_t sensor) {
+  if (sht3x_sensors[sensor].valid) { sht3x_sensors[sensor].valid--; }
 
   TwoWire& myWire = I2cGetWire(sht3x_sensors[sensor].bus);
   if (&myWire == nullptr) { return false; }   // No valid I2c bus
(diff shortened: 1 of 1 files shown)


tasmota/tasmota_xsns_sensor/xsns_14_sht3x.ino
```diff
@@ -104,29 +106,32 @@ bool Sht3xRead(uint32_t sensor, float &t, float &h) {
   if ((Sht3xComputeCrc(&data[0], 2) != data[2]) || (Sht3xComputeCrc(&data[3], 2) != data[5])) {
     return false;
   }
-  t = ((float)(((data[0] << 8) | data[1]) * 175) / 65535.0) - 45.0;
+  float t;
+  float h;
+  t = ((((data[0] << 8) | data[1]) * 175) / 65535.0) - 45.0;
```
s-hadinger, line 111: You can save 48 bytes by changing all `double` literals to `float` literals. I.e. `65536.0f` instead of `65536.0`...

When you don't add the `f` suffix, all calcucations are done in `double` precision instead of `float` single precision, which generally brings unwanted libraries and takes more time and space. For temp/humi, single precision is more than enough

eku: Is there anything to be said against integer arithmetic by using tenths or hundredths as the basis everywhere?

s-hadinger: That would be even better. For ex:

```C
uint16_t changeUIntScale(uint16_t inum, uint16_t ifrom_min, uint16_t ifrom_max,
                                       uint16_t ito_min, uint16_t ito_max)
```

is a easy way to convert unsigned 16 bits from range a..b to c..d.

So if I'm not mistaken, it should be something like:

```C
t = changeUIntScale((data[0] << 8) | data[1], 0, 65536, -45, 130);
```

s-hadinger: Hmmm, the example above would lose precision. The alternative would be to have `int16_t t_10` as the temperature in 1/10th of Celsius, and:

```C
t = changeIntScale((data[0] << 8) | data[1], 0, 65535, -450, 1300);
```

1/100th of Celsius would overflow the `int16_t` size.

Grumium: Good call. While the resolution is 1/100th, the accuracy of these sensors is at best 0.1 °C (SHT45) and typically 0.2 °C (SHT40). Do we sacrifice resolution for code size? What are your thoughts?

s-hadinger: I generally tend to prefer integers, so yes in 1/10th resolution should be enough.

The way I do in Zigbee to print the value is:
```C
              char buf[12];
              dtostrf(t_100 / 100.0f, 2, 1, buf);
```

But `float` is fine as well

eku: > 1/100th of Celsius would overflow the `int16_t` size.

One decimal place should be sufficient.

Grumium: Great. Who creates a new PR? I can do it tomorrow the earliest.

Grumium: @s-hadinger @eku 
I have another question regarding this. I just played around with different data types and I noticed 
that int16_t allocates more memory than int32_t since a conversion happens.
I think this is briefly touched in https://tasmota.github.io/docs/Sensor-API/#uint8_t-or-uint32_t

Wouldn't it be recommended to change, apart from `t_10`->int32_t and `h_10`->int32_t, e.g., `sht3x_count` and `sht3x_addressed`, and `data` to **u**int32_t as well and only preserve the `uint8_t`s in the struct?

And what is actually the best practice regarding explicity defining signed/unsigned integers over a simple `int`?

joba-1: Some of my best? practices:
* Type int is always signed. Prefer unsigned if you don’t need negative values.
* Don't use int or unsigned if you need more than 16bit (for portability). Use int32_t and friends for specific size requirements.
* On tasmota prefer smallest data over smaller flash over most performant. RAM is the most constrained resource.

sfromis: While 1 decimal may be "sufficient" on a temperature, I'd not support removing precision from what the sensor provides. While 2 decimals will be more than the accuracy you can expect from a sensor, the 2nd decimal is pretty good for trends, meaning that a bot of temperatures does not get so jagged, and you can also spot small trends if you're after fine-grained control. Also even if the sensor may only have a accuracy of 0.5.

Grumium: Thank you @joba-1 and @sfromis for the clarification.
With reference to the example here, we can probably conclude that there is no better way of calculating the temperature without sacrificing precision and this is as good as it gets:
`float t = ((((data0 << 8) | data1) * 175) / 65535.0f) - 45;` 

In terms of relative humidity, I think there is some potential without sacrificing any physically meaningful precision:

`int32_t h_100 = ((((data[3] << 8) | data[4]) * 12500) >> 16) - 600;` and
`int32_t h_100 = ((((data[3] << 8) | data[4]) * 10000) >> 16);`
saves 32 bytes over 
`float h = (((data[3] << 8) | data[4]) * 100) / 65535.0f;` and
`float h  = ((((data[3] << 8) | data[4]) * 125) / 65535.0f) - 6;`

Agreed?

arendst: Yes

PR #21841, opened 2024-07-25
Neopool relay state detection and reporting fixes

tarvip:
## Description:

This fixes reporting some Neopool relay states to MQTT.
Relay state calculation was wrong, at least in my case heating relay state was not properly detected without this fix.
Also, fixed keys used in json for reporting values, there was redundant space (I understand for some people it can be breaking change, e.g some dashboards can stop working, if they are already using key that contains space).

previously (without this PR) it was like this:
```
"Relay":{"State":[0,1,0,0,0,0,1],"Aux":[0,0,0,1],"Acid":0,"Redox":0,"Chlorine":0,"Heating ":0}
```

now it is:
```
"Relay":{"State":[0,1,0,0,0,0,1],"Aux":[0,0,0,1],"Acid":0,"Redox":0,"Chlorine":0,"Heating":1}
```

(in my case AUX4 relay is used for heating)

## Checklist:


curzon01:
well done, json key were definitely wrong, thx - leave it that way

PR #21897, opened 2024-08-02
improved read access for SDM630

f-reiling:
read access improved by reading several registers in one cycle -> higher update rate of meter values

## Description:

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.0.4

### DIFF
--- a/tasmota/tasmota_xnrg_energy/xnrg_10_sdm630.ino
+++ b/tasmota/tasmota_xnrg_energy/xnrg_10_sdm630.ino
@@ -39,6 +39,7 @@
 #include <TasmotaModbus.h>
 TasmotaModbus *Sdm630Modbus;
 
+#ifdef SDM630_HIGH_UPDATE_RATE
 struct sSdm630RequestConfig{
   uint16_t startAddress;
   uint8_t  registerToRead; // according to spec: max 80 register can be read a once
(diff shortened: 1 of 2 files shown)


arendst:
And the impact on compiled file size is?

tasmota/tasmota_xnrg_energy/xnrg_10_sdm630.ino
```diff
@@ -95,117 +92,72 @@ void SDM630Every250ms(void)
...
 
-        case 22:
-//          Energy->import_active[0] = value;
           EnergyUpdateTotal();
           break;
       }
 
       Sdm630.read_state++;
-      if (sizeof(sdm630_start_addresses)/2 == Sdm630.read_state) {
+      if ( (sizeof(sdm630ReqConf)/sizeof(sdm630ReqConf[0])) == Sdm630.read_state) {
```
eku, line 152: Just wondering why we do not habe a macro `NELEM` for this. Saw this on many POSIX systems.

arendst: Well I have a macro called ``nitems`` which does just that ;-)

f-reiling:
> And the impact on compiled file size is?

compared to current implementation +1.6k

eku:
> read access improved by reading several registers in one cycle -> higher update rate of meter values

What real-life problem is to be solved here?

f-reiling:
> > read access improved by reading several registers in one cycle -> higher update rate of meter values
> 
> What real-life problem is to be solved here?

It depends on the application. 
* for some use-cases it might be sufficient to have an update rate of approx. 5s per value.
* In case of a fast power controller (as i am currently working on), it is better to have a higher update rate
* as not all values are read in same cycle, the data is partly inconsistent (e.g. U*I != P) -> new implementation ensures consistent data

eku:
> > And the impact on compiled file size is?
> 
> compared to current implementation +1.6k

You might want to define unions containing both the floats and the byte array for several values togetherand iterate over it. Should save you the repeated calls to `convBufToFloat`.

f-reiling:
this will not work on esp32 platform as modbus transmits MSB first. therefore the order of bytes needs to be changed

eku:
> this will not work on esp32 platform as modbus transmits MSB first. therefore the order of bytes needs to be changed

My suggestion concerned the list of elements and not their value. I realise that the byte order has to be changed.

f-reiling:
i added a define to select between both implementations, default is the old one.
i rechecked impact on code/RAM. it seems that the addional code usage is approx. 0k1

arendst:
Thx.

PR #22030, opened 2024-08-27
Added define to adjust calibration parametrs for bl09xx

PavelSheremetev:
## Description:

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.0.4

### DIFF
--- a/tasmota/tasmota_xnrg_energy/xnrg_14_bl09xx.ino
+++ b/tasmota/tasmota_xnrg_energy/xnrg_14_bl09xx.ino
@@ -44,26 +44,48 @@
 
 //#define DEBUG_BL09XX
 
+#ifndef BL0939_PREF
 #define BL0939_PREF                 713       // =(4046*1*0,51*1000)/(1,218*1,218*(390*5+0,51)) = 713,105
+#endif
+#ifndef BL0939_UREF
 #define BL0939_UREF                 17159     // =(79931*0,51*1000)/(1,218*(390*5+0,51)) = 17158,92
+#endif
+#ifndef BL0939_IREF
 #define BL0939_IREF                 266013    // =(324004*1)/1,218 = 266013,14
+#endif
 
+#ifndef BL0940_PREF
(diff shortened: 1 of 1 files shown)


Jason2866:
Why?

PavelSheremetev:
> Why?

I design devices that use the BL0940 chip. When monitoring consumption with a current transformer or current shunt, different calibration values need to be used in the xnrg_14_bl09xx.ino file. This pull request allows entering calibration values in the user_config_override.h file, which is much more convenient as it eliminates the need to fork the Tasmota repository.

arendst:
A result of changing the default calibration offsets is that no pre-compiled Tasmota OTA binary can be used.

Can't you just use the current offsets and then configure the correct VoltageSet, CurrentSet and PowerSet? What's the issue here other than one time conveniently no calibration needed but scewing up any future updates?

PavelSheremetev:
> A result of changing the default calibration offsets is that no pre-compiled Tasmota OTA binary can be used.

I don't understand so deeply the nuances of Tasmoto firmware, but similarly the possibility to specify calibration parameters in [ade7880](https://github.com/arendst/Tasmota/blob/d8f01ed4c02b4edd84d554b4ef5f50f26e12569e/tasmota/tasmota_xnrg_energy/xnrg_23_ade7880.ino#L47-L88) module is already involved.
Also BL09XX_WRITE_COMMAND and BL09XX_READ_COMMAND differ from the operating values specified in BL0940 [datasheet](https://www.belling.com.cn/media/file_object/bel_product/BL0940/datasheet/BL0940_V1.1_en.pdf). 
I just didn't want to create a repository fork for my devices just because of this.
Thanks.

barbudor:
@arendst I don't see that PR breaking anything and changing any default values used in pre-built Tasmota
It only allows default values to be overriden in 'user_config_override.h' for a custom build which makes sense for someone manufacturing his own device in reliable way that makes manual calibration not needed.

Issue #21200, opened 2024-04-16
Range Extender driver `xdrv_58_range_extender` does not compile with core 3.0.0


Jason2866:
Details of failing can be seen in this GH Actions run https://github.com/Jason2866/Tasmota-build/actions/runs/8704050149/job/23871521878

@joba-1 @sillyfrog can you have a look and adopt to IDF5.1 / Arduino core 3.0.0?

joba-1:
At first glance:  tcpip_adapter API discontinued and replaced by ESP-NETIF :(
See https://docs.espressif.com/projects/esp-idf/en/release-v5.1/esp32/migration-guides/release-5.x/5.0/networking.html?highlight=tcp#tcp-ip-adapter

No idea how much changed. Maybe it is more or less search&replace ... and testing.

Could you please make it easy for me and send a platformio.ini or (a link to) instructions to compile locally with this new stuff? What I have working is VS Code on Linux with PlatformIO Extension and a cloned tasmota. I guess I just have to replace some platform or framework entries?

Jason2866:
Do a fresh clone of Tasmota in a different folder. Don't use your existing setup it may conflicts since the platformio setup has changed. You can use this env
```
[env:tasmota32-rangeextender]
extends                     = env:tasmota32_base
build_flags                 = ${env:tasmota32_base.build_flags}
                              -DFIRMWARE_TASMOTA32
                              -DUSE_WIFI_RANGE_EXTENDER
                              -DUSE_WIFI_RANGE_EXTENDER_NAPT
lib_ignore                  =
                              TTGO TWatch Library
                              Micro-RTSP
```

sillyfrog:
I noticed this has been marked stale. I have this on my radar, but it'll be a little bit before I'll have time to look at it properly... Let me know if you get very far in the meantime. Thanks!

joba-1:
sure, I‘ll let you know of any progress (or problems, in case you have an idea). Earliest next week, nothing new so far.

sfromis:
Unstale, builds still failing

sfromis:
Any hope of getting this feature refactored to match the API change?

sfromis:
*unstale*

sfromis:
Tickle

fdaubercy:
Hi.
I always have to run my rangeextender with a version older than the current one. Compiling with this option still doesn't work.
Frédéric

Jason2866:
@fdaubercy Thats why this issue is open ;-)

[...]

PR #22198, opened 2024-09-24
ported range_extender to ESP-IDF v5

pharaocurt:
## Description:

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.0.5

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_58_range_extender.ino
@@ -94,8 +94,10 @@ const char kDrvRgxCommands[] PROGMEM = "Rgx|" // Prefix
 #ifdef USE_WIFI_RANGE_EXTENDER_NAPT
                                        "|"
                                        "NAPT"
+#ifdef USE_WIFI_RANGE_EXTENDER_PORTADD
                                        "|"
                                        "Port"
+#endif
 #endif // USE_WIFI_RANGE_EXTENDER_NAPT
                                        "|"
                                        "Clients"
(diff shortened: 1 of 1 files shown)


pharaocurt:
ported xdrv_58_range_extender to work with current ESP-IDF v5.
i had do add new #define USE_WIFI_RANGE_EXTENDER_PORTADD, which must not be defined for correct linking, else an undefined reference error will occur: ip_portmap_add. This function is defined in lwip/lwip_napt.h, but only compiled with defined option IP_NAPT_PORTMAP on generation of arduino-esp32 library. Therefore it is currently not possible to add a portmap to enable access of a port from main network on the range extended network without added route to extended network at main gateway.
If arduino-esp32 library will be enhanced by IP_NAPT_PORTMAP, my changes with USE_WIFI_RANGE_EXTENDER_PORTADD can be removed, but this feature could not be verified currently.

pharaocurt:
ported xdrv_58_range_extender.ino to ESP-IDF v5. Tested to work correctly on ESP32 with and without NAPT, but prevoius full features could not be reached.
I had to add definition #define USE_WIFI_RANGE_EXTENDER_PORTADD, which must not be defined for final correct linking, else:
undefined reference to `ip_portmap_add'. This function is defined in "lwip/lwip_napt.h", but only compiled and available with option IP_NAPT_PORTMAP, which seems beeing undefined in tasmotas arduino-esp32 library. With enhanced arduino-esp32 library, the #ifdef USE_WIFI_RANGE_EXTENDER_PORTADD and #endif can be removed to include full previous functionallity, but this is currently untested.

sfromis:
Were you trying to fix the pending issue about the Range Extender failing compile?
https://github.com/arendst/Tasmota/issues/21200

pharaocurt:
Yes, i think the Range Extender failing compile can be solved (with the limitation i mentioned above) by these few changes in my try range_extender branch. But i fail in creation of a correct pull request.

Issue #22282, opened 2024-10-13
shutter1#position triggers every second


kugelkopf123:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._
I upgraded my Shelly2.5 from v9.x to the latest dev-branch today. Reason: Had sometimes issues with HW Watchdog and wrong shutterposition after that. The update was done via serial not OTA. 
Since then shutter1#position in rule3 triggers every second. 
Because I thought it was due to the big version jump, I deleted the shellys via serial and reflashed them.

There are other shutters running in the house (v9.x) with the same rule and there are no problems there. 

Rule3: `on SHUTTER1#Position do var1 %value% endon; on shutter#moved do backlog publish stat/%topic%/RESULT {‘ShutterPosition1’:%var1%} endon`

Even if you shorten it to:
`Rule3 on Shutter1#position do var1 %value% endon`

the rule is triggered every second.

By the way, it doesn't matter whether rule3 or rule1 or 2.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [ ] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [ ] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Device used (e.g., Sonoff Basic): Shelly2.5
- [ ] Tasmota binary firmware version number used: 14.2.0.6 (tasmota)
  - [ ] Pre-compiled
  - [x] Self-compiled
- [x] Flashing tools used: VisualStudioCode
- [x] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:

14:11:54.634 CMD: Backlog Template; Module; GPIO 255
14:11:54.656 RSL: RESULT = {"NAME":"Shelly 2.5BT","GPIO":[320,0,34,0,224,129,0,0,640,128,608,225,3456,4736],"FLAG":0,"BASE":18}
[...]

PR #22353, opened 2024-10-24
Optimized behavior to publish shutter data with sensor request

stefanbode:
## Description:
Before the shutterposition and other data where published every second as sensor data. Building rules on this data was difficult. With the change the data to shutter data is now only published every second, if the shutter is moving

#22282

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241023

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_27_esp32_shutter.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_27_esp32_shutter.ino
@@ -195,6 +195,7 @@ struct SHUTTERGLOBAL {
   bool     callibration_run = false;         // if true a callibration is running and additional measures are captured
   uint8_t  stopp_armed = 0;                  // Count each step power usage is below limit of 1 Watt
   uint16_t cycle_time = 0;                   // used for shuttersetup to get accurate timing
+  bool     sensor_data_reported = false;     // ensure that shutter sensor data reported every sedond is only reported if shutter is moving and there is a change.
 } ShutterGlobal;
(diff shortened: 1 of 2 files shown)


jziolkowski:
@stefanbode can you please elaborate more about the reasoning behind that? I did not observe the data being published every 1s, and only each teleperiod (StatusSNS) and while movement (RESULT)

stefanbode:
This is only on the sensor bus that is used for rules and to update the webpage. You do not see this normally. That's the reason i was also not aware. But when you create a rule an just print the value you will see the rules gets triggered every second. This was the same with e.g. shutterposition

jziolkowski:
Ok, so is there some middle ground? So that the shutter positions can be still obtained with status?

Maybe a solution would be to update the existing StatusSHT so that it returns everything available about shutters? not only their config, but the current position etc?

OR find a way to have that data available with telemetry, as it was before?

stefanbode:
There should be no change in the telemetry reporting. If so this is a bug and needs a fix

jziolkowski:
yep, it's gone from StatusSNS

jziolkowski:
as of 13.0.x it works, on latest builds not anymore. I can't tell which revision exactly made the regression, but here we are :)

stefanbode:
Accepted as regression. Will fix

jziolkowski:
Any plans on fixing that?

stefanbode:
Oh yes. Just missed it. Work on it today

PR #22370, opened 2024-10-27
New feature Auto Clear Remote Temp for MiElHVAC

### LINKED ISSUE #22365: MiElHVAC Remote Temp and reverting to internal temp
### PROBLEM DESCRIPTION
After sending command with remote temperature to miel hvac unit the unit correctly reports the new room temperature BUT 
the unit never reverts to internal temp.(waiting time of over 10minutes)
Tried sending  0 temp  but instead of reverting to internal temp sensor it shows 10.
(cmnd/tasmota_xxx/HVACRemoteTemp) 

only power cycling the unit fixes the problem

### REQUESTED INFORMATION

- [ ] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [ ] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
  - [ ] Pre-compiled
  - [ ] Self-compiled
```lua
  Configuration output here:


```
```lua
  Rules output here:


```
```lua
  STATUS 0 output here:


```
```lua
  Console output here:


```

### TO REPRODUCE

### EXPECTED BEHAVIOUR
_A clear and concise description of what you expected to happen._

### SCREENSHOTS

### ADDITIONAL CONTEXT

grzegorz914:
## Description:
* This PR add auto clear remote temperature function
* This function is called on first run and also `10sec` after the remote temperature stop updating its value
* Send `clear` manually is also available
* The auto clear time is in (ms), default is `10000ms`, can be changed with command `HVACRemoteTempClearTime`

**Related issue (if applicable):** fixes #22365

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241023


mamrai1:
I think 10sec are not enough. Maybe set the time to follow teleperiod ? Or 60 sec are more realistic

grzegorz914:
Why is not realistic? you can call Remote Temp every 1 sec or quickly.

mamrai1:
I mean realistic for an Hvac operation…the default in swicago library i think is 3 minutes…

grzegorz914:
10 sec is very long time, the microcontroller respond time is in milliseconds. Where You see the problem to send data for example every 5 sec. I think the data need to be operated as quickly as possible.

mamrai1:
Ok. Let imagine 2 scenarios:
1. The remote temperature sensor is on the same microcontroller: yes 10 sec is more than enough but
2. The remote temperature sensor  is on another controller (in the same room of course) and sends the data via the mqtt server or directly on the controller on the unit: it does make sense to send the readings at teleperiod ….

grzegorz914:
I use PLC to communicate with my microcontrollers, the 10sec is very long time where my loop in code is 7ms. Update data every few seconds is not a problem and also is nothing for CPU load. I don’t understand where is the problem, You can send the Remote Temp every 1sec, 2sec, 500ms etc. This technically is not a problem.

mamrai1:
Many users use wifi/mqtt…  Would be ok only a check in the mqtt loop if there is disconnection from the mqtt server?

mamrai1:
Also in looked at mitsubishi2mqtt project and the time is set at 5 minutes !  (lines ~ 1350)

https://github.com/gysmo38/mitsubishi2MQTT/blob/master/src/mitsubishi2mqtt/mitsubishi2mqtt.ino

grzegorz914:
This is wrong, in 5min the temp can changed from 20 to 25 grad. What happened if set temp is set to 23? nothing, just heating because the remote temp was not refreshed.

mamrai1:
I do agree that in some situations the 5min is too long but also in some  situations the 10sec is too short. :)

grzegorz914:
What is the benefit to get longer time? 
Why the 10sec time is not OK?,  because for me shorter time is always better because gat more real time data to the device.

grzegorz914:
I can make the auto clear time configurable but I have not enough skill with Tasmota to store it in non-volatile memory. In other case will always revert to default after restart.  May be someone can help.

arendst:
Make it a command.

This can then be used in a persistent rule during restart/boot and set the desired parameter like it was persistent.

This way no additional resources are neede.

PR #22435, opened 2024-11-05
Use standard `var` spelling for reactive power unit

scop:
## Description:

Note: this is a monkey change, I'm not 100% sure of the effects, please review with care. Home Assistant has switched from `VAr` to `var` a while back.

There are a few more occurrences of `VAr` in the tree, but changing those did not seem safe (or I was even more unsure what doing so might affect than the ones I included here).

- https://en.wikipedia.org/wiki/Volt-ampere#Volt-ampere_reactive
- https://engineering.electrical-equipment.org/energy-efficiency-kvar/kvar-or-kvar.html

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241030


Jason2866:
There is a difference for
Unit name: Var
Unit symbol: var
regarding https://de.wikipedia.org/wiki/Var_(Einheit)
So imho leaving as it is.

s-hadinger:
The international symbol is `VA` not `va`. I suggest that we stick to international official units.

Issue #22267, opened 2024-10-11
Bluetooth BT BLE devives (EQ3) not connecting properly


AleXSR700:
### PROBLEM DESCRIPTION
_A clear and concise description of what the problem is._

I do not know if this is related to issue https://github.com/arendst/Tasmota/issues/21445, but after upgrading my esp32/esp32c3 devices to 14.2.0.6, my eQ3 devices are no longer connecting reliably.

I have 8 Xiaomi devices and 8 EQ3s. The Xiaomis (running ATC firmware) seem to work but the EQ3s just show `20:44:07.951 EQ3: 001A2208C9FA: trv operation failed - retrying -12` (with the respective device ID).

Interestingly, I see all 8 EQ3 TRVs right after a restart and they briefly show the correct temperatures in the WebUI. But then the connection starts failing and all values are set to 0.

Does anybody have similar issues?

EDIT: Reverted to 13.4.0 (http://ota.tasmota.com/tasmota32/release-13.4.0/tasmota32-bluetooth.bin) and it works again. So seems that something is broken in 14.2.0.6 again.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [ x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x ] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x ] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [x ] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x ] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x ] Device used (e.g., Sonoff Basic): esp32 and esp32c3
- [x ] Tasmota binary firmware version number used: 14.2.0.x
  - [x ] Pre-compiled
  - [ x] Self-compiled
- [x ] Flashing tools used: OTA
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:


```
- [ ] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```lua
  Rules output here:


```
- [x ] Provide the output of this command: `Status 0`:
```lua
  STATUS 0 output here:
[...]

PR #22453, opened 2024-11-09
MI32: Prevent active BLE read with unencrypted MI-format beacons

thewebface:
## Description:

I have various LYWSD03 sensors running ATC firmware that are sending unencrypted MI-format beacons. This is causing Tasmota to constantly try and make active read connections to them. These connections are not necessary as the beacons are unencrypted.

Currently, Tasmota skips the active connection for encrypted beacons if the key is known, this PR make it also skip if the beacons are unencrypted.

I can see the same code is present for the MHOC401 sensor and have applied the same fix, although I haven't tested on this device.

Note: I noticed this issue due to regular reboots caused by `BLE disconnect taking > 60s`. These no longer happen running this branch.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241030

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_62_esp32_mi_ble.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_62_esp32_mi_ble.ino
@@ -792,7 +792,7 @@ int genericSensorReadFn(int slot, int force){
       break;*/
     case MI_LYWSD03MMC:
       // don't read if key present and we've decoded at least one advert
-      if (MIBLEsensors[slot].needkey == KEY_REQUIRED_AND_FOUND && !force) return -2;
+      if ((MIBLEsensors[slot].needkey == KEY_NOT_REQUIRED || MIBLEsensors[slot].needkey == KEY_REQUIRED_AND_FOUND) && !force) return -2;
       res = MI32Operation(slot, OP_READ_HT_LY, LYWSD03_Svc, nullptr, LYWSD03_BattNotifyChar);
       break;
     case MI_LYWSD02MMC:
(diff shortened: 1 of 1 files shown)


Jason2866:
@btsimonh Are you fine with this PR?

btsimonh:
hmm...  the MI code is not something I am overly familiar with.  The basic code seems fine :).

However, @thewebface - do your devices respond on the battery indicator `LYWSD03_BattNotifyChar`?

If so, this could be a good opportunity to check if https://github.com/arendst/Tasmota/pull/22328/commits/b6eeb4fdd13371727552c241e052e7ea6238517f has an impact on getting the battery from MI devices....
(see also @Staars 's last comment in https://github.com/arendst/Tasmota/issues/22267)

Would it be possible to investigate if the devices respond correctly when response is set to false, or it they need the default true?

(i.e., just try with `bool response = false;` replaced with `bool response = true;` and report if your devices respond one way or the other or both ways)

IF they respond with true, but not with false, we'll need to make some allowance in the BLE32 API to allow the EQ3 driver to specify it wants false, and leave default true for battery requests to MI devices.

thewebface:
I've not been doing any active reads at all yet as the sensors are returning the correct battery state in the beacons. I can give it a try though

thewebface:
Unfortunately my devices don't appear to support active read of battery state
```
failed - svc not on device?
```

Jason2866:
Is the PR ready to merge? For me it looks like there are open questions. If so please change to Draft

thewebface:
Correct me if I'm wrong @btsimonh but I don't think any of the outstanding questions relate to this PR specifically so as far as I'm aware it should be good to go. I've been running this branch for 2 weeks now and it's been stable.

PR #22466, opened 2024-11-11
MS5837 console calibration options and code cleanup

vtHydroponics:
## Description:

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [X] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241030


vtHydroponics:
A question for the MS5837 calibration via the console... I need a value to be persistently stored for the pressure offset to be retained between resets. Automatic recalibration of the pressure offset would not be possible due to the requirement of both sensors to be at the same elevation for calibration. Thank you.

arendst:
What's the size of the value (uint8_t, uint16_t, uint32_t, int) - keep it as small as possible?

vtHydroponics:
the size of the value is a 4 byte float

arendst:
Use variable `Settings->ms5837_pressure_offset`

PR #22529, opened 2024-11-22
CheapPower module

dr-m:
## Description:
CheapPower module

This fetches electricity prices and chooses the cheapest future time slots. Currently, the following price data sources have been implemented:
* `cheap_power_dk_no.tapp`: Denmark, Norway (Nord Pool)
* `cheap_power_elering.tapp`: Estonia, Finland, Latvia, Lithuania (Nord Pool via [Elering](https://elering.ee/en), ex VAT)
* `cheap_power_fi.tapp`: Finland (Nord Pool)
* `cheap_power_se.tapp`: Sweden (Nord Pool); assuming Swedish time zone
* `cheap_power_smard.tapp`: https://smard.de (Central Europe; local time)
* `cheap_power_uk_octopus.tapp`: United Kingdom (Octopus Energy)

See `cheap_power/README.md` for more details.

To use:
* copy the `cheap_power_*.tapp` for your data source to the file system
* invoke the Tasmota command `CheapPower1`, `CheapPower2`, … to
   * download prices for some time into the future
   * automatically choose the cheapest future time slots (default: 1)
   * to schedule `Power1 ON`, `Power2 ON`, … at the chosen slots
   * to install a Web UI in the main menu

In case the prices cannot be downloaded, the download will be retried in 1, 2, 4, 8, 16, 32, 64, 64, 64, … minutes until it succeeds.

The user interface in the main menu consists of the following buttons:
⏮ moves the first time slot earlier (or wraps from the beginning to the end)
⏭ moves the first time slot later (or wraps from the end to the beginning)
⏯ pauses (switches off) or chooses the optimal slots
🔄 requests the prices to be downloaded and the optimal slots to be chosen
➖ decreases the number of slots (minimum: 1)
➕ increases the number of slots (maximum: currently available price slots)

The status output above the buttons may also indicate that the output is paused until further command or price update:
(0≤1)0.299‥2.38 ¢/kWh	⭘
Or it may indicate the duration until the next time slot and its price:
(1≤1)0.299‥2.38 ¢/kWh (0.299)	⭙ 5:05
The first number indicates the number of active or scheduled time slots, and the second number indicates the maximum number of time slots per day (default 1).

The two quantities around the ‥ are the minimum and maximum known prices from the current time onwards.

The scheduled slots are also indicated by red bars in the graph of the available current and future prices.
## Checklist:


dr-m:
> Thanks. Is it valid only for Finland? If so, I believe it should show in the name of the file like "cheap_power_Finland"

The current version is useful only for Finland, but it could be fairly easily extended to cover at least the entire [Nord pool](https://www.nordpoolgroup.com/) area. The data is produced by ENTSO-E, which based on https://transparency.entsoe.eu/load-domain/r2/totalLoadR2/show seems to cover most of Europe, including Türkiye and Ukraine. Some of these countries might not implement dynamic pricing for end consumers yet, but I believe that it will come.

Ideally, someone would run a public service that is based on some code like https://github.com/oysteinjakobsen/fetch-day-ahead-price or https://github.com/JaccoR/hass-entso-e so that end users can save themselves the trouble of registering and configuring an API key with ENTSO-E. Besides, this API could be too resource intensive to implement in Tasmota.

The hypothetical service would deliver the known prices starting from the currently active slot in a uniform format. The command would take a parameter to the command to specify the price area (such as an area of Norway):
```
CheapPower2 NO4
```
This could be translated into a simple URL like `http://e-prices.example.com/no4` or `http://no4.e-prices.example.com` if such a service existed, or it could use country specific services (see below). I’d prefer HTTP instead of HTTPS, because SSL and TLS are constantly evolving, which could require frequent firmware updates.

The second parameter could also be a full URL, for example pointing to server in the LAN, which could run an ENTSO-E interface and cache to serve multiple devices. This would in no means be limited to ENTSO-E or Europe.

Another thinkable enhancement would be a third parameter to specify the desired number of slots to choose per day, like this:
```
CheapPower1 DE 5
```
Some heating could need to run for multiple hours per day. There are some plans to narrow the price slots from 60 to 15 minutes in the future. In that case, even my deployment would require 2 to 4 such slots per day.

A quick search turned up some further open JSON data sources:
 * Sweden: https://mgrey.se/espot/api
 * Norway: https://www.hvakosterstrommen.no/strompris-api
 * Denmark: https://www.elprisenligenu.dk/elpris-api

These could be implemented fairly easily, after asking the operators if this kind of automated access is okay with them. The format of the URL and the data would likely vary between any area-specific JSON data sources. I realize that to reduce the memory footprint, it could make sense to split this interface into separate modules that would be loaded on demand.

For Estonia, I only found https://elektrihind.ee/borsihind/ which does not seem to include any public JSON based interface. The German [Fraunhofer-Institut für Solare Energiesysteme](https://www.ise.fraunhofer.de/) is running https://energy-charts.info/charts/price_spot_market/chart.htm with a nice country selection, but apparently without any raw data interface that is suitable for this kind of use. There is a CSV export function that seems to spit out data for the current week.

I think that this needs to start somewhere. If it helps, I can make the country parameter mandatory in the first version, and reject anything else than `FI`. Possibly I could refactor the parsing and implement dynamically loaded parser for two data sources, one of them being the `FI` data source.

dr-m:
I was in contact with the provider of the Swedish price data, and now there is a simpler URL https://mgrey.se/espot?format=json&domain=SE1&date=2024-12-13 that will return the price for a single zone for the given day. Unfortunately, when I tried accessing this HTTPS server in the Berry console of Tasmota 14.1.0, I got an error, I suppose due to some TLS or SSL incompatibility. It is not possible to enable plain HTTP support on this server.

I could make the URL pattern configurable, so that this service could be reached via (say) `http://router.lan/espot` which would be a proxy for the external HTTPS server, for example by [nginx reverse proxy](https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/). In that way, it could be claimed that this feature is not specific to a particular country.

s-hadinger:
Hmmm. I probed `mgrey.se:443` with openssl and they support only the following ciphers:
```
Testing ECDHE-ECDSA-AES256-GCM-SHA384... YES
Testing ECDHE-ECDSA-CHACHA20-POLY1305... YES
Testing ECDHE-ECDSA-AES128-GCM-SHA256... YES
```

I will evaluate the impact of supporting `ECDHE-ECDSA-AES128-GCM-SHA256` in addition to `ECDHE-RSA-AES128-GCM-SHA256`

s-hadinger:
Please try with the latest version which includes https://github.com/arendst/Tasmota/pull/22649

I have now enabled ECDSA:

```berry
wc = webclient()
print(wc.begin('https://mgrey.se/espot?format=json&domain=SE1&date=2024-12-13'))
print(wc.GET())
print(wc.get_string())
print(wc.close())
```

Output shows:
```
<instance: webclient()>
200
{"date":"2024-12-13","SE1":[{"hour":0,"price_eur":1.35,"price_sek":15.54,"kmeans":0},{"hour":1,"price_eur":1.35,"price_sek":15.54,"kmeans":0},{"hour":2,"price_eur":1.35,"price_sek":15.52,"kmeans":0},{"hour":3,"price_eur":1.36,"price_sek":15.6,"kmeans":0},{"hour":4,"price_eur":1.4,"price_sek":16.07,"kmeans":0},{"hour":5,"price_eur":1.38,"price_sek":15.87,"kmeans":0},{"hour":6,"price_eur":1.31,"price_sek":15.13,"kmeans":0},{"hour":7,"price_eur":1.38,"price_sek":15.87,"kmeans":0},{"hour":8,"price_eur":1.37,"price_sek":15.8,"kmeans":0},{"hour":9,"price_eur":1.5,"price_sek":17.21,"kmeans":1},{"hour":10,"price_eur":1.37,"price_sek":15.76,"kmeans":0},{"hour":11,"price_eur":1.43,"price_sek":16.43,"
nil
```

dr-m:
Thank you a lot, @s-hadinger! I upgraded to a development snapshot (b3b969978283df8da352e9d7580642caaeddbb28 is 1 commit ahead of the merge 615c6763f68a889fd22c67e913a29f0b8934d307 of #22649):
```
Version 14.4.0.1(b3b9699-tasmota32)-3_1_0(2024-12-14T23:37:22)
```
I am glad to see that also sahkotin.fi now is accessible with https. My upcoming update will revise that URL as well.

sfromis:
FTR, most example projects with Tasmota+Berry are published on Github repositories owned by the creator. I've collected a list of repositories including Berry code:
http://sfromis.strangled.net/tasmota/berry/github-repositories

dr-m:
That URL is timing out for me. Yes, I was uncertain if this would be the appropriate place. I was expecting to a link to a package directory at https://tasmota.github.io/docs/Tasmota-Application/.

I spent quite a bit of time debugging today, trying to figure out what I am doing wrong when implementing support for a second data source. It turns out that the problem is directly caused by switching `http` to `https`. 555dc048467b6f99274bd274ea998529fd1dcfa6 is a simple rebase of the original implementation, only replacing the `.bec` file inside the `.tapp` file with straight `.be`. That works fine.

As soon as I change the code to use `https` instead of `http` (adding just one `s`) and try to execute `CheapPower FI`, there will be no response to the Tasmota console; it seemingly hangs. Much of the time, after a few seconds, the device will recover, and the console will show some output indicating that it was reset. Twice, I had to reset it by pressing the button on the device. I’m running a `tasmota32.bin` of Tasmota 14.4.0.1 (b3b9699-tasmota32).

Unfortunately, there is no serial console connection to my only Tasmota equipped device, so I’m afraid I am unable to debug this deeper. When I was experimenting with the Berry console, it seemed that
```berry
      var data = json.load(wc.get_string())
      wc.close()
```
could run out of memory (end up with `data=nil`) while
```berry
      var data = wc.get_string()
      wc.close()
      data = json.load(data)
```
might allow the Berry garbage collector to free some memory earlier. However, when I tried to revise the program like this, it would still not work (cause the device to be reset or to lose the WLAN connection).

I would appreciate it if you could check if the anomaly is reproducible for you.

s-hadinger:
With the URL above `https://mgrey.se/espot?format=json&domain=SE1&date=2024-12-13`, the payload is 1404 bytes.

Once loaded the JSON takes 3.5KB which is not huge but still significant

sfromis:
Well, of course using `https` will have to use more memory than unencrypted `http`, but with a string size of "only" 1395 bytes, that should not be a pain point.

I had no trouble getting past the point of fetching the data, using:
`import cheap_power cheap_power.update()`
Unsurprisingly, it failed later (when not having a channel where to turn on power), but I suppose that the test got "far enough"? I added a couple of `print` to see the used URL, and size of data retrieved.

My test was using a recent build of tasmota 14.4.1.1 (a bit newer than yours), on an ESP32-S3 with PSRAM available. Switching to an ESP32-C3 with less RAM (and no PSRAM) made no difference, it still had no trouble https-fetching the data.

While deferring the `json.load` till after closing the connection would in principle allow a bit earlier garbage collection, I'd not expect the still open `webclient` instance to take up a lot of RAM, and with the moderate amount of data, I'd not even expect RAM to be "the issue", as confirmed by it also working on an ESP32-C3 with lower resources.

Of course, not knowing or replicating your test case, I can't be sure if my test got far enough for whatever issue you had.

In general, I like to reduce scope of test cases to "zoom in". As long as I can prune to code to be shorter I know that errors still occurring has to be within what's left.

dr-m:
I had tried to add several `print` statements as well, but the problem was that there was no output in the Tasmota console of the web UI after I executed the command `CheapPower` (or `CheapPower FI` in the version that I did not publish yet), until the eventual restart messages, where the first few messages had a timestamp close to 00:00 and the last ones with the current time. The hang or crash is reproducible just by replacing the `http:` with `https:` in the published code, installing the `.tapp` file, and executing `CheapPower` from the Tasmota console right after restart. With `http:` this code has been stable, running for several weeks.

I will try to narrow down this "test case" for reproducing the anomaly. That will take a few days, though. I think I should use an `https` URL that is expected to return a constant dataset, instead of something that depends on the time of the day.

barbudor:
Print goes nowhere
Either use Arduinos's 'Serial.print' to send to serial
Or tasmota's 'AddLog' for the webconsole

sfromis:
Of course, `Serial.print` or `AddLog` does not apply to coding in Berry, where you instead have `print()` or `log()`. As a crash-proof technique, you could also write it to a file opened for append, and `flush()` the buffer after writes.

If the failure is first time in a code path, you could also create a runtime error at a certain point in the code, if you get that, you know that it was safe that far.

If the failure is *not* first time in the code path, it might be a memory leak, which you can check by following free memory. If it keeps depleting, something is rotten.....

sfromis:
The Berry language does allow such:
`throw 'crashing'`
Of course, also trivial runtime errors like division by zero
`print(1/0)`

barbudor:
> Of course, `Serial.print` or `AddLog` does not apply to coding in Berry, where you instead have `print()` or `log()`. As a crash-proof technique, you could also write it to a file opened for append, and `flush()` the buffer after writes.

Sorry
I'm too tired, too much work

dr-m:
I debugged this a little. I removed the `.tapp` and input the code with the following modification to the Berry Console:
```diff
--- tasmota/berry/modules/cheap_power/cheap_power.be	2024-12-15 21:00:30.860176636 +0200
+++ tasmota/berry/modules/cheap_power/cp.be	2024-12-24 09:25:48.127720826 +0200
@@ -1,9 +1,6 @@
 import webserver
 import json
 
-var cheap_power = module("cheap_power")
-
-cheap_power.init = def (m)
 class CheapPower
   var prices # future prices for up to 48 hours
   var times  # start times of the prices
@@ -20,7 +17,7 @@
     "<td style='width:25%'><button onclick='la(\"&op=2\");'>🔄</button></td>"
     "<td style='width:25%'><button onclick='la(\"&op=3\");'>⏭</button></td>"
     "</tr></table>"
-  static var URL0 = 'http://sahkotin.fi/prices?start=', URL1 = '&end='
+  static var URL0 = 'https://sahkotin.fi/prices?start=', URL1 = '&end='
   static var URLTIME = '%Y-%m-%dT%H:00:00.000Z'
 
   def init()
@@ -29,15 +26,9 @@
   end
 
   def start(idx)
-    if idx == nil || idx < 1 || idx > tasmota.global.devices_present
-      tasmota.log(f"CheapPower{idx} is not a valid Power output")
-      tasmota.resp_cmnd_failed()
-    else
       self.channel = idx - 1
       tasmota.add_driver(self)
       self.update()
-      tasmota.resp_cmnd_done()
-    end
   end
 
   def power(on) tasmota.set_power(self.channel, on) end
@@ -45,9 +36,8 @@
   # fetch the prices for the next 24 to 48 hours
   def update()
     var wc = webclient()
-    var rtc = tasmota.rtc()
-    self.tz = rtc['timezone'] * 60
-    var now = rtc['utc']
+    self.tz = 120 * 60
+    var now = 1734822000
     var url = self.URL0 +
       tasmota.strftime(self.URLTIME, now) + self.URL1 +
       tasmota.strftime(self.URLTIME, now + 172800)
@@ -149,6 +139,6 @@
     tasmota.web_send_decimal(status)
   end
 end
-return CheapPower()
-end
-return cheap_power
+var cheap_power = CheapPower()
+cheap_power.start(1)
+print(cheap_power.chosen)
```
Right after rebooting the Shelly Pro 2 into Tasmota 14.4.0.1 (b3b9699-tasmota32), when I input the code with the above modification to the Berry Console, the `print` statement will display `2`. (This should be deterministic, because I hard-coded the URL by hard-coding the `var now` to a past date.)

To my surprise, the code would run just fine, even though I’m now using `https`. I tried again without the penultimate (3rd) hunk, that is, letting it use the current time, and it still seems to work, with `size(cheap_power.prices)` being 16 and `cheap_power.chosen`  being 15 at the moment. The source of the instability would seem to be the `tasmota.add_cmd()` or the `tasmota.resp_cmnd_done()`.

sfromis:
AFAICT, the "culprit" was the call hierarchy of 
> command handler
> cheap_power.start
> self.update
> self.schedule_chosen
> self.power
> tasmota.set_power
> Power command

The thing is that issuing commands can be quite problematic when being invoked from an existing Tasmota core-related callback (especially a command callback), and `tasmota.set_power` will internally be issuing a `Power` command (as visible in the console).

In many cases, you can use a workaround of breaking the command hierarchy, using something like `tasmota.add_timer(0, self.update)`, deferring the function call to right after returning from the current call hierarchy.

dr-m:
@sfromis Thank you for the advice; I will try that. I was thinking of `tasmota.set_timer()`; maybe you meant that by `tasmota.add_timer()`.

For the record, f7fc7320e68dec8a26c0f5e04b424d95de78b33a is my current development, implementing an interface for Swedish prices. With `http://sahkotin.fi` it works, but with `https` it will cause the system to hang as soon as `CheapPower1 FI` is invoked. The code worked fine with `https` when I used a similar patch that I had posted above (not installing any Tasmota command hooks). I tried to make the test conditions as deterministic as possible, by restarting the system after updating the file using the file manager, and right after restart, invoke `CheapPower1 FI` from the console.

ryancdotorg:
It would be nice if this could be made to support Octopus Energy in the UK as well. Octopus's API returns larger text results, though - looks like about 15KB.

dr-m:
@ryancdotorg Great idea. I hope that implementing support for more markets would make this pull request more ‘eligible’ to be merged to the Tasmota repository.

I searched the web, came across https://gist.github.com/TerryE/55e413ce59b40a7233df9d76ad5821e6 and checked the output of https://api.octopus.energy/v1/products/. I successfully ran the following in the Tasmota Berry console of my Shelly Pro 2:
```berry
import json
var wc=webclient()
wc.begin("https://api.octopus.energy/v1/products/AGILE-24-10-01/electricity-tariffs/E-1R-AGILE-24-10-01-B/standard-unit-rates/")
var rc=wc.GET()
var s=wc.get_string()
wc.close()
print(s)
s=json.load(s)
print(s)
```
The URL in the above snippet seems to return the price information starting from the furthest available time in the future, in descending order of time, at 30-minute intervals. The interface could conveniently use a fixed URL and simply assume that it will return all data down to the current timestamp, and some time in the past (which our parser would ignore).

The `B` in the URL would be the tariff zone. That as well as the two occurrences of `AGILE-24-10-01` would have to be a parameter to the Tasmota command. Something like `CheapPower UK AGILE-24-10-01 B`? Is this what you had in mind? How often would the identifier such as `AGILE-24-10-01` have to be updated? Every time the electricity contract is renewed?

sfromis:
I'd say that support for multiple data sources would probably best be implemented by some sort of plug-in architecture, with the core functionality for controlling energy usage by optimum time would have a standardized interface to data source plugins. Maybe something like a near future list of time stamps (typically but not necessarily hourly) and price. Anyone could create their own data gathering function for their local market(s).

Of course, mixing in PV and buffer batteries (like from an EV) can quickly be used to make the situation more complicated, if that's what you want.... Etc.

dr-m:
I think that more complicated control solutions, such as those involving PV, battery banks or EV charging, are better implemented in a larger environment, such as Home Assistant. This control only makes sense for rather simple use cases.

I’d like a plugin architecture where the `cheap_power.tapp` would load a market specific `.be` file, which would return `prices[]` and `times[]` for the main UI. Can arrays be passed "by reference" and modified by the callee?

Another thing that I need to figure out a simple enough GUI and logic for allowing multiple cheap slots per day to be used. Nord Pool is supposed to move from 60-minute to 15-minute slots in the near future.

sfromis:
In Berry, when you pass objects (including lists), it is always by reference, with the callee getting full access. The caller should not depend on it being unmodified.

A more generalized approach, not even depending on it being on the same device, could be to pass it as JSON. Yes, using `json.dump` and `json.load` also works for plain lists (arrays).

ryancdotorg:
Regarding Agile Octopus, yes, the tariff region would need to be manually selected. In theory, it's possible to determine based on lat/lon but that seems like too much magic to me.

There's another endpoint that can be used to discover the current correct tariff identifier, but it's a bit of a faff.

FWIW can see both sides of the argument as to whether this makes sense to have this built into Tasmota.

sfromis:
As I see what was suggested, it was adding Berry sample code for users to pick up, not having it "built into" Tasmota. And lots of users have published their Berry projects via Github, without them being submitted to the common Tasmota repository.

ryancdotorg:
Ah, thanks for clarifying.

dr-m:
Right, it probably does not make sense to include a "non-core component" like this in any pre-built Tasmota firmware image.

I created this pull request because the only directory of Tasmota Berry projects or `.tapp` files that I could find is https://tasmota.github.io/docs/Tasmota-Application/#tapps. Everything that is listed there is included in the Tasmota repository. I just found https://github.com/tasmota/Berry_playground which I think would be useful to list in https://tasmota.github.io/docs/Berry/. Is this what you @sfromis had in mind?

I can imagine that running some more complex Berry or Tasmota code in a CI pipeline could be good for some regression testing of the core code. If I understood it correctly, the current `.github/workflows/` is only compiling code for various targets, not running any tests (say, with `qemu` emulating an ESP32 device).

sfromis:
I was not trying to "say what to do", but as mentioned in the beginning of this thread, many other personal projects using Tasmota+Berry are in separate Github repositories, where I've created a list of ones I've noticed. Quite a long list:
http://sfromis.strangled.net/tasmota/berry/github-repositories
That list is very basic, and does not try to highlight "juicy" projects, except through code size being a proxy for "something interesting going on".

When it comes to listing such projects directly in the Berry docs, I don't have much of a stance either way, but a few such repositories are already linked from:
https://tasmota.github.io/docs/Berry-Cookbook/

Unfortunately https://github.com/tasmota/Berry_playground did not really work out, and not much is happening there. I can't say that I see that structure as a very suitable for acting as a combined repository, especially when creators already using Github can have their own places.

Most of what's stored in the core Tasmota project repository are in the direction of useful tools closely tied to Tasmota, or "educational" programmin examples, but still a mixed bag.
https://github.com/arendst/Tasmota/tree/development/tasmota/berry

All 3 places are already linked at the bottom of the mentioned cookbook page.

dr-m:
Sorry, I had forgotten about http://sfromis.strangled.net/tasmota/berry/github-repositories because it used to time out for me. It is accessible now.

I think that this might serve as an "educational" example of implementing a modular Tasmota application, once I have implemented the suggested pluggable interface to a few data sources.

dr-m:
I think that the most straightforward way is to implement price source specific derived classes. I successfully did that for Finland and Sweden. For UK, what I currently have causes a system restart.

I was trying to structure the code like this:
```berry
var cheap_power = module("cheap_power")

cheap_power.init = def (m)
import cheap_power_base

  class CheapPower: CheapPowerBase
    # …
  end
  return CheapPower()
end
return cheap_power
```
In the file `cheap_power_base.be` I would have just the base class definition and its dependencies, like this:
```berry
import webserver
import json

class CheapPowerBase
# …
end
```
The `.tapp` file (containing `autoexec.be`, `cheap_power_base.be` and only one derived class) would load fine, but as soon as I try to execute the command `CheapPower`, it would fail like this:
```
15:48:07.257 CMD: CheapPower
15:48:07.378 BRY: Exception> 'syntax_error' - /cheap_power.tapp#/cheap_power_fi.be:6: 'CheapPowerBase' undeclared (first use in this func...
15:48:07.380 stack traceback:
15:48:07.382 	<unknown source>: in function `_anonymous_`
15:48:07.383 	<unknown source>: in function `exec_cmd`
15:48:07.385 	<unknown source>: in function `event`
15:48:07.387 RSL: RESULT = {"Command":"Error","Input":"CHEAPPOWER"}
```
If I load the code to the Berry console without the `module` decoration, and if I replace the `import` statement with the contents of `cheap_power_base.be`, it will work as I intended.

Is there a way to have multiple classes declared in the same Berry module, without them having to reside in the same `.be` file? The documentation focuses on using, not writing, modules.

sfromis:
The way I'd handle a plugin structure would be simpler, and needing no subclassing, favoring the *composition over inheritance* principle. The plugin would not need to extend anything, as Berry is duck-typed.

It can work well to simply pass an instance of the plugin having the relevant method(s). Obviously, it would fail runtime if the plugin does not have properly defined methods.

In simple cases, you'd not even need an instance, as you could simply pass a function doing the work, callable by the base. 

The inherent need to agree on parameters and such is no different in this type of extension architecture, as you get no compile-time validation of parameters or return values in either case.

Here's a simple example of a class delegating the variable part of the work to an "action" function passed to the class initialization. This also has the detail of the class using a default action if nothing was passed as parameter.
http://sfromis.strangled.net/tasmota/berry/examples/slider.be

dr-m:
Thank you @sfromis, I will check your encapsulation example and see if that could lead to a better solution.

I figured out how to make modules and class inheritance work. The base class is in a module on its own, and each subclass (specific to a data source) derives from the base class. I understand that this could lead to wasting memory for one singleton object of the base class; the derived class would use another singleton object. The singleton of the base class is only being used for `classof(cheap_power_base)`; any actual data will be stored in the `cheap_power` singleton.

To avoid code bloat on the end device, I ended up creating area specific `cheap_power_*.tapp` files that will always include the same `autoexec.be` and `cheap_power_base.be` and the area specific `*/cheap_power.be`.

I fixed the interface to the Swedish data source so that we can fetch the prices for the current day and the next day (in 2 requests). There was a bug in `update()`; it is now invoking `prune_old()` before `find_cheapest()`.

I tested this for all 3 data sources (Finland, Sweden, Octopus Energy in the UK) on my Shelly Pro 2.

dr-m:
The example http://sfromis.strangled.net/tasmota/berry/examples/slider.be defines one function pointer, which is being assigned to an anonymous function that is defined when an object is instantiated. That is a good solution for overloading a single function, in scenarios where one would use a lambda expression in C++11 or later.

For overloading multiple functions (in this case: `start()`, `url()`, `parse()`), I find that it is clearer to use a derived class and named member functions. The `CheapPowerBase::start()` only needs to be overloaded for parameterized data sources.

It looks like I could rather easily include interfaces for Norway and Denmark, possibly using a unified parser, because the format only differs in the field name (`NOK_per_kWh` and `DKK_per_kWh`). I successfully tested the following with `webclient` and `json.load()`:
https://www.hvakosterstrommen.no/api/v1/prices/2025/01-06_NO5.json
https://www.elprisenligenu.dk/api/v1/prices/2025/01-06_DK2.json

sfromis:
For the multi-function case, I'd actively want to **not** use a derived class, but just a class without any inheritance hierarchy, and then pass an instance of that class. Not much different from the case of a single function.

This is the generally recommended principle of "composition over inheritance" now recognized as a cleaner solution in most cases. Easier to work with when no requirements for tight coupling.

Trivial example to illustrate:
```
class base
  var plugin
  def init(plugin)
    self.plugin = plugin
  end
  def run(parm)
    self.plugin.put(parm)
    var returned = self.plugin.get()
    print(returned)
  end
end
class plugin
  var data
  def put(data)
    self.data = data
  end
  def get()
    return self.data
  end
end
plug = plugin()
base(plug).run('abc')
```

dr-m:
Thank you. Currently, there is some tight coupling in the overloaded member function `start()`. It should not be too difficult to refactor that.

dr-m:
The encapsulation approach (d1632c7ff308d9a92546bef89bb04f87cd872122) is cleaner than the one based on a derived class (dddfe01dce69ecd6798f3f5e92e4136785f908e4). I figured out that `import cheap_power_base` can actually return the class, instead of returning an instance.

dr-m:
Sorry, I had introduced a bug when I refactored some code into `prune_old()`; the scheduled time slot could move around when `web_sensor()` is removing old price history. I’m also replacing `static PAST` with `var past`, because that parameter actually depends on the data source (currently -1800 seconds or -½ hours for Octopus Energy and -1 hour for all other sources).

dr-m:
I found [a nice CSV data source  for Estonia, Finland, Latvia and Lihuania](https://dashboard.elering.ee/assets/api-doc.html#/nps-controller/getPriceAsCSVUsingGET) with timestamps in the POSIX `time_t` format. There is also a JSON interface, but it does not seem to allow filtering for one country. Parsing CSV seems to be this straightforward:
```berry
for line: string.split(s,'\n')
  for column:string.split(line,';')
    print(column)
  end
end
```

dr-m:
I implemented a price graph, with the chosen time slot highlighted in red. The time unit is now determined automatically based on the difference of the first two timestamps in the downloaded data.

For now, this remains limited to one time slot per day. I came up with a design: Add ➕ and ➖ buttons for changing the maximum number of slots, and let the ⏮ and ⏭ buttons move the first available slot. The logic would choose up to the specified amount of slots (limited by the available future price information), starting from the specified first slot, or from the current time.

dr-m:
For many Central European countries,  I found an electricity price data source in CSV format https://www.smard.de/home/downloadcenter/download-marktdaten/ that apparently requires POST requests. The usage conditions https://www.smard.de/home/datennutzung seem to be acceptable. I came across some documentation https://github.com/bundesAPI/smard-api that explains how to use GET requests for some of the data, but I couldn’t find any parameters for retrieving price data.

dr-m:
e536ffd1666bf70eb8b6310dcb707df2a0ebc0b5 implements the choice of multiple time slots per day, employing a reusable implementation of a [binary heap](https://en.wikipedia.org/wiki/Binary_heap) both for choosing the N cheapest time slots and for sorting them by time, to keep the scheduling logic simple. The duration of a time slot will be updated automatically when downloading the data. The price graph displays the currently scheduled time slots with a red background.

In other words, this should be ready for the upcoming 15-minute MTU transition of the EU electricity market.

dr-m:
I figured out the minimal POST request format for [smard.de](https://www.smard.de/home/downloadcenter/download-marktdaten/).
```berry
var wc = webclient()
wc.begin('https://www.smard.de/nip-download-manager/nip/download/market-data')
var rc=wc.POST('{"request_form":[{"format":"CSV","moduleIds":[8004169],"region":"DE","timestamp_from":1738018800000,"timestamp_to":1738191600000,"type":"discrete","language":"de"}]}')
var data = rc == 200 ? wc.get_string() : nil
wc.close()
```
All these parameters seem to be mandatory. The timestamps are 1000 times the POSIX `time_t` (milliseconds from the Unix epoch). The response will include two timestamp columns apparently in German local time, and the price in €/MWh. I found the following data sources:
| moduleIds | legend |
| --------- | ------ |
| 8004169 | Deutschland/Luxemburg |
| 8004170 | Österreich |
| 8000251 | ~~DE/AT/LU~~ |
| 8005078 | ~~∅ Anrainer DE/LU~~ |
| 8000252 | ~~Dänemark 1~~ |
| 8000253 | ~~Dänemark 2~~ |
| 8000254 | Frankreich |
| 8000255 | Italien (Nord) |
| 8000256 | Niederlande |
| 8000257 | Polen |
| 8000258 | ~~Schweden 4~~ |
| 8000259 | Schweiz |
| 8000260 | Slowenien |
| 8000261 | Tschechien |
| 8000262 | Ungarn |
| 8004996 | Belgien |
| 8004997 | ~~Norwegen 2~~ |

I ~~striked through~~ those data sources that are missing data or seem to be redundant. For Denmark, Sweden and Norway, the already implemented GET data sources include a currency conversion, while these prices are in €/MWh and do not cover all price zones of those countries.

Implementing this data source would require a little refactoring so that the POST method can be supported.

@s-hadinger, do you think that covering the UK and most of the EU would justify the choice of a generic name `CheapPower`? I see that we’d still be missing some of the south, at least Portugal, Spain, and Greece.

dr-m:
I tested d0072eb5455a7a12c62e8205060df02cd1f199a9 with a few of the https://smard.de price sources AT, BE, CH, CZ, DE, FR, HU, IT, LU, NL, PL, SI. The currency unit in the main menu status display is a hyperlink to the data source, now also for `cheap_power_dk_no.tapp`, which provides an interface to two similar data sources.

dr-m:
Relying on exceptions for bounds checking in `heapify()` turned out to be a bad idea. Occasionally, when explicitly specifying a start slot and multiple slots, if the last slot was the cheapest, it would not be chosen. Implementing explicit bounds checks seems to have solved this problem. I also updated the status display so that it displays the minimum, maximum, and the next scheduled price.

dr-m:
I implemented one last cleanup of `find_cheapest()` and split the [C++11 `std::make_heap`](https://cplusplus.com/reference/algorithm/make_heap/) inspired  code into a separate module, to simplify "live refinement" of `find_cheapest()` in the Berry console:
```berry
import cheap_power
import binary_heap
var make_heap = binary_heap.make_heap, remove_heap = binary_heap.remove_heap
```
After these steps, one can paste a slightly edited copy of `def find_cheapest(first)` to the Berry console, replacing `self` with `cheap_power`, and it’ll be possible to invoke this function like `find_cheapest()` or `find_cheapest(5)` or whatever.

dr-m:
I’ve filed berry-lang/berry#459 for the `binary_heap` module that I created as part of this. I plan to return to this one once that part has been settled.

dr-m:
While the `binary_heap` module is yet to be merged to Berry and updated into Tasmota, I thought that I would update this. I learned that `while` loops can be more efficient than `for` loops that iterate `range` objects. I have a feeling that it now takes less time to pick the cheapest price slots.

sfromis:
The reason for `for` with `range` indeed being slower is that it needs to allocate/free a `range` object every time the execution path hits the `for`. And object allocation costs a lot more than incrementing/testing an index value.

For that reason, it can make quite some difference when it is inner loops, but not so much a factor for outer loops not entered again and again.

dr-m:
The `binary_heap.be` module here is identical to https://github.com/berry-lang/berry/blob/master/modules/binary_heap.be. I’m looking forward to the next update of Berry in Tasmota, so that I can avoid duplicating that file in this pull request.

dr-m:
There has been no update to Berry in Tasmota since my last comment.

dr-m:
ENTSO-E switched to 15-minute pricing cycle on 2025-10-01, a few months behind the original schedule, which was about the same time this ticket was auto-closed. The software continues to work for me without any modification. The price sources for Sweden, Denmark, and Norway use a format that does not allow to report 15-minute intervals. I did not test the https://www.smard.de APIs, but I don’t see a reason why they wouldn’t work, because their web user interface is visualizing quarter-hourly price data.

My intention is to reimplement this as an OpenWrt [LuCI](https://github.com/openwrt/luci) application. The server part (in the [`ucode`](https://ucode.mein.io) language) will be very compact. I should force myself to write the browser part. It would fetch and visualize JSON price data and offer a few controls, such as temporarily pausing the load or adjusting the number of active slots. For permanent settings (the data source, the number of active slots, and the commands to execute, such as invoking `curl` to send a `POWER` command to Tasmota) there would be a configuration section, like in [luci-app-example](https://github.com/openwrt/luci/tree/master/applications/luci-app-example).

PR #22538, opened 2024-11-23
[Shutter] Fix UI inverted mode

SteWers:
## Description:
Fix UI for shutter in inverted mode

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
@@ -1424,6 +1424,7 @@ void HandleRoot(void) {
 
       int32_t ShutterWebButton;
       uint32_t shutter_button_idx = 1;
+      uint32_t shutter_button_idx_temp;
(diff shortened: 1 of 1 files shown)


arendst:
Was this an issue before the new GUI was introduced too?

SteWers:
May be. There were several issues with the inverted mode, as you have to think completely in the other direction. Something felt wrong, when I saw the new (very nice) UI. So I investigated both modes. Normal mode worked correct.
In line 1436 was the wrong flag requested for inverted mode. After changing it, the buttons were displayed at the right position, but commanding the wrong direction. The rest of the code is for solving that. Now it works in both modes, but maybe there is a "nicer" way to solve it.

arendst:
Let's ask @stefanbode what he thinks is the correct solution.

SteWers:
Good idea.
I had a quick look at the release 14.3.0. The buttons were not swapped. As they were not near to the slider, it was no problem. 
image
(Shutter 1 is not inverted, shutter 2 is inverted)

But now the slider should move towards the button, which is pressed.
image
(Same configuration as above with the fix applied)

stefanbode:
Great thanks. Agree to all findings. Nice change.

PR #22543, opened 2024-11-24
Shitf595 dynamic device count

jziolkowski:
## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_60_shift595.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_60_shift595.ino
@@ -88,7 +88,7 @@ void Shift595SwitchRelay(void) {
 }
 
 void CmndShift595Devices(void) {
-  if ((XdrvMailbox.payload > 0) && (XdrvMailbox.payload * 8 <= MAX_RELAYS_SET - TasmotaGlobal.devices_present)) {
+  if ((XdrvMailbox.payload > 0) && (XdrvMailbox.payload * 8 <= MAX_RELAYS_SET - Shift595->first)) {
     Settings->shift595_device_count = (1 == XdrvMailbox.payload) ? SHIFT595_DEVICE_COUNT : XdrvMailbox.payload;
     TasmotaGlobal.restart_flag = 2;
   }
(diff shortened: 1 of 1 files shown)


barbudor:
What if not all your relays are on the 595 ?
Like if you have a display which has an on/off on the same device ?

arendst:
Once I receive my 595 based relay board I will look at the code, haven’t seen it yet, and make it work like the mcp23xxx one with virtual relays.

barbudor:
@arendst Which board have you ordered ?

jziolkowski:
@barbudor https://www.aliexpress.com/item/1005007370916864.html this one

jziolkowski:
> What if not all your relays are on the 595 ? Like if you have a display which has an on/off on the same device ?

That's why I reference devices_present in the condition. Since to my knowledge GPIO-configured relays take precedence over drivers, that should be enough to check if the driver can add more relays (especially that if used together with PCF and MCP which have lower driver number and I assume will bump the devices_present counter before 595 init)

jziolkowski:
>  make it work like the mcp23xxx one with virtual relays.

What do you mean by that? Like a configurable driver, or being able to use only a part of the outputs?

arendst:
As said, I haven’t looked at the driver yet, but it should just add relays like extending normal gpuo relays. Come to think about it I may be misguided by the fact thaat the 595 driver only supports outputs. In that case forget my remarks.

Noschvie:
> https://www.aliexpress.com/item/1005007370916864.html this one

this board uses an ESP8266, is there a board with an ESP32 too? Thanks!

jziolkowski:
> > https://www.aliexpress.com/item/1005007370916864.html this one
> 
> this board uses an ESP8266, is there a board with an ESP32 too? Thanks!

Short answer: no.

I replaced it with an ESP32-C3-12F. But I did that because I need ESP32 with its 16 shutters support.

If you want to only have a ton of relays, you're fine with ESP8266

barbudor:
@Noschvie There are the Dingtian boards
Up to 32 inputs and 32 outputs but they are more expensives
https://www.aliexpress.com/item/1005007002132841.html

There's a Tasmota driver for it

Beware that if you want to flash Tasmota you need to tell them when ordering otherwise they ship with their own firmware and they blow the fuses to prevent reflashing.

jziolkowski:
16 vs 32 relays, tho

barbudor:
8, 16 or 32 relays

jziolkowski:
> 373*123*43mm

wow... almost 40cm.

arendst:
I'll merge this now.

Issue #18670, opened 2023-05-16
Shutter Interlocked relays (both) go ON after reboot


kbaras:
### PROBLEM DESCRIPTION
Hi!

I have a serious problem with the latest Tasmota release. 
I have a shutter configured with two relays, with  ShutterMode1 and the relays are interlocked (1,2). As I am using inverted relays the configuration has been set accordingly (with relay_i both for relay1 and relay2). It works fine at normal operation (the interlock as well, so if one of the relays is turned ON and the other is clicked to turn it ON than the interlock automatically turns OFF the other relay so you are unable to turn ON both - which is the expected behaviour).

Problem comes after I restart Tasmota (plugged out the power source and plugged in again) then both of the shutter relays are ON after startup! After clicking on either of the relays the interlock intervenes and it turns OFF the other relay so from this point it works fine until the next restart...

I have a separate inverted relay configured additionally and it does not start with ON state.

Initially the PowerOnState was 3, but I have also tried with 0 but it did not help...

  STATUS 0 output:
18:20:30.680 CMD: Status 0
18:20:30.687 RSL: STATUS = {"Status":{"Module":18,"DeviceName":"Tasmota Wemos Kitchen shutter & sinklight","FriendlyName":["Tasmota","Tasmota2","Tasmota3"],"Topic":"tasmotawemoskitchen","ButtonTopic":"0","Power":3,"PowerOnState":0,"LedState":1,"LedMask":"FFFF","SaveData":1,"SaveState":1,"SwitchTopic":"0","SwitchMode":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"ButtonRetain":0,"SwitchRetain":0,"SensorRetain":0,"PowerRetain":0,"InfoRetain":0,"StateRetain":0,"StatusRetain":0}}
18:20:30.723 RSL: STATUS1 = {"StatusPRM":{"Baudrate":115200,"SerialConfig":"8N1","GroupTopic":"tasmotas","OtaUrl":"http://ota.tasmota.com/tasmota/release/tasmota.bin.gz","RestartReason":"External System","Uptime":"0T00:00:21","StartupUTC":"2023-05-16T17:20:09","Sleep":50,"CfgHolder":4617,"BootCount":17,"BCResetTime":"2023-05-16T17:34:21","SaveCount":44,"SaveAddress":"F8000"}}
[...]

PR #22548, opened 2024-11-25
[Shutter] Fix PowerON

SteWers:
## Description:

I have on some devices, not on all, the problem, that both shutter relais are switches on after power on. This doesn't happen after a software restart. This PR fixes the problem.
@stefanbode: Please have a look at it, if I did it the right way. 

## Checklist:

### DIFF
--- a/tasmota/tasmota_support/support_tasmota.ino
+++ b/tasmota/tasmota_support/support_tasmota.ino
@@ -466,9 +466,9 @@ void SetPowerOnState(void)
   for (uint32_t i = 0; i < TasmotaGlobal.devices_present; i++) {
 #ifdef ESP8266
     if (!Settings->flag3.no_power_feedback &&  // SetOption63 - Don't scan relay power state at restart - #5594 and #5663
-        !TasmotaGlobal.power_on_delay &&       // SetOption47 - Delay switching relays to reduce power surge at power on
+        !TasmotaGlobal.power_on_delay          // SetOption47 - Delay switching relays to reduce power surge at power on
 #ifdef USE_SHUTTER
-        !Settings->flag3.shutter_mode          // SetOption80 - Enable shutter support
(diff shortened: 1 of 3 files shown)


stefanbode:
Let’s keep this on hold. Here you disable the power switch on restart. May be a good idea if shutters defined and a shutter relay but should not kick in in any other situation. I do not understand why this happens NOW. It was not an issue the last 5 years.

stefanbode:
Thanks for all the investigation efforts. We can curate the symptoms but I would better understand the root cause and fix this. If there is no shutterrelay defined it should not matter is so80 is 0 or 1. why should it mess with relay 1 and 2 if the shutterrelay can also be anything else. this is crazy first. Do both relays go ON even with the interlock defined? I did not get this. It should be absolutely impossible. There is a setoption where Tasmota reads the state of the relay and want to set it. This is default and I disabled it on shutters as far as I remember.

SteWers:
I totally agree with you to find the root cause. Both relays go ON even when interlock is activated.
Your hint with the setoption was the right way. I think you mean SO63. When I set it to 1 the problem is away. I read about it, but as the documentation only refers to _warm start_, I haven't tested with it. I'll do some tests tomorrow and close the PR.
It would be nice, if you investigate why it isn't set, when shutter is enabled. Interesting interaction between SO80 and SO63...

SteWers:
Found this #18670. Same problem also solved with SO63. An integration in the shutter driver is desirable.

SteWers:
Let's try this way.

arendst:
Before merging this I suggest to move `ShutterInit()` from FUNC_PRE_INIT to FUNC_INIT and see if this solves the issue.

By moving it to FUNC_INIT it is executed AFTER the Initial power state is set. This may well solve the issue observed.

Pls give it a try.

SteWers:
I tried your suggestion but unfortunately it doesn't solve the problem.

arendst:
Thx for testing

PR #22596, opened 2024-12-06
CSE7761 reads also the negative power

josefpavlik:
## Description:

The chip CSE7761 mounted i.e. in Sonoff POWCT is able to read the sense of the power. It's able to distinct if you consume power from the grid or if you feed the power into the grid. However the actual Tasmota reads always the absolute value.

This patch reads the also the Power Factor registers form the chip and when it's negative (you supply to the grid), inverts the sign of the power. Negative power means feed the grid, positive power means consume from the grid.
The Current and Voltage are always positive, change only the sign of the Active Power.

I've also set the resolution of Voltage resolution to one decimal digit and resolution of frequency to 2 decimal digints. It was 0 decimal digits. 

The second commit makes report values when the relay is off. In many applications the relay does not switch off the load, so it have no sense to do not report the results when the relay is off.

Tested with Sonoff POWCT with one channel only. The patch should handle both channels of the chip, but this is not tested.

## Checklist:
  - [done] The pull request is done against the latest development branch
  - [done] Only relevant files were touched
  - [done] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241117
  - [done] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

### DIFF
--- a/tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino
+++ b/tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino
@@ -33,6 +33,11 @@
  * Based on datasheet from ChipSea and analysing serial data
  * See https://github.com/arendst/Tasmota/discussions/10793
  * https://goldenrelay.en.alibaba.com/product/62119012875-811845870/GOLDEN_GI_1A_5LH_SPST_5V_5A_10A_250VAC_NO_18_5_10_5_15_3mm_sealed_type_all_certificate_compliances_class_F_SPDT_Form_available.html
+ * 
+ * options:
+ * negativepower 0 - calculate negative power as positive (default)
+ * negativepower 1 - calculate negative power as negative
+ * negativepower 2 - calculate negative power as zero
 \*********************************************************************************************/
 
 #define XNRG_19                       19
@@ -481,7 +486,7 @@ void Cse7761GetData(void) {
     CSE7761Data.current_rms[0], CSE7761Data.current_rms[1],
     CSE7761Data.active_power[0], CSE7761Data.active_power[1]);
 
-  if (true || Energy->power_on) {  // Powered on <xxx] ALWAYS
+  if (Energy->power_on) {  // Powered on 
     for (uint32_t channel = 0; channel < Energy->phase_count; channel++) {
       if (0 == channel) {
         // Voltage = RmsU * RmsUC * 10 / 0x400000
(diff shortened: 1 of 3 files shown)


tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino
```diff
@@ -635,6 +650,8 @@ void Cse7761DrvInit(void) {
     Energy->use_overtemp = true;                 // Use global temperature for overtemp detection
     TasmotaGlobal.energy_driver = XNRG_19;
   }
+  Settings->flag2.frequency_resolution = 2;      // 0.01Hz resolution
```
arendst, line 653: Pls do not do this. Although I understand why you do this, some users may not like it. They can always use the correct commands to change these parameters.

tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino
```diff
@@ -481,7 +481,7 @@ void Cse7761GetData(void) {
     CSE7761Data.current_rms[0], CSE7761Data.current_rms[1],
     CSE7761Data.active_power[0], CSE7761Data.active_power[1]);
 
-  if (Energy->power_on) {  // Powered on
+  if (true || Energy->power_on) {  // Powered on <xxx] ALWAYS
```
arendst, line 484: Doesn't command `SO21 1` do the same?

tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino
josefpavlik: I've thinking about some option like:

tasmota/tasmota_xnrg_energy/xnrg_19_cse7761.ino
josefpavlik: Yes, sorry :-)

josefpavlik:
Hi
I've reverted the patch with forced read. I've verified that the option 21 works.
However, I've add new command **negativepower**
options are:
0 for absolute value. It's default, works as standared Tasmota
image
1 for negative value of negative power - the power is reported with negative sign and it's subtract from the energy. Energy counters goes down.
image
2 for zero value of negative power - when the power is negative, goes to the grid, the device shows zero and does not update the energy counter.
image

The config is stored in Energy->Settings.flag, i've reserved 2 bits in this flag for this option.

Please review this patch and tell me if it can be accepted. If you want to rename the command, obviously I can do it.

arendst:
I don't see why we need command `negativepower`.

For negative power we already have the option to monitor this as `energy_export`. Recently command `SetOption162` was added to not substract negative power from Energy Today but only add it to energy_export.

So all we need to do is allow the CSE7761 driver to enable option `Energy->local_energy_active_export`.

Based on your work I will try to add your negative power detection in the current driver.

arendst:
I did implement it minimally and it works for the Sonoff CT.

Alas, it doesn't work well for the Sonoff DualR3. The build the hardware in such a way that it now reports negative PowerFactor which should have been positive.

I'll work on a solution tomorrow.

arendst:
Just merged your idea in the current CSE7761 driver.

It uses negative values always for Sonoff CT depending on the current flow / CT location.
For the Sonoff Dual R3 nothing changed.

In order to not decrement Energy Today execute command `SetOption162 1`

Thx again for your work.

PR #22774, opened 2025-01-06
Use 75% of the buffer for speeds higher than 115200

mrkev-gh:
Higher speeds need larger buffer. Processing takes longer, so let's put the threshold at 75% of the buffer size to have some headroom. E.g. for 1Mbit with 2.5kB buffer 2048 characters would be roughly still around 20ms, and reserve additional 512 chars / 5ms to handle the buffer after the threshold is reached

## Description:

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.0.241206

I do not expect 8266 would be able to handle 1Mbit+ speeds.
To use with TCP Serial Bridge TCP_BRIDGE_BUF_SIZE needs to be increased.
Tested on ESP32-C3 and 1.5Mbit (rockchip serial console speed) with a 4kB buffer

### DIFF
--- a/lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
+++ b/lib/default/TasmotaSerial-3.6.0/src/TasmotaSerial.cpp
@@ -172,10 +172,13 @@ void TasmotaSerial::Esp32Begin(void) {
     // At 19200, 120 chars are ~60ms
     // At 76800, 120 chars are ~15ms
     uart_set_rx_full_threshold(m_uart, 120);
-  } else {
+  } else if (m_speed == 115200) {
     // At 115200, 256 chars are ~20ms
     // Zigbee requires to keep frames together, i.e. 256 bytes max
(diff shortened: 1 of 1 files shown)


s-hadinger:
Looks ok to me, but note that I never use serial above 115200

mrkev-gh:
@s-hadinger Thanks for the review. The only outstanding thing is the need to redefine the TCP_BRIDGE_BUF_SIZE. Would you be open to dynamic buffer allocation in TCP Bridge? E.g. for >115200 to allocate the size based on TCPBaudRate setting to keep 20ms worth of chars and 5ms for processing, so...assuming 8N1 that would be something like 25/(1/(baudrate/9)*1000) bytes. , for 1,5Mbit that's 4.07kB total - roughly what I'm successfully running with on ESP32. I don't have anything else to really test it on, but OTOH the current 256 byte buffer is insufficient for anything over 115200.

PR #22807, opened 2025-01-12
I2S: add command i2sloop

Staars:
## Description:

New command `i2sloop` for endless looping of  sound file, can work gapless for opus format.
Syntax equal to `i2splay`, e.g.:
`i2sloop2 brown.opus` - loops a sound file that could contain brown noise to build your personal noise machine.

The development revealed a bug (memory leak) in the framework, that was introduced some time ago. Using the stack is better anyway in the scope of Tasmota, as we already allocate memory for the task. Thus avoiding multiple allocations later in this task will additionally improve performance a tiny bit.

More small refactoring steps to reduce code duplication. 

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.1.250109


sfromis:
Noticed one detail, not specific to this improvement. That use of index 2 (seems to be for `decoder_type`) also on the existing `I2SPlay` command is not mentioned in the [docs](https://tasmota.github.io/docs/I2S-Audio/#commands).

PR #22822, opened 2025-01-15
Berry  now removes any existing rule with same pattern and same id (only if id is not nil)

s-hadinger:
## Description:

Berry, make sure that call to `tasmota.add_rule(pattern, id)` removes any pre-existing rule with same `pattern` and same `id` (only if `id` is not `nil`)

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.1.250109


sfromis:
Nice....

I was a bit surprised to see one detail, `self.remove_rule(f, id)` when the prototype is
`def remove_rule(pat, id)`

I'd have expected the call to be 
`self.remove_rule(pat, id)`
when the function is still matching on the trigger pattern, in addition to the id:
`if self._rules[i].trig.rule == pat && self._rules[i].id == id`

How can that work?

s-hadinger:
You are right, I wanted a quick fix but it doesn't work. Reverting

PR #22850, opened 2025-01-19
Remove user ca headers

GMagician:
Since local CA files are user made files add them to gitignore (as already done for user_config_override.h)

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.1.1.250109

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ data
 unpacked_fs
 unpacked_boards
 tasmota/user_config_override.h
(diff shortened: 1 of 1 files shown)


s-hadinger:
Ok for me. Thanks

GMagician:
Since 'local_ca_descriptor.h' content is constant it should also be present by default on repo. It's not included until USE_LOCAL_CERT is defined so no code increase until really required

PR #22953, opened 2025-02-05
Fix reboot with no private key

GMagician:
When USE_MQTT_AWS_IOT is defined but private key is not defined, if mqtt connection is done, system will raise an exception and reboot. On web page it seems that it's working, just slower then usual. Opening console you can see something weird. This seems to fix.

## Checklist:

### DIFF
--- a/lib/lib_ssl/bearssl-esp8266/src/ssl/ssl_ccert_single_ec.c
+++ b/lib/lib_ssl/bearssl-esp8266/src/ssl/ssl_ccert_single_ec.c
@@ -59,7 +59,7 @@ cc_choose(const br_ssl_client_certificate_class **pctx,
 	scurve = br_ssl_client_get_server_curve(cc);
 
 	if ((zc->allowed_usages & BR_KEYTYPE_KEYX) != 0
-		&& scurve == zc->sk->curve)
+		&& zc->sk && scurve == zc->sk->curve)
 	{
 		int x;
 
--- a/lib/lib_ssl/tls_mini/src/WiFiClientSecureLightBearSSL.cpp
+++ b/lib/lib_ssl/tls_mini/src/WiFiClientSecureLightBearSSL.cpp
@@ -1000,7 +1000,7 @@ bool WiFiClientSecure_light::_connectSSL(const char* hostName) {
     // ============================================================
     // Set the EC Private Key, only USE_MQTT_AWS_IOT
     // limited to P256 curve
-    br_ssl_client_set_single_ec(_sc.get(), _chain_P, 1,
+    br_ssl_client_set_single_ec(_sc.get(), _chain_P, _chain_P ? 1 : 0,
                                 _sk_ec_P, _allowed_usages,
                                 _cert_issuer_key_type, &br_ec_p256_m15, br_ecdsa_sign_asn1_get_default());
   #endif // USE_MQTT_AWS_IOT


Noschvie (approved):
Tested these changes, works perfect! Thanks!

GMagician:
Ok I suspcet I have to post a PR in https://github.com/earlephilhower/bearssl-esp8266...please let me know

lib/lib_ssl/bearssl-esp8266/src/ssl/ssl_ccert_single_ec.c
```diff
@@ -59,7 +59,7 @@ cc_choose(const br_ssl_client_certificate_class **pctx,
 	scurve = br_ssl_client_get_server_curve(cc);
 
 	if ((zc->allowed_usages & BR_KEYTYPE_KEYX) != 0
-		&& scurve == zc->sk->curve)
+		&& zc->sk && scurve == zc->sk->curve)
```
s-hadinger, line 62: You shouldn't need to change the underlying Bearssl library. Otherwise it should be reported upstream. But I doubt the problem is in the library.

lib/lib_ssl/tls_mini/src/WiFiClientSecureLightBearSSL.cpp
```diff
@@ -1000,7 +1000,7 @@ bool WiFiClientSecure_light::_connectSSL(const char* hostName) {
     // ============================================================
     // Set the EC Private Key, only USE_MQTT_AWS_IOT
     // limited to P256 curve
-    br_ssl_client_set_single_ec(_sc.get(), _chain_P, 1,
+    br_ssl_client_set_single_ec(_sc.get(), _chain_P, _chain_P ? 1 : 0,
```
s-hadinger, line 1003: This is not right. You cannot set a client ec and say it has zero instances. The test must be upwards.  Tasmota shouldn't call this function at all if no EC cert is present

s-hadinger (changes requested):
Thanks for working on this bug, but it should be in Tasmota to not call Bearssl to set a client cert if there is none. You shouldn't patch Bearssl to accept invalid calls

Noschvie:
written on Discord:
Had a look to tasmota\tasmota_xdrv_driver\xdrv_02_9_mqtt.ino / Line number 1852 / "file does not exist, do nothing"
“Doing nothing” is probably the problem; the file should be initialized with a default value.

s-hadinger:
I will look at it next week

lib/lib_ssl/bearssl-esp8266/src/ssl/ssl_ccert_single_ec.c
GMagician: The problem is you can get till here with sk == NULL. The problem may be the call below but read there other note

lib/lib_ssl/tls_mini/src/WiFiClientSecureLightBearSSL.cpp
GMagician: The probles here is that if you call once this (br_ssl_client_set_single_ec), then internal vtable is initialized `(cc->client_auth_vtable = &cc->client_auth.single_ec.vtable;` and once it is initialized all calls are done and it may reach code in library (above) once you kill your key you are in trouble

lib/lib_ssl/bearssl-esp8266/src/ssl/ssl_ccert_single_ec.c
GMagician: However this is a bad "library" behaviour, it's not wise to not validate variables

s-hadinger: Yes and no. You are not supposed to pass a null pointers anyways for your certificate

lib/lib_ssl/tls_mini/src/WiFiClientSecureLightBearSSL.cpp
s-hadinger: The problem is that you shouldn’t call this function if you don’t have an EC client certificate.

GMagician:
> I will look at it next week

I'm trying to help with this but I see something that it seems confused:
when USE_MQTT_AWS_IOT is define `br_ssl_client_set_single_ec` is always called by `WiFiClientSecure_light::_connectSSL` so it seems that client certificate is mandatory (otherwise null pointer use).
In xdrc_02_98_mqtt you clear user and password when USE_MQTT_AWS_IOT is defined so this enforce that when certificate is present no user/password is required but some lines above this you have:
```c++
#if defined(USE_MQTT_TLS) && defined(USE_MQTT_AWS_IOT)
    // don't enable MQTT for AWS IoT if Private Key or Certificate are not set
    if (Mqtt.mqtt_tls) {
      if (0 == strlen(SettingsText(SET_MQTT_PWD))) {     // we anticipate that an empty password does not make sense with TLS. This avoids failed connections
        Mqtt.allowed = false;
      }
    }
#endif
```

Comment says something already read but code does somethin different. it forces a passord to be present.

If USE_MQTT_AWS_IOT means client certificate then above code may be the point where change code and remove password check and add pk/certificate presence check. If this may be a solution I'm ready help

PR #22970, opened 2025-02-09
Utilize ESP-IDF's LCD driver for pushing pixels to RGB displays with ESP32-S3

treitmayr:
This commit is an optimization of code and possibly performance for pushing pixel data to the frame buffer of an RGB display with an ESP32-S3. It uses ESP-IDF's LCD driver for doing so and replaces multiple code paths in the previous implementation. No code for other ESP32 variants has been changed.

## Description:

**Related issue (if applicable):** None, but see discussion arendst/Tasmota#22553

## Checklist:
        N/A - Only code inside USE_ESP32_S3 sections was changed, "tasmota-display" was checked to compile fine.

### DIFF
--- a/lib/lib_display/UDisplay/uDisplay.cpp
+++ b/lib/lib_display/UDisplay/uDisplay.cpp
@@ -1225,12 +1225,19 @@ Renderer *uDisplay::Init(void) {
     _panel_config->de_gpio_num = de;
     _panel_config->pclk_gpio_num = pclk;
 
+    // assume that byte swapping of 16-bit color is done only upon request
+    // via display.ini and not by callers of pushColor()
+    // -> swap bytes by swapping GPIO numbers
+    int8_t *par_db8 = lvgl_param.swap_color ? par_dbl : par_dbh;
     for (uint32_t cnt = 0; cnt < 8; cnt ++) {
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks, let me review it first

s-hadinger:
WOW !!!
The result is spectacular. With `DisplayRotate 0`, the performance is twice as fast with a whopping 14400 pix/ms. Speed is slightly slower with rotation, but far better than the non-optimized code we had previously.

@treitmayr you are a hero!

(I'm waiting for CI to finish before merging)

treitmayr:
That's what I suspected without having measured it myself, especially as I use the display in portrait orientation (where I currently get up to 8000 pix/ms).
Thanks for merging and the complement ;-).

sfromis:
Great!

I'm also seeing refresh at boot go from 8000-9142 pix/ms up to 12800-16000 pix/ms, and time from 7-8 ms down to 4-5 ms per 64000 pixels.

Jason2866:
Not the correct place. Tasmota uDisplay driver is slow compared to implementations in other projects. For sure that's not the only piece of code which needs review.

Jason2866:
Not using double buffers and for SPI driven displays DMA mode is not used.

PR #22972, opened 2025-02-09
Allow acl in mqtt when client certificate is in use

GMagician:
Acl in mqtt needs user name if 'use_identity_as_username/use_subject_as_username' is not true. It also seems that login in mqtt doesn't accept an user with an empty password, then reenabled both

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
@@ -57,7 +57,7 @@ const char kMqttCommands[] PROGMEM = "|"  // No prefix
   D_CMND_MQTTFINGERPRINT "|"
 #endif
   D_CMND_MQTTUSER "|" D_CMND_MQTTPASSWORD "|" D_CMND_MQTTKEEPALIVE "|" D_CMND_MQTTTIMEOUT "|" D_CMND_MQTTWIFITIMEOUT "|"
-#if defined(USE_MQTT_TLS) && defined(USE_MQTT_AWS_IOT)
+#if defined(USE_MQTT_TLS) && (defined(USE_MQTT_AWS_IOT) || defined(USE_MQTT_CUSTOM_CERT))
   D_CMND_TLSKEY "|"
 #endif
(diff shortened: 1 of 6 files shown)


s-hadinger:
The problem here is that Tasmota sets non-empty user/password for MQTT. This change would immediately break any existing configuration when using Client Certificate, and not setting user/password to empty strings.

I'm afraid there should be some option or additional setting to force user/password when using client cert

tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
```diff
@@ -1190,12 +1190,9 @@ void MqttReconnect(void) {
     MqttNonTLSWarning();
   }
 #ifdef USE_MQTT_AWS_IOT
-  // re-assign private keys in case it was updated in between
+  // re-assign private key in case it was updated in between
   if (Mqtt.mqtt_tls) {
     if ((nullptr != AWS_IoT_Private_Key) && (nullptr != AWS_IoT_Client_Certificate)) {
```
s-hadinger, line 1195: As indicated in comment, this would break any existing client cert configuration where user/password have been left to default values. AWS IOT will refuse connection if you send non-empty user/password

GMagician: ok then I think it will be useful to add a new define to let use custom certificate with mosquitto acl compatibility
What do you think about USE_MQTT_CUSTOM_CERT?

GMagician: OT: I saw that web doesn't accept ';' char in password. Maybe some validation is required?

GMagician:
> The problem here is that Tasmota sets non-empty user/password for MQTT. This change would immediately break any existing configuration when using Client Certificate, and not setting user/password to empty strings.
> 
> I'm afraid there should be some option or additional setting to force user/password when using client cert

I missed this....sorry. maybe may I recycle INCLUDE_LOCAL_CERT?

GMagician:
Replaced by #22998

PR #22998, opened 2025-02-12
Allow acl in mqtt when client certificate is in use

GMagician:
I messed up my previous PR, here is a copy with a new commit.

This is a change to let client certificate to be used with mosquitto with different use cases

## Checklist:

### DIFF
--- a/BUILDS.md
+++ b/BUILDS.md
@@ -14,8 +14,8 @@ Note: the `minimal` variant is not listed as it shouldn't be used outside of the
 | USE_HOME_ASSISTANT        | -     | - / - | -     | -     | -     | -     |
 | USE_TASMOTA_DISCOVERY     | x     | x / x | x     | x     | x     | x     |
 | USE_MQTT_TLS\*            | -     | - / x | -     | -     | -     | -     |
+| USE_MQTT_CLIENT_CERT      | -     | - / - | -     | -     | -     | -     |
 | USE_MQTT_AWS_IOT          | -     | - / - | -     | -     | -     | -     |
-| USE_MQTT_CUSTOM_CERT      | -     | - / - | -     | -     | -     | -     |
 | USE_4K_RSA                | -     | - / - | -     | -     | -     | -     |
 | USE_TELEGRAM              | -     | - / - | -     | -     | -     | -     |
 | USE_KNX                   | -     | - / x | x     | -     | -     | -     |
--- a/BUILDS.md
+++ b/BUILDS.md
@@ -15,6 +15,7 @@ Note: the `minimal` variant is not listed as it shouldn't be used outside of the
 | USE_TASMOTA_DISCOVERY     | x     | x / x | x     | x     | x     | x     |
 | USE_MQTT_TLS\*            | -     | - / x | -     | -     | -     | -     |
 | USE_MQTT_AWS_IOT          | -     | - / - | -     | -     | -     | -     |
+| USE_MQTT_CUSTOM_CERT      | -     | - / - | -     | -     | -     | -     |
 | USE_4K_RSA                | -     | - / - | -     | -     | -     | -     |
 | USE_TELEGRAM              | -     | - / - | -     | -     | -     | -     |
 | USE_KNX                   | -     | - / x | x     | -     | -     | -     |
--- a/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
+++ b/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
@@ -44,7 +44,7 @@ uint32_t *stack_thunk_light_save = NULL;  /* Saved A1 while in BearSSL */
 uint32_t stack_thunk_light_refcnt = 0;
 
 //#define _stackSize (5600/4)
-#if defined(USE_MQTT_AWS_IOT) || defined(USE_MQTT_AWS_IOT_LIGHT) || defined(USE_MQTT_AZURE_IOT)
+#if defined(USE_MQTT_AWS_IOT) || defined(USE_MQTT_AWS_IOT_LIGHT) || defined(USE_MQTT_AZURE_IOT) || defined(USE_MQTT_CUSTOM_CERT)
   #define _stackSize (5300/4)   // using a light version of bearssl we can save 300 bytes
 #else
   #define _stackSize (4800/4)   // no private key, we can reduce a little, max observed 4300
--- a/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
+++ b/lib/lib_ssl/tls_mini/src/StackThunk_light.cpp
@@ -44,7 +44,7 @@ uint32_t *stack_thunk_light_save = NULL;  /* Saved A1 while in BearSSL */
 uint32_t stack_thunk_light_refcnt = 0;
 
 //#define _stackSize (5600/4)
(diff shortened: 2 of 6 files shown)


Noschvie:
Hello
using these settings and a "self-signed" certificate,

```
// 2.2. Advanced: Using Full certificate validation
#ifndef USE_MQTT_TLS
#define USE_MQTT_TLS
#define USE_MQTT_TLS_CA_CERT              // Force full CA validation instead of fingerprints, slower, but simpler to use.
#define USE_MQTT_AWS_IOT                  // This will include LetsEncrypt CA, as well as our CA, in tasmota_ca.ino for verifying server certificates
#define USE_MQTT_TLS_FORCE_EC_CIPHER      // Force Elliptic Curve cipher (higher security) required by some servers (automatically enabled with USE_MQTT_AWS_IOT)
#define INCLUDE_LOCAL_CERT
#endif
```

With this PR I have to add (or replace `USE_MQTT_AWS_IOT`)

`#define USE_MQTT_CLIENT_CERT
`

isn't it? Thanks

s-hadinger:
Yes that's the plan

Noschvie:
Ok, now I replaced USE_MQTT_AWS_IOT with USE_MQTT_CLIENT_CERT and it works as before, great!

ps: the next challenge is to get and use an LE certificate...

GMagician:
is USE_MQTT_TLS_FORCE_EC_CIPHER still used? I can't find it in source code...

s-hadinger:
No it's not used anymore. Tasmota supports only ECDHE with P256 curve. It used to support RSA but it's removed. 

Don't confuse with server certificates that need to be RSA signed (2048 or 4096 bits)

arendst:
What is the (breaking) impact by this PR of changing the define USE_MQTT_AWS_IOT to USE_MQTT_CLIENT_CERT????

To me it seems USE_MQTT_AWS_IOT doesn't do what it used to do. In fact it is only used in `xdrv_02_9_mqtt.ino` WITHIN define USE_MQTT_CLIENT_CERT. So this seems a breaking change to me instead of deprecated; It just doesn't work with the old defines!!

Any comment much appreciated.

s-hadinger:
Maybe I overlooked, but there is:

```
#if defined(USE_MQTT_AWS_IOT) && !defined(USE_MQTT_CLIENT_CERT)
  #define USE_MQTT_CLIENT_CERT                   // USE_MQTT_AWS_IOT requires USE_MQTT_CLIENT_CERT
#endif
```

that shouls ensure that `USE_MQTT_AWS_IOT` automatically enables `USE_MQTT_CLIENT_CERT`

arendst:
Ah, I missed that one while searching for USE_MQTT_AWS_IOT.

OK. Thx. Will update changelogs accordingly.

Noschvie:
The problem with the boot loop if the TLS key (TLSKey1, TLSKey2) has not yet been supplied still exists, right?

GMagician:
I posted a PR for that but required a change in libs and has been discharged. Probably can simply stop the reconnection or disabled mqtt. I may create a PR but I need some guide because I don't want it is rejected again

GMagician:
I posted a PR for that but required a change in libs and has been discharged. Probably can simply stop the reconnection or disabled mqtt. I may create a PR but I need some guide because I don't want it is rejected again

Noschvie:
Had a look to tasmota\tasmota_xdrv_driver\xdrv_02_9_mqtt.ino / Line number 1852 / "file does not exist, do nothing"
“Doing nothing” is probably the problem; the file should be initialized with a default value.

s-hadinger:
Thanks, this got indeed out of sight.

The conservative option would be to consider that when `USE_MQTT_CLIENT_CERT` is defined, no MQTT connection is started unless the certificate is provided. This would prevent any connection using password with such compile option.

In principle I don't like it because it imposes a behavior at compile time. I would prefer to have a "certification-capable" Tasmota version that can do both; although the pre-compiled would do only password authentication.

I'm afraid that to do this in a clean way, there is no other option than adding a new `SetOption` to force client cert. Am I missing something?

GMagician:
By what I recall about loop problem, if you ask for use certificate and don't give a certificate system will continue to reboot due to a null pointer use. Solution I proposed with other PR was to protect the null pointer use and let system works even with no certificate, but that required a lib change...

GMagician:
By what I recall about loop problem, if you ask for use certificate and don't give a certificate system will continue to reboot due to a null pointer use. Solution I proposed with other PR was to protect the null pointer use and let system works even with no certificate, but that required a lib change...

Noschvie:
There may be a workaround for the "required library change" as part of xdrv_02_9_mqtt.ino such as initializing the TLSKey1 and TLSKey2 fields?

GMagician:
AFAIK nothing has been done to solve it. You have to enable tls custom client certs, enable mqtt
#define MQTT_TLS_ENABLED        true
#define MQTT_PORT               8883
#define USE_MQTT_TLS
#define USE_MQTT_CLIENT_CERT
#define OMIT_LETS_ENCRYPT_CERT
#define OMIT_AWS_CERT
#define INCLUDE_LOCAL_CERT

You may miss it is rebooting because it seems to work, but response slooooowly

GMagician:
AFAIK nothing has been done to solve it. You have to enable tls custom client certs, enable mqtt
#define MQTT_TLS_ENABLED        true
#define MQTT_PORT               8883
#define USE_MQTT_TLS
#define USE_MQTT_CLIENT_CERT
#define OMIT_LETS_ENCRYPT_CERT
#define OMIT_AWS_CERT
#define INCLUDE_LOCAL_CERT

You may miss it is rebooting because it seems to work, but response slooooowly

s-hadinger:
I still don't see any crash. I couldn't enable `#define INCLUDE_LOCAL_CERT` since I don't have any additional CA.

Noschvie:
It seems to be fixed, I tested it with 14.5.0.2 and no further boot loops occur. Great!

```
// 2.2. Advanced: Using Full certificate validation
#ifndef USE_MQTT_TLS
#define USE_MQTT_TLS
#define USE_MQTT_TLS_CA_CERT                   // Force full CA validation instead of fingerprints, slower, but simpler to use.  (+2.2k code, +1.9k mem during connection handshake)
#define USE_MQTT_AWS_IOT                       // This will include LetsEncrypt CA, as well as our CA, in tasmota_ca.ino for verifying server certificates
#define USE_MQTT_TLS_FORCE_EC_CIPHER           // Force Elliptic Curve cipher (higher security) required by some servers (automatically enabled with USE_MQTT_AWS_IOT) (+11.4k code, +0.4k mem)
#define INCLUDE_LOCAL_CERT
#endif
```

GMagician:
> I still don't see any crash. I couldn't enable `#define INCLUDE_LOCAL_CERT` since I don't have any additional CA.

I'll try to replicate but pay attention that you don't get a "crash" if you don't look at console it just seems a slowed down system

GMagician:
I confirm that reboot issue is stil present. To see it you have to setup a working tasmota/mosquitto system. Once you have it you have to delete tlskey from filesystem and reboot.
Looking at console you will see a continuos increase of number in red circle.
immagine
But everything seems to work, just a slow response.

Tasmota setup requires TLS active, ca certificate (not sure if mandatory), client certificate, no user/password (certificate CN is used by mosquitto but I suspect it doesn't matter if both user and passwords are used).

PR #23020, opened 2025-02-17
Add support for LVGL transitions

treitmayr:
## Description:

This PR enables support for the LVGL transition feature which creates continuous adjustments from one style to another to be performed over a given time duration.

For this the object type `lv.style_transition_dsc` was enabled, and it can be fully initialized via its constructor (which in turn calls the native LVGL function `lv_style_transition_dsc()`). The PR currently only includes the actual code changes as well as an updated LVGL berry example, but does not include the changed generated files. I can add the generated files as soon as all kinks and quirks found in the code during the review were straightened out.

One thing to observe is that passing a `style_transition_dsc` object to `lv.style.set_transition()` does not add a reference to the object, therefore it also has to be kept in some variable to prevent its garbage collection. I am not sure how this is usually solved.

Also please check for any issues in the `lv_berry.c` code because I implemented this glue code mostly be looking up examples in other Tasmota source files. In a stripped down version we could leave out the function `list_to_zbytes()` and require the user to create an `lv.style_prop_arr` object as the first parameter for initialization.

As for the example, it works very well with just copy/pasting it to the Berry console. It also shows the two different ways of creating the `lv.style_transition_dsc` object.

## Checklist:

### DIFF
--- a/lib/libesp32_lvgl/lv_binding_berry/src/embedded/lvgl_extra.be
+++ b/lib/libesp32_lvgl/lv_binding_berry/src/embedded/lvgl_extra.be
@@ -112,10 +112,10 @@ end
 class lv_style_prop_arr : bytes
   def init(l)
     if type(l) != 'instance' || !isinstance(l, list)  raise "value_error", "argument must be a list" end
-    super(self).init(size(l) * 4)
+    super(self).init(size(l))
 
     for e: l
-      self.add(int(e), 4)
+      self.add(int(e))
     end
   end
 end
--- a/lib/libesp32_lvgl/lv_binding_berry/src/lv_berry.c
+++ b/lib/libesp32_lvgl/lv_binding_berry/src/lv_berry.c
@@ -259,6 +259,143 @@ int lv0_lvobj__void_call(bvm *vm, fn_lvobj__void func) {
 }
 
 /*********************************************************************************************\
+ * Support for lv_style_transition_dsc `init()`
+ *
+ * Either encapsulate the pointer passed as `comptr` as arg1
+ * Or allocate a new empty style structure in memory. In this case, it is never freed.
+\*********************************************************************************************/
+
+static uint8_t *list_to_zbytes(bvm *vm, int index);
+
+int lv_be_style_transition_dsc_init(bvm *vm) {
+  int argc = be_top(vm);
+  lv_style_transition_dsc_t *trans = NULL;
+  // default values
+  lv_style_prop_t *props = NULL;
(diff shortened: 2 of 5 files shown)


s-hadinger:
It will need to wait for 1 day or 2. Version 14.5.0 is getting out, I will merge right after.

treitmayr:
Thank you for the quick review! No hurry, please - good to see a new release coming.

At some point maybe you could let me know about your opinion regarding the `list_to_zbytes()` code path. It adds convenience at the cost of more code. And although it took me quite a while to finish this function, it may not be worth it after all when it is only used in the (probably rare) case someone uses transitions.
Before the merge please also let me know if I shall provide the updated generated files or you do that yourself (not sure how that usually is done).

lib/libesp32_lvgl/lv_binding_berry/src/embedded/lvgl_ctypes.py
```diff
@@ -35,6 +35,7 @@
 lv_dir = ct.u8
 lv_coord_t = ct.i32
 lv_value_precise = ct.i32
+lv_style_prop_t = ct.u8
```
s-hadinger, line 38: This is not actually useful

treitmayr: Removed it, you are right, it is not of any use.

treitmayr:
I just realized that there is a second example for LVGL transitions ("gummy button") which I now adjusted as well.

lib/libesp32_lvgl/lv_binding_berry/src/embedded/lvgl_extra.be
```diff
@@ -112,10 +112,10 @@ end
 class lv_style_prop_arr : bytes
   def init(l)
     if type(l) != 'instance' || !isinstance(l, list)  raise "value_error", "argument must be a list" end
-    super(self).init(size(l) * 4)
+    super(self).init(size(l))
```
s-hadinger, line 115: This one is a bug indeed

s-hadinger:
Overall excellent job. My main concern is that class `lv_style_transition_dsc` already exists as a bytes buffer. Hence creating the C class instead masks the access to the internal fields.

Ideally I would like to call your constructor on top of the bytes() mapping.

I'm wondering if this could be even more simplified by doing a super class in Berry just to override the constructor.

Stay tuned

s-hadinger:
Bump. It will review this

s-hadinger:
Still need more time to review

PR #23028, opened 2025-02-19
Add GNSS & NTRIP drivers

catalinsanda:
## Description:

Add NMEA GNSS driver over serial and implement NTRIP support (Caster & Server) for high precision positioning.

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.1.1.250203


catalinsanda:
First time contributing to a C/C++ project, moreover I haven't done any serious work in those languages in the last 20 years, so please be gentle :).

This PR allows Tasmota to integrate with high prevision GNSS receivers like the UM980 (on which I tested), to fetch positioning data and also stream RTCM messages to casters and/or clients allowing them centimeter level precision.

There are a few ways one could achieve this using existing software, but all seemed too complicated. So, because I'm a big user of Tasmota, I set about to create these drivers and simplify the user experience. 
I asked the community what they think of the idea [here](https://github.com/arendst/Tasmota/discussions/21688), and even though there wasn't much engagement I wasn't deterred. 

Let me know what additional info I can provide.

tasmota/include/tasmota_template.h
```diff
@@ -230,6 +230,9 @@ enum UserSelectablePins {
   GPIO_TM1640CLK, GPIO_TM1640DIN,       // TM1640 (16 x seven-segment LED controler)
   GPIO_TWAI_TX, GPIO_TWAI_RX, GPIO_TWAI_BO, GPIO_TWAI_CLK,  // ESP32 TWAI serial interface
   GPIO_C8_CO2_5K_TX, GPIO_C8_CO2_5K_RX, // C8-CO2-5K CO2 Sensor
+#ifdef ESP32
```
s-hadinger, line 233: Is this feature only available on ESP32?

tasmota/language/en_GB.h
```diff
@@ -531,6 +531,10 @@
 #define D_SENSOR_DALI_TX                  "Dali TX"
 #define D_CONFIGURE_DALI                  "DALI"
 
+// xdrv_91_esp32_gnss.ino
```
s-hadinger, line 534: You need to copy this section in all languages files (use search/replace multi-line) to make it at once.

tasmota/language/en_GB.h
```diff
@@ -1315,4 +1319,16 @@
 #define D_CHARGING                        "Charging"
 #define D_CAPACITY                        "Capacity"
 
+// xdrv_93_esp32_gnss_ntrip.ino
```
s-hadinger, line 1322: You need to copy this section in all languages files (use search/replace multi-line) to make it at once.

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+#endif
+
+#ifdef USE_NTRIP
+extern void ProcessRTCMMessage(const uint8_t *data, size_t length);
+#endif
+
+/*********************************************************************/
+/*                          DEFINES                                  */
+/*********************************************************************/
+#define XDRV_92 92
```
s-hadinger, line 42: @arendst Should it be driver 92 or any other number?

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+#define D_CMND_GNSS "GNSS"
```
s-hadinger, line 43: This define should be moved to "i18n.h"

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+#define RTC_UPDATE_INTERVAL (30 * 60) // Every 30 minutes
```
s-hadinger, line 44: In general, avoid general names for defines. Please use a unique prefix. Arduino merges all "*.ino" files into one huge C++ file, and collisions are likely

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+#define INVALIDATE_AFTER 30           // 30 seconds
```
s-hadinger, line 45: Same for all others, please use a prefix

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+  uint32_t lastDataTime;
+  bool isActive;
+};
+#endif
+
+/*********************************************************************/
+/*                   CLI COMMANDS                                    */
+/*********************************************************************/
+
+void CmndGNSSBaudrate(void);
```
s-hadinger, line 96: Due to the magic of Arduino, you are not strictly required to make these early declarations. But they don't hurt either.

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+  char date[32];
+  char last_valid_time[32];
+  NMEACounters nmea_counters;
+  uint32_t total_nmea_sentences;
+#ifdef USE_AUX_WEBSERVER
+  uint32_t stream_clients;
+#endif
+} GNSSData;
+
+static TinyGPSPlus gps;
```
s-hadinger, line 136: "static" modifier has almost no effect here, due to the fact that Arduino merges anyways all .INO files into one giant C++ file. Hence use (again) a prefix to avoid name collisions

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+static GNSSParser gnssParser;
+static HardwareSerial *gnssSerial = nullptr;
+
+static uint32_t last_rtc_sync = 0;
```
s-hadinger, line 140: It is preferred to declare a unique 'struct' for all globals for each driver. It makes compiled code more compact and avoids name collisions

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+/*********************************************************************/
+#include <queue>
+#include <array>
+#include <cstring>
+#include <TinyGPSPlus.h>
+#include <GNSSParser.h>
+#include <vector>
+#include <algorithm>
+
+#ifdef USE_AUX_WEBSERVER
```
s-hadinger, line 31: Please add a line for USE_AUX_WEBSERVER in "my_user_config.h", with a description of the effect

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+#include <ESPAsyncWebServer.h>
+#endif
+
+#ifdef USE_NTRIP
```
s-hadinger, line 35: Same, you need to add a description in "my_user_config.h"

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+      }
+    }
+  }
+};
+#endif // USE_AUX_WEBSERVER
+
+/*********************************************************************/
+/*                  SERIAL CONFIG STRING MAPPING                     */
+/*********************************************************************/
+const SerialConfigMap kSerialConfigs[] = {
```
s-hadinger, line 242: Avoid such big maps, they eat up flash space. Any alternative you can think of?

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+void rtcmInitializeCasterEndpoint(AsyncWebServer *web_server);
+#endif
+static void initAsyncServer();
+static void cleanupInactiveClients();
+static void broadcastSerialData(const uint8_t *buffer, size_t length);
+
+/*********************************************************************/
+/*                   ASYNC WEBSERVER RESPONSE CLASS                  */
+/*********************************************************************/
+class GNSSAsyncStreamResponse : public AsyncWebServerResponse
```
s-hadinger, line 182: Hmmm. I'm not familiar with "AsyncWebServerResponse". Where does this class come from?

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+
+    char filename[20];
+    snprintf_P(filename, sizeof(filename), PSTR(TASM_FILE_DRIVER), XDRV_92);
+
+    AddLog(LOG_LEVEL_DEBUG, PSTR("GNSS: Saving settings to file %s"), filename);
+
+#ifdef USE_UFILESYS
+    if (!TfsSaveFile(filename, (const uint8_t *)&GNSSSettings, sizeof(GNSSSettings)))
+    {
+      AddLog(LOG_LEVEL_INFO, D_ERROR_FILE_NOT_FOUND);
```
s-hadinger, line 313: In general, "AddLog" needs to add a 3 letters prefix indicating which module generated the log. Here probably "GNS: "

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+    }
+#else
+    AddLog(LOG_LEVEL_INFO, D_ERROR_FILESYSTEM_NOT_READY);
```
s-hadinger, line 316: Same

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+    GNSSSettingsLoad();
+  }
+
+  if (gnssSerial)
+  {
+    delete gnssSerial;
+    gnssSerial = nullptr;
+  }
+
+  gnssSerial = new HardwareSerial(1);
```
s-hadinger, line 367: Please use TasmotaSerial class instead. It abstract all the details of serial management. It may as well make it working on ESP8266.

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+    ascii_line[sizeof(ascii_line) - 1] = 0;
+
+    for (size_t i = 0; i < line_length; i++)
+    {
+      uint8_t byte = buffer[offset + i];
+      snprintf(&hex_line[i * 3], 4, "%02X ", byte);
+      ascii_line[i] = (byte >= 32 && byte <= 126) ? byte : '.';
+    }
+
+    AddLog(LOG_LEVEL_DEBUG_MORE, PSTR("GNSS Raw [%04X]: %-48s  |%s|"),
```
s-hadinger, line 547: Add prefix in log

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
```diff
@@ -0,0 +1,876 @@
...
+/*********************************************************************/
+/*             TASMOTA WEBSERVER DISPLAY (USE_WEBSERVER)             */
+/*********************************************************************/
+#ifdef USE_WEBSERVER
+void GNSSShow(bool json)
+{
+  if (json)
+  {
+    ResponseAppend_P(
+        PSTR(",\"GNSS\":{\"Valid\":%d,\"DateTime\":\"%s %s\",\"Lat\":%f,\"Lon\":%f,\"Alt\":%f,"
```
s-hadinger, line 561: You can make it work for ESP8266 as well, but need to use the special extensions

https://tasmota.github.io/docs/Sensor-API/#extension-to-vsnprintf 

It may also be useful to limit and control the number digits in the output

tasmota/tasmota_xdrv_driver/xdrv_93_esp32_gnss_ntrip.ino
```diff
@@ -0,0 +1,1643 @@
...
+#include <algorithm>
+#include <vector>
+#include <base64.hpp>
+#include <AsyncTCP.h>
+#ifdef USE_AUX_WEBSERVER
+#include <ESPAsyncWebServer.h>
+#endif
+
+#define XDRV_93 93
+#define D_CMND_NTRIP "NTRIP"
```
s-hadinger, line 35: Move to "i18n.h"

tasmota/tasmota_xdrv_driver/xdrv_93_esp32_gnss_ntrip.ino
```diff
@@ -0,0 +1,1643 @@
...
+/*********************************************************************/
+#include <algorithm>
+#include <vector>
+#include <base64.hpp>
+#include <AsyncTCP.h>
+#ifdef USE_AUX_WEBSERVER
+#include <ESPAsyncWebServer.h>
+#endif
+
+#define XDRV_93 93
```
s-hadinger, line 34: Should this be a separate driver to 92? Can it works on its own? Or would it make sense to merge two drivers?

tasmota/tasmota_xdrv_driver/xdrv_93_esp32_gnss_ntrip.ino
```diff
@@ -0,0 +1,1643 @@
...
+#define D_CMND_NTRIP "NTRIP"
+#define MESSAGE_STATISTICS_COUNT 10
+#define NTRIP_CLIENTS 2
+
+/*********************************************************************/
+/*                           GLOBAL DEFINES                          */
+/*********************************************************************/
+
+#define CLIENT_TIMEOUT 300000
```
s-hadinger, line 43: Overall same comments as before, use a prefix to avoid name collisions

s-hadinger (changes requested):
Thank you for your valuable addition. I made some comments mainly linked to the way Arduino merges C++ files. So "static" does not work, and "#define" need to use prefix to avoid name collisions

Jason2866:
Please move the gnss lib in folder `lib_div`

tasmota/include/tasmota_template.h
catalinsanda: Haven't tried, but I'm not sure that an ESP8266 has the necessary resources to support an async web server and the speed to stream the serial data without falling behind.

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
catalinsanda: I'll look at how this is handled elsewhere in the codebase, maybe I can find something to piggyback on.

tasmota/tasmota_xdrv_driver/xdrv_92_esp32_gnss.ino
catalinsanda: It's from here: https://github.com/ESP32Async/ESPAsyncWebServer/blob/main/src/ESPAsyncWebServer.h

tasmota/tasmota_xdrv_driver/xdrv_93_esp32_gnss_ntrip.ino
catalinsanda: Yes, will merge the two drivers. 
The initial thinking was that the NTRIP data could have been fed by a different driver, but that's not going to happen.

Jason2866:
Adding an additional Webserver (AsyncWebServer) in a driver is a no go. Please change the code to work with the Tasmota Webserver only.

catalinsanda:
That's how I initially started, unfortunately (unless I got something wrong) Tasmota is using https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WebServer/src/ESP8266WebServer.h, which is a **one simultaneous connection** web server. 

This driver needs to support long lived (hours, days) persistent connections, for two use-case: downloading raw GNSS Ephemeris data from the receiver, which are then sent to a PPP service for precise positioning and also to function as a NTRIP Caster. 

In the initial implementation I tried using the built in driver (after some tweaking of its code), but it only supported one connection and when the client was connected the UI would freeze.

Jason2866:
Understand the problematic, if there is no other way to solve this, maybe not an enhancement  that fits. Honestly idk if there is a chance for a merge.

catalinsanda:
What's the rationale behind it? It's not a driver which will be part of the main binaries, and whoever enables it will know from the start that it opens a new server on 8080.

Jason2866:
Tasmota design choices. Not adding drivers where complete new concepts are introduced and furthermore not useable for others.

s-hadinger:
I would strongly suggest to move the Web part to Berry. I have implemented `webserver_async` which offers Async Web server and should suit your needs. It has the benefit of not tweaking with FreeRTOS tasks and does not impact the stability of Tasmota. I can provide support if you choose this path.

s-hadinger:
I still don't undestand how `#include <ESPAsyncWebServer.h>` would actually work, since the library is not part of Tasmota.

catalinsanda:
Thanks for your comments guys.

@Jason2866 - The line seems very arbitrary, what's the difference between using a library that starts a FTP server (which is part of Tasmota) and one which starts a Web Server (because the current one has limitations for this use-case)? "Tasmota design choices." sounds a lot like "Because I f... say so", this seems very unprofessional, but anyway, your house your rules and I can respect that. If this doesn't get merged I'll just keep a fork and publish my own binaries for whoever is interested. 

@s-hadinger - Don't think that porting this on top of Berry will provide any benefit to the user, and that's what I'm after, it will just bypass this faux restriction of not adding another WebServer. I was hoping that his driver is enabled by adding a bunch of defines in `user_config_override.h`, so the users are aware of what they're doing, but the more I think of it the more it makes more sense to build my own binaries, because most users don't know or can't be bothered with building their own custom images. 
Regarding the header files, you're right, I installed that library locally using `pio` and haven't added it to the lib folder. My bad!

So, to conclude, if this is a no go, please feel free to close this PR.

PR #23123, opened 2025-03-08
Let Tasmota compile with USE_BERRY but without USE_WEBCLIENT

GMagician:
Despite comments say you can disable USE_WEBCLIENT when you habe USE_BERRY, if you do that you get a compile error because 'StreamBeBytesWriter' is not defined.

## Checklist:


s-hadinger:
Sorry but this looks half baked. If your intent is to avoid any outgoing request, then becloader needs to be disabled as well.  
I'm not even sure `USE_WEB_CLIENT` makes sense anymore, or it should be reconsidered more globally. I don't see the value so unless you're willing to do this global change, `USE_WEB_CLIENT` option will be removed (I.e. Always included)

PR #23124, opened 2025-03-08
Remove misleading comment in config

GMagician:
As per title

## Checklist:


s-hadinger:
Sorry again but it is not the right way to do it. If we conclude that this option is not optional anymore, then it should be completely removed. Not just a change in comments. 

I can do this later if you prefer

GMagician:
@s-hadinger if may be of some help I may add a PR where I remove all reference of USE_WEBCLIENT in code (and also in header file)

PR #23177, opened 2025-03-22
remove malformed clutter from webcam status0 JSON response

DavidPletcher:
A regression was added in commit 01154e949, which prepends clutter to the status0 JSON status message.

Example:
```
$ curl 'http://webcam/cm?cmnd=Status0' -u admin:password
{s}Webcam Frame rate{m}0 FPS{e}{"Status":{"Module":0,"DeviceName":"... (edited for brevity)
```

The response is not properly formatted JSON and breaks client software.

The problem occurs because a new case statement was added for the purpose of injecting webcam stats into the JSON status message, but a break statement is missing and execution falls through to the following case, which prepends garbage to the output buffer.

With this one-line fix in place, the output is properly formatted:
```
$ curl 'http://webcam/cm?cmnd=Status0' -u admin:password
{"Status":{"Module":0,"DeviceName":"... (edited for brevity)
```

Note that the prior case for FUNC_EVERY_SECOND is also missing a break statement. That looks wrong to me but it's unrelated to fixing this issue and I'm inclined to punt that concern to others who are more familiar with this code.

## Description:

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_81_esp32_webcam.ino
@@ -1577,6 +1577,7 @@ bool Xdrv81(uint32_t function) {
       WcUpdateStats();
     case FUNC_JSON_APPEND:
       WcSensorStats();
+      break;
     case FUNC_WEB_SENSOR:
       WcStatsShow();
       break;


arendst:
Thx. You're right. I'll merge this and chk your other suggestion.

DavidPletcher:
Thank you!

Tasmota is great, thanks for the awesomeness!

PR #23181, opened 2025-03-22
HASPmota autostart when `pages.jsonl` exists

s-hadinger:
## Description:

HASPmota changes:
- HASPmota starts automatically if a file named `pages.jsonl` exists. It is possible to override it if you start haspmota in `autoexec.be`
- Added `be_post_autoxec_code` as a string literal containing code executed after `autoexec.be` has completed
- `haspmota` has a new attribute `started`, so you can call `haspmota.start()` multiple times, although it is initialized only once (subsequent calls are ignored).

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - Matter prepare for ICD cluster (#23158)
 - Berry `re.dump()` (#23162)
 - Telnet server using command `Telnet <0|1|port>[,<IP filter>]`
+- HASPmota autostart when `pages.jsonl` exists
 
 ### Breaking Changed
 - Berry remove `Leds.create_matrix` from the standard library waiting for reimplementation (#23114)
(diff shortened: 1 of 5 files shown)


sfromis:
Arguably, starting HASPmota unconditionally when `pages.jsonl` exists may be a bit of an "inconvenient surprise" for users taking advantage of current flexibility for some ways of initializing, like when deliberately delaying `import haspmota haspmota.start()` until having data sources (like from MQTT subscriptions) available, just commenting out the start call for purposes of debugging, work-in-progress or ... whatever.

Of course, this can be mitigated by renaming to something `my.jsonl`, and then use the **undocumented** parameters `dark` and `templ_name` on an explicit `haspmota.start`. I'd not expect many to have noticed these in the source code, and fewer to use them.

Not a big deal, just a "devils advocate" observation from the sideline :wink:

s-hadinger:
Good remarks, but I thought on the contrary that the barrier for entry for HASPmota was too high in the very common use case.

For users that already initalize HASPmota in autoexec.be, it does not change anything because the autorun is done after autoexec.be.

I also need to change the documentation accordingly.

PR #23195, opened 2025-03-27
Use normal gamma table for fading

Qowy:
## Description:

Fading uses a modified gamma table that results in much higher values for low brightnesses.
According to @s-hadinger on Discord this was done to improve the appearance of turning on lights so that they do not take too long.
Im a not sure why that would be needed.
If I want my lights to be at the specified brightness faster I reduce "speed" 

More importantly, the existing code breaks (or hinders) DIMMER ! since it calculates wrong dimming values when stopping the fade. However this can be fixed separately. (https://github.com/arendst/Tasmota/pull/23196)

I also have a feeling this is why I could never get multiple DIMMER + look smooth for the lower half of the Brightness %.

Question is, if this should maybe be implemented as a "setoption"?

I am still trying to wrap my head around the visual implications of the current behavior.

I think it means that if doing (DIMMER=DIMMER+1) from 1 to 50 looks visually linear
Fading from 1 to 50 in the same amount of time looks different. And that is not what is intended is it?

I think I noticed that.

So I guess we are back to square one. Why was this needed in the first place?

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_04_light.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_04_light.ino
@@ -2047,14 +2047,14 @@ int32_t ChannelWhite_when_PWMCT(void) {
 // Calculate the Gamma correction, if any, for fading, using the fast Gamma curve (10 bits in+out)
 uint16_t fadeGamma(uint32_t channel, uint16_t v) {
   if (isChannelGammaCorrected(channel)) {
-    return ledGammaFast(v);
+    return ledGamma10_10(v);
   } else {
     return v;
   }
 }
 uint16_t fadeGammaReverse(uint32_t channel, uint16_t vg) {
   if (isChannelGammaCorrected(channel)) {
-    return leddGammaReverseFast(vg);
+    return ledGammaReverse(vg);
   } else {
     return vg;
   }


stefanbode:
Interesting post, because I observed the same strange behavior on the low light. I have a ambient lightning in the kitchen that is most times in the low light area and behaves a bit rocky. I will create a version with your two changes and give it a try if it looks better. Thx

Qowy:
> Interesting post, because I observed the same strange behavior on the low light. I have a ambient lightning in the kitchen that is most times in the low light area and behaves a bit rocky. I will create a version with your two changes and give it a try if it looks better. Thx

Make sure to that the 3rd change is already in dev branch https://github.com/arendst/Tasmota/pull/23194

The original discussion is this: https://github.com/arendst/Tasmota/discussions/23105

s-hadinger:
I did some testing with WS2812 LED (8 bits) and `Fade 1` with `Sleep 10` (5 seconds). I'm still not convinced by the result. When powering down, the light almost stutter before turning off. Same problem when turning on, it looks almost like it freezes for half a second before starting again increasing light.

Mabye the curve of GammaFast is too steep, but I think we need to cheat a little.

s-hadinger:
Close in favor of #23230

Thank you for your contribution

PR #23197, opened 2025-03-27
fixes for ina226 driver

protectivedad:
## Description:

adjust current/voltage timing to be consistent
fix missing comma in json display

I discuss the timing issue in [discussion](https://github.com/arendst/Tasmota/discussions/23189). The short version is that the current INA226 config has the driver getting only the last 2.352 ms of voltage and 32.976 ms of current data for every second. The time spans don't match and most of the data from the last second is ignored.

## Checklist:
  - [X] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [X] The code change is tested and works with Tasmota core ESP32 V.3.1.3.250302

### DIFF
--- a/tasmota/tasmota_xsns_sensor/xsns_54_ina226.ino
+++ b/tasmota/tasmota_xsns_sensor/xsns_54_ina226.ino
@@ -78,7 +78,11 @@
 
 #define INA226_REG_CONFIG                       (0x00)    // Config register
 #define INA226_RES_CONFIG                       (0x4127)  // Config register at reset
-#define INA226_DEF_CONFIG                       (0x42FF)  // Our default configuration
+// 0x4B27 = 0100 101 (256 - averages) 100 (1.1 ms - vbusct) 100 (1.1 ms - vshct) 111 (continuous - mode)
+// driver will sample 256 * (1.1 ms + 1.1 ms) = 563.2 ms of data per second
+#ifndef INA226_DEF_CONFIG
+#define INA226_DEF_CONFIG                       (0x4B27)  // Our default configuration
+#endif
 #define INA226_CONFIG_RESET                     (0x8000)  // Config register reset bit
 
 #define INA226_REG_SHUNTVOLTAGE                 (0x01)
@@ -460,7 +464,7 @@ bool Ina226CommandSensor()
     // Full scale vbus is volatile (saved in RAM)
     dtostrfd((Ina226Info[device].vbus_lsb*32768.0), 2, fs_vbus_str);
     // Send json response
-    Response_P(PSTR("{\"Sensor54-device-settings-%d\":{\"SHUNT_R\":%s,\"FS_I\":%s\"FS_V\":%s}}"),
+    Response_P(PSTR("{\"Sensor54-device-settings-%d\":{\"SHUNT_R\":%s,\"FS_I\":%s,\"FS_V\":%s}}"),
       device + 1, shunt_r_str, fs_i_str, fs_vbus_str);
   }
 


tasmota/tasmota_xsns_sensor/xsns_54_ina226.ino
```diff
@@ -78,7 +78,7 @@
 
 #define INA226_REG_CONFIG                       (0x00)    // Config register
 #define INA226_RES_CONFIG                       (0x4127)  // Config register at reset
-#define INA226_DEF_CONFIG                       (0x42FF)  // Our default configuration
+#define INA226_DEF_CONFIG                       (0x47FF)  // Our default configuration
```
barbudor, line 81: Can you please take the opportunity of that change to enclose that #define line the #ifndef/#endif as suggested in the discussion
That will allow anyone to override that value if needed without touching the tasmota base code

barbudor (changes requested):

barbudor:
I still don't fully get the impact/usefullness of that change

Can you please write exactly what is the impact of that configuration change on the conversion process ?
It would be even better to comment the code to state exactly what this value involve in terms of conversion speed and averaging count for both current and voltage so we understand without haveing to pop out the INA datasheet.
Thanks

protectivedad:
> I still don't fully get the impact/usefullness of that change
> 
> Can you please write exactly what is the impact of that configuration change on the conversion process ? It would be even better to comment the code to state exactly what this value involve in terms of conversion speed and averaging count for both current and voltage so we understand without haveing to pop out the INA datasheet. Thanks

First I will admit some of it is just my engineering/physics bias that the number displayed should be accurate to the time frame it is being polled at and that the two numbers should be for the exact same time period. For a steady state it make absolutely no difference. The difference happens when current/voltages jump around.

I'd ask you:
You don't think that having the sample times for voltage and current being different was a mistake? I'm curious what is the benefit to only using the last 2.352 ms of voltage and 32.976 ms of current for every second?

For example my scooter scenario when I accelerate/decelerate. I can't think of why you wouldn't time things to the sample rate it just seems like proper design and it costs nothing to do. It also conforms to the advice in the INA226 [datasheel](https://www.ti.com/lit/ds/symlink/ina226.pdf?ts=1743002794620&ref_url=https%253A%252F%252Fsearch.brave.com%252F). In "6.4.1 Averaging and Conversion Time Considerations" they give an example of reading the device in 5ms increments and setting the device timing to be close to that. They also recommend having the largest sample and averaging for the application timing. Since our application samples at every second the driver should configure the INA226 chip to do the same.

I did have the question that I never got to about what is the best 1 second default. 64 sample at 8.244 each, 128 samples at 4.156, etc. Since you seem only concerned with the steady state almost never changing I figured the longest conversion time would provide the most noise reduction and would be the best. It was also the default for the current.

barbudor:
Power reading from INA devices is useless as they are not calculating in floating point.
I have removed it from INA219/3221 a long time ago
It is much more efficient and accurate do to the multiplication V * I on the ESP using floats.

> You don't think that having the sample times for voltage and current being different was a mistake?
You've have misunderstood me. Indeed I don't see the point in having2 different values
It was the choice of the original coder which I never checked before because I don't use many 226 (mainly 219 and 3221) and I was too lazy to decompose the hex value into its meaning (you may see that in the 219/3221 drivers, it is made readable.
So yes having same duration is more coherent.

The question still being what is the best durations for a generalTasmota user ?
There's always a compromise to do between duration and accuracy. TI doesn't give much details on the design of the ADC part. It's most likely a SAR but we don't know if there's a sample & hold stage in front.
As exposed in the datasheet section 6.4.1, short conversion times invole more noise. So ideally you may want to have the highest duration and the highest number of avering, but conversion time should remains short regarding the variation speed of the signal.

You still haven't written clearly what is you choice in terms of averaging and measurment time. And it would be niceto add that as a comment in the code above the define

- 64 * (8.244ms+8.244ms) would be above 1 sec
- 16 * (8.244ms+8.244ms) would be above 264 msec
- 64 * (4.156ms+4.156ms) would be 531 ms
But long measurment time requires more stable signals. Other options could be
- 256 * (1.1ms+1.1ms) would be 563 ms and probably cover more use cases

Anyhow, with the addition of the #ifdef/#endif, anyone can now build it's own version easily

protectivedad:
> Power reading from INA devices is useless as they are not calculating in floating point. I have removed it from INA219/3221 a long time ago It is much more efficient and accurate do to the multiplication V * I on the ESP using floats.

"useless" is a strong word. I agree it is more efficient to do what you did. My initial thought was why bother going through the trouble of reading the power value and then multiplying by a lsb when you can just use the values you have already read and multiply? Then I thought that for me at least I'd want the value from the chip. I can ignore it if I want and I always have the V and I to calculate what is currently presented as the chip value.

I setup a test bench with power supply set at 4V a shunt and a battery charge tester drawing 1 amp. I am using this setup to test the accuracy of adding the amp-hour and watt-hour functions, like in the INA3221 driver. After an hour on a steady draw the chip power values are indeed off from my tester.

I'll remove the change if you feel people don't want the actual chip values just a convenient display of the V * I.
> 
> > You don't think that having the sample times for voltage and current being different was a mistake?
> 
> You've have misunderstood me. Indeed I don't see the point in having2 different values It was the choice of the original coder which I never checked before because I don't use many 226 (mainly 219 and 3221) and I was too lazy to decompose the hex value into its meaning (you may see that in the 219/3221 drivers, it is made readable. So yes having same duration is more coherent.

My bad.
>
> The question still being what is the best durations for a generalTasmota user ? There's always a compromise to do between duration and accuracy. TI doesn't give much details on the design of the ADC part. It's most likely a SAR but we don't know if there's a sample & hold stage in front. As exposed in the datasheet section 6.4.1, short conversion times invole more noise. So ideally you may want to have the highest duration and the highest number of avering, but conversion time should remains short regarding the variation speed of the signal.

I'm looking for guidance really. I picked 64 * (8.244 + 8.244) because the amp measurement was configured to 8.244 ms and it most closely hits 1 second. It is of course a little over, which means that eventually you read the same number twice. As you shorten the duration and increase the sample you get farther and farther from 1s. Doesn't really matter to me. Just thought the default should most accurately represent the drivers sampling time by default.

> 
> You still haven't written clearly what is you choice in terms of averaging and measurment time. And it would be niceto add that as a comment in the code above the define
> 
>     * 64 * (8.244ms+8.244ms) would be above 1 sec
> 
>     * 16 * (8.244ms+8.244ms) would be above 264 msec
> 
>     * 64 * (4.156ms+4.156ms) would be 531 ms
>       But long measurment time requires more stable signals. Other options could be
> 
>     * 256 * (1.1ms+1.1ms) would be 563 ms and probably cover more use cases

This is where I just don't "get it". For half the time you have no idea what the system you are measuring is doing. I understand it might and probably does make no difference for most implementations but isn't it better to have a number that samples what you are measuring across the whole 1 second? Though this is a lot better then the 970 ms that is ignored in the original configuration.

> 
> 
> Anyhow, with the addition of the #ifdef/#endif, anyone can now build it's own version easily

That's very true. Let me know your thoughts on my points, I can update the PR anyway you think is best. As you point out as long as the #ifdef is there I can always override it for my purposes. If you feel the ` * 256 * (1.1ms+1.1ms) would be 563 ms and probably cover more use cases` is best I will make it so and add the comment.

protectivedad:
Sorry for taking the long way around. I've set the default to [barbudor's](https://github.com/barbudor) suggestion and added the comments. I see the issue with the power and have reverted it back. I do think there is value in displaying the INA226 information instead of a helper function.

barbudor:
LGTM

Thanks for your understanding

barbudor (approved):

PR #23205, opened 2025-03-28
SML: CRC for binary SML

tombueng:
## Description:
In  binary mode ('s') and v1 SML has CRC only at the end of each file. So the whole file needs to be buffered in order to parse it later. Parsing and then discarding on crc-error would also have been an option. 
Buffersize (1024) and crcmode (15=auto-detect) MUST be given via special option line e.g. "1,=soC,1024,15". See Updates docs on USE_SML_CRC.
Will collect all file data into a buffer scanning for the start sequence, then for the stop sequence. If found, check crc and if fine, send to SML_Decode method in slices each beginning with the SYNC byte. 

Docs PR here https://github.com/tasmota/docs/pull/1471

## Checklist:
  - [ X] The pull request is done against the latest development branch
  - [ X] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [ X] The code change is tested and works with Tasmota core ESP32 V.3.1.3.250302 (mine: 3_1_3/5.3.2.250228)


gemu2015:
thanks for this useful pr.
comments:
1. since we default to 5 meters it is not acceptable to allocate another 5 k of buffers. please allocate only when an SML meter is actually defined.
2. preferably we want to eliminate #defines as much as possible. think about making this dependent on descriptor special options. and then enable it per default on ESP32. (since on ESP8266 every k of RAM counts)
3. i saw very large SML dumps which exceeded 1k. probably make the special option about the buffers size, being zero disabling the CRC option

gemu2015:
another remark,
can you probably put all sml crc related memory in a separate struct and allocate it only when needed. (saves another 90 bytes) as said on an esp8266 every byte of RAM counts

tombueng:
> thanks for this useful pr. comments:
> 
> 1. since we default to 5 meters it is not acceptable to allocate another 5 k of buffers. please allocate only when an SML meter is actually defined.
> 2. preferably we want to eliminate #defines as much as possible. think about making this dependent on descriptor special options. and then enable it per default on ESP32. (since on ESP8266 every k of RAM counts)
> 3. i saw very large SML dumps which exceeded 1k. probably make the special option about the buffers size, being zero disabling the CRC option
Thank you @gemu2015 for looking into it!

1 -> Will check, though it would only be allocated for meters that haven been configure.
2 -> If we want to have this enabled by default on ESP32 we will need some kind of crc-detection because the meters seem to use all kind of differnt crc algos and polys. I will check what we can do here. 
3 -> will

gemu2015 (commented):
let me suggest some changes.
we should have another special option instead of #defines
check this subroutine:
char *SpecOptions(char *cp, uint32_t mnum)
and add this case
```
case 'A':
	cp += 2;
	mp->so_sml_crc = (mp->so_sml_crc & 0xf000) | (strtol(cp, &cp, 10)&0xfff);
	if (*cp == ',') {
		cp++;
		mp->mp->so_sml_crc = (mp->so_sml_crc & 0x0fff) | (strtol(cp, &cp, 10)<<12);
	}
        break;
```
define a uint16_t so_sml_crc in special option section of meter desc.
Initialize this var with zero to disable crc mode

if a user wants to enable he just has to add this to the descriptor
1,=soA,1024,0
1. parameter being the buffer size,
2. parameter being the crc checksum type

also i meant to have a variable in the meter struct 
struct SML_CRC_DATA *sml_crc_data;

and allocate and initialize that struct when a binary sml is defined in descriptor.

thus we only have an overhead of 6 bytes per meter to the previous version.

gemu2015 (commented):
looks good! 
since crc type has only 4 bits auto_crc should be 15 not 16.

tombueng:
After running for a few days I can now confirm that the wrong readings that occurred one or twice a day have disappeared. So it seems to be the transmission that was causing the problem. I suspect it is due to high fluctuations in power that induce into the reading head or even into the Smartmeter itself. I will also do some testing with shielding the reading device.

tombueng:
Now, with CRC check:
image

Before, without CRC check:
image

gemu2015:
@tombueng 
i also tested the crc check and it works well.
thank you very much
@arendst 
pr can be merged

tombueng:
Ready

tombueng:
Docs PR here https://github.com/tasmota/docs/pull/1471

tombueng:
Docs have been merged, Byteflip modes still missing in docs.

PR #23206, opened 2025-03-29
Expose esp_http_server to Berry

jetpax:
## Description: Add WS server to Berry

## Checklist:
  - [ ✓] The pull request is done against the latest development branch
  - [ ✓] Only relevant files were touched
  - [ ✓] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ x ] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [ ✓] The code change is tested and works with Tasmota core ESP32 V.3.1.3.250302
  - [ ✓] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

I saw a few discussions around adding a websocket server and wanted one myself, so as I've found the esp-idf http_server to be pretty solid, I adapted it for Tasmota/Berry. Obviously it will not work on esp8266, but hopefully others may find it useful.

- provides Tasmota/Berry with a  fast, async, full featured http/websocket/files server capable of serving complex applications if required.

- only adds 3 files, and requires no new libraries beyond what is included in esp-idf. 

- doesn't currently support https/wss, but could easily be modified to do so.

(To avoid task switching issues, all received data touched by Berry VM is put in a queue in the esp_http_server context, and is read by Berry in the FastLoop, ie Berry VM context). 

It consists of three primary components working together:

1. **HTTP Server** - Handles web requests and serves as the foundation for WebSocket and file serving
2. **WebSocket Server** - Manages real-time bidirectional communication between the browser and Berry VM
3. **File Server** - Serves static web content and resources to the browser (server directly from the esp-idf context)

To try it out you'll need to add the following to your user_config_override.h

```
#ifndef USE_BERRY_WSSERVER
#define USE_BERRY_WSSERVER
#endif
#ifndef USE_BERRY_HTTPSERVER
#define USE_BERRY_HTTPSERVER
#endif
#ifndef USE_BERRY_WEBFILES
#define USE_BERRY_WEBFILES
#endif
```

Here's a simple example script showing the 3 parts working together

```
# Debugging script for WebSocket/HTTP server integration
import httpserver
import string
import wsserver
import json
import webfiles

# Configure log level
var LOG_LEVEL = 2

# Global state
var ws_clients = {}
var msg_count = 0
var log_countdown = 10

# HTTP root handler
def http_root_handler(uri)
  tasmota.log("DEBUG: HTTP root handler called")

  # Get system information
  var memory = tasmota.memory()
  var rtc = tasmota.rtc()
  var uptime = tasmota.millis() / 1000
  
  # Create HTML content
  var html = string.format(
    "<html><head><title>Async HTTP/WebSocket/Webfiles Demo</title>" +
    "<style>body{font-family:Arial,sans-serif;margin:20px;line-height:1.6}" +
    "h1{color:#0066cc}h2{color:#009900}table{border-collapse:collapse;width:100%%}" +
    "th,td{text-align:left;padding:8px;border:1px solid #ddd}" +
    "th{background-color:#f2f2f2}tr:nth-child(even){background-color:#f9f9f9}" +
    "</style></head><body>" +
    "<h1>Async HTTP/WebSocket/Webfiles Demo</h1>" +
    "<p>System is running normally. You can connect to the WebSocket at <code>ws://[host]:8080/ws</code></p>" +
    
    "<h2>Memory Information</h2>" +
    "<table><tr><th>Metric</th><th>Value</th></tr>" +
    "<tr><td>Heap Free</td><td>%d KB</td></tr>" +
    "<tr><td>Stack Low</td><td>%.2f KB</td></tr>" +
    "<tr><td>PSRAM Free</td><td>%d KB</td></tr>" +
    "<tr><td>Fragmentation</td><td>%d%%</td></tr>" +
    "</table>" +
    
    "<h2>Time Information</h2>" +
    "<table><tr><th>Metric</th><th>Value</th></tr>" +
    "<tr><td>UTC Time</td><td>%d</td></tr>" +
    "<tr><td>Local Time</td><td>%d</td></tr>" +
    "<tr><td>Uptime</td><td>%.1f seconds</td></tr>" +
    "</table>" +

    "<h2>WebSocket Clients</h2>" +
    "<table><tr><th>Total Clients</th><th>Messages Processed</th></tr>" +
    "<tr><td>%d</td><td>%d</td></tr>" +
    "</table>" +
    
    "<script>setTimeout(function() { document.location.reload(); }, 5000);</script>" +
    "</body></html>",
    
    memory["heap_free"],
    memory["stack_low"],
    memory["psram_free"],
    memory["frag"],
    rtc["utc"],
    rtc["local"],
    uptime,
    size(ws_clients),
    msg_count
  )
  
  # Simply send the response
  return httpserver.send(html)
end

# Start servers
tasmota.log("⚡ Starting servers", LOG_LEVEL)
wsserver.stop()
httpserver.stop()

# Start HTTP server
httpserver.start(8080)
httpserver.on("/", http_root_handler)

# Start WebSocket server
wsserver.start("/ws", httpserver)

# WebSocket event handlers
def ws_connect_handler(client, event, message)
  # Log event information
  tasmota.log(string.format("⚡ WS_CONNECT: client=%d", client), LOG_LEVEL)
  
  # Handle new connection
  tasmota.log("⚡ Client " + str(client) + " connected", LOG_LEVEL)
  
  # Get system information
  var system_info = {
    "status": "connected",
    "msg": "Hello from WebSocket server",
    "memory": tasmota.memory(),
    "rtc": tasmota.rtc()
  }

  var send_result = wsserver.send(client, json.dump(system_info))
  tasmota.log("⚡ Welcome message send result: " + str(send_result), LOG_LEVEL)

  
  # Track client
  ws_clients[client] = { 
    "connected": true,
    "messages_received": 0,
    "messages_sent": 1
  }
  
  
  msg_count += 1
  return true
end

def ws_disconnect_handler(client, event, message)
  # Handle disconnection
  tasmota.log("⚡ Client " + str(client) + " disconnected", LOG_LEVEL)
  
  # Remove client from tracking
  if ws_clients.contains(client)
    var client_info = ws_clients[client]
    tasmota.log(string.format("⚡ Client %d stats: received=%d, sent=%d", 
                client, client_info["messages_received"], client_info["messages_sent"]), LOG_LEVEL)
    ws_clients.remove(client)
  end
  
  return true
end

def ws_message_handler(client, event, message)
  # Handle incoming message
  tasmota.log("⚡ Message received from client " + str(client) + ": " + message, LOG_LEVEL)
  
  # Update client stats
  if ws_clients.contains(client)
    ws_clients[client]["messages_received"] += 1
  else
    # Client not in our list, add it
    ws_clients[client] = {
      "connected": true,
      "messages_received": 1,
      "messages_sent": 0
    }
  end
  
  # Create response
  var response = {
    "status": "received", 
    "echo": message,
    "memory": tasmota.memory(),
    "rtc": tasmota.rtc(),
    "uptime": tasmota.millis() / 1000,
    "clients": size(ws_clients)
  }
  
  if wsserver.is_connected(client)
    var response_json = json.dump(response)
    var send_result = wsserver.send(client, response_json)
    tasmota.log("⚡ Response send result: " + str(send_result), LOG_LEVEL)
    
    if ws_clients.contains(client)
      ws_clients[client]["messages_sent"] += 1
    end
  end
  
  msg_count += 1
  return true
end

# Register WebSocket event handlers
wsserver.on(wsserver.CONNECT, ws_connect_handler)
wsserver.on(wsserver.DISCONNECT, ws_disconnect_handler)
wsserver.on(wsserver.MESSAGE, ws_message_handler)


# Register for queue processing
tasmota.add_fast_loop(/-> httpserver.process_queue())

# Set up file server
var file_path = "/files"  # The directory that contains your web files
var serve_path = "/files/"     # The URL path prefix to serve files from


# Initialize file server
var files_result = webfiles.serve(file_path, serve_path)
if files_result
  tasmota.log("⚡ File server started at " + serve_path + " serving from " + file_path, LOG_LEVEL)
else
  tasmota.log("⚡ Failed to start file server", LOG_LEVEL)
end

tasmota.log("⚡ HTTP/WS/webfiles debug test initialized", LOG_LEVEL)

```
 I'm new to Tasmota so apologies it if it not added in the 'right' way, but I tried to follow what others seem to have done as far as possible.

### DIFF
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -40,6 +40,17 @@ be_extern_native_module(gpio);
 be_extern_native_module(display);
 be_extern_native_module(energy);
 be_extern_native_module(webserver);
+#ifdef USE_HTTPSERVER
+be_extern_native_module(httpserver);
+
+#ifdef USE_WSSERVER
+be_extern_native_module(wsserver);
+#endif // USE_WSSERVER
+
+#ifdef USE_WEBFILES
+be_extern_native_module(webfiles);
+#endif // USE_WEBFILES
+#endif // USE_HTTPSERVER
 be_extern_native_module(flash);
 be_extern_native_module(path);
 be_extern_native_module(unishox);
@@ -170,6 +181,15 @@ BERRY_LOCAL const bntvmodule_t* const be_module_table[] = {
 #ifdef USE_WEBSERVER
     &be_native_module(webserver),
 #endif // USE_WEBSERVER
+#ifdef USE_HTTPSERVER
+    &be_native_module(httpserver),
+#ifdef USE_WSSERVER
+    &be_native_module(wsserver),
+#endif // USE_WSSERVER
+#ifdef USE_WEBFILES
+    &be_native_module(webfiles),
+#endif // USE_WEBFILES
+#endif // USE_HTTPSERVER
 #ifdef USE_ZIGBEE
     &be_native_module(zigbee),
     &be_native_module(matter_zigbee),
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -40,17 +40,17 @@ be_extern_native_module(gpio);
 be_extern_native_module(display);
 be_extern_native_module(energy);
 be_extern_native_module(webserver);
-#ifdef USE_HTTPSERVER
+#ifdef USE_BERRY_HTTPSERVER
 be_extern_native_module(httpserver);
 
-#ifdef USE_WSSERVER
+#ifdef USE_BERRY_WSSERVER
 be_extern_native_module(wsserver);
-#endif // USE_WSSERVER
+#endif // USE_BERRY_WSSERVER
 
-#ifdef USE_WEBFILES
+#ifdef USE_BERRY_WEBFILES
 be_extern_native_module(webfiles);
-#endif // USE_WEBFILES
-#endif // USE_HTTPSERVER
+#endif // USE_BERRY_WEBFILES
+#endif // USE_BERRY_HTTPSERVER
 be_extern_native_module(flash);
 be_extern_native_module(path);
 be_extern_native_module(unishox);
@@ -181,15 +181,15 @@ BERRY_LOCAL const bntvmodule_t* const be_module_table[] = {
 #ifdef USE_WEBSERVER
     &be_native_module(webserver),
 #endif // USE_WEBSERVER
-#ifdef USE_HTTPSERVER
+#ifdef USE_BERRY_HTTPSERVER
     &be_native_module(httpserver),
-#ifdef USE_WSSERVER
+#ifdef USE_BERRY_WSSERVER
     &be_native_module(wsserver),
-#endif // USE_WSSERVER
-#ifdef USE_WEBFILES
+#endif // USE_BERRY_WSSERVER
+#ifdef USE_BERRY_WEBFILES
     &be_native_module(webfiles),
-#endif // USE_WEBFILES
-#endif // USE_HTTPSERVER
+#endif // USE_BERRY_WEBFILES
+#endif // USE_BERRY_HTTPSERVER
 #ifdef USE_ZIGBEE
     &be_native_module(zigbee),
     &be_native_module(matter_zigbee),
--- a/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
@@ -17,8 +17,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifdef USE_BERRY
-#ifdef USE_WEBSERVER
+#ifdef USE_WEBFILES
 
 #ifndef LOG_LOCAL_LEVEL
 #define LOG_LOCAL_LEVEL ESP_LOG_INFO
@@ -387,5 +386,4 @@ module webfiles (scope: global) {
 @const_object_info_end */
 #include "be_fixed_webfiles.h"
 
-#endif // USE_WEBSERVER
-#endif // USE_BERRY 
+#endif // USE_WEBFILES
--- a/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
@@ -17,7 +17,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifdef USE_WEBFILES
+#ifdef USE_BERRY_WEBFILES
 
 #ifndef LOG_LOCAL_LEVEL
 #define LOG_LOCAL_LEVEL ESP_LOG_INFO
@@ -389,4 +389,4 @@ module webfiles (scope: global) {
 @const_object_info_end */
 #include "be_fixed_webfiles.h"
 
-#endif // USE_WEBFILES
+#endif // USE_BERRY_WEBFILES
--- a/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
@@ -280,61 +280,64 @@ static int w_webfiles_serve(bvm *vm) {
         
         ESP_LOGI(TAG, "Setting up file server with base path: %s, uri prefix: %s", path, prefix);
         
-        // Update base path
-        strlcpy(base_path, path, sizeof(base_path));
-        
-        // Update URI prefix
-        strlcpy(uri_prefix, prefix, sizeof(uri_prefix));
-        
-        // Check if directory exists
         struct stat st;
-        if (stat(base_path, &st) != 0 || !S_ISDIR(st.st_mode)) {
-            ESP_LOGE(TAG, "Base path is not a valid directory: %s", base_path);
+        if (stat(path, &st) != 0 || !S_ISDIR(st.st_mode)) {
+            ESP_LOGE(TAG, "Input path is not a valid directory: %s", path);
             be_pushbool(vm, false);
-            be_pop(vm, be_top(vm) - initial_top);
-            return 1;
+            // Clean up stack before returning
+            while (be_top(vm) > initial_top) {
+                be_pop(vm, 1);
+            }
+            be_return (vm); // Return directly
         }
-        
-        // Check if HTTP server is running
+
         httpd_handle_t server = be_httpserver_get_handle();
         if (!server) {
             ESP_LOGE(TAG, "HTTP server not running");
             be_pushbool(vm, false);
-            be_pop(vm, be_top(vm) - initial_top);
-            return 1;
+            while (be_top(vm) > initial_top) {
+                be_pop(vm, 1);
+            }
+            be_return (vm);
         }
+
+        strlcpy(base_path, path, sizeof(base_path));
+        strlcpy(uri_prefix, prefix, sizeof(uri_prefix));
         
-        // Register URI handler with wildcard support
-        // The * at the end allows it to match all URIs that start with prefix
-        char uri_pattern[64];
-        snprintf(uri_pattern, sizeof(uri_pattern), "%s*", prefix);
-        
-        ESP_LOGI(TAG, "Registering URI handler with pattern: %s", uri_pattern);
-        
+
+        // Use a static buffer for the URI pattern
+        static char registered_uri_pattern[64]; // ADD static keyword
+        // Ensure it's null-terminated even if snprintf truncates
+        registered_uri_pattern[sizeof(registered_uri_pattern) - 1] = '\0';
+        snprintf(registered_uri_pattern, sizeof(registered_uri_pattern), "%s*", prefix);
+
+        ESP_LOGI(TAG, "Registering URI handler with pattern: %s", registered_uri_pattern);
+
         httpd_uri_t uri_handler = {
-            .uri      = uri_pattern,
+            // Point to the static buffer
+            .uri      = registered_uri_pattern, // Use the static buffer
             .method   = HTTP_GET,
             .handler  = webfiles_handler,
             .is_websocket = false,
             .user_ctx = NULL
         };
-        
+
         esp_err_t ret = httpd_register_uri_handler(server, &uri_handler);
         if (ret != ESP_OK) {
             ESP_LOGE(TAG, "Failed to register URI handler: %d", ret);
             be_pushbool(vm, false);
             be_pop(vm, be_top(vm) - initial_top);
-            return 1;
+            be_return (vm);
         }
         
         be_pushbool(vm, true);
         be_pop(vm, be_top(vm) - initial_top);
-        return 1;
+        be_return (vm);
     }
     
     be_pushbool(vm, false);
     be_pop(vm, be_top(vm) - initial_top);
-    return 1;
+    be_return (vm);
 }
 
 // webfiles.serve_file(file_path, uri) -> bool
@@ -352,7 +355,7 @@ static int w_webfiles_serve_file(bvm *vm) {
             ESP_LOGE(TAG, "File not found: %s", file_path);
             be_pushbool(vm, false);
             be_pop(vm, be_top(vm) - initial_top);
-            return 1;
+            be_return (vm);
         }
         
         // TODO: Implement custom handler for specific files
@@ -361,12 +364,12 @@ static int w_webfiles_serve_file(bvm *vm) {
         ESP_LOGW(TAG, "serve_file not yet implemented");
         be_pushbool(vm, false);
         be_pop(vm, be_top(vm) - initial_top);
-        return 1;
+        be_return (vm);
     }
     
     be_pushbool(vm, false);
     be_pop(vm, be_top(vm) - initial_top);
-    return 1;
+    be_return (vm);
 }
 
 // Module definition
--- a/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
+++ b/lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
@@ -1,0 +1,391 @@
+/*
+  be_webfiles_lib.c - Static file server for Berry using ESP-IDF HTTP server
+
+    Copyright (C) 2025 Jonathan E. Peace
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifdef USE_BERRY
+#ifdef USE_WEBSERVER
+
+#ifndef LOG_LOCAL_LEVEL
+#define LOG_LOCAL_LEVEL ESP_LOG_INFO
+#endif
+
+#include "be_constobj.h"
+#include "be_mapping.h"
+
+// ESP-IDF includes
+#include "esp_log.h"
+#include "esp_http_server.h"
+#include "esp_vfs.h"
+
+// External Berry/Tasmota includes
+extern httpd_handle_t be_httpserver_get_handle(void);
+extern bool httpserver_queue_message(int type, int client_id, 
+                                const char* data, size_t len, void* user_data);
+
+// Tag for logging
+static const char *TAG = "WEBFILES";
+
+// Default base path for files
+static char base_path[64] = "/files";
+
+// URI prefix for the file server
+static char uri_prefix[32] = "/";
+
+// Maximum file path length
+#define FILE_PATH_MAX 128
+
+// Scratch buffer size for file transfer
+#define SCRATCH_BUFSIZE 4096   // 4KB scratch buffer for chunks
+
+// Static buffer for file sending - fixed allocation
+static char scratch_buffer[SCRATCH_BUFSIZE];
+
+// MIME Type Mapping
+static const struct {
+    const char *extension;
+    const char *mimetype;
+} mime_types[] = {
+    {".html", "text/html"},
+    {".htm", "text/html"},
+    {".js", "application/javascript"},
+    {".mjs", "application/javascript"}, // ES modules
+    {".css", "text/css"},
+    {".png", "image/png"},
+    {".jpg", "image/jpeg"},
+    {".jpeg", "image/jpeg"},
+    {".gif", "image/gif"},
+    {".ico", "image/x-icon"},
+    {".svg", "image/svg+xml"},
+    {".json", "application/json"},
+    {".txt", "text/plain"},
+    {".md", "text/markdown"},
+    {".wasm", "application/wasm"}, // WebAssembly
+    {".map", "application/json"}, // Source maps
+    {".woff", "font/woff"},
+    {".woff2", "font/woff2"},
+    {".ttf", "font/ttf"},
+    {".otf", "font/otf"},
+    {".bin", "application/octet-stream"},
+    {NULL, NULL}
+};
+
+// Get MIME type based on file extension
+static const char* get_mime_type(const char *path) {
+    const char *ext = strrchr(path, '.');
+    if (!ext) return "text/plain";
+    
+    for (int i = 0; mime_types[i].extension; i++) {
+        if (strcasecmp(mime_types[i].extension, ext) == 0) {
+            return mime_types[i].mimetype;
+        }
+    }
+    return "text/plain";
+}
+
+// Build full path including base path
+static const char* get_full_path(char *dest, const char *uri, size_t destsize) {
+    size_t base_len = strlen(base_path);
+    size_t prefix_len = strlen(uri_prefix);
+    size_t uri_len = strlen(uri);
+    
+    // Handle query parameters and fragments in URI
+    const char *query = strchr(uri, '?');
+    if (query) {
+        uri_len = query - uri;
+    }
+    const char *fragment = strchr(uri, '#');
+    if (fragment && (!query || fragment < query)) {
+        uri_len = fragment - uri;
+    }
+    
+    // Skip the URI prefix to get the relative path
+    const char *relative_path = uri;
+    if (prefix_len > 1 && strncmp(uri, uri_prefix, prefix_len) == 0) {
+        relative_path = uri + prefix_len - 1; // -1 because we want to keep the leading slash
+        uri_len -= (prefix_len - 1);
+    }
+    
+    // Check if path will fit in destination buffer
+    if (base_len + uri_len + 1 > destsize) {
+        ESP_LOGE(TAG, "Path too long");
+        return NULL;
+    }
+    
+    // Construct full path
+    strcpy(dest, base_path);
+    if (base_len > 0 && base_path[base_len-1] == '/' && relative_path[0] == '/') {
+        // Avoid double slash
+        strlcpy(dest + base_len, relative_path + 1, uri_len);
+    } else {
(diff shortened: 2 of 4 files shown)


Jason2866:
Have you metrics about Flash / RAM usage by adding?

jetpax:
No, but if you are interested you can get flash usage by building it with pio run, and RAM usage by running the script above and opening the web page.

Jason2866:
@jetpax  Well that was a friendly way to ask you to do. Without this info the PR will not be considered to be merged ;-)

jetpax:
Apologies @Jason2866, not at my dev station right now, but what exactly do you require in order to consider merging this PR (might be an idea to add that to the checklist for newcomers like me in the future?)

Jason2866:
Increase of flash usage (compile with and without the PR). Same for RAM. Enough to do only for the ESP32.
For a "simple" driver this is not needed since looking at the code an assumption is possible. This PR is not a simple driver.

jetpax:
**BEFORE**

```
// #ifndef USE_WSSERVER
// #define USE_WSSERVER
// #endif
// #ifndef USE_HTTPSERVER
// #define USE_HTTPSERVER
// #endif
// #ifndef USE_WEBFILES
// #define USE_WEBFILES
// #endif
```

```
Total sizes:
Used static IRAM:   69398 bytes ( 288746 remain, 19.4% used)
      .text size:   68371 bytes
   .vectors size:    1027 bytes
Used stat D/IRAM:   69404 bytes ( 272356 remain, 20.3% used)
      .data size:   19388 bytes
      .bss  size:   50016 bytes
Used Flash size : 1938722 bytes
           .text: 1315918 bytes
         .rodata:  622548 bytes
Total image size: 2027508 bytes (.bin may be padded larger)
==================================================================================
SUCCESS] Took 31.30 seconds with
==================================================================================

Environment    Status    Duration
-------------  --------  ------------
tasmota32s3    SUCCESS   00:00:31.297

```


**AFTER**

```
#ifndef USE_WSSERVER
#define USE_WSSERVER
#endif
#ifndef USE_HTTPSERVER
#define USE_HTTPSERVER
#endif
#ifndef USE_WEBFILES
#define USE_WEBFILES
#endif
```

```

Total sizes:
Used static IRAM:   69454 bytes ( 288690 remain, 19.4% used)
      .text size:   68427 bytes
   .vectors size:    1027 bytes
Used stat D/IRAM:   73860 bytes ( 267900 remain, 21.6% used)
      .data size:   19484 bytes
      .bss  size:   54376 bytes
Used Flash size : 1987398 bytes
           .text: 1348786 bytes
         .rodata:  638356 bytes
Total image size: 2076336 bytes (.bin may be padded larger)
================================================================================
[SUCCESS] Took 32.72 seconds
==================================================================================

Environment    Status    Duration
-------------  --------  ------------
tasmota32s3    SUCCESS   00:00:32.723
```

Noschvie:
@jetpax please have a look to the [eKey notification api](https://developer.ekey.net/t/documentation-notification-api/66)
Do you think that the functions of your PR are suitable to support this eKey webhook API?
Thanks!

jetpax:
@Noschvie I don't see why not as it just seems to be an incoming webquery, which could be handled by Berry. The API doesn't seem very clearly spelled out tho. 

But as mentioned , https is not currently enabled, although that is quite simply done.

Note that for https, per espressif

> The initial session setup can take about two seconds, or more with slower clock speed or more verbose logging. Subsequent requests through the open secure socket are much faster (down to under 100 ms).

so i guess if your application is OK with that you should be fine.

s-hadinger:
Wow, you did a lot of work. Having websockets is definitely interesting.

But I'm worried that it adds 48KB of Flash for this, this is a lot. For simple async web server, I did a full version in Berry that takes only 8.3KB and can be loaded dynamically - although it does not support websockets.

I'm ok to merge since it does not add any code by default, but I doubt we include it in any pre-compiled build due to its size.

lib/libesp32/berry/default/be_modtab.c
```diff
@@ -40,6 +40,17 @@ be_extern_native_module(gpio);
 be_extern_native_module(display);
 be_extern_native_module(energy);
 be_extern_native_module(webserver);
+#ifdef USE_HTTPSERVER
```
s-hadinger, line 43: I'm worried that `USE_HTTPSERVER` is misleading since the Arduino web server is enabled with other flags.

Maybe it would be clearer to rename `USE_BERRY_HTTPSERVER`

lib/libesp32/berry/default/be_modtab.c
```diff
@@ -40,6 +40,17 @@ be_extern_native_module(gpio);
...
+be_extern_native_module(httpserver);
+
+#ifdef USE_WSSERVER
```
s-hadinger, line 46: Same comment `USE_BERRY_WSSERVER`

lib/libesp32/berry/default/be_modtab.c
```diff
@@ -40,6 +40,17 @@ be_extern_native_module(gpio);
...
+be_extern_native_module(wsserver);
+#endif // USE_WSSERVER
+
+#ifdef USE_WEBFILES
```
s-hadinger, line 50: Same comment `USE_BERRY_WEBFILES`

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
```diff
@@ -0,0 +1,389 @@
...
+
+#ifndef LOG_LOCAL_LEVEL
+#define LOG_LOCAL_LEVEL ESP_LOG_INFO
+#endif
+
+#include "be_constobj.h"
+#include "be_mapping.h"
+
+// ESP-IDF includes
+#include "esp_log.h"
```
s-hadinger, line 30: We don't use esp-idf logging in Tasmota. It doesn't hurt but it will be disabled at compile time.

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
```diff
@@ -0,0 +1,389 @@
...
+}
+
+//checks if a .html, .css, or .js file exists with a .br suffix and serve that Brotli-compressed version instead
+
+static esp_err_t webfiles_handler(httpd_req_t *req) {
+  char filepath[FILE_PATH_MAX];
+  char brotli_filepath[FILE_PATH_MAX];
+  FILE *file = NULL;
+  struct stat file_stat;
+  bool use_brotli = false;
```
s-hadinger, line 166: Is it really worth enabling brotli? What is the code impact? Also ESP32 CPU is not very fast, LAN is generally faster than compression.

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
```diff
@@ -0,0 +1,389 @@
...
+  
+  // Check if it's a regular file
+  if (!S_ISREG(file_stat.st_mode)) {
+      ESP_LOGE(TAG, "Not a regular file: %s", filepath);
+      httpd_resp_send_err(req, HTTPD_404_NOT_FOUND, "Not a file");
+      return ESP_FAIL;
+  }
+  
+  // Open the file for reading
+  file = fopen(filepath, "r");
```
s-hadinger, line 220: We need to check that using `fopen` is compatible with universal file system, especially when it comes to dealing with SD Card

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
```diff
@@ -0,0 +1,389 @@
...
+        
+        ESP_LOGI(TAG, "Setting up file server with base path: %s, uri prefix: %s", path, prefix);
+        
+        // Update base path
+        strlcpy(base_path, path, sizeof(base_path));
+        
+        // Update URI prefix
+        strlcpy(uri_prefix, prefix, sizeof(uri_prefix));
+        
+        // Check if directory exists
```
s-hadinger, line 289: Wouldn't it be better to update `base_path` after you check that it's valid?

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
```diff
@@ -0,0 +1,389 @@
...
+        
+        // Register URI handler with wildcard support
+        // The * at the end allows it to match all URIs that start with prefix
+        char uri_pattern[64];
+        snprintf(uri_pattern, sizeof(uri_pattern), "%s*", prefix);
+        
+        ESP_LOGI(TAG, "Registering URI handler with pattern: %s", uri_pattern);
+        
+        httpd_uri_t uri_handler = {
+            .uri      = uri_pattern,
```
s-hadinger, line 315: Passing a string allocated on the stack to a handler sounds really bad. Didn't you see a crash here?

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
```diff
@@ -0,0 +1,389 @@
...
+            .is_websocket = false,
+            .user_ctx = NULL
+        };
+        
+        esp_err_t ret = httpd_register_uri_handler(server, &uri_handler);
+        if (ret != ESP_OK) {
+            ESP_LOGE(TAG, "Failed to register URI handler: %d", ret);
+            be_pushbool(vm, false);
+            be_pop(vm, be_top(vm) - initial_top);
+            return 1;
```
s-hadinger, line 327: This is not the appropriate Berry exit. It should be `be_return(vm);`

lib/libesp32/berry_tasmota/src/be_wsserver_lib.c
```diff
@@ -0,0 +1,1452 @@
...
+// This handler is called directly by the ESP-IDF HTTP server framework
+// Any Berry VM operations must be queued for the main task
+static esp_err_t ws_handler(httpd_req_t *req) {
+    ESP_LOGI(TAG, "Handling WebSocket frame");
+    
+    // Start with basic HTTP data
+    int client_slot = -1;
+    
+    // Initialize WebSocket frame structure
+    httpd_ws_frame_t ws_pkt;
```
s-hadinger, line 241: Did you try:
`httpd_ws_frame_t ws_pkt = {0};`

This is normally a standard construct to zero a structure

s-hadinger (commented):
Thanks, this is a huge piece of code. This is an excellent start.

I couldn't review in details yet. I made some comments and request.

I'm worried that all Berry functions end with `return 1;`. Please replace with the idiomatic `be_return(vm);`

s-hadinger:
> Note that for https, per [espressif](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/protocols/esp_https_server.html)
> 
> > The initial session setup can take about two seconds, or more with slower clock speed or more verbose logging. Subsequent requests through the open secure socket are much faster (down to under 100 ms).

This slowdown is due to TLS session negociation. But I understand this code does not support HTTPS. Anyways, adding HTTPS server would include mbedTLS and increase significantly the Flash size

Jason2866:
> > Note that for https, per [espressif](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/protocols/esp_https_server.html)
> > > The initial session setup can take about two seconds, or more with slower clock speed or more verbose logging. Subsequent requests through the open secure socket are much faster (down to under 100 ms).
> 
> This slowdown is due to TLS session negociation. But I understand this code does not support HTTPS. Anyways, adding HTTPS server would include mbedTLS and increase significantly the Flash size

mbedtls support (ciphers) is mostly disabled in Tasmota Arduino framework since it needs a **lot** of flash space. That's a general design decision which will not changed.

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
jetpax: The brotli stuff is only to support brotli compressed files (transparanetly) so that eg CSS/SVG/JS/HTML can be stored in the filesystem compressed and are automatically decompressed in the client. There is no compression running on the ESP32

jetpax:
@Jason2866 I actually dont need tls for my application, but I see that WolfSSL is an[ option in esp-idf](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/protocols/esp_tls.html#underlying-ssl-tls-library-options). Would that be a possible way forward?

lib/libesp32/berry/default/be_modtab.c
jetpax: I did think about that, but was following be_webserver_lib.c which uses USE_WEBSERVER,  and realised that the  exposed functionality  could also be used by C code if desired.  Happy to change if you think its the best thing to do.

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
jetpax: Yes, appreciate that, but obv was useful (essential) for debugging. In fact I left it in because I knew it would be disabled for most builds :)

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
jetpax: Appreciate you help on this

lib/libesp32/berry_tasmota/src/be_wsserver_lib.c
jetpax: Agreed!

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
jetpax: my bad, fixed.

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
jetpax: oops, classic dangling pointer, fixed

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
jetpax: Agreed, fixed.

jetpax:
Many thanks @s-hadinger, recommended fixes applied. 

Just for reference, here's brief overview of how the code fits together. 

###  Startup Process

#### HTTP Server (be_httpserver_lib.c)

When `httpserver.start(port)` is called from Berry:
- Creates the HTTP server using ESP-IDF's `httpd_start`
- Initializes the message queue (`xQueueCreate`)
- Registers a fast loop callback with Tasmota
- Returns a handle to be used by other components

**Message Queue Initialization:**
- Creates a queue with a capacity of 10 messages
- Sets up message types for WebSocket, file, and web requests
- Prepares the infrastructure for thread-safe communication

#### WebSocket Server (be_wsserver_lib.c)

When `wsserver.start(ws_path)` is called:
- Registers a URI handler for WebSocket connections with the HTTP server
- Sets up connection tracking structures for clients
- Prepares callbacks that will be invoked when events occur

**WebSocket Event Registration:**
- Registers Berry callbacks for:
  - Connection events (CONNECT)
  - Disconnection events (DISCONNECT)
  - Message events (MESSAGE)

#### File Server (be_webfiles_lib.c)

When `webfiles.serve(file_path, serve_path)` is called:
- Registers a URI handler for file serving with the HTTP server
- Sets up the base path for serving files from the filesystem
- Configures MIME type mappings for different file extensions
- Handles wildcard URI patterns to catch all paths under the prefix


**URI Registration Process:**
- Creates a URI pattern with wildcard (`serve_path + "*"`) to match all paths with the prefix
- Registers the pattern with ESP-IDF's HTTP server via `httpd_register_uri_handler()`
- File server operates independently of the queue mechanism used by WebSocket for efficiency

**MIME Type Handling:**
- Automatically detects appropriate MIME types based on file extensions
- Sets proper HTTP headers for content type, caching, and CORS
- Provides optimized handling for static assets vs. HTML content


### Message Processing Flow

1. **HTTP Server Context (ESP-IDF Thread)**:
   - Client requests arrive and are handled by HTTP server task
   - Instead of directly calling Berry VM functions, handlers queue messages
   - Message data is copied to ensure it remains valid across contexts
   - Messages are queued with type information (WebSocket, File, Web)

2. **Berry VM Context (Main Tasmota Task)**:
   - The Berry app registers a fast loop function that runs ~200 times per second
   - In each iteration, it checks for messages with `httpserver.has_messages()`
   - If messages exist, it calls `httpserver.process_queue()` to process one message
   - Based on message type, the appropriate handler is called:
     - `be_wsserver_handle_message` for WebSocket messages
     - `be_httpserver_process_file_request` for file requests
     - `be_httpserver_process_web_request` for web requests

3. **Berry Callback Invocation**:
   - The type-specific handlers call the corresponding Berry callbacks
   - For WebSocket messages, `be_wsserver_handle_message` calls the global `wsserver_dispatch` function
   - The Berry application processes the message and may send responses
   - All Berry VM operations occur in the main task context, preventing stack corruption

This architecture ensures that all Berry code runs in the appropriate context to ensure thread safety.

### Production Side:
- HTTP server tasks (running potentially at higher priority) receive client requests
- Instead of directly calling Berry VM functions, they enqueue messages

### Consumer Side:
- The main Tasmota task processes the queue via the fast_loop mechanism
- Berry VM operations only happen in the Berry VM context

jetpax:
> Wow, you did a lot of work. Having websockets is definitely interesting.
> 
> 
> 
> But I'm worried that it adds 48KB of Flash for this, this is a lot. For simple async web server, I did a full version in Berry that takes only 8.3KB and can be loaded dynamically - although it does not support websockets.
> 
> 
> 
> I'm ok to merge since it does not add any code by default, but I doubt we include it in any pre-compiled build due to its size.

I actually implemented a web socket server in Berry but couldnt get it to converge, and it actually did get quite big on its own, even without file serving. Then I tried  with PsychicHTTP, and the Cpp/C  interface was unwieldy, so finally settled on esp_http.

I understand your concern about code size but for those niche applications where folk are trying to push Tasmota to the limit I think it has a place.

Also in passing I wanted to say I think the adoption of Berry in Tasmota was a genius move; it has taken a great project to another level, and given it a unique position in the world of connected gadgets. My congrats and thanks to everyone on the Tasmota team!

lib/libesp32/berry/default/be_modtab.c
jetpax: Changed to USE_BERRY_HTTPSERVER

lib/libesp32/berry/default/be_modtab.c
jetpax: Chnaged to USE_BERRY_WSSERVER

lib/libesp32/berry_tasmota/src/be_webfiles_lib.c
s-hadinger: I tried brotli serving, but unfortunately I discovered the hard way that brotli is disabled when using HTTP. brotli requires HTTPS. This seems to make any brotli support useless.

s-hadinger:
I'm ok to merge but as mentioned, it's too heavy to be included in pre-compiled builds (I'm repeating myself here for new readers that may have missed this above).

s-hadinger:
I'm curious about brotli support for pre-compiled assets. My tests show the brotli requires HTTPS and is disabled in browsers for HTTP. Do you confirm?

s-hadinger:
I did a small change in #23256 to avoid 500 bytes to be used in flash when the feature is not enabled.

sfromis:
The story about Brotli not used on plain old http (instead of https) seems to be a deliberate decision to avoid various middleware (like proxies) screwing up a compression format they don't know about.

Noschvie:
How can this feature be used?
Used the defines from the first post to compile the firmware, but `import httpserver` leads to `BRY: Exception> 'import_error' - module 'httpserver' not found`

What else is missing here? Thanks

sfromis:
The source code shows that the Berry module `httpserver` depends on:
`#define USE_BERRY_HTTPSERVER`

This commit shows the name change: https://github.com/arendst/Tasmota/pull/23206/commits/908146012a7596e5ef56789876457a53d3bea26b

Noschvie:
> The source code shows that the Berry module `httpserver` depends on: `#define USE_BERRY_HTTPSERVER`
> 

oh... got it, thanks again!

jetpax:
> I'm curious about brotli support for pre-compiled assets. My tests show the brotli requires HTTPS and is disabled in browsers for HTTP. Do you confirm?

Yes good catch, I'll switch it to gzip instead, its almost as good and probably more relevant to our use case

Noschvie:
@jetpax Is there a way to support POST requests like this? Thanks

`curl -X POST -H "Content-Type: application/json" -d "{ ... }" http://192.168.7.118:8080/api/`

Noschvie:
@jetpax thanks again for this feature, I tested it and it works perfectly!

Noschvie:
Hello @jetpax 
The latest development seems to have a problem with this feature, resulting in the following error message:

```
lib/libesp32/berry_tasmota/src/be_webfiles_lib.c: In function 'w_webfiles_serve':
lib/libesp32/berry_tasmota/src/be_webfiles_lib.c:329:14: error: 'httpd_uri_t' {aka 'struct httpd_uri'} has no member named 'is_websocket'
  329 |             .is_websocket = false,
      |              ^~~~~~~~~~~~
*** [.pio/build/tasmota32-httpsrv/libb62/berry_tasmota/be_webfiles_lib.c.o] Error 1

```

s-hadinger:
Yes indeed, let me do a quick fix

s-hadinger:
Fixed

Noschvie:
@s-hadinger Thanks for the fix, but unfortunately this causes the following error:

```
lib/libesp32/berry_tasmota/src/be_wsserver_lib.c:100:31: error: unknown type name 'httpd_ws_frame_t'
  100 | static void log_ws_frame_info(httpd_ws_frame_t *ws_pkt);
      |                               ^~~~~~~~~~~~~~~~
lib/libesp32/berry_tasmota/src/be_wsserver_lib.c:123:31: error: unknown type name 'httpd_ws_frame_t'
  123 | static void log_ws_frame_info(httpd_ws_frame_t *ws_pkt) {
      |                               ^~~~~~~~~~~~~~~~
```

s-hadinger:
This won't work since esp-idf is now compiled without `#define CONFIG_HTTPD_WS_SUPPORT`

@Jason2866 What would it take to have `#define CONFIG_HTTPD_WS_SUPPORT` optional ?

Jason2866:
@s-hadinger will have a look. Should work to set the Option with HybridCompile `custom_sdkconfig =...`

s-hadinger:
@Jason2866 Confirmed, adding `custom_sdkconfig        = CONFIG_HTTPD_WS_SUPPORT=y` makes the compilation succesful

@Noschvie I will add some guarding directives so that the compilation fails if this parameter is not specified (it will be more explicit). But keep in mind that it will not compile in the standard build.

Jason2866:
@s-hadinger will add the option in Arduino Lib Builder. Seems it has no impact in flash usage. Will be "in" Platform release in May

s-hadinger:
Even better thanks

Noschvie:
Thank you so much for your commitment!

Noschvie:
@s-hadinger  Adding `custom_sdkconfig = CONFIG_HTTPD_WS_SUPPORT=y` results in a successful compilation.

However, importing httpserver leads to the following error:
`BRY: Error executing compiled code (import_error - module 'httpserver' not found).`

What else needs to be done? Thanks!

Noschvie:
@jetpax Do you have any idea why this feature no longer works? Thanks!

sfromis:
Not trying to answer fully, I can [repeat myself](https://github.com/arendst/Tasmota/pull/23206#issuecomment-2783953707) from over a year ago.

> The source code shows that the Berry module `httpserver` depends on:
> `#define USE_BERRY_HTTPSERVER`
> 
> This commit shows the name change: > https://github.com/arendst/Tasmota/pull/23206/commits/908146012a7596e5ef56789876457a53d3bea26b

Noschvie:
I'm already using these settings and it compiles without errors. It's just that the module can't be loaded.

```
#ifndef USE_BERRY_WSSERVER
 #define USE_BERRY_WSSERVER
#endif
#ifndef USE_BERRY_HTTPSERVER
 #define USE_BERRY_HTTPSERVER
#endif
#ifndef USE_BERRY_WEBFILES
 #define USE_BERRY_WEBFILES
#endif
```

sfromis:
Well, doing a build with those 3 `#define` symbols, using `import httpserver` works for me.

Noschvie:
Everything is fine now, the solution lay in omitting `custom_sdkconfig = CONFIG_HTTPD_WS_SUPPORT=y`
Thanks again!

PR #23259, opened 2025-04-07
Alpinejs 282

s-hadinger:
## Description:

Add AlpineJS 2.8.2 available under `/alpinejs` URL. It is stored in `gzip` format and served as compressed file to the browser (most browser accept gzip compressed files).

It is not enabled by default for now and requires `#define USE_ALPINEJS`. The goal is to make it enabled in non-minimal builds in the near future when testing is done.

Added flash size: +8.8KB (no ram impact)

## Checklist:


arendst:
Using your PR I decided to free the claimed device driver id due to deminishing amount of free id's.

It's now an include file.

PR #23271, opened 2025-04-10
Mitsubishi electric HVAC set temp should be reported as a float

echo-bravo-yahoo:
## Description:

There's a description of the issue in the commit message, but the heart of it is that the set temp for mitsubishi electric HVAC systems is often a float, but is always reported on the RESULT topic as an int. I have not yet had a chance to try this on an ESP8266, and I do not have an ESP32 to try it with.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_44_miel_hvac.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_44_miel_hvac.ino
@@ -899,7 +899,7 @@ miel_hvac_cmnd_settemp(void)
 		update->temp05 = miel_hvac_deg2temp(degc);
 	}
 
-	ResponseCmndNumber(degc);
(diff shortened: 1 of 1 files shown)


arendst:
Providing a PR without even do a test compile (which fails) isn't really helpful.

Anyway, I managed to change the temp to float while decreasing code size considerably on the way. Use command `TempRes` to control amount of decimals.

echo-bravo-yahoo:
Sorry, I left it in draft with unchecked items in the checklist because it _wasn't_ ready for merge. I looked in README, CONTRIBUTING, API, and BUILDS, and didn't find any information on _how_ to run the test suite. I did compile to confirm that worked, and was planning on flashing an ESP8266 today. Next time, would you prefer I just cut a ticket?

Jason2866:
Discussions is a perfect place for

PR #23315, opened 2025-04-18
xdrv_04_light: force light refresh every second

mringwal:
## Description:

This PR triggers a period refresh of all lights every second. This fixes the issue of electric noise causing some WS2812 LEDs to turn on although they are turned off in Tasmota. By refreshing them every second, the WS2812 always reflect the current requested state.

See Discussion: https://github.com/arendst/Tasmota/discussions/23180

This is a direct fix/hack. I'm open for other approaches, e.g. adding a custom command which would refresh the lights from a Berry Script.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_04_light.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_04_light.ino
@@ -3460,6 +3460,10 @@ bool Xdrv04(uint32_t function)
       case FUNC_EVERY_50_MSECOND:
         LightAnimate();
         break;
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks for the PR, although I don't agree to force a refresh every second to all displays. Your problem is specific, and not the case for general WS2812, so doing this for every LED strip would generate more problems than it solves.

The alternative would be to have a new command like `PixelRefresh` or whatever name is suitable, then you can have a **rule** that calls this function every second. The result would be the same for you, and would not impact other users.

PR #23347, opened 2025-04-27
Added Wireguard VPN

s-hadinger:
## Description:

Added support for WIreguard VPN for both ESP8266 and ESP32, based on the works done on ESPHome and https://github.com/droscy/esp_wireguard.

Wireguard needs to self-compile and `#define USE_WIREGUARD`, which adds between 22KB and 29KB of Flash, as well as 0.9KB of RAM.

Configuration is contained in a file names `wireguard.conf`, example:
```
[Interface]
PrivateKey = <priv_key>
Address = 192.168.254.6/24
DNS = 9.9.9.9, 149.112.112.112

[Peer]
PublicKey = <pub_key>
PresharedKey = <psk>
Endpoint = <endpoint>:<port>
AllowedIPs = 0.0.0.0/0
PersistentKeepalive = 25

[Tasmota]
Netmask = 0.0.0.0
AutoConnect = 0
```

Tasmota-specific attributes:
- `[Tasmota]/Netmask`: contains the mask used to redirect traffic to Wireguard instead of standard network. Default is `0.0.0.0` which means that all traffic is redirected to VPN. Set to `255.255.0.0` if you want only `192.168.x.x` to be directed to VPN (assuming the `Address` is within `192.168.x.x`)
- `[Tasmota]/AutoConnect`:  if `true`, Wireguard tunnel is created as soon as a network connection is up, and auto-reconnects if networks goes off and on again.

Commands:
- `WGConnect`: establish connection, shows `WG : Wireguard peer UP` when peer is up (i.e. handshake was succesful)
- `WGStop`:  stop Wireguard connection, shows `WG : Wireguard peer DOWN` when peer is down. If `AutoConnect` was `true`, `WGStop` sets it to `false` to avoid auto-reconnect (note: configuration is not changed, so `AutoConnect` will be enabled again after a reboot).

Documentation will be soon added.

## Checklist:


ptath:
> Wireguard needs to self-compile and #define USE_WIREGUARD

Done, but

> Configuration is contained in a file names wireguard.conf

Please tell us where (and when) this file should be located =)

`WGConnect` and `WGStop ` commands throw an error  

```
RSL: RESULT = {"Command":"Unknown","Input":"WGCONNECT"}
```

PR #23352, opened 2025-04-28
Optional Wifi strength indicator in WebUI status line

s-hadinger:
## Description:

Optional Wifi strength indicator in WebUI status line, requires `#define USE_WEB_STATUS_LINE_WIFI` (not included in pre-compiled binaries).

Adds 0.5KB of Flash.

Example:
image

When hovering the cursor on the Wifi icon, you get additional information
image

## Checklist:


sfromis:
What about also indicating Wifi activity? Basically the function of `LedLink` also without having a physical led.

s-hadinger:
Why not, but keep in mind it's refreshed every 2 seconds - so no real-time. Thus blinking would not work. What do you have in mind?

sfromis:
While physical `LedLink` does have the "now" advantage, what I meant was indicating activity in the last `WebRefresh` interval, which since very [recently](https://github.com/arendst/Tasmota/commit/6344c46d0817783353a7c84ec935fb8189bcfc36) can be down to 400 ms, basically slow blink. I'm assuming that the indicator updating happens by this beat.

Of course, not a big deal, just an idea. And it's not a thing which would be straightforward to do from Berry events `web_status_line_left` / `web_status_line_right`, unlike indicating status info like memory consumption.

s-hadinger:
Ah, I missed that `WebRefresh` could be set as low as 400ms.

Displaying free heap is as simple as:
```berry
class MemSticker
   def init()
     tasmota.add_driver(self)
   end
   def web_status_line_left()
     import webserver
     webserver.content_send(f"<span>&nbsp;{tasmota.memory('heap_free')}k</span>")
   end
end
MemSticker()
```
image

sfromis:
Yeah, that was what I meant about free memory and Berry events. No such access to Wifi activity.

Jason2866:
The free heap is redundant info. It is easy available in info page. Like the small usefull wifi and mqtt info icons, but imho such icons should be used as less as possible.

s-hadinger:
Agreed. Free heap is mostly a debugging feature that provides real-time memory evolution. It’s not useful to most users

PR #23462, opened 2025-05-22
Enhancement: WS2812 address leds by range 

ptrooms:
## Description: Enhances ws2812 addressing to facilitate colouring ledranges using the existing `led` command

**Related issue (if applicable):** individual addressing leds is painfully slow

## Checklist:

This PR facilitaties to address leds in ranges. For example one canb control by the `led` command the ranges via the following parameters:

- `ledX` - led 1 to/included X%`Pixels` ( X % Pixels e.g. remainder after integer division by `Pixels`)
- `led1xxxyyy` - where xy is number to control led _start_ (xxx / Pixels : inger division) until/including _end_ (yyy % Pixels : remainder). If xxx < yyy then the addressing is reversed using Pixels as offset. Note: 000<number<999  is to be fixed with leading zeroes.

Example: assume Pixels = 10 ; then the follwoing command:
- **led**5 100000 - normal addresses led5 to Red
- **led**14 001000 - led 1  thru 4 are coloured  Green
- **led1**005007 000010 - led 5 thru 7 is coloured blue
- **led1**009008 101010 - led 9 backward thru 8 is made white

By this one can quickly color a range of leds, specially handy to create VU-bars starting at 1, last led of a range inside the string of leds. 
Note: In case the command is used without color-paramter, the color value of the first led addressed is returned.

Hope this enhancement helps and inspires. In case of questions or improvements, let me know.


sfromis:
That syntax breaks existing use cases for the `Led` command, as `Led14 001000` currently means to set pixel number 14 to green.

Common Tasmota command formats will separate values by blank or comma, instead of stipulating fixed width and leading zeros for formats. JSON arguments are also used for "less simple" argument use cases.

ptrooms:
> That syntax breaks existing use cases for the `Led` command, as `Led14 001000` currently means to set pixel number 14 to green.
> 
> Common Tasmota command formats will separate values by blank or comma, instead of stipulating fixed width and leading zeros for formats. JSON arguments are also used for "less simple" argument use cases.

No. 

To explain: my routine uses the (number of Tasmota defined) `Pixels` to (s)elect if and when the number exceeds this value and thus must transition to bar-addressing.
F.e. Suppose number of `Pixels` is set to 10, then led**14** 000100 will coulor led 1 thru 4 to blue. If Pixels was (default 512) then led14 on a ws2814 strip with 10 leds, wouldtry to colour unexisted led14.
Technically the number in command **led1**_017023_ 001000 exceeds technically a WS2814 strip of f.e. 40-leds (Pixels =40) but in this case, it will colour the leds 17 thru 23 into Green.

Of course one could introduce e new commmand and format etc.etc. which will give more work and imo is really not needed when one understand the operational syntax(is) of the number in relation to `Pixels`.

sfromis:
The `PulseTime` example is very different, as the command syntax does not vary depending on factors not part of the command itself. Hence command parsing will always be the same, before starting to apply the command to current configuration. Anyway, what makes it "less understandable" is to just have more digits without any indication of multiple fields.

And using comma or blank as delimiters in commands is used in parsing of quite a few standard Tasmota commands, specifically including the `Led` command which already now supports multiple color values for consecutive pixels - did you verify that this still works? Also commands like `Buzzer`, `DevGroupShare`, `Interlock`, `TimedPower`, `DspLine`, `PWMFrequency`, `TimeDST`, `UFSRename`, etc.

ptrooms:
Bump.
Let me know if and how to proceed of progress things .... able to adress leds in strings, works great.

arendst:
Glad it works for you. I'll take a look in July how to move this forward.

arendst:
kick

PR #23504, opened 2025-06-02
Update xsns_102_ld2402.ino, command table broken

sfromis:
Command pointer table broken after refactoring https://github.com/arendst/Tasmota/commit/3a28938b9153d01e7ccef0a03d981251ced56123

The error with a comma before an entry instead of after, to match the other pointers, looks very obvious to me. Hence, I'm submitting this PR without having the hardware, and thus not able to do a runtime test.

For context, I'm playing with extracting available commands from the source code, doing simple parsing of the reasonably recognizable format of the command table entries, leading to me noticing such an anomaly.

## Description:

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.1.3.250504


pkkrusty:
Verified with hardware that this PR fixes the problem and works as expected.

PR #23597, opened 2025-06-25
CSS uses named colors variables

s-hadinger:
## Description:

No functional change, now CSS style starts with the following (actual values are taken from `WebColor` settings):

```css
:root{--c_bg:#252525;--c_frm:#4f4f4f;--c_ttl:#eaeaea;--c_txt:#eaeaea;--c_txtwrn:#ff5661;
--c_txtscc:#008000;--c_btn:#1fa3ec;--c_btnoff:#08405e;--c_btntxt:#faffff;--c_btnhvr:#0e70a4;
--c_btnrst:#d43535;--c_btnrsthvr:#931f1f;--c_btnsv:#47c266;--c_btnsvhvr:#5aaf6f;--c_in:#dddddd;
--c_intxt:#000000;--c_csl:#1f1f1f;--c_csltxt:#65c115;}
```

When colors needs to be added to CSS elements, it is done like this:

```css
a{color:var(--c_btn);
```

Adds ~250 bytes to flash

## Checklist:


arendst:
What made you decide to change from (double %% to escape formatting)
```
const char HTTP_HEAD_STYLE1[] PROGMEM =
  ...
  "input{width:100%%;box-si ...

WSContentSend_P(HTTP_HEAD_STYLE1);
```
to (single % as not a format)
```
const char HTTP_HEAD_STYLE1[] PROGMEM =
  ...
  "input{width:100%;box-si ...

WSContentSend_P(PSTR("%s"), HTTP_HEAD_STYLE1);
```
?

By doing this it introduces different representations of the the use of the function `WSContentSend_P()` (as we do ` WSContentSend_P(HTTP_HEAD_STYLE_ZIGBEE);` which has no formatting) AND add 8 bytes for every case `WSContentSend_P(PSTR("%s"),..)` is used.

I prefer the status quo on this so if no compelling reason I'll revert the double quotes.

s-hadinger:
Ah yes indeed. It was to prevent unwanted formatting issues. I will revert.

s-hadinger:
Actually I realized that there is no WSContentSend_P variant that just outputs the string to the webclient without any formatting. This would incur less cycle since the snprintf is done for nothing. Such function could potentially reduce the code size

arendst:
You're right. If you manage to decrease code size I'm all in ;-)

PR #23601, opened 2025-06-27
Update xsns_16_tsl2561.ino

pamribeirox:
On detect make sure the device ID is the correct to avoid misdetection of other devices using the same I2C address

## Description:

## Checklist:


sfromis:
This change breaks TSL2561 support, at least in some cases. https://github.com/arendst/Tasmota/discussions/23694

According to one TSL2561 data sheet:
image
Source: https://eu.mouser.com/datasheet/2/588/TSL2561_DS000110_3_00-2066792.pdf

Looks like a new revision TSL2561T got a different part number.

PR #23819, opened 2025-08-23
Add ESP32 ROM SHA Hardware Acceleration to BearSSL

Staars:
## Description:

# Add ESP32 ROM SHA Hardware Acceleration to BearSSL

## Summary
This PR adds a new source file, `_sha_rom_idf5x.c`, to Tasmota’s BearSSL library, enabling hardware‑accelerated SHA hashing on ESP32 devices via the ROM SHA engine.  
Until now, BearSSL in Tasmota relied solely on software implementations for all hash algorithms.

## Details
- **New file:** `_sha_rom_idf5x.c`
- **Functionality:**
  - Integrates ESP32 ROM SHA acceleration for **SHA‑1**, **SHA‑224**, **SHA‑256\***, **SHA‑384**, and **SHA‑512**.
  - Uses a single global ROM SHA context, serialized with a global spinlock to prevent concurrent hardware access.
  - Falls back to BearSSL’s software implementation when hardware acceleration is unavailable or unsafe (e.g., SHA‑256 without resume support).
- **Integration:**
  - Hardware acceleration is enabled at build time with setting in `platformio_tasmota32.ini`:
    ```bash
    -DUSE_SHA_ROM
    ```
    Without this flag, BearSSL will continue to use its software hashing code.
- **Linker override note:**
  - The leading underscore in the filename (`_sha_rom_idf5x.c`) is **intentional**.
  - This ensures the linker overrides specific BearSSL functions with the hardware‑accelerated versions provided here.
- **Concurrency note:**
  - Only **one hash context** can be active at a time.
  - Interleaving operations from multiple contexts will corrupt state.
- **Platform scope:**
  - **ESP32:** Hardware acceleration enabled as described above.
  - **ESP8266 / other targets:** No change; software BearSSL remains in use.

## Benefits
- Significant performance improvement for supported SHA algorithms on ESP32.
- Reduced CPU load during hashing operations.
- **Smaller firmware size** — saves approximately **2.5 kB to 6.5 kB** compared to the pure software implementation.

## Limitations
- SHA‑256 acceleration is disabled on ESP32 variants without `SOC_SHA_SUPPORT_RESUME`.
- Not safe for concurrent hashing across multiple contexts.

## Testing
- Verified correct hash outputs against known test vectors.
- Benchmarked performance improvements and measured firmware size reduction over software implementation.

Created with kind assistance from Chat-GPT5.0.

## Checklist:

### DIFF
--- a/lib/lib_ssl/bearssl-esp8266/src/hash/_sha_rom_idf5x.c
+++ b/lib/lib_ssl/bearssl-esp8266/src/hash/_sha_rom_idf5x.c
@@ -1,0 +1,393 @@
+/* _sha_rom_idf5x.c — BearSSL drop‑in with ESP32 ROM SHA acceleration
+ *
+ *  ⚠ TEST‑ONLY IMPLEMENTATION — SERIALIZED VIA GLOBAL SPINLOCK ⚠
+ *
+ *  Overview:
+ *  ---------
+ *  - Provides BearSSL hash functions backed by the ESP32 ROM SHA engine.
+ *  - Uses a single global ROM SHA_CTX instance.
+ *  - Access is serialized with a global portMUX spinlock.
+ *
+ *  Concurrency Warning:
+ *  --------------------
+ *  - This locking only prevents simultaneous entry into the ROM SHA engine.
+ *  - It does NOT make multiple concurrent hash contexts safe.
+ *  - Interleaving operations from different contexts will corrupt state.
+ *  - Only one hash context may be active at any given time.
+ *
+ *  Platform Notes:
+ *  ---------------
+ *  - ESP32 (ESP_PLATFORM): Uses ROM SHA acceleration as described above.
+ *  - ESP8266 / other targets: This translation unit compiles to nothing;
+ *    stock BearSSL symbols are provided instead.
+ *
+ *  SHA‑256 Limitation on ESP32:
+ *  ----------------------------
+ *  - The ESP32 ROM SHA‑256 implementation shares internal state with the
+ *    hardware engine; without pause/resume support, state cannot be saved
+ *    and restored safely.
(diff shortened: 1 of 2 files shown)


Staars:
## SHA256 speed – ESP32‑C3

### Software Only (No Hardware Acceleration)

| Size (B) | Iterations | Elapsed (s) | KB/s    |
|---------:|-----------:|------------:|--------:|
| 1024     | 1000       | 0.924       | 1082.25 |
| 4096     | 250        | 0.572       | 1748.25 |
| 8192     | 125        | 0.513       | 1949.32 |

---

### With Hardware Acceleration

| Size (B) | Iterations | Elapsed (s) | KB/s    |
|---------:|-----------:|------------:|--------:|
| 1024     | 1000       | 0.409       | 2444.99 |
| 4096     | 250        | 0.122       | 8196.72 |
| 8192     | 125        | 0.074       | 13513.5 |

There is a good amount of overhead in this simple Berry benchmark, but the ratio should be something like that and the difference in the length of the whole benchmark is definitely perceptible.

Another note: Matter still works with HW acceleration.

Jason2866:
why not enabling by default?

s-hadinger:
Nice work. Let's do some testing. However I
I'm nkt sure it has a measurable effect on TLS connections

Staars:
Good point. Probably the only way to know, if this produces unwanted side effects.

I  do not expect earth shattering effects in real world speed up either, but binary size decrease alone is something I like.

s-hadinger:
Code size reduction is always good!

Jason2866:
So let us enable this by default. Okay?

s-hadinger:
Let's do it. We can always revert in case of problem

PR #23836, opened 2025-08-26
core/net: add AsyncHttpClientLight + Berry async webclient; keep HTTPClient.h enums; replace old backend by default

TheHexaMaster:
## Description:

Introduce a self-contained async HTTP/HTTPS client for ESP32 (AsyncHttpClientLight) and integrate it directly into the Berry webclient module. The old HttpClientLight backend is replaced by default (no feature flag). Code still include <HTTPClient.h> only for the enum/constant types (e.g., followRedirects_t, transferEncoding_t, HTTPC_TE_*).

**Why**

- Non-blocking GET/POST for telemetry/RPC without freezing the Berry VM.
- Lower TIME_WAIT pressure and safer connection reuse.
- TLS public-key pinning, optional RSA-only hardening.
- Keep API/error-code compatibility while modernizing internals.

**What changed**

**AsyncHttpClientLight (new core client)**

- HTTP via lwIP sockets (IClientLite + non-blocking AsyncTcpAdapter) with SO_LINGER (RST-close) to reduce TIME_WAIT.
- HTTPS via BearSSL WiFiClientSecure_light, with:
- tls_pin_pubkey(fp1[,fp2]) (SHA-1 pubkey fingerprints, one or two pins)
- tls_set_rsa_only(true) optional hardening.
- Redirects (strict): 301/307 for GET/HEAD; 302/303 switch to GET; configurable limit (default 5).
- Chunked & identity decoding; reuse honored unless prior body wasn’t drained (auto reconnect).
- FreeRTOS worker (WDT-friendly): asyncGETStart(), asyncPOSTStart(...), asyncState(), asyncAbort().
- Error codes aligned with HTTPClientLight (−1..−11).
- Async in-RAM body buffer capped ~16 KB (heap safety). Use sync streaming for larger bodies.

**Berry webclient (updated bindings)**

- New methods: async_get_start(), async_post_start(str|bytes), async_state(), async_abort().
- async_state() publishes last_code (HTTP or negative error) and last_rx (received bytes), and frees hidden holds for byte payloads.
- Headers added via add_header() persist per instance; they reset on begin() and after redirects (re-add if needed).
- Sync API unchanged: GET/POST/PUT/PATCH/DELETE, get_string/bytes, write_file/flash.

**Dependency note**

<HTTPClient.h> remains included only for enum/constant types. No Arduino HTTPClient source code is copied; transport, parsing and async infra are independent.

**Compatibility**

- Existing Berry scripts using sync APIs remain compatible.
- OTA/firmware workflows (sync GET + streaming) are unaffected.

**Operational notes**

- Call async_state() at least once after an async job finishes to publish results and clean up before reusing the same instance.
- Add Content-Type once after begin() (e.g., application/json); re-add after begin() or a redirect.
- Don’t call close() while an async job is running.

**Typical use cases:**

**Secure periodic telemetry over HTTP(S) + local MQTT**

- Use webclient for encrypted REST telemetry to the cloud while keeping MQTT for LAN/HA use.
- Pattern: create one client per endpoint, begin() once, add Content-Type: application/json, then call async_post_start(payload) from your scheduler and occasionally async_state() to publish last_code/last_rx and clean up.
- Keeps the Berry VM responsive; no long blocking writes. For large uploads/downloads, fall back to sync streaming (write_file, get_bytes).

**RPC long-polling (e.g., ThingsBoard) without tight polling loops**

- With async GET you can keep a long-poll open (e.g., ?timeout=60) instead of reconnecting every few seconds.
- Pattern: async_get_start() on /rpc?timeout=60, periodically call async_state():
- When async_state() returns DONE, parse the body, execute the command (optionally via async POST reply), then immediately start the next long-poll (async_get_start() again).
- When it returns RUNNING, do nothing (connection is open); VM stays free for other tasks.
- This reduces connection churn, lowers TIME_WAIT pressure, and prevents VM stalls seen with previous sync blocking mode.

**Risk**

- ESP32-only paths; HTTPS relies on existing BearSSL light.
- Behavior changes confined to the webclient backend; public sync APIs preserved.

**Test plan**

- Sync: GET large binary → write_file(); verify size/hash; chunked and identity.
- Async: burst async_post_start() with periodic async_state(); check last_code/last_rx, stable heap, no WDT.
- Redirects: verify 301/302/303/307 and header reset behavior.
- TLS: good vs bad fingerprint; RSA-only against ECDSA cert (should fail).

## Checklist:

### DIFF
--- a/lib/libesp32/HttpClientLight/src/AsyncHttpClientLight.cpp
+++ b/lib/libesp32/HttpClientLight/src/AsyncHttpClientLight.cpp
@@ -1,0 +1,1572 @@
+/*
+ * AsyncHttpClientLight.cpp  - Created on: 26.08.2025
+ *
+ * Copyright (C) 2025 by Martin Macák <macak@msb.sk>
+ *
+ * Async HTTP/HTTPS client for ESP32 (WDT-friendly)
+ *
+ * Features
+ * - HTTP: lwIP socket transport (IClientLite + AsyncTcpAdapter) with SO_LINGER (RST-close) to reduce TIME_WAIT pressure
+ * - HTTPS: BearSSL::WiFiClientSecure_light
+ * - API and error codes compatible with Arduino HTTPClient/HTTPClientLight semantics
+ *
+ * Async extensions
+ * - asyncGETStart(), asyncPOSTStart(): launch a request in the background (FreeRTOS worker), non-blocking for the caller
+ * - asyncState(): query status; after ASYNC_DONE you can use getString()/getSize()/writeToStream() without further I/O
+ * - asyncAbort(): best-effort cancellation of an in-flight async request
+ * Note: in async mode the response body is stored in an internal RAM buffer (hard cap kAsyncBodyLimit).
+ *
+ * Acknowledgments: API/error-code compatibility inspired by Arduino HTTPClient / Tasmota HTTPClientLight.
+ * This file contains an independent implementation; no source code was copied.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+
+
+#include "AsyncHttpClientLight.h"
+#include <base64.h>
+#include "WiFiHelper.h"
+#include <ctype.h>
+
+// --------- Utils (SHA1 pin parsing) ------------------------------------
+static inline int hexval_local(char c) {
+  if (c >= '0' && c <= '9') return c - '0';
+  c = toupper((unsigned char)c);
+  if (c >= 'A' && c <= 'F') return 10 + (c - 'A');
+  return -1;
+}
+static bool parse_sha1_pin_local(const char* in, uint8_t out[20]) {
+  if (!in) return false;
+  char buf[41]; int n = 0;
+  for (const char* p = in; *p && n < 40; ++p) {
+    if (*p == ':' || *p == ' ') continue;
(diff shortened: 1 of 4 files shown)


TheHexaMaster:
ping :)

arendst:
@s-hadinger do we need this? As it stands it changes berry webclient only so that's why I ask you.

s-hadinger:
Yes. We have already discussed on WhatsApp. I will review it

Jason2866:
@s-hadinger Is this part used in safeboot? We cant accept any code size increase in safeboot.

TheHexaMaster:
> @s-hadinger Is this part used in safeboot? We cant accept any code size increase in safeboot.

In default safeboot build flags in platformio_tasmota32.ini is defined:

`[safeboot_flags]
lib_ignore                  = ${esp32_defaults.lib_ignore}
                              ESPmDNS
                              LinkedList
                              ESP Mail Client
                              IRremoteESP8266
                              NeoPixelBus
                              OneWire
                              EEPROM
                              EEPROM 24C128_256_521
                              MFRC522
                              JPEGDEC
                              universal display Library
                              ESP8266Audio
                              ESP8266SAM
                              FFat
                              SD
                              SD_MMC
                              UdpListener
                              Berry
                              Berry mapping to C
                              Berry Tasmota mapping
                              Berry int64 implementation for 32 bits architecture
                              Berry Matter protocol implementation
                              Berry animation library for WS2812 leds
                              Micro-RTSP
                              re1.5
                              DHT sensor library
                              ccronexpr`  
                              
I personally did not investigate safeboot build process, but im assuming in safeboot partition the whole berry is disabled and even no C bindings are made. The new asynchttp hooks are made in xdrv_52_3_berry_webserver.ino used by berry. Other native parts of tasmota code using legacy httpclientlight are untouched. @s-hadinger let me know if im wrong.

TheHexaMaster:
ping

s-hadinger:
Hi I finally got there, sorry for the delay. This awesome, but I'm not comfortable in replacing everything the was fully tested with this new implementation without keeping a fallback solution. Can you make it as an option with a feature-flag?

TheHexaMaster:
> Hi I finally got there, sorry for the delay. This awesome, but I'm not comfortable in replacing everything the was fully tested with this new implementation without keeping a fallback solution. Can you make it as an option with a feature-flag?

Sure Stephan, i can make it with user-config override selectable option. I think better then to close and refuse this request, ill refactor later against latest version to ensure compatibility. Thank you.

PR #23847, opened 2025-08-28
Update web UI styles and HTML syntax

SteWers:
## Description:
Merged HTTP_HEAD_STYLE2 into HTTP_HEAD_STYLE1 and removed redundant style files to decrease code size. Updated button and form HTML across web UI to use consistent markup and reduce HTML syntax warnings. Adjusted various style and separator elements for improved HTML5 compliance and consistency.
## Checklist:


tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
```diff
@@ -2594,7 +2591,7 @@ void HandleWifiConfiguration(void) {
                   for (uint32_t k = 0; k < 4; ++k) {
                     WSContentSend_P(PSTR("<i class='b%d%s'></i>"), k, (num_bars < k) ? PSTR(" o30") : PSTR(""));
                   }
-                  WSContentSend_P(PSTR("</span></div></div>"));
+                  WSContentSend_P(PSTR("</div></span></div>"));
```
arendst, line 2597: Wrong. The original is correct

arendst:
While it may comply to HTML5 it definitly changes the layout (changing location of ``<p>`` vs ``</p>``) Also removing ``<p>`` for single line html code makes programming menagement a nightmare.

So, as it stands it won't be merged as "HTML5compliance" != correct layout.

tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
SteWers: Hm. I will investigate again after my vacation.

tasmota/html_uncompressed/HTTP_HEAD_STYLE1.h
```diff
@@ -1,5 +1,4 @@
 const char HTTP_HEAD_STYLE1[] PROGMEM =
```
arendst, line 1: I understand why you want to merge these files into one file. The problem is that it most likely fails on an ESP8266 due to lack of memory.

As the ESP8266 has scarce memory I had to jump through all kinds of hoops to produce a stable GUI. Many code had to be written in Javascript to be executed on the client to expand to full HTML code. Also the use of chunked HTML pages allowed generating the HTML.

To keep memory use at a minimum the chunked buffer is around 1k RAM. The webserver lib will double this. If you merge these files then the single file will be over 3k RAM generated from UNISHOX (as that's the default state) bypassing the chunked buffer, as designed, and be doubled by the webserver. This will most likely lower free memory below the as stable considered 10k RAM.

So do not change these files.

tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
SteWers: I'd a look at the resulting HTML code.

This is the code before the PR:
```html
1:	<div title='-25 dBm (100%)'>
2:		<a href='#p' onclick='c(this)'>SSID</a>
3:		<span class='q'>
4:			<div class='si'>
5:				<i class='b0'></i><i class='b1'></i><i class='b2'></i><i class='b3'></i>
6:			</span>
7:		</div>
8:	</div>
```
We see opening DIV-SPAN-DIV and closing SPAN-DIV-DIV. The lines 6 and 7 have to be swapped.

After the PR it looks like this and the order is correct:
```html
1:	<div title='-25 dBm (100%)'>
2:		<a href='#p' onclick='c(this)'>SSID</a>
3:		<span class='q'>
4:			<div class='si'>
5:					<i class='b0'></i><i class='b1'></i><i class='b2'></i><i class='b3'></i>
6:			</div>
7:		</span>
8:	</div>
```

tasmota/html_uncompressed/HTTP_HEAD_STYLE1.h
SteWers: I wasn't aware of the memory problem. The merge is reverted. The unnecessary `</style><style>` combination is removed anyway.

SteWers:
> While it may comply to HTML5 it definitly changes the layout (changing location of `<p>` vs `</p>`) Also removing `<p>` for single line html code makes programming menagement a nightmare.

This does not only apply to HTML5. `<form>...</form>` between is not allowed between `<p>` and `</p>`. The browser (Firefox) marks this as error:

image

***

`<p><form id=but3 style="display: block;" action='cn' method='get'><button>Einstellungen</button></form></p>`

is interpreted by the browser as:

image


So this PR simplifies and corrects the HTML code. I can not see a nightmare, when using `<p></p>` to create a space between the buttons. There are no (visible) changes of the layout.

tasmota/tasmota_xdrv_driver/xdrv_01_9_webserver.ino
arendst: You're correct.

arendst:
Thx.

arendst:
Pls fix this:

image

arendst:
No worries. I'm fixing it now together with another change I want (button group spacing)

arendst:
Fixed:

image

Issue #23874, opened 2025-09-05
Wireguard stays in failed state if DNS-Request fails


fleinze:
### PROBLEM DESCRIPTION
When wireguard is started (manually or automatically at startup) a dns-request is sent if the endpoint address requires it. If this dns-request fails or times out wireguard stays in a failed state until it is manually restarted with `backlog wgstop;wgconnect`
Other Services like NTP or MQTT retry after failed DNS Requests.
It would be great to add dns retries to wireguard as it would make it more resilient.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [x] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Problem is not scripter related, in this case open a discussion and tag  gemu2015
- [x] Device used (e.g., Sonoff Basic): ZB-GW03-V1.2
- [x] Tasmota binary firmware version number used: 15.0.1.3
  - [ ] Pre-compiled
  - [x] Self-compiled
- [x] Flashing tools used: OTA Update
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:


```
- [ ] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```lua
  Rules output here:


```
- [ ] Provide the output of this command: `Status 0`:
```lua
  STATUS 0 output here:


```
- [x] Set `weblog` to 4 and then, when you experience your issue, provide the output of the Console log:
```lua
00:00:02.284 ETH: IPv4 10.xx.xx.156, mask 255.255.255.0, gateway 10.xx.xx.1
00:00:02.507 WIF: Checking connection...
00:00:02.507 WIF: Attempting connection...
00:00:03.141 WIF: Connecting to AP1 XXXxxx Channel 6 BSSId xx:xx:xx:xx:xx:xx in mode HT40 as tasmota-zbtest...
[...]

PR #23885, opened 2025-09-07
detect invalid DNS IP in WiFiHelper::scrubDNS()

fleinze:
## Description:

The issue is that with the current version of lwIP `WiFiHelper::scrubDNS()` does not work correctly. If `dns_getserver()` returns the IP 0.0.0.0 it is not recognised as invalid IP and saved for later restore. Therefore I added an extra check. Fixes the underlying issue of #23874

## Checklist:


fleinze:
found a better solution, closing PR

s-hadinger:
I'm OK for the extra test. But I'm not convinced by the change in ifdefs. ScrubDns is only useful when ipv6 is enabled, ipv4 only shouldn't need such behavior. Did you find it useful for ipv4 only compiles?

fleinze:
Yes, I also had the issue of overwritten DNS servers with ipv4 only compiles, hence I added the extra ifdefs. 
The DNS issue arises with dual ethernet and wifi ESPs, in my case it is a Zigbee-Gateway.
After startup the gateway gets an ethernet IP (and DNS) first with DHCP. Then Wifi starts and overwrites the DNS configuration. Any DNS requests that are made after Wifi startup and Wifi DHCP now fail. This happens with normal and IPV4-only builds. With the fixed scrubDNS() it works.

fleinze:
See #23886

PR #23886, opened 2025-09-07
Fix scrubDNS()

fleinze:
## Description:

There is an issue with `WiFiHelper::scrubDNS()` where it saves invalid DNS-IPs (0.0.0.0). This is because the static cast from `IPAddress` to `ip_addr_t`does not work correctly in the current version of the esp-idf framework. This PR fixes this. Also fixes the underlying issue of #23874

## Checklist:


fleinze:
I also had the issue of overwritten DNS servers with ipv4 only compiles, hence I added the extra ifdefs.
The DNS issue arises with dual ethernet and wifi ESPs, in my case it is a Zigbee-Gateway.
After startup the gateway gets an ethernet IP (and DNS) first with DHCP. Then Wifi starts and overwrites the DNS configuration. Any DNS requests that are made after Wifi startup and Wifi DHCP now fail. This happens with normal and IPV4-only builds. With the fixed scrubDNS() it works.

eku:
What's the relation between `has_v6` and `USE_IPV6`? The code tests for `has_v6` in branches outside `USE_IPV6`.

Jason2866:
You have probably issues since trying to use `Initialized 'tasmota-zbtest.local'` Using mDNS extension .local is **not** supported with Tasmota. It does not work and creates issues.

fleinze:
Yes, `has_v6` outside of `USE_IPV6` is always set `false`. I could edit it out but it should not matter in the final binary as modern compilers usually detect when an if condition is always true.

fleinze:
@Jason2866 tasmota-zbtest is the hostname of the ESP32. I never made any config nor any `#define` regarding mDNS, I don't know why this ends up in the log.

fleinze:
mDNS-log comes from support_network.ino: 
`AddLog(LOG_LEVEL_INFO, PSTR(D_LOG_MDNS "%s '%s.local'"), (Mdns.begun) ? PSTR(D_INITIALIZED) : PSTR(D_FAILED), TasmotaGlobal.hostname);`

Jason2866:
Ahh, right. It is added for Matter.

eku:
> Yes, `has_v6` outside of `USE_IPV6` is always set `false`. I could edit it out but it should not matter in the final binary as modern compilers usually detect when an if condition is always true.

To my understanding `has_v6` should only exist if `USE_IPV6` is defined and the test for its value should me moved inside `USE_IPV6`. That would make the code more readable.

fleinze:
Ok, I will adjust my code accordingly.

fleinze:
found a small error in the code, need to do another commit

fleinze:
Please review

lib/default/WiFiHelper/src/WiFiHelper_ESP32.cpp
```diff
@@ -84,27 +84,29 @@ void WiFiHelper::scrubDNS(void) {
...
-    const IPAddress ip_dns = IPAddress(dns_getserver(i));
+    const ip_addr_t* ip_dns = dns_getserver(i);
     // Step 1. save valid values from DNS
-    if (!ip_addr_isany_val((const ip_addr_t &)ip_dns)) {
-      if (ip_dns.type() == IPv4 && (has_v4 || !has_v6)) {
-        ip_dns.to_ip_addr_t(&dns_save4[i]);       // dns entry is populated, save it in v4 slot
-      } else if (has_v6) {
-        ip_dns.to_ip_addr_t(&dns_save6[i]);       // dns entry is populated, save it in v6 slot
+    if (ip_dns && !ip_addr_isany_val(*ip_dns)) {
+      if (IP_IS_V4(ip_dns)) {
```
eku, line 98: Can `ip_dns` be a v6-Adress if support is not compiled in?

fleinze:
You are right, this check is not needed if v6 is not compiled in. I changed the code to reflect this.

fleinze:
Can someone please restart the CI checks? It failed because of a dependency timeout, not because of the code change.

s-hadinger:
Sorry for the delay, I finally took time to review. Let's merge, we'll see if it causes problems and we can revert in such case.

Issue #23853, opened 2025-08-30
MQTT TLS connection is broken


e9hack:
### PROBLEM DESCRIPTION
Commit "Add ESP32 ROM SHA Hardware Acceleration to BearSSL (#23819)" breaks MQTT TLS connections.

I have compiled in my root certificate. If connection is tried with SetOption132 as 0, the result is:   

01:57:35.935 MQT: Attempting connection...
01:57:35.983 MQT: TLS connection error: 55
01:57:35.983 MQT: Connect failed to mosquitto.xxx.yyy:8883, rc -2. Retry in 20 sec

If connection is tried with SetOption132 as 1, the result is:

01:56:38.902 MQT: Attempting connection...
01:56:39.026 MQT: TLS connection error: 27
01:56:39.027 MQT: Connect failed to mosquitto.xxx.yyy:8883, rc -2. Retry in 30 sec

If I comment out -DUSE_SHA_ROM in platformio_tasmota32.ini, both connection variants work.

Test device is a CurrySmarter T10 wall socket with an esp32c3.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [ ] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [ ] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [x] Problem is not scripter related, in this case open a discussion and tag  gemu2015
- [x] Device used (e.g., Sonoff Basic): CurrySmarter T10
- [x] Tasmota binary firmware version number used: 15.0.1.3 (tasmota32)
  - [ ] Pre-compiled
  - [x] Self-compiled
- [ ] Flashing tools used: _____
- [x] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
[...]

PR #23902, opened 2025-09-12
Use HAL instead of ROM for SHA HW acceleration

Staars:
## Description:

Complete rewrite of the concept to use a drop in replacement for the software path for SHA operation in BearSSL.
This should solve TLS related failure, which were based on the impossible interaction with multihash.c of BearSSL.

We win a bit less of flash space compared to ROM, which will vary depending on your build environment, in my used environment of a ESP32-C3 this saves about 1kB.

Does still use `-DUSE_SHA_ROM`, which we may refactor later, if no severe new issues arise.

Related to: https://github.com/arendst/Tasmota/issues/23853

## Checklist:


s-hadinger:
Awesome, this is huge!

PR #23911, opened 2025-09-15
Rework WiFi strength indicator

SteWers:
## Description:
- Consolidate WiFi strength indicator in web status line and WiFi setting
- Highlight connected SSID/AP in settings
- Simplify `WifiGetRssiAsQuality`
## Checklist:


Jason2866:
Mhh, resulting in needing more flash space.

SteWers:
Yes, a bit. But when activating `USE_WEB_STATUS_LINE_WIFI` it needs much less than before.

arendst:
I see what you try to achieve but as it adds 268 bytes when not enabled it's a pity.

Also "consolidating" the WiFi symbol on top left status and WiFi Settings screen still breaks wildly used berry provided signal strength symbol like ie LoRaWan:

image

So for now I prefer to keep it as it is.

PR #23914, opened 2025-09-16
Add rule variable TIMESTAMP_MS

andrewjessop:
## Description:

Add new variable for rule evaluation: `TIMESTAMP_MS` which will give the timestamp in millisecond resolution.

e.g.:
```
Rule1 ON IrReceived#Data DO WebQuery http://192.168.4.1/data POST [Content-Type:text/plain]%value%,%deviceid%,%timestamp_ms% ENDON
```

## Checklist:


andrewjessop:
Question to reviewers: what is the etiquette for updating corresponding docs?  Do you raise a corresponding PR and link them?

I'm guessing if accepted, I should add my new variable here: https://github.com/tasmota/docs/blob/master/docs/Rules.md?plain=1#L231

sfromis:
Beware that in general such a variable has rather limited value in rules, outside of your very narrow example. You cannot use it in calculations, which are single-precision floating point, losing more than just the milliseconds. Due to this, I think that having it available would lead to confusion.

PR #23114, opened 2025-03-06
Berry remove 'Leds.create_matrix' from the standard library waiting for reimplementation

s-hadinger:
## Description:

See #23072

Berry `Leds.create_matrix` is currently inconsistent, broken and not used. We are removing this code from the standard solidified library. It's not clear whether we need a full-generic solution, or bespoke version per display type (like Ulanzi).

## Checklist:
  - [x] The pull request is done against the latest development branch
  - [x] Only relevant files were touched
  - [x] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.3.250302
  - [x] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #23935, opened 2025-09-21
add Berry module: Matrix

Staars:
## Description:

Introduce new native Matrix driver for Berry, after the old Berry implementation was removed: https://github.com/arendst/Tasmota/pull/23114
This new version beats the Berry approach in performance and functionality by far and allows to write much shorter Berry application code.

Implements hardware agnostic MatrixCore with support for mono, RGB, and RGBW pixel buffers,
serpentine addressing, and multiple construction modes (bitplane, external buffer, or allocated).

My test object was a Ulanzi clock, which can now use smooth animated screen transitions, text scrolling and fast image operations for little icons, that can be animated too.
I have written a ton of test code, which in that process formed the Berry API. 

Exposes a Berry API for:
- Pixel access via get()/set()
- Buffer operations: clear(), blit(), scroll()
- Brightness scaling and optional tinting
- Mono→color expansion

All operations are in-place with integer math for performance.

## Checklist:

### DIFF
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -260,6 +260,7 @@ be_extern_native_class(AXP2102);
 be_extern_native_class(OneWire);
 be_extern_native_class(Leds_ntv);
 be_extern_native_class(Leds);
+be_extern_native_class(Matrix);
 be_extern_native_class(AudioGenerator);
 be_extern_native_class(AudioFileSource);
 be_extern_native_class(AudioOutputI2S);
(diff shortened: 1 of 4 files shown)


Staars:
Title is not is not completely correct. It is a class ,not a module.

s-hadinger:
Thanks for the implementation. I'm just worried that `Matrix` is a bit too generic. Shouldn't the name contain explicitly leds? Like `Leds_Matrix` or `LedsMatrix`?

Staars:
Good question regarding the name.
The class will probably only be used with some kind of relatively low-count displays (one level below LVGL), which in real life will most of the time be LEDs. But it is not tightly bound to the `Leds` class in Berry. It should be possible to handle every driver, that can provide a pointer to its frame buffer as bytes() in Berry, which of course is only `Leds` right now.
In that sense, a more generic name for it would be: pixel matrix
As I like short names - maybe: `pixmat` ?

But I have no overly strong option about it and can live with whatever name you prefer.

Jason2866:
I will take the Red Pill ;-)

s-hadinger:
`pixmat` is perfect

Jason2866:
the ci error is related to a bug in build system, not from the PR.

PR #23940, opened 2025-09-22
Berry extension manager

s-hadinger:
## Description:

First version of the Tasmota Extension Manager:
Tasmotaext_wifi_running

This feature is for internal testing first, and requires `#define USE_EXTENSION_MANAGER`. Once testing is complete, it should be enabled in default builds.

Documentation is here: https://tasmota.github.io/docs/Tasmota-Extension/

Repository for extensions is here: https://github.com/tasmota/Tasmota-Extensions

## Checklist:


MG695:
Hi,

It looks good as, if I understand well, this can "containerize" my Berry scripts so that I can start/stop (and free memory) them at will, which is my need.
However, looks like only option to do so is to make them public in the online store.
Having this public online store is a great option but I don't need/want to make public my scripts. Any possibility to also have a "local/private store" to manage local/private extensions/scripts ?

Would it be possible to soon add this feature in development binaries so that it can be easily tested without having to build your own firmware ?

Thx

s-hadinger:
Let's start with this first "online store" and we can extend if there is use for it.

Currently the online store is only a discovery service. The installation is only done by copying the `.tapp` file in the `/.extensions/` folder, nothing more.

You can also store the `.tapp` file in the root directory, and it is run anyways (but you can't unload it from the UI)

PR #23949, opened 2025-09-25
BearSSL: use ROM bigint functions in BearSSL

Staars:
## Description:

This change reduces firmware size by replacing BearSSL C implementations of Curve25519 and P‑256 arithmetic with thin wrappers around the ROM’s built‑in big‑integer and ECC functions on supported ESP32 chips (basically all but "old" plain ESP32).
When these functions are hardware accelerated (seems not to be the case on the ESP32-C6 - I will investigate deeper), this will also be faster. Greatest winner in my tests: ESP32-C3.

Keeps the external API (ec_*) unchanged, so higher‑level Ed25519/X25519 and ECDSA/ECDH code continues to work transparently.

Added _ec_c25519_m15.c: Curve25519 scalar multiplication and field operations implemented as calls into the ROM m15 engine.
Added _ec_p256_m15.c: P‑256 scalar multiplication and field operations using the same ROM interface.

These are drop in replacements, that will be activated by default with: `-DUSE_SHA_ROM`
This makes it easy to test for potential problems by simply deleting a file.
If we do not run into unsolvable problems, I would suggest to make a cleaner integration into BearSSL later including the use of a more correct macro name.

## Checklist:

### DIFF
--- a/lib/lib_ssl/bearssl-esp8266/src/ec/_ec_c25519_m15.c
+++ b/lib/lib_ssl/bearssl-esp8266/src/ec/_ec_c25519_m15.c
@@ -1,0 +1,308 @@
+/*
+ * _ec_c25519_m15.c — BearSSL Curve25519 (X25519) implementation using ESP32 ROM-backed Montgomery arithmetic
+ *
+ * This file provides a fast Montgomery ladder implementation for Curve25519 scalar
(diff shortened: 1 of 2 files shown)


Jason2866:
Size decrease in all build variants around 5k Awesome!

Staars:
We must fill the resulting gap ASAP 😀 .

PR #23959, opened 2025-09-28
Adjust the timer configuration page

SteWers:
## Description:
Reduce the number of timers displayed in the configuration page to ensure better appearance on small screens, such as mobile phones.
Use `#define MAX_GUI_TIMERS 10` to set the value of the displayed timers. In this example, there are 10 timers visible.

grafik

This also saves 32 bytes of flash.
## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
@@ -805,15 +805,15 @@ const char HTTP_TIMER_SCRIPT6[] PROGMEM =
   "}"
   "wl(it);";
 const char HTTP_TIMER_STYLE[] PROGMEM =
-  ".tl{float:left;border-radius:0;border:1px solid #%06x;padding:1px;width:6.25%%;}";  // COLOR_FORM, Border color needs to be the same as Fieldset background color from HTTP_HEAD_STYLE1 (transparent won't work)
+  ".tl{border:1px solid var(--c_frm);padding:1px;width:28px;}";  // COLOR_FORM, Border color needs to be the same as Fieldset background color from HTTP_HEAD_STYLE1 (transparent won't work)
 const char HTTP_FORM_TIMER1[] PROGMEM =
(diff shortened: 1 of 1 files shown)


Noschvie:
Wouldn't it be easier to insert a line break after the 8th button?

SteWers:
Yes, that would be possible and I tried it. But it breaks the layout in all builds. It is also possible to break autmatically after the last button, which fits in the space depending on the screen size. Thats all not very nice. So I decided to go this way.

Jason2866:
It is a hack not a solution. It cripples the GUI to adjust all existing timers. What is the way for setting the non visible timers? Same as before for small displays, use the Console.

SteWers:
It is a solution for fitting the page in small displays, when not all 16 timer are needed. Of course all timers can be configure via console. Do you think, an automatic "multi line" display is better?

image

Jason2866:
Ugly, but better than let the user guess where are the "other" timers and open issues because GUI is wrong.
Probably more thinking for a better solution needed before changing anything here.

SteWers:
OK. Changed to automatic wrap.
Further changes can be done, when a better idea comes up.

arendst:
What is the gain regarding a smaller window by changing the amount of tabs while the 7 day checkboxes is the largest item on the form?

image

How does your full timer parameter form look like?

arendst:
I've been working hard on making the timer settings window compliant for every language. They all have different length of strings making it hard to satisfy layouts.

What about this for a smaller, phone friendly layout like I use for LoRaWan:

image

SteWers:
> while the 7 day checkboxes is the largest item on the form

Oh yes. I tested in german and there are 3-letter-days only two letters. 🤔
The timer settings in LoRaWan are looking very nice. The layout should be also used for the normal timer setting.

arendst:
Released new timer parameter screen just now.

SteWers:
> Released new timer parameter screen just now.

Very nice. Thanks.

PR #23975, opened 2025-10-03
Tweaks at timer setting page

SteWers:
## Description:
Here are some tweaks for the timer settings page:
- Correct syntax error
- Remove some ineffective html elements
- Normalize language settings
- Some changes in german language file
## Checklist:


Jason2866:
When changes in sort order in language file is done, it has to be done in all files.

SteWers:
All language files are now sorted in the same way.

sfromis:
As the symbol `D_TIMER_TIME` is specific to the Timers feature, and is part of a group of similar-named symbols, shortening it to just `D_TIME` does not look like an improvement.

It is common (while not universal) that i18n symbols for specific features are named to have a feature name prefix in the symbol names, instead of being the English-language version of the word(s).

Also, same word in English, used different places, may not always have the same translation in other languages, as this might depends on the specific context of usage.

SteWers:
`D_TIMER_TIME`  is NOT specific to timers only. It is also used in `xdrv_01_9_webserver.ino`. The use should be general and not related to a single driver.

SteWers:
> may not always have the same translation in other languages

That's a valid point. I think, naming it to `D_TIME_OF_DAY` makes the usage clear for all languages.

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
```diff
@@ -803,15 +803,15 @@ const char HTTP_TIMER_SCRIPT6[] PROGMEM =
...
 const char HTTP_FORM_TIMER1[] PROGMEM =
-  "<fieldset style=text-align:center;'>"
+  "<fieldset style='text-align:center;'>"
   "<legend style='text-align:left;'><b>&nbsp;" D_TIMER_PARAMETERS "&nbsp;</b></legend>"
   "<form method='post' action='" WEB_HANDLE_TIMER "' onsubmit='return st();'>"
   "<br><label><input id='e0' type='checkbox'%s><b>" D_TIMER_ENABLE "</b></label><br><br><hr>"
   "<input id='t0' value='";
 const char HTTP_FORM_TIMER2[] PROGMEM =
-  "' hidden><div id='bt'></div><br><br><br><br><br>"
+  "' hidden><div id='bt'></div><br>"
```
arendst: Why oh why. Removing these ends up in 
image
 instead of 
image

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
```diff
@@ -803,15 +803,15 @@ const char HTTP_TIMER_SCRIPT6[] PROGMEM =
   "}"
   "wl(it);";
 const char HTTP_TIMER_STYLE[] PROGMEM =
-  ".tl{float:left;border-radius:0;border:1px solid var(--c_frm);padding:1px;width:12.5%%;}";  // COLOR_FORM, Border color needs to be the same as Fieldset background color from HTTP_HEAD_STYLE1 (transparent won't work)
+  ".tl{border-radius:0;border:1px solid var(--c_frm);padding:1px;width:12.5%%;}";  // COLOR_FORM, Border color needs to be the same as Fieldset background color from HTTP_HEAD_STYLE1 (transparent won't work)
```
arendst: Do not remove the float. If less buttons are shown it will float center while I want it to be left justified to analog tabs.

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
```diff
@@ -830,24 +830,24 @@ const char HTTP_FORM_TIMER3[] PROGMEM =
...
+                             "\x4F\xC3\xD0\xEC\xF0\xCB\xE3\xF1\x1E\x68\x18\x0E\x32\x37\x41\x37\x19\x3F\x4C\x6F"
+                             "\x86\x1E\x0F\x41\xC7\xE1\xDB\x3B\x0E\xB3\xD0\xEC\x1C\x7E\x03\x03\xC9\xC7\xC1\x37"
+                             "\x27\x3D\x0E\xC1\x34\xCC\xFC\x3D\x07\x7C\xFC\x3D\x04\xD3\x35\xA2\x1F\x0F\xB1\x3C"
+                             "\x7D\xAE\xB1\xAC\xF8\x7D\x89\xE3\xEC\x10\xAE\x9B";
 #define  HTTP_FORM_TIMER4       Decompress(HTTP_FORM_TIMER4_COMPRESSED,HTTP_FORM_TIMER4_SIZE).c_str()
 #else
 const char HTTP_FORM_TIMER4[] PROGMEM =
   "<span><select style='width:60px;' id='ho'></select></span>"
-  "&nbsp;" D_HOUR_MINUTE_SEPARATOR "&nbsp;"
+  " " D_HOUR_MINUTE_SEPARATOR " "
```
arendst: Nothing wrong with &nbsp;

arendst:
So in the end it takes a lot of time to verify the impact of your changes and the end result just doesn't makes much different.

The only valid change is a missing single quote which I'll add now.

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
SteWers: This works only in combination with the change in line 806.

image

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
SteWers: As discussed in #23959 there are always 16 buttons. So this only wastes flash. See also line 814.

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
SteWers: Yes, nothing wrong, but the result is the same and saves flash.

SteWers:
> it takes a lot of time to verify the impact of your changes

Sorry for that.

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
SteWers: Reverted

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
SteWers: Reverted

SteWers:
I brought the old html style back.

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
arendst: Alas, yesterday I came across this issue again. Changing `&nbsp;` to ` ` resulted in no visible space and hence no visible difference in table columns (in yesterday's case)

So let's end the discussion of replacing it by a space ' ' and revert to '&nbsp;` for once and for all.

SteWers:
Changed ' ' to '&nbsp;'

tasmota/tasmota_xdrv_driver/xdrv_09_timers.ino
```diff
@@ -888,11 +889,7 @@ void HandleTimerConfiguration(void)
 #else
   WSContentSend_P(HTTP_FORM_TIMER3);
 #endif  // USE_SUNRISE
-#ifdef USE_UNISHOX_COMPRESSION
```
arendst, line 891: Why is this removed? The compressed code is smaller than the original so that's why it is here.

arendst: Oh I fear you compressed the current HTTP_FORM_TIMER4[]. In that case the language specific D_HOUR_MINUTE_SEPARATOR is hardcoded in the compressed code and is wrong for other languages. Pls revert.

SteWers: Oh yes, you're right. I now use the same compressed and uncompressed code of HTTP_FORM_TIMER4. This makes the  code simpler and should avoid such mistakes in future.

arendst: But as compression is all about saving code space it was there for a purpose not being ease of coding. So pls recreate the correct compressed code and reuse the define.

SteWers: One does not exclude the other.

PR #24030, opened 2025-10-18
force mqtt 8883 port when TLS

GMagician:
mqtt uses port 8883 when tls is used, help to not forget it

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino
@@ -29,9 +29,11 @@
 #endif
 WiFiClient EspClient;                     // Wifi Client - non-TLS
 
-#ifdef USE_MQTT_AZURE_IOT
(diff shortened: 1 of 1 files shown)


sfromis:
I'd say that you should not **force** a port number, as port 8883 is only a commonly used default value, without it being the only valid option for TLS. As I read the code from before your suggested change, Azure IoT has a restriction requiring port number 8883, but standard brokers (like Mosquitto) has this as a configuration option.

[Other code](https://github.com/arendst/Tasmota/blob/623dea884f71bf1d0db0f273b8c1af1d924e114f/tasmota/tasmota_xdrv_driver/xdrv_02_9_mqtt.ino#L180-L191) in the driver you changed also sees port 443 and 8443 as alternate TLS port numbers, when issuing a warning about non-TLS on a port number expected to be used with TLS.

This looks like those port numbers are seen out there. While it may seem odd to use port 443 (generally reserved for https), using this port number can be a trick to get a connection on networks with severe port restrictions.

GMagician:
That's fair, I only saw 8883 on mosquito but that may be due to my limited view.

PR #24063, opened 2025-10-26
TuyaMCU: handle WIFI_SELECT / WIFI_RESET without soft-locking MCU

thekeymeister:
## Description:

If an MCU initiated WiFi reset is carried out on some Tuya devices, the MCU will not respond to any commands from Tasmota until it has determined a cloud connection has been established. This also occurs during initial setup of the device. At present, Tasmota will reset the WiFi settings, but doesn't give the MCU the status updates it expects, essentially soft locking it.

This will initially ACK the WIFI_RESET/WIFI_SELECT by sending a WIFI_STATE with an empty payload.

It then establishes the requested pairing mode from the received WIFI_RESET/WIFI_SELECT (0x00 = EZ mode, 0x01 = AP mode)

WIFI_STATEs are then sent to simulate the process of establishing a connection to Tuya servers, in the following order.

0x00 (Entered EZ pairing mode) or 0x01 (Entered AP pairing mode)
0x02 (Configured, but not connected)
0x03 (Router connected)
0x04 (Cloud connected)

Upon completion, the MCU is satisfied that a connection is established, and will allow control of the MCU via Tasmota. Without this, the device can be stuck and essentially becomes 'listen only'.  We then continue with the existing functionality of resetting the current Tasmota WiFi settings.

As well as properly handling the WiFi reset procedure, this allows for initial setup to be done without ever having to connect to a Tuya server.

I've tested this on two YH002 blind motors (module swapped WBR3 to ESP-12S) - one previously used, one brand new out of the box.  Both working fine.  I've been unable to test with an ESP32 device as I haven't got any.

I never formally created an issue, but my [comment on this page](https://blakadder.com/tuya-covers/#comment-6440618288) explains the initial issue.
Basically, my 5 year old pressed some buttons on my Zemismart YH002 blind motor which started the WiFi reset procedure, and the device stopped responding to all commands from Tasmota, and would only report changes in position that were initiated by button presses on the unit itself.

I saw a [post suggesting ESPhome may have handled this situation](https://github.com/esphome/issues/issues/3081) so moved these across to that in hope, however that didn't work for me either, so I delved deeper to fix the underlying issue.  I've already had a PR merged for the same for ESPhome [here](https://github.com/esphome/esphome/pull/10822), I felt it made sense to try and implement a fix here too.

## Checklist:


tasmota/tasmota_xdrv_driver/xdrv_16_tuyamcu_v1.ino
```diff
@@ -1097,12 +1097,36 @@ void TuyaNormalPowerModePacketProcess(void)
...
+
+      // Establish pairing mode - WIFI_RESET is assumed to be AP mode
+      uint8_t first = 0x00;                   // EZ default
+      if (is_select) {
+        if (payload_len >= 1 && Tuya.buffer[6] == 0x01) {
+          first = 0x01;                       // AP
+        }
+      } else {
+        first = 0x01;                         // RESET -> AP
+      }
```
eku:
```suggestion
      uint8_t first = 0x01;
      if (is_select && !(payload_len >= 1 && Tuya.buffer[6] == 0x01))
          first = 0x00;
      }
```

sfromis (commented):
The last change [462d40d](https://github.com/arendst/Tasmota/pull/24063/commits/462d40de184cf44809eb1408c35cc3b398b575f7) breaks the code, as removing the if leaves the ending "}" from the else part unmatched.

arendst:
Will fix asap.

PR #24189, opened 2025-12-04
Fix Analog gauge glitches on FreeRTOS for xdrv_92_vid6608 driver

petrows:
Original version assumed position and command as thread safe, but it is not true. Wrong readings lead to random move back and forth. Patch was made and tested on real device.

## Description:

* Add mutex to protect vid6608 from update/command
* Fix verious glitches

## Checklist:


petrows:
I have pushed small optimization: do not create mutex object, if no drives are defined and driver is idle mode.

petrows:
I have tested fully assembled device in ~48 hours in a row with changes from this PR: looks good so far. No glitch moves anymore, drives are stable and maintain proper position.

PR #24192, opened 2025-12-06
Berry `tasmota.micros()` to get time in microseconds

s-hadinger:
## Description:

New function to get current time in microseconds to get finer measurement than tasmota.millis(). Be careful that the counter changes sign every hour.

Also changed `tasmota.millis()` to a static method.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_tasmota.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_tasmota.ino
@@ -119,16 +119,24 @@ extern "C" {
   int32_t l_millis(struct bvm *vm);
   int32_t l_millis(struct bvm *vm) {
     int32_t top = be_top(vm); // Get the number of arguments
-    if (top == 1 || (top == 2 && be_isint(vm, 2))) {  // only 1 argument of type string accepted
+    if (top == 0 || (top == 1 && be_isint(vm, 1))) {  // only 1 argument of type string accepted
       uint32_t delay = 0;
       if (top == 2) {
-        delay = be_toint(vm, 2);
+        delay = be_toint(vm, 1);
       }
       uint32_t ret_millis = millis() + delay;
       be_pushint(vm, ret_millis);
       be_return(vm); // Return
     }
     be_raise(vm, kTypeError, nullptr);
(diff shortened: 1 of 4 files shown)


Jason2866:
@s-hadinger This changes are breaking the haspmota Watch 480x480 Demo TAPP.
Probably not just this demo.

tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_tasmota.ino
```diff
@@ -119,10 +119,10 @@ extern "C" {
   int32_t l_millis(struct bvm *vm);
   int32_t l_millis(struct bvm *vm) {
     int32_t top = be_top(vm); // Get the number of arguments
-    if (top == 1 || (top == 2 && be_isint(vm, 2))) {  // only 1 argument of type string accepted
+    if (top == 0 || (top == 1 && be_isint(vm, 1))) {  // only 1 argument of type string accepted
       uint32_t delay = 0;
       if (top == 2) {
```
arendst, line 124: I guess this needs be `if (top == 1) {`

s-hadinger: Normally no, because it is now a static method, so it does not have the default `self` first argument.

Let's revert this PR so we can identify the breaking change. I suppose that there is some code that expects 'tasmota.millis()' to being not static

arendst: I see a miscommunication coming.

The change from line 122 is in your failing PR. My suggestion is in line 124 which will never execute when line 122 is changed that way.

In the meantime @Jason2866 tested my change with success so....

Jason2866: `micros` is really short. Makes this sense in general in Berry?

s-hadinger: Yes it does. I’m doing profiling for leds animation on a very slow esp32s2 so it magnifies performance issues. In such cases, milliseconds is too coarse

PR #24214, opened 2025-12-11
fix displayrotate for i2c

Staars:
## Description:

Wrong logic persisted from the early stages of the huge refactoring.

We must return false from the panel to signal the calling display instance, that the Renderer is responsible for this job.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.7


arendst:
I was about to make the same change but merely on a trial and error basis comparing the way the other functions were overriden.

Anyway, I tested the fix too and it works. Hope no more other remnants of display changes come to light as I'm about to release v15.2.0 soon.

Jason2866:
Well, no release today for sure. Github Actions is "sick"

PR #24272, opened 2025-12-28
[Zigbee] Add device separators

SteWers:
## Description:
Add device separators for Zigbee, according to other drivers.

Before PR:

After PR:

## Checklist:


s-hadinger:
Thanks but I’m not a big fan of this. It clutters the UI and doesn’t bring much value

SteWers:
Of course there are always pros and cons. In my opinion, the separation makes it more readable. It is also used in other drivers. With this PR we get the same look and feel.

s-hadinger:
Let’s put it on hold. I want to test it with 15+ device and check the UI does not turn into a big prison with horizontal bars. I’m afraid you’ll have to wait for a week

SteWers:
OK. I have no experience with this, but I believe the prison bars are vertical. 😉

arendst:
I'm no fan either of having a bar at every single sensor. 

Just the way these sensors are displayed is a perfect grouping already.

sfromis:
Where `<hr>` lines are now used with sensor drivers is when each driver will sometimes have 6 values (or even more) to report, thus getting the sensor output grouped when having multiple. This has no horizontal lines between the values, and with more than one value per line, there may also be a header line in the group. Here it would not be an improvement to avoid the separators.

With Zigbee it is instead a one/two-line graphically distinct structure getting cluttered with a bunch of separators. With a much larger structure than this, I might use a low-contrast variation of the background color to aid horizontal eye-movements, and actively avoid a structure leading to many `<hr>` lines on a page. But neither in this simple Zigbee case.

SteWers:
These are all valid arguments. Therefore, I will close this PR. The idea with the different backgrounds sounds interesting. Perhaps I experiment with it a bit.

PR #24284, opened 2026-01-01
Fix Breathe animation to support ValueProviders passed for color

aschober:
## Description:
Fix Breathe animation to support ValueProviders passed for color.

An issue with BreatheAnimation was discovered when using it in a Template with a parameter for `color` which then passes a ClosureValueProvider as `color`. Appears to have been introduced in https://github.com/arendst/Tasmota/pull/24246 which requires that `color` has type `int`. Added fix to allow either an `int` or a ValueProvider.

Note I updated `breath_animation_test.be`, but I haven't been able to figure out the correct command or environment to run the tests. I've gotten to running the below command using the `berry` built to run `./solidify_all.be`, but likely I'm missing something else obvious.
```bash
$ ./lib/libesp32/berry/berry -s -g -m lib/libesp32/berry_animation/src -e 'import global import tasmota def log(x,l) tasmota.log(x,l) end ' ./lib/libesp32/berry_animation/src/tests/breathe_animation_test.be
Testing BreatheAnimation...
Imported animation module
type_error: 'nil' value is not callable
stack traceback:
        <unknown source>: in function `main`
```

## Checklist:


s-hadinger:
Thanks a lot, it just collides with another change. Let me merge as-is, the solidification will normally be solved by CI

PR #24300, opened 2026-01-03
[Zigbee] Several code optimizations without functional changes

SteWers:
## Description:
- Added SPACE between light color and value regarding to other value presentations
- Removed the 'json' parameter from ZigbeeShow, as it is not used
- Simplified code to make it more readable, efficient and smaller
## Checklist:


tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_A_impl.ino
```diff
@@ -235,14 +235,14 @@ void zigbeeZCLSendCmd(class ZCLFrame &zcl) {
...
 // multiplier == 1: ignore
 void ZbApplyMultiplierForWrites(double &val_d, uint32_t multiplier, uint32_t divider, int32_t base) {
-  if (0 != base) {
-    val_d = val_d - base;
+  if (base) {
+    val_d -= base;
   }
-  if ((0 != multiplier) && (1 != multiplier)) {
-    val_d = val_d / multiplier;
+  if (multiplier > 1) {
```
s-hadinger, line 241: I don't think this is right, multiplier can be negative

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_5_2_converters.ino
```diff
@@ -730,17 +730,17 @@ void ZCLFrame::applySynonymAttributes(Z_attribute_list& attr_list) {
         attr_list.removeAttribute(&attr);
       } else {
         attr.setKeyId(syn.new_cluster, syn.new_attribute);
-        if ((syn.multiplier != 1 && syn.multiplier != 0) || (syn.divider != 1 && syn.divider != 0) || (syn.base != 0)) {
+        if ((syn.multiplier > 1) || (syn.divider > 1) || (syn.base)) {
           // we need to change the value
           float fval = attr.getFloat();
-          if (syn.multiplier != 1 && syn.multiplier != 0) {
-            fval = fval * syn.multiplier;
+          if (syn.multiplier > 1) {
```
s-hadinger, line 736: multiplier can be negative, and this test breaks it

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_5_2_converters.ino
```diff
@@ -730,17 +730,17 @@ void ZCLFrame::applySynonymAttributes(Z_attribute_list& attr_list) {
...
+            fval *= syn.multiplier;
           }
-          if (syn.divider != 1 && syn.divider != 0) {
-            fval = fval / syn.divider;
+          if (syn.divider > 1) {
+            fval /= syn.divider;
           }
-          if (syn.base != 0) {
-            fval = fval + syn.base;
+          if (syn.base) {
```
s-hadinger: I'm not a big fan of this syntax as it hides the value we're testing for. This is good for booleans, but makes code less readable to my opinion

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_A_impl.ino
```diff
@@ -235,14 +235,14 @@ void zigbeeZCLSendCmd(class ZCLFrame &zcl) {
 // multiplier == 0: ignore
 // multiplier == 1: ignore
 void ZbApplyMultiplierForWrites(double &val_d, uint32_t multiplier, uint32_t divider, int32_t base) {
-  if (0 != base) {
-    val_d = val_d - base;
+  if (base) {
```
s-hadinger: Same, I prefer to keep the explicit `0`

s-hadinger (changes requested):
Thanks, but I believe it introduces a bug for multiplier

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_A_impl.ino
SteWers: No. It is a unsigned int32. So that will be no problem.

s-hadinger: Ah yes you are right

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_5_2_converters.ino
SteWers: Same as above. Negative multiplier seems to be an old syntax. Because of that I did the PR in docs.

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_5_2_converters.ino
SteWers: In this case it maybe better readable. The resulting code should be the same.

tasmota/tasmota_xdrv_driver/xdrv_23_zigbee_A_impl.ino
SteWers: In this case it maybe better readable. The resulting code should be the same.

SteWers:
@s-hadinger Is there anything else you mention to change here?

PR #24340, opened 2026-01-13
Save some ram in esp8266

GMagician:
This will exclude the mutex variable not used by esp8266, this should save some ram bytes.

Also removed a double initialization. Variable is initialized at the beginning of the function but it's also reinitialized inside 'if' and 'else' branches (I double checked code but give a third check)

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.1.8


arendst:
Just to be safe I wouldn't remove the "global" init but remove the one after `if ((WL_CONNECTED == WiFi.status()) && WifiHasIP()) {`

PR #24349, opened 2026-01-16
BearSSL: use ROM functions for bigint on ESP32 too

Staars:
## Description:

Depending on the firmware this saves up to 6 kB flash by using ROM functions for big integer multiplication on the "old" ESP32, which are now hardware accelerated.
Positive side effect is faster execution, which in my synthetic Berry test goes down from 1015 to 730 ms.
Network speed is super hard to benchmark and needs repeated and very large series, but in my best effort I measured 12,5% faster HTTPS connection when using Tasmotas extension store.

As there is no documentation or relevant example code about this, I had to trial-and-error a lot of things. That's why I left some hints and comments to easily switch this feature off for testing, if suspicious things occur.

You can remove the `//` to turn this feature off in both changed files:
`#if defined(ESP_PLATFORM) && !defined(ESP8266)// && !defined(CONFIG_IDF_TARGET_ESP32)`

## Checklist:

### DIFF
--- a/lib/lib_ssl/bearssl-esp8266/src/ec/_ec_c25519_m15.c
+++ b/lib/lib_ssl/bearssl-esp8266/src/ec/_ec_c25519_m15.c
@@ -21,7 +21,7 @@
  */
 
 #if defined(USE_SHA_ROM)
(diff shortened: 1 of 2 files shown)


s-hadinger:
Well done!

Issue #24345, opened 2026-01-15
[Matter] Connection loss after Wifi Reconnect: Interface index drift (st1 -> st2) causes "error sending packet"


utessel:
### PROBLEM DESCRIPTION
When the ESP32 loses Wifi connection (due to low RSSI) and reconnects, the LwIP stack assigns a new interface index (e.g., ipv6 ... %st2).
However, the Matter stack seems to keep a reference to the old interface (%st1).
This results in Matter being unable to send packets, while the Tasmota device itself is reachable via IPv4/IPv6 ping on the new interface.

### REQUESTED INFORMATION
_Make sure your have performed every step and checked the applicable boxes before submitting your issue. Thank you!_

- [ x] Read the [Contributing Guide and Policy](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) and [the Code of Conduct](https://github.com/arendst/Tasmota/blob/development/CODE_OF_CONDUCT.md)
- [x ] Searched the problem in [issues](https://github.com/arendst/Tasmota/issues)
- [ x] Searched the problem in [discussions](https://github.com/arendst/Tasmota/discussions)
- [ x] Searched the problem in the [docs](https://tasmota.github.io/docs/FAQ)
- [ ] Searched the problem in the [chat](https://discord.gg/Ks2Kzd4)
- [ ] Problem is not scripter related, in this case open a discussion and tag  gemu2015
- [ x] Device used (e.g., Sonoff Basic): Nous A8T
- [x ] Tasmota binary firmware version number used:15.2.0
  - [ x] Pre-compiled
  - [ ] Self-compiled
- [ ] Flashing tools used: _____
- [ ] Provide the output of command: `Backlog Template; Module; GPIO 255`:
```lua
  Configuration output here:


```
- [ ] If using rules, provide the output of this command: `Backlog Rule1; Rule2; Rule3`:
```lua
  Rules output here:


```
- [ ] Provide the output of this command: `Status 0`:
```lua
  STATUS 0 output here:


```
- [ ] Set `weblog` to 4 and then, when you experience your issue, provide the output of the Console log:
```lua
  Console output here:


```

### TO REPRODUCE
_Steps to reproduce the behavior:_

### EXPECTED BEHAVIOUR
Matter stack should update its interface binding/scope ID upon WIFI_EVENT_STA_CONNECTED or when a new IPv6 address is assigned.

Logs (Syslog Level 4):
Device loses connection and reconnects on new Interface st2:

[...]

PR #24358, opened 2026-01-17
Fix ESP32 reset during WiFi reconnect when mDNS is active

utessel:
## Description:

Explicitly call MDNS.end() before WiFi.disconnect() in support_wifi.ino. This prevents a race condition or crash where the mDNS service attempts to query or use the network interface while it is being torn down.

**Related issue (if applicable): #24345
During investigation of that issue I had lots of reboots: 
I used to stop the wifi and started it again lots of times.
It looks like a race condition with my test and mdns: 
Calling MDNS.end() before calling WiFi.disconnect(true) was a simple solution and it seems to be a good place for me.

In the end, it is not required to fix my problem of the Issue: see the second PR for this.

## Checklist:
  - [ x] The code change is tested and works with Tasmota core ESP32 V.3.1.9

### DIFF
--- a/tasmota/tasmota_support/support_wifi.ino
+++ b/tasmota/tasmota_support/support_wifi.ino
@@ -318,7 +318,7 @@ void WifiBegin(uint8_t flag, uint8_t channel) {
 #endif
 
 #ifdef USE_WIFI_RANGE_EXTENDER
-  }
+  if (WiFi.getMode() != WIFI_AP_STA || !RgxApUp()) {  // Preserve range extender connections (#17103)
 #endif  // USE_WIFI_RANGE_EXTENDER
 
 #if defined(ESP32) && defined(USE_DISCOVERY)
--- a/tasmota/tasmota_support/support_wifi.ino
+++ b/tasmota/tasmota_support/support_wifi.ino
@@ -318,8 +318,12 @@ void WifiBegin(uint8_t flag, uint8_t channel) {
 #endif
 
 #ifdef USE_WIFI_RANGE_EXTENDER
-  if (WiFi.getMode() != WIFI_AP_STA || !RgxApUp()) {  // Preserve range extender connections (#17103)
+  }
 #endif  // USE_WIFI_RANGE_EXTENDER
+
+#if defined(ESP32) && defined(USE_DISCOVERY)
+  MDNS.end();
+#endif
   WiFi.disconnect(true);  // Delete SDK wifi config
   delay(200);
   WifiSetMode(WIFI_STA);  // Disable AP mode


s-hadinger:
I’ve seen such crashes indeed. Thanks for the troubleshooting

s-hadinger:
Do you confirm that mDNS starts again when wifi is enabled again ? In such case does Matter still work?

What happens when you stop wifi but Ethernet is enabled? Does this change shuts down mDNS for Ethernet too?

utessel:
I can confirm that Matter works when I stopped and started Wifi. Even when the Lifetick was tried while it was off. In fact, the fix is for a corner case. For ethernet? I did not test that. Interesting question. My main problem happened this night again on one device, in a different variant, also related to mdns :-(
I will have a look into these aspects.

utessel:
Looking into: I expect, yes, mdns will be stopped for all. But we are her in the starting part of a wifi. And there we have a disconnect call. With a delay. That looks like other race conditions happened.
After that, the network is started. As long as it gets an IP, that will use mdns begin and start it again. 
Anyhow: The interrupt of mdns on ethernet in case of a reboot seems to be worse. The race of mdns here is that code tries to send something while we are "disconnecting".
I didn't track down to find the "sending" problem, as I expected there might several of things like this.

s-hadinger:
My main concern is that devices with both Wifi and Ethernet active. When you turn off Wifi, I fear that mDNS is turned down although Ethernet is still active

utessel:
I cannot test that. So the concern is correct. 
at least, if matter is used, the mdns part will cause reboots: at least: often (in my case: >50%. Very annoying if you search for a different problem...) 

But the fix is in the WifiBegin, as that Wifi.disconnect there triggered the problem.

So this fix does not change anything when turning wifi off.

s-hadinger:
I think we still have a major problem with devices with both wifi and ethernet. I will test next week.

s-hadinger:
I had Claude Opus 4.6 do a complete analysis of how esp_mdns works:

Here's how it actually works under the hood:

`mdns_init()` creates a single global mDNS server (`s_server`), starts a service task, registers event handlers for WiFi, Ethernet, and IP events, and enables PCBs on any interfaces that already have IPs. It's a one-time call.

`mdns_free()` destroys everything — the task, the queue, all PCBs on all interfaces, all services, and unregisters all event handlers. It's a nuclear option. There's no "free just WiFi" — it kills mDNS entirely.

The per-interface lifecycle is handled automatically by `handle_system_event_for_preset()` in `mdns_netif.c`. When WiFi disconnects (`WIFI_EVENT_STA_DISCONNECTED`), it only disables the STA PCBs (v4 and v6). When Ethernet disconnects (`ETHERNET_EVENT_DISCONNECTED`), it only disables the ETH PCBs. They're completely independent. When an interface gets an IP again (`IP_EVENT_STA_GOT_IP`, `IP_EVENT_ETH_GOT_IP`), the corresponding PCB is re-enabled automatically.

If you need to manually control a custom netif (not the default STA/AP/ETH), the API provides `mdns_register_netif()`, `mdns_unregister_netif()`, and `mdns_netif_action()` for that. But for standard WiFi STA + Ethernet, the predefined event handlers do everything.

s-hadinger:
Hence I'm reverting this PR because it does more harm than it solves problem:

If the device has both ETH and Wifi, and wifi is turned off, mdns is lost for ETH and the device crashes at first MDNS call for ETH.

PR #24359, opened 2026-01-17
Matter: Fix connectivity loss during roaming and fix UDP stop bug

### LINKED ISSUE #24345: [Matter] Connection loss after Wifi Reconnect: Interface index drift (st1 -> st2) causes "error sending packet"

utessel:
## Description:

**Related issue (if applicable):** fixes #24345

I debugged my problem:

The solution I found was to handle the error, as a retry will not help in such a case.
In case of sending fail error, I check for changes and call a method to stop the udp.
That didn't work, as stop does not exist? 
But close does. 

## Checklist:
  - [ x] The code change is tested and works with Tasmota core ESP32 V.3.1.9

### DIFF
--- a/lib/libesp32/berry_matter/src/embedded/Matter_UDPServer.be
+++ b/lib/libesp32/berry_matter/src/embedded/Matter_UDPServer.be
@@ -109,7 +109,7 @@ class Matter_UDPServer
   # Stops the server and remove driver
   def stop()
     if self.listening
-      self.udp_socket.stop()
+      self.udp_socket.close()
       self.listening = false
       # tasmota.remove_driver(self)
       tasmota.remove_fast_loop(self.loop_cb)
@@ -170,6 +170,10 @@ class Matter_UDPServer
     else
       if tasmota.loglevel(3)
         log(format("MTR: error sending packet to '[%s]:%i'", packet.addr, packet.port), 3)
+      end
+      # Trigger check for network interface change
+      if self.device.handle_send_error
+        self.device.handle_send_error()
       end
     end
     return ok
--- a/lib/libesp32/berry_matter/src/embedded/Matter_UDPServer.be
+++ b/lib/libesp32/berry_matter/src/embedded/Matter_UDPServer.be
@@ -202,6 +202,10 @@ class Matter_UDPServer
         else
           self.packets_sent.remove(idx)
           log(format("MTR: .          (%6i) Unacked packet '[%s]:%i' msg_id=%i", packet.session_id, packet.addr, packet.port, packet.msg_id), 3)
+          # If a packet is unacked, check if network interface changed
+          if self.device.handle_send_error
+            self.device.handle_send_error()
+          end
         end
       else
         idx += 1
--- a/lib/libesp32/berry_matter/src/embedded/Matter_zz_Device.be
+++ b/lib/libesp32/berry_matter/src/embedded/Matter_zz_Device.be
@@ -57,6 +57,7 @@ class Matter_Device
   var probe_sensor_time               # number of milliseconds to wait between each `read_sensors()` or `nil` if none active
   var probe_sensor_timestamp          # timestamp for `read_sensors()` probe (in millis())
                                       # if timestamp is `0`, this should be scheduled in priority
+  var last_ip6local                   # keep track of ipv6 link local address to detect interface change (st1 -> st2)
 
 
   #############################################################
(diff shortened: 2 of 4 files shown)


utessel:
Had to fix another variant: When the Interface changed, the code was able to send (to the right address!) if it was a GU Address, but it did not receive anything. The situation would not fix on its own. As the "handle error" checks for the local link, this also helps: If the local link string is still the same, nothing happens, so that is safe.

utessel:
Note: All these problems happen for life ticks, where the initiative (send udp) starts from the tasmota device and the controller sends the ack. 
The tested case in the closed issue was the other way around. 
Maybe it is related to the controller implementation? Google Home in my case.

utessel:
I just found another issue: The fix works almost already, but not in the case when the AP changed: I had an exception in mdns then. I think I know the solution, but have to test that first.

utessel:
I had to build a test env to enforce the AP changing. Not sure if all cases are working. I even got a race from the web ui if the AP changes while it is open. My Idea for that: hack the Scanning Loop so it decides for a different one if I want. I did not find a command to enforce an BSSID. 
But a more simple test was to call the handle send error method from the berry console: that triggered the exception i had seen (reliable): Start wants to use add service, but that is still there. So removing it, and giving start a cleaner environment did help.
Testcode:
var m = matter_device
m.last_ip6local = "fe80::1"
m.handle_send_error()

utessel:
Sorry. I had "luck": My test env with the changing AP showed another issue with my approach:
My last commit fixed the berry exception.
But in case of AP change, now I get an exception in "mdns_service_remove_for_host". 
That is really annoying:
The "react in case of error" approach does not work reliable.

utessel:
Some more investigation showed:
When the AP switches back to the first one, matter works again. Sometimes...
But in relevant percentages, the change of AP triggered a reboot: there seem to be several race conditions in the lower layer wifi/network code.
Just as note to my "future self":
The test worked like this: In a network with at least two APs visible with the same SSID.
The code:
In WifiBeginAfterScan, I changed the Wifi.best_network_db and the rssi_scan to -100 if the bssid is a specific value: (I used first byte == global variable). 
To trigger that, I hacked the wifitest command to get a number: 
That sets the global variable byte, and sets the Wifi.scan_state to 2. 
Then the scan will start. And depending on the global variable, the AP will change.

PR #24368, opened 2026-01-19
some display fixes

gemu2015:
## Description:

1. st77xx controllers now work
2. parallel displays push colors now work with right colors with display text
3. use latest jpeg decoder, old version still in source because one user gets strange missing pictures which do not occur with previous implementation 
4. now push pixels number of pixels changed to 32 bit to be able to push complete display
5. fix e-display full and partial update (also with old driver)

none of my waveshare e-papers works with the new driver
there were heavy changes in this section. i could not fix this in the short run and temporarily use the old driver version for these displays.

a user reports OLED color displays also no longer working

## Checklist:

### DIFF
--- a/lib/lib_deprecated/Xlatb_RA8876-gemu-1.0/RA8876.cpp
+++ b/lib/lib_deprecated/Xlatb_RA8876-gemu-1.0/RA8876.cpp
@@ -947,7 +947,7 @@ void RA8876::setAddrWindow(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1) {
 }
 
 // pixel color is swapped in contrast to other controllers
-void RA8876::pushColors(uint16_t *data, uint16_t len, boolean not_swapped) {
+void RA8876::pushColors(uint16_t *data, uint32_t len, boolean not_swapped) {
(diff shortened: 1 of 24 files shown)


sfromis:
Looking at the reasons for a build error with `xdrv_81_esp32_webcam_task_motion.ino` failing to find function `esp_jpg_decode` and type `jpg_scale_t`, I found that this PR means that `esp_jpg_decode.h` may no longer be included in builds, with the symbol `USE_NEW_JPG` hardcoded in both `xdrv_13_display.ino` and `support_jpeg.ino` to control exclusion of old function calls. This looks like a breaking change to `xdrv_81_esp32_webcam_task_motion.ino` when not having similar control of new/old the new function calls, and not doing an include of the old functions now removed elsewhere.

gemu2015:
indeed the "old" version needs this folder in /lib/lib_esp32_div
maybe the author of the motion webcam driver adapts to the "new" jpeg driver which should be faster then the old one.
[ESP32_JPDEC.zip](https://github.com/user-attachments/files/24842703/ESP32_JPDEC.zip)

PR #24378, opened 2026-01-23
Remove movmem_P

GMagician:
## Description:

If I didn't have an oversight all memory in these calls is from ram to ram so no need to use _P version. This should also speed up code and maybe reduce footprint if function is not linked

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.1.9


arendst:
The funny thing is it adds 16 bytes to the code instead of lowering code size. :-)

arendst:
Implemented.

PR #24403, opened 2026-02-01
Move custom envs on top

GMagician:
## Description:

Move custom envs on top. There is an high probability that if someone add them he will often choose them, so putting in top of list helps.

## Checklist:

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -19,11 +19,11 @@ src_dir                     = tasmota
 lib_dir                     = lib/default
 boards_dir                  = boards
 build_cache_dir             = .cache
-extra_configs               = platformio_tasmota32.ini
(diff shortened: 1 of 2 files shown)


sfromis:
How does it benefit a user to move cenv first in the search order? 
And I'm not sure that cenv should override same name in regular envs, potentially resulting in other builds being thrown out of whack.

GMagician:
> How does it benefit a user to move cenv first in the search order? And I'm not sure that cenv should override same name in regular envs, potentially resulting in other builds being thrown out of whack.

When you are in platformio you may choose which env to use (in bottom status bar). When you click you have a long list of environments and with default platform.ini your custom are at the end. With this change you see them at top

immagine

Jason2866:
PR not accepted, no change which may damage setup.

PR #24421, opened 2026-02-06
[Berry] Provide WiFi-SSID

SteWers:
## Description:
Provide WiFi-SSID to Berry
## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_tasmota.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_tasmota.ino
@@ -281,6 +281,7 @@ extern "C" {
       if (show_rssi) {
         be_map_insert_int(vm, "rssi", rssi);
         be_map_insert_int(vm, "quality", WifiGetRssiAsQuality(rssi));
+        be_map_insert_str(vm, "ssid", SettingsTextEscaped(SET_STASSID1 + Settings->sta_active).c_str());
       }
     }
     be_pop(vm, 1);


sfromis:
Berry code can already get the currently in-use SSID.
`tasmota.cmd('_State')['Wifi']['SSId']`

Of course, it may not even be connected via Wifi, but Ethernet. Or not at all.

SteWers:
Thanks for the hint. I wasn't aware of that and changed the Extensions-PR (https://github.com/tasmota/Tasmota-Extensions/pull/1/commits/2f2c3052ba8391011591a5ce6e248e2feb249022).

s-hadinger:
Actually I m not opposed to it. Please reopen and I will merge. This will make less console noise than issuing a command

SteWers:
PR is reopened and can be merged.

sfromis:
Adding such a field still gets into considerations about how to handle `Wifi 0`. The SSID configuration will still point at the dormant AP, while networking may (or not) be via Ethernet. But returning the configured AP instead of active AP looks misleading to me.

Of course, then "Wifi Heap Sticker" gets no signal quality either, and would show no connection. Arguably, the sticker could be expanded to be "Network Heap Sticker", if the point is to include networking details, instead of merely Wifi signal strength.

Still, it would make most sense for `tasmota.wifi()` to return active info, in line with commands `State` and `Status 5`, instead a mix including configured but not active details. The `State` command will only return AP names if connected via Wifi. Adding other "missing" fields could also make sense, if going for "completeness".

SteWers:
There is always the active ssid reported. It is requested in the same way as on the info page. Also `tasmota.wifi('ssid')` only returns a (!= nil) value, when wifi is connected.
BTW: I cannot see a ssid with `Status 5`

SteWers:
I looked at the `State` command. The SSID is determined in the same way there:
https://github.com/arendst/Tasmota/blob/c71530bfd88be91273e9f4cd28cd56d48382c515/tasmota/tasmota_support/support_tasmota.ino#L922

PR #24436, opened 2026-02-10
Updated LVGL Panel using binary stream

mdaskalov:
## Description:
I've rewritten the LVGL Panel to send updates using a binary stream.

The stream uses the following 10-byte header:

2("LV") + 2(x) + 2(y) + 2(w) + 2(h) 

and then sends the unchanged data of the LVGL callback.

On the client side, a pre-calculated lookup table is used for the RGB565 to ARGB8888 conversion. The image data is updated incrementally, and the canvas is updated only once at the end.

The stream could be used in external applications because the screen size is included in the HTTP headers.

## Checklist:


s-hadinger:
Thanks this is very cool, but I don't get why you removed the async HTTP server part. Isn't there a risk that Tasmota freezes if the client does not read the content fast enough?

mdaskalov:
Initially, I've implemented it with the tcpclientasync module, but I think it's slightly slower with it. 
In any case, we need to block the server side; otherwise, we run out of memory pretty fast. 
I also had "tasmota.yield()" in the loops, but removed it, as it doesn't help much. 
I am testing with a 480x480 display, and it updates smoothly. With a 240x280 display, there are no freezes at all.

You can optimize it further, of course!

s-hadinger:
Ah yes indeed, the memory problem makes async less interesting as a whole. Let me do some testing.

mdaskalov:
Both hdr and pixels get overwritten with each paint callback. 

So if we don't use copies on write the submitted data gets corrupted... 

With copy, we get pretty fast out of memory. Maybe it could work, if we block only when copy fails - but I haven't tried this.

For testing I use my analog [clock](https://github.com/mdaskalov/AnalogClock) which updates the second hand each 50ms.

On a 480x480 display we have hiccups, even if the JS client doesn't process at all - so the bottleneck is obviously on the server side.

TheHexaMaster:
Tested on ESP32 P4 Board with DSI 1024x600 display and works excellent. Perfect job @mdaskalov

mdaskalov:
I also made tests with delays in the stream processing on the client side... 

The client always manages to stream data without blocking the server. The data gets buffered and comes on the next `reader.read()` at once. Even with five streams running at the same time, the server never gets blocked.
Currently, we reserve data only for one full-screen-update, so if more data was buffered, the stream will stop anyway because of `RangeError`. 

This said, I also tried again the async client, and found a nice feature to write using `offset` and `len` implemented in #24076.

Unfortunately, there is probably an uncovered bug in this feature, so I changed it.

Using write with offset, `introspect` is called only once per LVGL paint callback, which makes it slightly faster as before.

s-hadinger:
Awesome, let's merge for now. It's definitely much faster.

However I'm missing the automatic resizing for the screen size (see screenshot)
image

PR #24451, opened 2026-02-14
core/net: add AsyncHttpClientLight + Berry Async WebClient (with build selector)

TheHexaMaster:
## Description:

Introduce a self-contained async HTTP/HTTPS client for ESP32 (AsyncHttpClientLight) and integrate it directly into the Berry webclient module with def selector #define USE_BERRY_WEBCLIENT_ASYNC. 

- Non-blocking GET/POST for telemetry/RPC without freezing the Berry VM.
- Lower TIME_WAIT pressure and safer connection reuse.
- TLS public-key pinning, optional RSA-only hardening.
- Keep API/error-code compatibility while modernizing internals.

**What changed**

**AsyncHttpClientLight **

- HTTP via lwIP sockets (IClientLite + non-blocking AsyncTcpAdapter) with SO_LINGER (RST-close) to reduce TIME_WAIT.
- HTTPS via BearSSL WiFiClientSecure_light, with:
- tls_pin_pubkey(fp1[,fp2]) (SHA-1 pubkey fingerprints, one or two pins)
- tls_set_rsa_only(true) optional hardening.
- Redirects (strict): 301/307 for GET/HEAD; 302/303 switch to GET; configurable limit (default 5).
- Chunked & identity decoding; reuse honored unless prior body wasn’t drained (auto reconnect).
- FreeRTOS worker (WDT-friendly): asyncGETStart(), asyncPOSTStart(...), asyncState(), asyncAbort().
- Error codes aligned with HTTPClientLight (−1..−11).
- Async in-RAM body buffer capped ~16 KB (heap safety). Use sync streaming for larger bodies.

**Berry webclient (updated bindings)**

- New methods: async_get_start(), async_post_start(str|bytes), async_state(), async_abort().
- async_state() publishes last_code (HTTP or negative error) and last_rx (received bytes), and frees hidden holds for byte payloads.
- Headers added via add_header() persist per instance; they reset on begin() and after redirects (re-add if needed).
- Sync API unchanged: GET/POST/PUT/PATCH/DELETE, get_string/bytes, write_file/flash.

**Dependency note**

<HTTPClient.h> remains included only for enum/constant types. No Arduino HTTPClient source code is copied; transport, parsing and async infra are independent.

**Compatibility**

- #define USE_BERRY_WEBCLIENT_ASYNC selector, w/o default sync webclient is used.
- Existing Berry scripts using sync APIs remain compatible.
- OTA/firmware workflows (sync GET + streaming) are unaffected.

**Operational notes**

- Call async_state() at least once after an async job finishes to publish results and clean up before reusing the same instance.
- Add Content-Type once after begin() (e.g., application/json); re-add after begin() or a redirect.
- Don’t call close() while an async job is running.

**Typical use cases:**

**Secure periodic telemetry over HTTP(S) + local MQTT**

- Use webclient for encrypted REST telemetry to the cloud while keeping MQTT for LAN/HA use.
- Pattern: create one client per endpoint, begin() once, add Content-Type: application/json, then call async_post_start(payload) from your scheduler and occasionally async_state() to publish last_code/last_rx and clean up.
- Keeps the Berry VM responsive; no long blocking writes. For large uploads/downloads, fall back to sync streaming (write_file, get_bytes).

**RPC long-polling (e.g., ThingsBoard) without tight polling loops**

- With async GET you can keep a long-poll open (e.g., ?timeout=60) instead of reconnecting every few seconds.
- Pattern: async_get_start() on /rpc?timeout=60, periodically call async_state():
- When async_state() returns DONE, parse the body, execute the command (optionally via async POST reply), then immediately start the next long-poll (async_get_start() again).
- When it returns RUNNING, do nothing (connection is open); VM stays free for other tasks.
- This reduces connection churn, lowers TIME_WAIT pressure, and prevents VM stalls seen with previous sync blocking mode.

**Risk**

- ESP32-only paths; HTTPS relies on existing BearSSL light.
- Behavior changes confined to the webclient backend; public sync APIs preserved.

**Test plan**

- Sync: GET large binary → write_file(); verify size/hash; chunked and identity.
- Async: burst async_post_start() with periodic async_state(); check last_code/last_rx, stable heap, no WDT.
- Redirects: verify 301/302/303/307 and header reset behavior.
- TLS: good vs bad fingerprint; RSA-only against ECDSA cert (should fail).

## Checklist:

### DIFF
--- a/lib/libesp32/AsyncHttpClientLight/library.json
+++ b/lib/libesp32/AsyncHttpClientLight/library.json
@@ -1,0 +1,17 @@
+{
+  "name": "Async HttpClient Light", 
+  "version": "1.0",
+  "description": "Async extension for HttpClientLight by Stephan Hadinger",
+  "license": "MIT",
+  "homepage": "https://github.com/arendst/Tasmota",
+  "frameworks": "arduino",
+  "platforms": "espressif32",
+  "authors":
+  {
+    "name": "Martin Macak (HexaMaster)",
+    "maintainer": true
+  },
+  "build": {
+    "flags": [ "-I$PROJECT_DIR/include" ]
+  }
+}
--- a/lib/libesp32/AsyncHttpClientLight/readme.txt
+++ b/lib/libesp32/AsyncHttpClientLight/readme.txt
@@ -7,7 +7,7 @@ TLS public-key pinning, optional RSA-only hardening.
 Keep API/error-code compatibility while modernizing internals.
 What changed
 
-AsyncHttpClientLight (new core client)
+AsyncHttpClientLight
 
 HTTP via lwIP sockets (IClientLite + non-blocking AsyncTcpAdapter) with SO_LINGER (RST-close) to reduce TIME_WAIT.
 HTTPS via BearSSL WiFiClientSecure_light, with:
@@ -30,8 +30,10 @@ Dependency note
 
 Compatibility
 
+#define USE_BERRY_WEBCLIENT_ASYNC selector, w/o default sync webclient is used.
 Existing Berry scripts using sync APIs remain compatible.
 OTA/firmware workflows (sync GET + streaming) are unaffected.
+
 Operational notes
 
 Call async_state() at least once after an async job finishes to publish results and clean up before reusing the same instance.
--- a/lib/libesp32/AsyncHttpClientLight/readme.txt
+++ b/lib/libesp32/AsyncHttpClientLight/readme.txt
@@ -1,0 +1,242 @@
+Introduce a self-contained async HTTP/HTTPS client for ESP32 (AsyncHttpClientLight) and integrate it directly into the Berry webclient module with def selector #define USE_BERRY_WEBCLIENT_ASYNC. 
+Code still include <HTTPClient.h> only for the enum/constant types (e.g., followRedirects_t, transferEncoding_t, HTTPC_TE_*).
+
+Non-blocking GET/POST for telemetry/RPC without freezing the Berry VM.
+Lower TIME_WAIT pressure and safer connection reuse.
+TLS public-key pinning, optional RSA-only hardening.
+Keep API/error-code compatibility while modernizing internals.
+What changed
+
+AsyncHttpClientLight (new core client)
+
+HTTP via lwIP sockets (IClientLite + non-blocking AsyncTcpAdapter) with SO_LINGER (RST-close) to reduce TIME_WAIT.
+HTTPS via BearSSL WiFiClientSecure_light, with:
+tls_pin_pubkey(fp1[,fp2]) (SHA-1 pubkey fingerprints, one or two pins)
+tls_set_rsa_only(true) optional hardening.
+Redirects (strict): 301/307 for GET/HEAD; 302/303 switch to GET; configurable limit (default 5).
+Chunked & identity decoding; reuse honored unless prior body wasn’t drained (auto reconnect).
+FreeRTOS worker (WDT-friendly): asyncGETStart(), asyncPOSTStart(...), asyncState(), asyncAbort().
+Error codes aligned with HTTPClientLight (−1..−11).
+Async in-RAM body buffer capped ~16 KB (heap safety). Use sync streaming for larger bodies.
+Berry webclient (updated bindings)
+
+New methods: async_get_start(), async_post_start(str|bytes), async_state(), async_abort().
+async_state() publishes last_code (HTTP or negative error) and last_rx (received bytes), and frees hidden holds for byte payloads.
+Headers added via add_header() persist per instance; they reset on begin() and after redirects (re-add if needed).
+Sync API unchanged: GET/POST/PUT/PATCH/DELETE, get_string/bytes, write_file/flash.
+Dependency note
+
+<HTTPClient.h> remains included only for enum/constant types. No Arduino HTTPClient source code is copied; transport, parsing and async infra are independent.
+
+Compatibility
+
+Existing Berry scripts using sync APIs remain compatible.
+OTA/firmware workflows (sync GET + streaming) are unaffected.
+Operational notes
+
+Call async_state() at least once after an async job finishes to publish results and clean up before reusing the same instance.
+Add Content-Type once after begin() (e.g., application/json); re-add after begin() or a redirect.
+Don’t call close() while an async job is running.
+Typical use cases:
+
+Secure periodic telemetry over HTTP(S) + local MQTT
+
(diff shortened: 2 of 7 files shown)


s-hadinger:
Wow impressive work. I merge now since it's an optional feature, and we'll need some additional tests.

Man, what a huge contribution you've made!!!

s-hadinger:
@TheHexaMaster I had a full review conducted by Claude Opus 4.8. Here are the findings.

Beyond the architecture part (one FreeRTOS task per request), do you agree to fix them?

---

## Architecture concern: a FreeRTOS task per request

Every operation — including **synchronous** `GET/POST/getString/writeToStream` — runs by
spawning a new FreeRTOS task (`WorkerTask`, 12 KB stack via `ASYNCHTTP_WKR_STACK`) and then
the caller busy-waits on a semaphore (`runJob`).

- For **sync** calls this is pure overhead: the main loop is blocked the whole time anyway
  (it spins on `xSemaphoreTake` + `delay(0)`), so you pay 12 KB stack + task create/teardown
  per request and gain nothing.
- For **async** it legitimately offloads blocking I/O to the worker.

The cost model (a fresh 12 KB-stack task for each job, low priority `tskIDLE_PRIORITY+1`) is
heavy for an ESP32 and atypical for Tasmota, which normally does cooperative networking on
the main loop / `fast_loop`. **This central design decision deserves a discussion before the
feature goes further.**

---

## Critical bugs (blocking)

### C1. Use-after-free on destruction while a job is in flight
`~AsyncHttpClientLight()` calls `end()` (which does `_transport->stop()` + `_transport.reset()`)
and deletes `_busy`/`_asyncDone`, but it **never aborts or joins a running worker task**.

If the Berry `webclient` object is GC'd or `deinit`'d while `async_state()` is still
`RUNNING` (script drops its reference, or calls `close()`/reuses incorrectly), the worker
keeps dereferencing `self->_transport`, `self->_asyncBody`, and the freed semaphores.
Guaranteed crash.

**Fix:** destructor (and `end()`/`deinit`) must set `_asyncAbortReq`, wait for the worker to
exit, then free.

### C2. Deadlock if response accessors are called while `RUNNING`
`getString()` / `writeToStream()` only take the fast path when `_asyncState == ASYNC_DONE`.
If called while still `RUNNING`, they fall through to `runJob()` →
`xSemaphoreTake(_busy, portMAX_DELAY)`. But `_busy` is only released inside `asyncState()` on
the *same* (now-blocked) main thread. The main loop hangs forever → WDT reset.

The docs say "don't do this," but a library should not deadlock the device on misuse.

### C3. No timeout/abort in the sync body reader → infinite loop
`writeToStreamDataBlock()` (used by sync `getString`, `writeToStream`, `get_bytes`,
`write_file`, `write_flash`) loops:

```cpp
while (_transport && _transport->connected() && (len > 0 || len == -1)) {
  int avail = _transport->available();
  if (avail <= 0) { delay(1); continue; }   // no timeout, no abort check
  ...
}
```

With an identity response, keep-alive, and no `Content-Length` (`_size == -1`), a server that
holds the socket open with no further data spins forever. The readme acknowledges this
"identity with no data" hazard, but only the **async** path (`readBodyToStreamString`) guards
it (abort flag + 16 KB cap). The sync path has neither a timeout nor an abort check.

### C4. Stale `Content-Length` on a 302/303 redirect-to-GET
In `Job_SendBuf`/`Job_SendStr`, each redirect iteration does **not** reset `_headers`;
`Content-Length` is only (re)added when a payload is present. On a 302/303 the code sets
`payload=nullptr`/`size=0` and switches to GET, so the previously injected
`Content-Length: N` stays in `_headers` and is sent on the bodyless GET. The server then
waits for a body that never arrives → hang/timeout.

Related: `beginInternal()` (reached via `setURL()` for absolute redirect targets) never clears
`_headers`, contradicting the readme's "headers reset after redirect."

### C5. Stale async body returned after a later sync call
`_asyncState` is never reset to `IDLE` once `DONE`, and sync `GET()` doesn't touch it. After
one async job, a subsequent **sync** `GET()` on the same instance populates `_size`/transport,
but `getString()` still sees `_asyncState == ASYNC_DONE` and returns the **old** `_asyncBody`.

**Fix:** any new request (sync or async) must invalidate the DONE fast-path.

---

## High / medium issues

### H1. Async POST from `bytes` does not copy the payload
`asyncPOSTStart(uint8_t*, size)` stores the raw pointer; the worker reads it later. The
binding guards GC by stashing the bytes object in `.__async_hold`, but nothing prevents the
script from **mutating/resizing** that `bytes` while the job runs, which can realloc the
buffer → worker reads freed memory. The `String` overload copies (safe); the buffer overload
should too, or the contract must be enforced.

### H2. Unbounded RAM growth for chunked sync responses
`wc_getstring`/`wc_getbytes` guard with `if (sz >= 32767)` using `getSize()`, but a chunked
response reports `_size == -1`, so the guard is bypassed and `Job_GetString` accumulates the
entire body into a `StreamString` with no cap → OOM on a large chunked download. (The async
path is capped at 16 KB; the sync path is not.)

### H3. IPv4-only HTTP transport
`AsyncTcpAdapter` uses `sockaddr_in` / `AF_INET` and `(in_addr_t)ip`. Tasmota supports IPv6;
the previous `WiFiClient`-based path did too. Plain-HTTP async over IPv6 will not work.

### M1. TLS transmit buffer effectively 85 bytes
`new BearSSL::WiFiClientSecure_light(16384, 0)` → `setBufferSizes(recv, xmit)` yields
`_iobuf_out_size = 0 + MAX_OUT_OVERHEAD(85)`. So a 16 KB receive buffer (heavy heap, allocated
up front) is paired with a near-zero send buffer (every TLS record carries almost no app data
— poor POST throughput). Tasmota normally uses balanced sizes (e.g. 1024/1024). Needs tuning.

### M2. `SO_LINGER` RST-close is unconditional and aggressive
`AsyncTcpAdapter::stop()` always sets `l_onoff=1, l_linger=0`, forcing a TCP RST on every close
to cut TIME_WAIT. That can discard unread data and surface as connection-reset on the peer.
Fine as an *option*, questionable as an unconditional default.

### M3. Default HTTPS is unauthenticated
With no pins, `BearSslAdapter` calls `setPubKeyFingerprint(any, any, allow_all=true)` — accepts
any server key (MITM-open). This matches existing Tasmota `_light` behavior, so it is not a
regression, but combined with SHA-1 pubkey pinning it should be documented clearly that
security requires explicit pinning.

s-hadinger:
Anyways, I have put the analysis here: https://github.com/arendst/Tasmota/blob/development/lib/libesp32/AsyncHttpClientLight/CODE_ANALYSIS.md

PR #24481, opened 2026-02-22
Add an independent MQTT client for use by Berry applications and extensions

aschober:
## Description:
- Introduce `mqttclient` Berry class (ESP32) for connecting to MQTT brokers with own credentials, TLS, LWT, and auto-reconnect separate from global MQTT client
- Berry `mqttclient` aims to match functionality of global MQTT client including backoff logic using Settings->mqtt_retry, mqtt_keepalive, and mqtt_socket_timeout
- Rename `retry_counter_delay` to `retry_counter_multiplier` in xdrv_02_9_mqtt for clarity and avoid redundant multiplication.

## Checklist:

### DIFF
--- a/lib/libesp32/berry/default/be_modtab.c
+++ b/lib/libesp32/berry/default/be_modtab.c
@@ -272,7 +272,9 @@ be_extern_native_class(udp);
 be_extern_native_class(webclient);
 be_extern_native_class(tcpclient);
 be_extern_native_class(tcpclientasync);
+#ifdef USE_BERRY_MQTTCLIENT
 be_extern_native_class(mqttclient);
+#endif // USE_BERRY_MQTTCLIENT
 be_extern_native_class(webserver_async);
 be_extern_native_class(tcpserver);
 be_extern_native_class(energy_struct);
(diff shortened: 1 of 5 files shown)


s-hadinger:
Thanks for your contribution, however I'm not sure why this would be useful.

At least guard it with `#define USE_xxx` to not make it in the standard builds

sfromis:
Speculating about potential use cases, maybe connecting to an external broker at the same time as having the regular connection to the inhouse broker. Of course, also for users liking to chop MQTT into separate networks, and only do controlled bridging. Or connecting to an external broker without the inherent security risk in the automatic Tasmota subscription for commands.

Not sure how many would realistically have such use cases.

Jason2866:
The use case what is it made for would be interesting!

s-hadinger:
In such case, the mqtt client would need to be handled only by Berry code, and would not inherit from anything in Tasmota drivers. That's why I'm asking for a use-case.

aschober:
Tasmota w/ Berry provides a flexible platform that can be both a standalone, programmable device while also able to join + add functionality to existing Home Assistant networks so I've been building an application on top of the platform.

Yes, my use-case is to have a Tasmota Application connect to an external broker for application-specific data at the same time as a regular connection to an inhouse broker. And, yes, the external broker should not see any of the automatically-generated Tasmota subscriptions, status, commands, etc as the scope of the external broker is only for application-specific messages and should not see other private device messages. 

I'm working on a Tasmota Application that connects to an external MQTT broker for external data and 1) is able to change state based on the received data and 2) provide the external data as a Sensor to Home Assistant. 

Added `#define USE_BERRY_MQTTCLIENT` but understood if this is not useful enough for inclusion in the main project.

s-hadinger:
Thanks, as long as the added code is guarded by `#ifdef USE_BERRY_MQTTCLIENT` I see no problem in including this code in the main repo.

PR #24501, opened 2026-02-26
webcam CSI: fix Ethernet

Staars:
## Description:

Use Tasmotas hardware agnostic helper function to retrieve the IP address.

Thanks to @sfromis for reporting the bug.

## Checklist:


sfromis:
Maybe a similar fix for MAC would be relevant. I'm seeing Wifi MAC "00:00:00:00:00:00", probably due to Wifi not being operational without connection to an ESP32-C6, but it would be quite ok to just run Ethernet-only. However, the missing MAC does also mean that the MQTT discovery payloads are like `tasmota/discovery/000000000000/config`, meaning no unique device ID available. While this is not breaking much (except for retained on these payloads), it would make sense to use the Ethernet MAC (which is valid) when not having Wifi available.

PR #24575, opened 2026-03-20
vid6608-xdrv-92: allow set max steps per-motor

petrows:
## Description:

This PR adds option to define max range per motor for Automotive VID6608 analog gauge driver. New defines:

* `VID6608_STEPS_DEFAULT` default value for all drives (12 * 320)
* `VID6608_STEPS_1` .. `VID6608_STEPS_4` custom override for particular drive

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.10 / V.3.3.7

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_92_vid6608.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_92_vid6608.ino
@@ -86,6 +86,38 @@
   #define VID6608_RESET_ON_INIT  true
 #endif
 
+/**
+ * @brief Defaine steps range
+ *
+ * Some drives can have another steps scale:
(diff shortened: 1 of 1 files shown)


s-hadinger:
Did you test on ESP8266? Just to make sure the const array in PROGMEM does not make the lib crash.

petrows:
Dear @s-hadinger , you are absolutely right! This code crashes on ESP8266, i tested it incorrectly on devboard - sorry for that. And thanks a lot for your attention.

I have reworked update, now it uses `memcpy_P`. Tested more and found another bug: there was incorrect calculation for `GaugePercent` command, as it has to operate with max-steps value per motor.

PR #24594, opened 2026-03-25
Clear default Wi-Fi gateway/subnet and reset DNS to 0.0.0.0

HMvdZijden:
Remove hardcoded `192.168.1.1` defaults for `WIFI_GATEWAY`, `WIFI_SUBNETMASK`, and `WIFI_DNS` in `my_user_config.h` to avoid unintended static IP configuration when DHCP is in use.


s-hadinger:
Why do those default value cause a problem? Blanking the values will not make them disappear but replaced with 0.0.0.0. What would be the behavior of Tasmota with such values? (I never tested)

arendst:
As long as WIFI_IP_ADDRESS is 0.0.0.0 there is only dhcp. If a user changes it he is supposed to change the defaults too unless they are correct.

Not an issue so won't merge.

PR #24599, opened 2026-03-28
Berry solidification made possible with Python runtime

s-hadinger:
## Description:

The Berry runtime was ported from `C` to `Python` which allows to move solidification stage into the toolchain. The goal is eventually to introduce new features like conditional compilation (smaller firmwares) and text localization.

The porting took less than a week using Kiro and Claude Opus 4.6. It generates a byte-to-byte compatible solidification source code. 

The solidification is not integrated in the toolchain yet.

As a side effect, `int64` support was ported to plain `C` code not using `lv_mapping` which makes it also faster.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.1.10 / V.3.3.7

### DIFF
--- a/lib/libesp32/berry/berry_port/__init__.py
+++ b/lib/libesp32/berry/berry_port/__init__.py
@@ -1,0 +1,3 @@
+# Berry Python Port
+# Port of the Berry scripting language interpreter from C to pure Python.
+# See: https://github.com/Skiars/berry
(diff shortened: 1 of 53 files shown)


Jason2866:
It is astonishing to see the spectrum from AI. From awesome jobs like this. Otoh it fails with damn simple stuff.

PR #24600, opened 2026-03-29
Add devcontainer and automations configuration

Thomas1208:
Sets up a reproducible development environment using the Ona devcontainer format.

**`devcontainer.json`**
- Base image: `mcr.microsoft.com/devcontainers/python:3` (Python is the PlatformIO runtime)
- VS Code extension: `platformio.platformio-ide`

**`.ona/automations.yaml`**
- `install-platformio`: installs PlatformIO CLI via pip on `postDevcontainerStart`
- `build`: manual task running `pio run -e ${PIO_ENV:-tasmota}`; defaults to the ESP8266 `tasmota` variant, override with `PIO_ENV=tasmota32` etc.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,11 @@
+{
+	"name": "Tasmota",
+	"image": "mcr.microsoft.com/devcontainers/python:3",
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"platformio.platformio-ide"
+			]
+		}
+	}
+}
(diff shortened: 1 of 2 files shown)


Jason2866:
With the change it is everytime needed to click away the Message window if it should start in devcontainer. Furthermore Tasmota is optimized for use with VSC extension pioarduino.

Does ONA provide the possibility to use without providing any payment data?

Furthermore the used Tasmota Platform espressif32 does not use `pip`. The use of `pip` is deprecated. The Platform uses `uv` only. Tasmota uses the fork pioarduino for the pio core. Not Platformio core. Some features will not work correctly using Platformio core.
See the Tasmota Github Actions how to setup the build environment.

PR #24709, opened 2026-05-05
Matter Zigbee bridge: route per-endpoint friendly names to correct endpoint

lePaulo:
## Summary

Fix multi-endpoint Zigbee device addressing in the Matter Zigbee bridge. The bridge mapped each Matter endpoint to a Zigbee `shortaddr` only, losing endpoint information. For multi-endpoint devices (e.g. 2-channel on/off relays), every Matter command routed to Zigbee endpoint 1 and incoming reports updated all Matter plugins targeting that device — making one channel uncontrollable and the other's state shared across both Matter endpoints.

## Changes

All in `lib/libesp32/berry_matter/src/embedded/Matter_z_Zigbee.be`:

### 1. Explicit `:N` syntax for per-endpoint targeting (per @s-hadinger's review)

Config value `"zigbee_device":"my_device:2"` parses to `device="my_device", endpoint=2`. Works for shortaddr (`"0xBE4B:2"`) or device-level friendly name (`"Nodon_1:2"`).

### 2. Per-endpoint friendly name auto-resolution (fallback)

When no explicit `:N` is given and the config string is a non-hex name, the mapper queries `ZbName <name>` once at first `resolve_zb_device` and walks the response to find the source endpoint. Preserves the existing Tasmota convention where per-endpoint friendly names are set via `ZbName <dev>,<name>,<ep>`.

### 3. Outgoing path

Pass the original config string verbatim in the `ZbSend` `Device` field — `ZbSend`'s `parseDeviceFromName` resolves per-endpoint friendly names natively. Adds explicit `"Endpoint":N` when known.

### 4. Incoming path

`Matter_Zigbee.attributes_final` filters by `attr_list._src_ep` when the mapper has a known endpoint. Note: `frame` is always `nil` at this call site (caller passes `nullptr`), so the filter relies on the attribute list's source endpoint exposed via the existing ctypes binding at [`xdrv_52_3_berry_zigbee.ino:314`](https://github.com/arendst/Tasmota/blob/development/tasmota/tasmota_xdrv_driver/xdrv_52_3_berry_zigbee.ino#L314).

### 5. Endpoint-aware boot probe

`probe_zb_values` now triggers a `ZbRead` per attribute the plugin tracks (`Power`/`Dimmer`/`CT`) for the specific endpoint. The response carries `_src_ep`, flows through `attributes_final`, and is routed to the right Matter plugin only.

The previous behavior (calling `pi.zigbee_received(nil, device.info())`) pushed the merged Z_Data of *all* endpoints into every plugin targeting the same shortaddr — so two Matter endpoints for the same Zigbee device both reflected the latest-touched endpoint's state at boot. Single-endpoint configs (no endpoint set) keep the cheap cached probe — no extra Zigbee traffic for the common case.

## Backwards compatibility

| `zigbee_device` value | Behavior |
|---|---|
| `"0x1234"` (hex string) | unchanged — defaults to first endpoint |
| `"my_device"` (device-level name) | unchanged — defaults to first endpoint |
| integer | unchanged — defaults to first endpoint |
| `"my_device:N"` (explicit) | NEW — routes to endpoint N |
| `"per_endpoint_name"` (set via `ZbName <dev>,<name>,<ep>`) | NEW — auto-resolves to N via ZbName lookup |

Single-endpoint devices and existing configs are unaffected.

## Tested on

A 2-channel Zigbee on/off device (`DeviceId 0x0100`) paired to a Tasmota Ethernet Zigbee gateway running Tasmota 15.4.0.

Setup:
\`\`\`
ZbName 0x1234,multi_relay
ZbName multi_relay,light_ep1,1
ZbName multi_relay,light_ep2,2
ZbLight multi_relay,0,1
ZbLight multi_relay,0,2
\`\`\`

Matter config (either form works):
\`\`\`berry
matter_device.plugins_config = {
  "3": {"type":"z_light0","zigbee_device":"multi_relay:1","name":"ep1"},
  "4": {"type":"z_light0","zigbee_device":"light_ep2","name":"ep2"}
}
\`\`\`

## Test plan

- [x] Multi-endpoint outgoing routing — Apple Home toggles each Matter endpoint, only the matching Zigbee endpoint fires
- [x] Multi-endpoint incoming filtering — physical toggle on either relay updates only the matching Matter endpoint
- [x] Boot state — endpoint-aware probe correctly populates per-endpoint state via ZbRead
- [x] Backwards compatibility — `zigbee_device:"0x1234"` and device-level friendly names still default to first endpoint
- [x] Both `:N` explicit syntax and per-endpoint friendly name auto-derivation produce the same `self.endpoint` filter behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)


s-hadinger:
Thanks this is a good addition, although I'm not in favour of just adding the index without a separator. I would rather have `"light_ep:2"` 
Can you make this change?

lePaulo:
Hi Stefan - sure. Good advice. Also thinking about '/' as a separator. What do you prefer ':' or '/'?
Cheers!
Paul

s-hadinger:
I tend to prefer `:` since it's only a numerical modifier, but if you insist on `/` I'm ok too

lePaulo:
Thanks for the review @s-hadinger.

I implemented your :N separator suggestion — explicit syntax is now supported, e.g. "zigbee_device":"my_device:2".

I also kept the existing per-endpoint friendly name auto-resolution (e.g. "zigbee_device":"light_ep2" where light_ep2 is set via ZbName 0x1234,light_ep2,2) as a fallback. 

Also added in this push: endpoint-aware boot probe via ZbRead, replacing the previous merged-state push. Without this, a multi-endpoint device's Matter plugins could all reflect the same (latest-touched) endpoint's state at boot. Tested on real hardware (NodOn SIN-4-2-20 dual-channel relays + ZB-GW03 Ethernet gateway) — Apple Home now shows correct per-endpoint state immediately after a power cycle.

Hope the code is self explanatory, but if you have any question or comment, don't hesitate!

Cheers!
Paul

Jason2866:
Please delete the solidified Berry file. It is build automatically

s-hadinger:
I removed the solidified file and merged

PR #24720, opened 2026-05-07
redirect the floating-point capable vfprintf / vfscanf symbols to picolibc's integer-only variant

Jason2866:
which saves around 6-7 kb flash space for all esp32 variants which uses `picolibc`. Only S2 does not use `picolibc` (bug in IDF) 

## Checklist:

### DIFF
--- a/pio-tools/picolibc_flags.py
+++ b/pio-tools/picolibc_flags.py
@@ -1,0 +1,26 @@
+#
+# Picolibc-specific linker flags.
+#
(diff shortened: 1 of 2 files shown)


s-hadinger:
I'm afraid this would break Berry floating point formatting that uses `snprintf()`, so doable for non-Berry variants (safeboot) but not the other variants

s-hadinger:
With #24725 merged, you can merge this one. No more issue with Berry

PR #24731, opened 2026-05-10
Fix default hostname generation to support single-specifier

aschober:
Fix default hostname generation to support single-specifier Format() patterns

- Route single-specifier hostnames through Format() so patterns like %NX (last N MAC hex chars) work correctly, while preserving the existing two-specifier %s-%04d behavior via snprintf
- Update comment in my_user_config.h to accurately describe chip ID rather than MAC

## Description:

## Checklist:


protectivedad:
This merge breaks being able to set WIFI_DEFAULT_HOSTNAME to "%s". Is there a way to keep the ability to set a hostname to single specifier that allows for the hostname to be just the MQTT topic?

arendst:
Ah, I see what you did. From the "original" two specifiers `%s-%04d` you only used `%s` while the code expected the second specifier. This worked as the code dropped the second and returned the topic only.

This change does a test for one or two specifiers and if only one specifier is found it performs a new action.

Let me chk if and how this can be resolved...

arendst:
Should be fixed now in latest dev branch (https://github.com/arendst/Tasmota/commit/cde8e62071570b996edaf0c1c9fc48e78b8f547e)

PR #24732, opened 2026-05-10
Add local file support to solidify-from-url

aschober:
Add local file support to solidify-from-url.py and fix URL rename alias bug

- Add support for local .be file paths (absolute or PROJECT_DIR-relative) in prepareBerryFiles used by custom_berry_solidify project option in solidify-from-url.py
- Fix URL path to use "/" as separator instead of os.path.sep, and pass the rename alias to addHeaderFile so the solidified header name matches the actual target filename

## Description:

## Checklist:

### DIFF
--- a/pio-tools/solidify-from-url.py
+++ b/pio-tools/solidify-from-url.py
@@ -3,32 +3,12 @@
 Import("env")
 
 import os
-import sys
 from genericpath import exists
 from os.path import join
 import subprocess
 from colorama import Fore, Back, Style
 import requests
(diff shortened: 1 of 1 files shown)


Jason2866:
Since you are already change. For coc there is no compiled binary used anymore. It has been ported to Python. Can you please adopt this too?

aschober:
@Jason2866 Updated `solifidify-from-url.py` to match how `gen-berry-structures.py` works.

PR #24738, opened 2026-05-13
Add descriptor-driven packed mono support for ST7305 RLCD

speccy88:
## Summary

This PR adds descriptor-driven support for the Waveshare ESP32-S3-RLCD-4.2 reflective LCD panel using a 400x300 1bpp SPI RAMWR display.

The important part of this revision is that the low-level SPI panel driver does **not** contain ST7305-specific controller code. Instead, it adds a small generic packed-monochrome transfer option to the uDisplay descriptor parser:

```ini
:F,2,4,3
```

For this descriptor, that means:

- pack pixels in 2-column x 4-row blocks
- stream bits row-major from bit 7 down to bit 0
- invert the final byte polarity
- reverse the Y scan direction

The ST7305-specific details remain in the descriptor file:

```text
tasmota/displaydesc/ST7305_RLCD_400x300_display.ini
```

## Why this approach is different from the first version

The first implementation worked on the hardware, but it was wired as a controller-specific mode inside the generic SPI panel driver:

- `:P,73`
- `UDISP_PIXFMT_ST7305_2X4`
- `updateFrameST7305_2X4()`

That was not aligned with the uDisplay driver design because it put a specific controller name and a controller-specific transfer path into the low-level panel code.

This version removes that approach. There are no ST7305-specific defines, constants, or function names in `lib/lib_display/UDisplay`. The low-level SPI panel now only knows about a generic descriptor-selected packed 1bpp transfer format. The descriptor decides whether to use it and which geometry/flags to use.

## Why this cannot be fixed only with existing display.ini fields

The panel does not use the existing monochrome OLED page layout. It accepts data through a TFT-style RAMWR flow, but one byte represents a 2x4 pixel block:

- bit 7 = row 0, column 0
- bit 6 = row 0, column 1
- bit 5 = row 1, column 0
- bit 4 = row 1, column 1
- bit 3 = row 2, column 0
- bit 2 = row 2, column 1
- bit 1 = row 3, column 0
- bit 0 = row 3, column 1

The existing descriptor commands can describe pins, init commands, address commands, rotation bytes, and color mode values, but they could not describe this framebuffer-to-transfer repacking step.

The new `:F,width,height,flags` descriptor field fills that gap without adding controller-specific code. For this display it is `:F,2,4,3`.

## Why this is less intrusive

- No ST7305-specific code in the low-level SPI panel driver.
- Existing descriptors are unaffected because packed mono is disabled unless `:F` is present on a 1bpp SPI descriptor.
- Existing `:P` behavior remains unchanged.
- Existing SPI TFT and SPI/I2C monochrome paths remain unchanged unless the descriptor opts in.
- The generic packed mono path is reusable for any future 1bpp RAMWR-style panel with a descriptor-declared block geometry.

The only shared correctness fixes are:

- 1bpp framebuffer allocation now rounds height up to the next 8-pixel page, matching the renderer's existing `x + (y >> 3) * width` addressing.
- `Renderer::drawFastVLineInternal()` uses 16-bit local `y`/`h` values so full-height clears above 255 pixels do not wrap.

Those are not ST7305-specific; they fix generic 1bpp behavior for taller displays.

## Hardware validated

Tested on:

- Board: Waveshare ESP32-S3-RLCD-4.2
- Display: ST7305 reflective LCD, 400x300 monochrome
- SPI pins:
  - CS GPIO40
  - SCK GPIO11
  - MOSI GPIO12
  - DC GPIO5
  - RESET GPIO41
  - no MISO
  - no backlight

Validated behavior:

- uDisplay model 17 initializes from `/display.ini`.
- `DisplayText [z]` clears the full 400x300 framebuffer.
- `DisplayText` text and primitives render correctly.
- LVGL renders text and objects correctly.
- Display off/on replays the framebuffer instead of exposing stale controller RAM.

## Build validation

Built successfully with:

```sh
~/.local/bin/pio run -e tasmota32s3-lvgl
```

The resulting firmware was flashed and tested on the hardware above.


Jason2866:
You have to find a different way to wire in the changes. Maybe a change of the descriptor ini parser?

speccy88:
Updated this PR to address the uDisplay architecture concern. The branch no longer adds ST7305-specific defines or controller-named functions in the low-level SPI panel driver. Instead, it adds a generic descriptor parser option, `:F,width,height,flags`, for packed 1bpp RAMWR-style panels. The ST7305 descriptor uses `:F,2,4,3` to describe its 2x4 packing, output polarity, and Y scan direction. This keeps the controller-specific behavior in the descriptor and leaves existing display configurations unchanged unless they opt in with `:F`.

PR #24759, opened 2026-05-17
VID6608: add hardware RMT support for Automotive gauge driver

petrows:
## Description:

Hey Folks, i would like to introduce my weekend project.

This PR updates the xdr-92 (the Automotive gauge stepper driver) for vid6608 chipset: add hardware acceleration for gauge drive using the [ESP-32 Remote Control Transceiver (RMT)](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/rmt.html). Feature can be enabled with the new `VID6608_RMT` config option, default is `false`.

### Now driver has 3 modes:

* Software, single-thread (ESP8266 only), using the old `arduino-vid6608` library; Very slow, but works everywhere.
* Software, multi-thread (ESP32 default), using the old `arduino-vid6608` library with FreeRTOS task; Moderate, works good on dual-core chips, works with issues on single-core chips (tested on ESP32C6).
* Hardware (ESP32 RMT) using the new `esp-32-vid6608-rmt` library; Best performance, but requires free RMT available.

### RMT driver notes

RMT driver gives really smooth movement, no glitches, even on weak chips. No `delay()` function calls and minimum load on SoC.

C0028-cut-2

Full video: https://youtu.be/dASliRoJWo4

Looks like it conflicts with WS2812b driver with default configuration (and likely other drivers) on weak platforms with reduced feature set, so RMT is disabled by default and requires explicit config option.

Acceleration curve is much more complex to implement for RMT driver, this implementation has very small one (compare to software variant), i will try to improve it in the future.

Documentation update: https://github.com/tasmota/docs/pull/1556

## Checklist:


petrows:
During work on this patch i have found another issue: #24760 ; currently it is not possible to switch (via user-config-h) the WS2812b driver to non-RMT mode, which is used by default.

s-hadinger:
This looks good to me. @arendst ok for merging?

PR #24761, opened 2026-05-17
VsCode: disable auto-trim whitespaces

petrows:
## Description:

Saving some files (source code, ini config) generates extremely annoying garbage diff with whitespace changes. Example of such commit: [2a94354](https://github.com/arendst/Tasmota/commit/2a9435449de7ae8939a2a4d57813936b5a38c22d#diff-14b190596cf1e1a8a1688aa4946790e3147e976e1b1a3d3d10d52a44d51b7227).

image

This PR disables such behavior for VsCode IDE.

## Checklist:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -74,5 +74,6 @@
         "commands": "platformio-ide.rebuildProjectIndex"
       }
     ],
-    "C_Cpp.dimInactiveRegions": false
+    "C_Cpp.dimInactiveRegions": false,
+    "editor.trimAutoWhitespace": false
   }


s-hadinger:
I believe auto-trim is useful. What you show is not gargabe diff, it's cleaned source code.

petrows:
Hey @s-hadinger , i agree with you that code should be trimmed and i have this option enabled in my IDE by default. Problem here is that such changes follow the another one - and they became part of PR - as unrelated changes. So they are not "garbage" itself, they are just in a wrong place.

My suggestion:
* Merge this PR to allow correct workflow as status-quo
* Prepare and open another PR later:
  * Trim white-spaces from common places, like config and header files
  * Enforce trim by changing option back: `"editor.trimAutoWhitespace": true`
  * Update the [CONTRIBUTING.md](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md) guideline with auto-trim allowance (currently is is explicitly prohibeted by rule: Pull requests / no.2)
  * Merge this as another single PR

I think this is a correct workflow here. What do you think? Also question for @arendst

s-hadinger:
I agree with the PR for triming and uddate in `CONTRIBUTING.md` but I don't see why we need to push a change for all users in between. Just pushing. global trimming PR should be fine.

petrows:
Hey @s-hadinger , you are right. I agree that easier would be to form this as a single PR. We should wait for @arendst , then please decline this PR - i will create a new one with all changes together.

arendst:
Yeah, auto trim should have been ON all the time.

PR #24787, opened 2026-05-25
Berry: use Tasmota Berry defines in coc

Staars:
## Description:

Adds visibility of `tasmota_defines_for_berry.h` to the Berry coc, which makes it possible to solidify optional code, that only makes sense in certain firmwares.
It is used with an optional third argument in the `@const_object_info_begin` section of the Berry file, that shall be solidified. Then  i.e. a certain class will only appear in a Berry module, if a build flag in an environment was set.

Second part of this PR is an example to only add WiFi commissioning in Matter via BLE, if the firmware allows this feature. Thus regular firmware sizes do not increase with unneeded overhead of special use cases.

## Checklist:

### DIFF
--- a/lib/libesp32/berry/tools/coc/coc
+++ b/lib/libesp32/berry/tools/coc/coc
@@ -16,7 +16,7 @@ class builder:
     def __init__(self, input_folders, output_folder, macro_files):
         self.output = output_folder
         self.input = input_folders
-        self.config = macro_files
+        self.config = list(macro_files)
         self.macro = None
         self.strmap = {}
         self.strmap_weak = {}
@@ -24,8 +24,9 @@ class builder:
         self.bytesmap = {}
 
         self.macro = macro_table()
+        self.config.append("../../../tasmota/tasmota_defines_for_berry.h")
         for path in self.config:
-            self.macro.scan_file(path)                
+            self.macro.scan_file(path)       
 
         for d in self.input:
             self.scandir(d)
--- a/lib/libesp32/berry_matter/src/be_matter_module.c
+++ b/lib/libesp32/berry_matter/src/be_matter_module.c
@@ -273,6 +273,11 @@ extern const bclass be_class_Matter_TLV;   // need to declare it upfront because
 #include "solidify/solidified_Matter_Plugin_9_Zigbee_Humidity.h"
 #include "solidify/solidified_Matter_Plugin_9_Zigbee_Occupancy.h"
 #include "solidify/solidified_Matter_Plugin_z_All.h"
+#if USE_MI_EXT_GUI
+#include "solidify/solidified_Matter_BTP.h"
+#include "solidify/solidified_Matter_zz_Device_BLE.h"
+// #warning ________________
+#endif //USE_MI_EXT_GUI
 #include "solidify/solidified_Matter_zz_Device.h"
 
 #include "be_fixed_matter.h"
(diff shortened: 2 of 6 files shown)


Staars:
Application example to use Matter commissioning via BLE, which would always be device specific in the real world as there is no single way to do it right.

```
#------------------------------------------------------------------------------
- Matter BLE - app-specific demo subclass
-
- All generic Matter-over-BLE behavior lives in `matter.Device_BLE`
- (lib/libesp32/berry_matter/src/embedded/Matter_zz_Device_BLE.be) and the
- pure BTP protocol is `matter.BTP`
- (lib/libesp32/berry_matter/src/embedded/Matter_BTP.be).
-
- This file only carries the demo light feedback used while commissioning a
- dev board.
-------------------------------------------------------------------------------#
import matter

class MATTER_BLE : matter.Device_BLE
    var have_light

    def on_ble_init()
        try
            import light
            light.set({'bri': 5, 'hue': 150, 'power': true, 'sat': 255})
            self.have_light = (light.get() != nil)
        except ..
            self.have_light = false
        end
    end

    def heart_beat()
        if self.commissioning.is_commissioning_open() == false  return end
        if self.have_light
            import light
            var l = light.get()
            var hue = (l['hue'] + 1) % 256
            light.set({'bri': l['bri'] == 50 ? 5 : 50, 'hue': hue})
        end
    end

    def on_commissioning_success(ctx)
        if self.have_light
            import light
            light.set({'bri':50,'hue':120})    # green
        end
    end

    # ctx = {'reason': <REASON_*>, 'detail': <str>, 'fabric_count': <int>}
    # base class already rearms BLE commissioning - we just show a hint colour
    def on_commissioning_failure(ctx)
        if self.have_light
            import light
            var hue = 0                       # red default
            if   ctx['reason'] == self.REASON_WIFI_FAILED    hue = 30   # orange
            elif ctx['reason'] == self.REASON_WIFI_TIMEOUT   hue = 45   # yellow-orange
            elif ctx['reason'] == self.REASON_PASE_FAILED    hue = 290  # magenta
            elif ctx['reason'] == self.REASON_WINDOW_TIMEOUT hue = 240  # blue
            end
            light.set({'bri':50,'hue':hue})
        end
    end
end

return MATTER_BLE()


# Example for autoexec.be with demo template for dev board
#-
if tasmota.cmd("template").tostring() != {'NAME':'NodeMCU-S3-12k','GPIO':[1,1,1,1,1,416,417,418,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,0,0,0,0,1,1,1,1,1,1,1,1,1,1,0],'FLAG':0,'BASE':1}.tostring()
   tasmota.cmd('backlog template {"NAME":"NodeMCU-S3-12k","GPIO":[1,1,1,1,1,416,417,418,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,0,0,0,0,1,1,1,1,1,1,1,1,1,1,0],"FLAG":0,"BASE":1}; module 0;')
end
if tasmota.cmd("SSId1")["SSId1"] == "" && tasmota.cmd("SSId2")["SSId2"] == ""
    if tasmota.cmd("so151")["SetOption151"] == "OFF"
        tasmota.cmd("so151 1")
    end
    # this is the state after first flash or reset 1
    tasmota.cmd("so115 1") # make sure BLE is on
    tasmota.set_timer(5000, /->load("this_file.be")) # let the system do some booting
end
-#
```

PR #24792, opened 2026-05-27
JsonParser: fix NULL dereference on allocation failure

joluxer:
## Checklist:
  - [v] The pull request is done against the latest development branch
  - [v] Only relevant files were touched
  - [v] Only one feature/fix was added per PR and the code change compiles without warnings
  - [v] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [v] The code change is tested and works with Tasmota core ESP32 V.3.3.8 from Platform 2026.05.50
  - [v] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

---

## Problem

`JsonParser::allocate()` calls `new jsmntok_t[_size]` without checking the result.
On platforms built with `-fno-exceptions` (ESP8266, and ESP32 in the default
Tasmota build), `new[]` returns `nullptr` on allocation failure instead of throwing.
Four locations in `JsonParser` subsequently dereference `_tokens` without a NULL
guard, producing a hard crash whenever the token buffer cannot be allocated.

Observed crash signature (confirmed via addr2line):

```
Exception 28 (LoadStoreError)
epc1 = 0x4024c3a0  ->  JsonParser::postProcess()
excvaddr = 0x00000014  ->  NULL + 5 * sizeof(jsmntok_t)
```

Trigger: heap exhausted during rule evaluation (e.g. after a WiFi reconnect), JSON
parsing attempted with insufficient memory, `allocate()` fails silently, first
token access crashes.

## Research

I tried to find the origin repository and the adapter code. I found the origin
project, but could not find the adapter code, which fails.

Thus I decided to do the fixes in the Tasmota files. 

## Fix

In files (`lib/default/jsmn-shadinger-1.0/src/JsonParser.cpp` and `JsonParser.h`):

Four paths corrected:

- `allocate()`: `if (!_tokens) { _size = 0; }` after `new[]` -- sets safe empty
  state; subsequent calls to `parse()` skip processing.
- `parse()`: `(_tokens != nullptr) &&` guard before `postProcess()`.
- `parse()` error path: `_token_len = 0` added -- implements the existing `TODO`
  comment; ensures `operator bool()` returns `false` after any parse failure.
- `getRootObject()` (.cpp) and `getRoot()` (.h): return `&token_bad` sentinel
  instead of `&_tokens[0]`.

Under normal conditions (sufficient heap): no behavior change.
Under allocation failure: `operator bool()` returns `false`; callers that check
the result handle it gracefully without crashing.

## Memory impact (ESP8266 release build)

| | Flash | RAM | IRAM |
|---|---|---|---|
| This fix | +16 B | 0 | 0 |

Minimal Flash delta from the added null guards; no static RAM or IRAM impact.

## Tests

Crash confirmed via `addr2line` against `firmware.elf` -- `excvaddr=0x14` is an
exact match for `_tokens[5]` with `_tokens == nullptr` and `sizeof(jsmntok_t) == 4`.
Build-tested on ESP8266.

### DIFF
--- a/lib/default/jsmn-shadinger-1.0/src/JsonParser.cpp
+++ b/lib/default/jsmn-shadinger-1.0/src/JsonParser.cpp
@@ -305,7 +305,7 @@ JsonParser::~JsonParser() {
 }
 
 const JsonParserObject JsonParser::getRootObject(void) const {
-  return JsonParserObject(&_tokens[0]);
+  return (_tokens != nullptr) ? JsonParserObject(&_tokens[0]) : JsonParserObject(&token_bad);
 }
 
 const JsonParserToken JsonParser::operator[](int32_t i) const {
@@ -408,11 +408,11 @@ void JsonParser::parse(char * json_in) {
   allocate();
   jsmn_init(&this->_parser);
   _token_len = jsmn_parse(&this->_parser, json_in, json_len, _tokens, _size);
-  // TODO error checking
-  if (_token_len >= 0) {
+  if ((_tokens != nullptr) && (_token_len >= 0)) {
     postProcess(json_len);
   } else {
-    this->free();   // invalid JSON
+    this->free();
+    _token_len = 0;
   }
 }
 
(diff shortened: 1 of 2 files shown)


s-hadinger:
Awesome. Thanks

PR #24793, opened 2026-05-28
tools: add ESP8266 crash stack dump decoder

joluxer:
## Checklist:
  - [v] The pull request is done against the latest development branch
  - [v] Only relevant files were touched
  - [v] Only one feature/fix was added per PR and the code change compiles without warnings
  - [v] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [v] The code change is tested and works with Tasmota core ESP32 V.3.3.8 from Platform 2026.05.50
  - [v] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

# Description

Adds `tools/decode-esp8266-stack.sh`, a shell script for resolving ESP8266 crash
output to function names and source locations.

My default build target is `tasmota-4M`, thus the internal default of the script for `tasmota-4M`.

## What it does

Reads an ESP8266 serial crash log (from file or stdin) and resolves:
- Exception program counters (`epc1`, `epc2`, `epc3`, `excvaddr`, `depc`)
- Stack frame addresses from the `>>>stack>>>` ... `<<<stack<<<` section

Each address is passed through `xtensa-lx106-elf-addr2line` and
`xtensa-lx106-elf-c++filt`. Addresses that do not resolve to code (data words
misidentified as stack addresses) are silently suppressed to keep output readable.

## Usage

```sh
# from a saved log file
./tools/decode-esp8266-stack.sh crash.log

# piped directly from a serial terminal
cat /dev/ttyUSB0 | ./tools/decode-esp8266-stack.sh

# other build environment
ELF_TARGET=tasmota-minimal ./tools/decode-esp8266-stack.sh crash.log
```

## Requirements

- `firmware.elf` at `.pio/build/<ELF_TARGET>/firmware.elf` (default: `tasmota-4M`).
  Override via `ELF_TARGET` environment variable.
- `xtensa-lx106-elf-addr2line` and `xtensa-lx106-elf-c++filt` from the PlatformIO
  toolchain. Located automatically under `~/.platformio/packages`; falls back to a
  `toolchain-bin/` directory next to the script, where one could link the binutils
- For source file names and line numbers: firmware built with `-g` in `build_flags`.
  Without `-g`, function names are still resolved.

## No firmware impact

The script is not part of the build process and does not
affect Flash, RAM, or IRAM.

### DIFF
--- a/tools/decode-esp8266-stack.sh
+++ b/tools/decode-esp8266-stack.sh
@@ -1,0 +1,115 @@
+#!/bin/sh
+# decode-esp8266-stack.sh — decode ESP8266 crash stack dump using xtensa-lx106-elf-addr2line
+#
+# Usage:
+#   ./decode-esp8266-stack.sh crash.log
+#   cat crash.log | ./decode-esp8266-stack.sh
+#
+# The firmware ELF is expected at:
+#   <repo-root>/.pio/build/<ELF_TARGET>/firmware.elf
+#
+# Override the build target via environment variable:
+#   ELF_TARGET=tasmota-minimal ./decode-esp8266-stack.sh crash.log
+#
+# NOTE: The ELF must contain debug info. Build with the debug_build_flags
+# option in platformio.ini (or a build_flags = -g override in your env):
+#
+#   [env:tasmota-4M]
+#   build_flags = ${common.build_flags} -g
+#
+# Without -g, addr2line will resolve addresses but cannot report source
+# file names and line numbers.
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+: "${ELF_TARGET:=tasmota-4M}"
+
+ADDR2LINE="$(find ~/.platformio/packages -name "xtensa-lx106-elf-addr2line" 2>/dev/null | head -1)"
+if [ -z "$ADDR2LINE" ]; then
+    ADDR2LINE="$SCRIPT_DIR/toolchain-bin/xtensa-lx106-elf-addr2line"
+fi
+if [ ! -x "$ADDR2LINE" ]; then
+    echo "ERROR: addr2line not found or not executable: $ADDR2LINE" >&2
+    exit 1
+fi
+
+CXXFILT="$(find ~/.platformio/packages -name "xtensa-lx106-elf-c++filt" 2>/dev/null | head -1)"
+if [ -z "$CXXFILT" ]; then
+    CXXFILT="$SCRIPT_DIR/toolchain-bin/xtensa-lx106-elf-c++filt"
+fi
+if [ ! -x "$CXXFILT" ]; then
+    echo "ERROR: c++filt not found or not executable: $CXXFILT" >&2
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


tools/decode-esp8266-stack.sh
```diff
@@ -0,0 +1,115 @@
...
+#
+#   [env:tasmota-4M]
+#   build_flags = ${common.build_flags} -g
+#
+# Without -g, addr2line will resolve addresses but cannot report source
+# file names and line numbers.
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+: "${ELF_TARGET:=tasmota-4M}"
```
eku, line 25: Is there any reason not to pass such parameters to the script as options rather than via the environment, as is standard practice for scripts?

tools/decode-esp8266-stack.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/sh
```
eku:
```suggestion
#!/usr/bin/env sh
```

tools/decode-esp8266-stack.sh
eku, line 1: The script uses non-sh features:

```
 shellcheck decode-esp8266-stack.sh 

In decode-esp8266-stack.sh line 69:
    local addr="$1" label="$2"
    ^-------------------^ SC3043 (warning): In POSIX sh, 'local' is undefined.


In decode-esp8266-stack.sh line 70:
    local result
    ^----------^ SC3043 (warning): In POSIX sh, 'local' is undefined.

For more information:
  https://www.shellcheck.net/wiki/SC3043 -- In POSIX sh, 'local' is undefined.

```

eku (changes requested):

tools/decode-esp8266-stack.sh
joluxer: Of cause this is possible. Optional script parameters are easier to pass in from environment variables, especially, if they have default values and need to be set.

This was the primary reason why I used that way, to pass the parameter to Bash.

Eventually an ENV parameter and a commandline parameter end up the same typing amount from my perspective. Thats why I do not care. 

Programming effort is higher for (optional) command line parameters, thats often the reason for me to fall back to ENV parameters.

tools/decode-esp8266-stack.sh
joluxer: Then the tool should rely on bash, instead of sh, shouldn't it?

Jason2866:
Will not be merged. For Tasmota pioarduino VSC extension is recommended and use of https://marketplace.visualstudio.com/items?itemName=Jason2866.esp-decoder
for crash decoding

tools/decode-esp8266-stack.sh
eku: If you switch to bash anyway, have a look at the [Advanced Bash-Scripting Guide](https://tldp.org/LDP/abs/html/internal.html#EX33).

tools/decode-esp8266-stack.sh
eku: Not necessarily. The task can also be performed using the standard shell, in which case you wouldn't need to use `local`.

PR #24794, opened 2026-05-28
build: copy firmware artifacts with ELF-extracted build timestamp (opt-in)

joluxer:
## Checklist:

# Description

During the course of console log examination along a serious number of rebuild-and-try
sessions I stumbled across the recurring problem, that I could not easily match
the console output to the ELF file, that produced the output and the crash 
dumps - neither manually nor automatically. The latter was the more painful problem for
my test series'.

A typical solution for this in my MCU world is, to take the firmware timestamp,
which is output at start. Such a timestamp is taken from the compiler while 
building some sort of version header or version build artifact. 

Tasmota has the same best practice of taking the build timestamp and putting it
to the console while booting. This is a lean solution which I aprecciate and tried
to keep, despite the fact, that the information is needed by my tools externally 
from the build artifact. 

## Suggested solution

Some solutions in MCU build systems create the timestamp at start of the build system 
or at start of the compilation stage or at start of the linking stage and inject them
into the build artifacts. This could rely in file system timestamps, on macro
defintions or other mechanisms, like binary patches, etc.

Not relying on file system timestamps, which are unreliable on windows anyways, a
different way of taking the build timestamp from the build process was my option.

Extracting the build timestamp from the resulting ELF file seems the least invasive
idea for the build system and the firmware. Some research showed, that the date part
of the timestamp is alread in a named symbol, the time part not yet - it was anonymous. 

So making a named symbol for the time part was a tiny step in the firmware itself.

Now both timestamp parts are extractable from the ELF file without guessing and an 
external tool could do the rest of the lifting. 

This PR adds `pio-tools/timestamp-firmware.py`, a post-build script that 
copies firmware artifacts to timestamped filenames. Opt-in; disabled by default; 
standard builds are unaffected.

Intended use: correlating captured logs (crash dumps in serial output, automated
test results) with the exact binary that produced them, when multiple builds and 
console logs are archived side by side.

## Enabling

Add to the `[tasmota]` section of `platformio_override.ini`:

```ini
append_timestamp = 1
```

A commented-out example is added to `platformio.ini` alongside the existing
`bin_dir`, `map_dir`, and similar options.

## Artifacts written

(example: `tasmota-4M-2026-05-01T19-02-06`):

```
build_output/firmware/tasmota-4M-2026-05-01T19-02-06.bin
build_output/firmware/tasmota-4M-2026-05-01T19-02-06.bin.gz
build_output/firmware/tasmota-4M-2026-05-01T19-02-06.factory.bin  (ESP32 only)
build_output/firmware/tasmota-4M-2026-05-01T19-02-06.elf
build_output/map/tasmota-4M-2026-05-01T19-02-06.map.gz
```

The originals are kept; the timestamped copies are additions.

---

## Timestamp extraction

The timestamp is read from the ELF symbol table via `nm --defined-only --demangle`,
not from the host clock, not from the file system. This guarantees that the filename 
timestamp matches exactly the one the firmware reports at boot (via `GetBuildDateAndTime()`).

`GetBuildDateAndTime()` in `support_rtc.ino` already stored `__DATE__` in a named
PROGMEM variable (`mdate_P`). This PR adds the equivalent for `__TIME__`:

```cpp
static const char mtime_P[] PROGMEM = __TIME__;
```

Both symbols are then visible to `nm` without a `-g` debug build:

- ESP8266: `PROGMEM` places the symbol in `.irom0.text` (type `t` in nm output);
  visible without `-g`.
- ESP32: `PROGMEM` is a no-op; the `static const` ends up in `.rodata` (type `d`);
  visible without `-g`.

Without the named variable, `PSTR(__TIME__)` produces an anonymous string that is
not locatable by symbol name, particularly on ESP32 where PROGMEM is a no-op.

`nm` is available as the platform-specific `<toolchain-prefix>-nm` alongside the
compiler; `timestamp-firmware.py` derives the path from `$CC` (replacing `-gcc`
with `-nm`), so no additional tool installation is required.

---

## Firmware change

**No firmware behavior change.** The Python script runs post-build and does not
modify firmware content. 

The only firmware change is the added `mtime_P` named variable in 
`support_rtc.ino`; the `snprintf_P` call is updated to reference it
instead of an anonymous `PSTR(__TIME__)`. The string content is identical.

---

**Memory impact (ESP8266, tasmota release build):**

For this type of source change, the final impact is highly dependent on the
version of the optimizers in use. It might introduce an additional stored pointer
or not to the assembly. The tested GCC version introduced the stored pointer.

| | Flash | RAM | IRAM |
|---|---|---|---|
| `mtime_P` named variable (always active) | +4 B | 0 | 0 |
| Post-build script | 0 | 0 | 0 |

+4 B Flash from naming the `__TIME__` string; the string was already present in
the binary as an anonymous literal.

---

## Tested

Build-tested on all available Tasmota targets (ESP8266: all 37 targets pass;
ESP32 Xtensa and RISC-V: all standard targets pass). Symbol extraction verified
on ESP8266 and ESP32 Xtensa without `-g`. RISC-V nm symbol extraction not
explicitly verified, but compilation passed and produced timestamped files too.

### DIFF
--- a/pio-tools/tasmotapiolib.py
+++ b/pio-tools/tasmotapiolib.py
@@ -37,6 +37,8 @@ BIN_DIR = "bin_dir"
 DISABLE_MAP_GZ = "disable_map_gz"
 # if set, an alternative path to put generated .map files, relative to project directory
 MAP_DIR = "map_dir"
+# if set to 1, copies output .bin/.bin.gz with build timestamp appended to the filename
+APPEND_TIMESTAMP = "append_timestamp"
 
 # === END AVAILABLE OVERRIDES ===
 
--- a/pio-tools/timestamp-firmware.py
+++ b/pio-tools/timestamp-firmware.py
@@ -1,0 +1,128 @@
+Import("env")
+
+import re
+import shutil
+import struct
+import pathlib
+import subprocess
+import tasmotapiolib
+
+_MONTHS = {"Jan":1,"Feb":2,"Mar":3,"Apr":4,"May":5,"Jun":6,
+           "Jul":7,"Aug":8,"Sep":9,"Oct":10,"Nov":11,"Dec":12}
+
+
+def _nm_path(env):
+    return env.subst("$CC").replace("-gcc", "-nm")
+
+
+def _elf_sections(data):
+    """ELF32 LE: return (sh_addr, sh_offset, sh_size) for all SHT_PROGBITS sections."""
+    e_shoff,     = struct.unpack_from("<I", data, 32)
+    e_shentsize, = struct.unpack_from("<H", data, 46)
+    e_shnum,     = struct.unpack_from("<H", data, 48)
+    sections = []
+    for i in range(e_shnum):
+        b = e_shoff + i * e_shentsize
+        sh_type,   = struct.unpack_from("<I", data, b + 4)
+        sh_addr,   = struct.unpack_from("<I", data, b + 12)
+        sh_offset, = struct.unpack_from("<I", data, b + 16)
+        sh_size,   = struct.unpack_from("<I", data, b + 20)
+        if sh_type == 1 and sh_addr and sh_size:  # SHT_PROGBITS with content
+            sections.append((sh_addr, sh_offset, sh_size))
+    return sections
+
+
+def _elf_string(data, sections, vma):
+    """Read null-terminated ASCII string at VMA from ELF data."""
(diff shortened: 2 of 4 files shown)


Jason2866:
@arendst fine for me from the Platformio integration side.

joluxer:
Are there any open questions?

arendst:
Thx.

PR #24799, opened 2026-05-29
rules: pre-populate decompressed rule cache before network stack init

joluxer:
# Checklist:

# Description

## Problem

`k_rules[]` holds the decompressed text for each active rule set as heap-allocated
`String` objects (800-1100 bytes each, depending on rule content). These were
populated lazily on the first rule evaluation, which occurs after WiFi, MQTT, and
WebServer have already allocated heap. The cache therefore landed in the middle of
an already-fragmented heap and remained there permanently, acting as fragmentation
anchors that split the free heap into smaller non-contiguous regions.

Observed on ESP8266: heap fragmentation at boot reached 25-50%, leaving as little
as 4-8 KB of contiguous free space despite 20+ KB total free heap. Transient
allocations (JSON parsing, MQTT payloads, rule evaluation temporaries) competed
for those fragments; under sustained rule activity this led to allocation failures.

## Fix

Pre-populate `k_rules[]` during `FUNC_PRE_INIT` (`RulesInit`), before any network
stack allocation. The cache then lands at low heap addresses in a clean contiguous
region, leaving the upper heap as a single large unfragmented block available for
transient allocations.

```cpp
#ifdef USE_UNISHOX_COMPRESSION
  if (!Settings->flag4.compress_rules_cpu) {
    for (uint32_t i = 0; i < MAX_RULE_SETS; i++) {
      if (GetRuleLen(i) > 0) { GetRule(i); }
    }
  }
#endif
```

Active only when `USE_UNISHOX_COMPRESSION` is defined (standard Tasmota build) and
`compress_rules_cpu` is off (SetOption93 = 0, the cache-active code path).
When SetOption93 = 1, rules are decompressed on each evaluation with no persistent
cache; this code path is unchanged.

## Observed effect (ESP8266, two nearly-full rule sets)

- Heap fragmentation at boot: 25-50% -> ~1%
- Contiguous free space after boot: 4-8 KB -> ~24 KB

The cache objects themselves are the same size as before; only their position in
the heap changes. Total heap consumption at steady state is unchanged.

## Memory impact (ESP8266, tasmota release build)

| | Flash | RAM | IRAM |
|---|---|---|---|
| Code change | +32 B | 0 | 0 |

0 static RAM: no new variables. 0 IRAM: rules code resides in flash.
Runtime heap: unchanged in total; allocation order changes as described above.
The +32 B Flash is the cost of moving from lazy to eager cache population.

## Tested

Developed and target-tested on ESP8266 (ESP-12F). Build-test on all targets passed
for all major targets.

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_10_rules.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_10_rules.ino
@@ -933,6 +933,17 @@ void RulesInit(void)
     }
   }
   Rules.teleperiod = false;
+
+#ifdef USE_UNISHOX_COMPRESSION
+  // Pre-populate k_rules[] cache here (FUNC_PRE_INIT), before WiFi/MQTT/WebServer
+  // allocate heap, so the persistent cache lands at low heap addresses instead of
+  // fragmenting the middle of the heap when first rule evaluation occurs later.
+  if (!Settings->flag4.compress_rules_cpu) {
+    for (uint32_t i = 0; i < MAX_RULE_SETS; i++) {
+      if (GetRuleLen(i) > 0) { GetRule(i); }
+    }
+  }
+#endif
 }
 
 void RulesEvery50ms(void)


s-hadinger:
Great suggestion, thanks

joluxer:
Thanks for the compliments.

Debugging a heap allocator and heap maps is very interesting but rarely funny. 

Nevertheless I'm very desperate for running a stable firmware when controlling a couple of kilowatts of electrical power. So yeah, I'm motivated to do the things, that make the firmware better.

And as MCUs are "my thing", I'm going in for making those little creatures running better firmware. And at least there is some fun in debugging heap maps...

arendst:
Thx indeed.

I'll change your PR to reduce code size with:
```
void RulesInit(void) {
  // indicates scripter not enabled
  bitWrite(Settings->rule_once, 7, 0);
  // and indicates scripter do not use compress
  bitWrite(Settings->rule_once, 6, 0);

  TasmotaGlobal.rules_flag.data = 0;
  for (uint32_t i = 0; i < MAX_RULE_SETS; i++) {
    if (0 == GetRuleLen(i)) {
      bitWrite(Settings->rule_enabled, i, 0);
      bitWrite(Settings->rule_once, i, 0);
    }
#ifdef USE_UNISHOX_COMPRESSION
    else {
      // Pre-populate k_rules[] cache here (FUNC_PRE_INIT), before WiFi/MQTT/WebServer
      // allocate heap, so the persistent cache lands at low heap addresses instead of
      // fragmenting the middle of the heap when first rule evaluation occurs later.
      if (!Settings->flag4.compress_rules_cpu) {  // SetOption93 - (Compress) Keep uncompressed rules in memory to avoid CPU load of uncompressing at each tick (1)
        GetRule(i);
      }
    }
#endif
  }
  Rules.teleperiod = false;
}
```

arendst:
Thx again.

joluxer:
I appreciate the quality of the suggested change. 

I'd like to note for the record that the heap block map approach and the analysis behind it originated in this PR - reaching this small result - involved a significant amount of investigative work.

I'm also open to discussion on the placement: moving the code results in a smaller footprint - that is great, I'm happy to apply that change here. A short review comment would have been the natural way to handle this - I would have updated the PR accordingly.

PR #24803, opened 2026-05-30
dali: allow using TM_DALI_SEND_TWICE for 3 byte frames

l-jonas:
This helps configuring input devices as they need frames of three bytes length and need the command twice. Just calling DaliSend twice is not enough for this due to the delay.

## Description:

## Checklist:
    - was tested on base of v15.3.0


arendst:
Thx.

PR #24821, opened 2026-06-05
Add Matter virtual IR HVAC thermostat support

macjl:
## Summary
- Add a generic Matter thermostat plugin.
- Add a virtual HVAC plugin that maps Matter Thermostat state to Tasmota IRHVAC JSON commands.
- Add virtual HVAC option endpoints for IRHVAC boolean options: Econo, Quiet, Turbo, Light, Filter, Clean, Beep, and iFeel.
- Attach option endpoints to the HVAC endpoint PartsList instead of exposing them as top-level aggregator parts.
- Expose the new virtual device types in the Matter UI.

## Motivation
This enables ESP32 Matter builds with IR support to expose an IR-controlled HVAC as a Matter thermostat while still using Tasmota IRHVAC for protocol-specific command generation.

## Validation
- Checked the final diff with `git diff origin/development..HEAD --check`.
- Tested the feature on ESP32 devices with Matter pairing and HVAC control.
- Verified the IRHVAC command path with virtual HVAC state and option changes.


macjl:
Still interested in getting this reviewed/merged; keeping the PR active. Happy to adjust anything needed.

sushantkumarkhobian-lab (commented):
Could parse_configuration() validate the vendor field earlier and emit a warning if it's missing or invalid? That would make configuration errors easier to diagnose instead of silently ignoring IRHVAC commands until runtime.

macjl:
Addressed the v.HVAC vendor diagnostics review in `b4decb73b`.

`parse_configuration()` now normalizes the `vendor` field and emits a level-2 Matter warning when it is missing/empty, instead of leaving the endpoint to silently ignore IRHVAC sends later.

For invalid vendors, the actual supported-protocol validation lives in the IRHVAC C++ command path. To avoid a parse-time validation call that could send IR or mutate IRHVAC stored state, `send_irhvac()` now inspects the synchronous IRHVAC command response and logs a level-2 warning when IRHVAC rejects the command, including the configured vendor and returned error.

Validation run:
- `git diff --check`
- Matter Berry solidification via `../berry/berry -s -g solidify_all.be`

s-hadinger:
Good one, thanks. GPT-5.5 noted a regression, please fix and I will merge:

## Blocking finding: root Descriptor PartsList no longer gets change notifications

`lib/libesp32/berry_matter/src/embedded/Matter_zz_Device.be`

`signal_endpoints_changed()` was changed to notify every active endpoint returned by `get_active_endpoints(true)`:

```berry
var eps = self.get_active_endpoints(true)
for ep: eps
  self.attribute_updated(ep, 0x001D, 0x0003, false)
end
```

Because `true` excludes endpoint `0`, the root endpoint is no longer notified when its Descriptor `PartsList` changes. The root endpoint still exposes a `PartsList` containing active endpoints, so subscribed controllers may not learn that endpoints were added/removed.

Suggested fix: keep the broader notification, but explicitly include root endpoint `0`:

```berry
self.attribute_updated(0x0000, 0x001D, 0x0003, false)
var eps = self.get_active_endpoints(true)
for ep: eps
  self.attribute_updated(ep, 0x001D, 0x0003, false)
end
```

s-hadinger:
There is also a minor issue:

## Conformance concern: LTNE advertised while LocalTemperature remains exposed

Files:
- `lib/libesp32/berry_matter/src/embedded/Matter_Plugin_2_Thermostat.be`
- `lib/libesp32/berry_matter/src/embedded/Matter_Plugin_9_Virt_HVAC.be`

`Matter_Plugin_Virt_HVAC.thermostat_feature_map()` adds `LTNE` when no local temperature matcher is configured:

```berry
if self.local_temp_matcher == nil
  feature_map = feature_map | 0x40  # LTNE: LocalTemperatureNotExposed
end
```

However, the Thermostat cluster declaration always includes `LocalTemperature` (`0x0000`) in the AttributeList, and `read_attribute()` still returns the attribute as `NULL` when no local temperature is available.

If `LTNE` means `LocalTemperature` is not exposed, this is inconsistent and may confuse stricter Matter controllers. Either do not advertise `LTNE` and keep returning nullable `LocalTemperature`, or suppress/unsupported-read `LocalTemperature` and remove it from AttributeList when `LTNE` is active.

s-hadinger:
Ok, I reviewed "LTNE advertised while LocalTemperature remains exposed", full support would require dynamic list of attribute and significant more code. So we can accept this small twist to conformance.

I will merge with the fix

PR #24854, opened 2026-06-20
Fix Matter On/Off metadata advertised to strict controllers

daleckystepan:
**Disclaimer**
Fully AI crafted, I am not expert in this topic but it fixes my issue. I cannot recommended merging but only try to understand the root cause.

**Problem**
Google Home on iOS crashed when opening device details or trying to toggle a Tasmota ESP32-C3 Matter light/switch. The device never received the On/Off command, so the failure was happening client-side while Google Home was reading/parsing Matter metadata.

Two inconsistencies were present:

AttributeList was generated from a packed byte array of 16-bit attribute IDs, but the loop used the raw byte length as the element count. This could advertise invalid extra attributes.
In [Matter_Plugin_0.be](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), each attribute ID is read with attr_list_bytes.get(idx * 2, -2), so the loop count must be size(attr_list_bytes) / 2.

The On/Off cluster advertised the Lighting feature through FeatureMap, but did not expose the Lighting-feature attributes and returned an empty AcceptedCommandList.
FeatureMap for cluster 0x0006 is 0x01, meaning LT / Lighting. With that feature set, controllers can expect the related On/Off attributes and commands to exist. Before the fix, [Matter_Plugin_2_Light0.be](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) only advertised/read attribute 0x0000 (OnOff), while [Matter_Plugin_0.be](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) always returned an empty AcceptedCommandList.

**Fix**
The working fix makes the advertised metadata match the implementation:

Fix AttributeList iteration count:
Report accepted On/Off commands for cluster 0x0006:
The command IDs are encoded as U4, matching Matter command-id width.

Add the Lighting-feature On/Off attributes to the cluster:
Add simple handlers for those attributes:
Add simple command handling for the Lighting-feature commands:
**Why This Fixes Google Home**
Google Home appears to read FeatureMap, AttributeList, and AcceptedCommandList before rendering the details or control UI. Before the patch, those values contradicted each other:

FeatureMap said the On/Off cluster supports Lighting behavior.
AttributeList did not include the Lighting attributes.
AcceptedCommandList said no commands were accepted.
In one path, AttributeList could also contain bogus entries due to the byte-count bug.
After the fix, Google Home receives a self-consistent On/Off cluster description and no longer crashes.

**Tested**
On ESP32-C3 running Tasmota over Matter with Google Home app on iPhone:

Before fix: Google Home crashed when opening device details or toggling the device; Tasmota did not receive the command.
Force Static Endpoints did not help.
AttributeList count fix alone did not help.
Adding the On/Off metadata/command consistency fix resolved the crash and allowed control.

### DIFF
--- a/lib/libesp32/berry_matter/src/embedded/Matter_Plugin_0.be
+++ b/lib/libesp32/berry_matter/src/embedded/Matter_Plugin_0.be
@@ -441,7 +441,7 @@ matter_device.events.dump()
     elif attribute == 0xFFFB            # AttributeList
       var acli = TLV.Matter_TLV_array()
       var attr_list_bytes = self.get_attribute_list_bytes(cluster)
-      var attr_list_bytes_sz = (attr_list_bytes != nil) ? size(attr_list_bytes) : 0
+      var attr_list_bytes_sz = (attr_list_bytes != nil) ? size(attr_list_bytes) / 2 : 0
       var idx = 0
       while idx < attr_list_bytes_sz
         acli.add_TLV(nil, 0x05 #-TLV.U2-#, attr_list_bytes.get(idx * 2, -2))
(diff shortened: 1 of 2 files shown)


s-hadinger:
Thanks, this looks good to me

va1dir:
Nice work!

This fix resolves a crash in the Google Home app that occurs when attempting to control a Tasmota device while the Public Preview program is activated inside google’s app. (Tested with iOS only)

This fix does not impact when the beta program is turned off, it’s working fine out of beta program.

Also, I have validated that the crash still happens on Tasmota 15.4.0 with the beta program enabled using another device.

IMG_5594

PR #24862, opened 2026-06-27
feat: frontend design proposal

pierrejochem:
## Description:

Little design changes to make it a bit modern:

ui_4
ui_3
ui_2
ui_1

## Checklist:


arendst:
It's too dark for me. What's the code size increase in light of esp8266?

Ah a quick look at the compressed files size increase shows at least 4k....

s-hadinger:
Agreed. It may look more modern but definitely more confusing as well

pierrejochem:
Thank you for your feedback :-) I will try some more stuff and have a deeper look at the file sizes

sfromis:
Note that a very dark theme removing visual cues cannot be assumed to be generally seen as an improvement. Having the chance, I'd certainly opt out of that change.

Still, you *can* already use `WebColor` to create a similar theme, except for the header.

pierrejochem:
After a lot of investments I figured out that it is really hard to have a fancy ui without increasing the file size dramatically. I will skip it for now but not forever ;-)

PR #24879, opened 2026-07-05
backlog: Contain flag mutation side-effects

joluxer:
## Description

### Context

The previous PR fixed the backlog timer: `BacklogLoop` now unconditionally
owns `backlog_timer` after each drain step. That fix depends on one invariant:
`backlog_mutex` is set and cleared exclusively by `BacklogLoop`.

That invariant currently has no enforcement mechanism -- it holds by
convention. `backlog_mutex`, `backlog_timer`, and `backlog_nodelay` sit in
`TasmotaGlobal_t`, writable from any `.ino` file without restriction. A 
driver or rule handler that touches these fields will silently break the
previous fix without any compiler or linker feedback.

This PR removes that exposure by making the fields inaccessible outside their
translation unit. All mutation now goes through explicit entry points; accidental
modification from unrelated code paths is structurally impossible.

---

#### Protection boundary
`backlog_mutex`, `backlog_timer`, `backlog_nodelay`, `backlog_delay_guard`,
`backlog_no_mqtt_response`, and the backlog queue are no longer reachable from
`.ino` scope. All mutation goes through explicit entry points in a new
`Backlog::` API.

#### Implementation
- `tasmota/tasmota_support/support_backlog.cpp` -- state in anonymous
  namespace
- `tasmota/include/support_backlog.h` -- public API declarations
- `tasmota/tasmota_support/support_backlog.ino` -- one-line adapter:
  `BacklogLoop()` forwards to `Backlog::Loop()`

`SuppressMqttResponse()` added to `support.ino` as designated mutation path
for `no_mqtt_response` from backlog context (replaces direct field access).
`SettingsParam(index)` added as bounds-checked reference getter for
`Settings->param[]` - a small bonus for the whole system.

All callers in `support_command.ino` and `xdrv_10_rules.ino` mechanically
converted to the `Backlog::` API. No logic changed.

---

### Memory impact (measured, this PR only)

| Target     | Flash  | RAM | IRAM |
|------------|--------|-----|------|
| tasmota32  | +88 B  | +8  | 0    |
| tasmota-4M | +312 B | +8  | 0    |

The flash delta is the cost of the enforcement boundary. Without it, the
invariant from the previous PR remains advisory. The RAM delta reflects
alignment differences between anonymous-namespace variables in a `.cpp`
translation unit and packed fields in `TasmotaGlobal_t`.

---

**No logic changed.**

**Requires pr/backlog-timer-fix:** this PR enforces at module level the timer-ownership invariant established by the previous fix.

**Tested:**
Build-tested on ESP8266 (tasmota-4M) and ESP32 (tasmota32). Target-tested on ESP32: completed.

## Checklist:

### DIFF
--- a/tasmota/include/support_backlog.h
+++ b/tasmota/include/support_backlog.h
@@ -1,0 +1,46 @@
+/*
+  support_backlog.h - Public interface for the Backlog command queue
+
+  Copyright (C) 2021  Theo Arends
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _SUPPORT_BACKLOG_H_
+#define _SUPPORT_BACKLOG_H_
+
+namespace Backlog {
+  void Init();
+
+  bool     IsEmpty();
+  bool     IsNodelay();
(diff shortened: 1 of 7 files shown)


s-hadinger:
Thanks, good for me. @arendst ok to merge?

PR #24881, opened 2026-07-05
backlog: queue diagnostics and byte-limit safety net

joluxer:
## Description

### Problem

The Backlog queue had no visibility into its own state. Depth, drain
counters, timer status, staged flags, and configuration values were
inaccessible at runtime, making it difficult to diagnose timing issues,
verify that sequences executed as intended, or confirm that behavioral
fixes from earlier PRs had the expected effect.

The queue also had no upper bound on heap consumption. Under sustained
MQTT load the queue can grow until a malloc failure leaves a partial
command sequence in flight, or until the node resets. Both outcomes can 
be considered worse than refusing the sequence before it enters the queue.

---

### Changes

Two commits, each independently buildable.

#### Commit 1: queue diagnostics

`Backlog16 N` - set the chunk size for paged queue content output
  (Backlog21-29, default 20 entries per page; runtime, non-persistent).

`Backlog17 0/1` - enable a per-drain-step trace log at INFO level.
  Each drained command is logged with the queue depth at that moment
  and the entry's flavor byte. Build flag `BACKLOG_TRACE_SOURCE` extends
  the trace with the originating command source per entry.

`Backlog20` - JSON statistics snapshot: current and counter-verified
  depth, byte usage, timer state, mutex state, staged and current
  behavioral flags, per-operation counters (drained / enqueued /
  inserted / discarded / mutex-skipped), high-watermark values, and
  runtime config (SO34, SO166, chunk size). Bytes/Discarded/MaxBytes/
  BytesLimit report 0 in this commit; the byte-limit commit fills them.

`Backlog21-29` - queue content as JSON, paged by chunk size. Each entry
  appears on its own log line and in the assembled MQTT/Web-Console
  response.

The entry header is generalized to support the optional
  `BACKLOG_TRACE_SOURCE` source byte without scattering conditionals
  across access sites.

#### Commit 2: queue byte limit

`Backlog18 N` - set the byte limit at runtime (0 resets to the
  compile-time default; clamped to a minimum confirmed stable on ESP8266
  under sustained load). Platform-dependent compile-time defaults in
  `tasmota.h`; override via `user_config_override.h`.
  Capacity at default settings: ESP8266 4096 B ~= 97 short commands;
  ESP32 32768 B ~= 780.

A whole-sequence-or-nothing gate in `CmndBacklog` pre-checks the full
  estimated cost before tokenising the string. A sequence that would
  exceed the limit is rejected intact; the node stays reachable and logs
  the rejection. Per-command fallback checks guard the IF/ENDIF insertion
  path and any caller that bypasses the sequence gate.

Accounting covers command strings, linked-list node overhead, and
  heap-allocator bookkeeping per allocation - the cost a string-length
  check alone would miss.

`Backlog20` now reports real byte counters and the active limit.

---

### Memory impact (measured, per commit)

Ref: pr/backlog-behavior tip (85ba5553)

| Commit             | Target     | Flash  | RAM | IRAM |
|--------------------|------------|--------|-----|------|
| diagnostics        | tasmota32  | +1484  | +32 | 0    |
| diagnostics        | tasmota-4M | +1380  | +36 | 0    |
| byte limit         | tasmota32  | +600   | +8  | 0    |
| byte limit         | tasmota-4M | +708   | +20 | 0    |
| PR9 total          | tasmota32  | +2084  | +40 | 0    |
| PR9 total          | tasmota-4M | +2088  | +56 | 0    |

**Requires pr/backlog-behavior:** per-entry flavor byte and queue-access API from that PR are the structural prerequisite for the diagnostics and byte-limit accounting here.

---

### Tested

Build-tested on ESP8266 (tasmota-4M) and ESP32 (tasmota32).
Target test on ESP8266 (ESP-12F): Backlog20 snapshot verified; Backlog17
drain trace verified; byte-limit rejection under sustained MQTT load
confirmed (30+ min without crash; limit adjustment under load drains the
queue cleanly without side effects).

## Checklist

### DIFF
--- a/lib/default/TasmotaLList/src/LList.h
+++ b/lib/default/TasmotaLList/src/LList.h
@@ -58,6 +58,13 @@ public:
   LList() : _head(nullptr) {}
   ~LList() { reset(); }
 
+  // Structural size of one list node (sizeof LList_elt<T>, excluding heap-allocator bookkeeping).
+  // Callers that account for queue memory must not compute this from LList_elt<T> directly:
+  // the node layout is an implementation detail of LList and may change independently of T.
+  // Exposing it here keeps the per-element cost contract stable while enforcing the principle
+  // that only LList owns — and therefore discloses — the size of its own internal nodes.
+  static constexpr size_t element_size = sizeof(LList_elt<T>);
+
   // remove elements
   T * removeHead(void);      // remove first element
   void reset(void);           // remove all elements
--- a/tasmota/include/support_backlog.h
+++ b/tasmota/include/support_backlog.h
@@ -1,0 +1,46 @@
+/*
+  support_backlog.h - Public interface for the Backlog command queue
+
+  Copyright (C) 2021  Theo Arends
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _SUPPORT_BACKLOG_H_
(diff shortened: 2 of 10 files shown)


arendst:
I strongly feel this feature needs to be enabled using a comple time ``#define``.

The code size impact is way to large compared to it's functionality and use-case. Most of the time it won't be used considering the fact Tasmota is used some ten years now and there was never felt a use to monitor/trace backlog use simply because normally only small backlogs are being used.

PR #24882, opened 2026-07-05
backlog feature: fast-lane queue for nodelay commands (USE_BACKLOG_FASTLANE)

joluxer:
## Description

### Situation

Commands queued via Backlog0/2 share the timer-gated queue with timed
sequences. The SO166 post-external-command window (100 ms) stalls the
whole queue, including entries that explicitly requested no delay. Under
sustained MQTT load this made the nodelay path unreliable.

---

### Changes

Opt-in: `USE_BACKLOG_FASTLANE` at build time, SetOption167 at runtime.

Backlog0/2 commands are routed to a separate fast-lane queue at enqueue
time and drained per BacklogLoop() call with no timer gate, within a
burst budget. At least one entry is always drained. The budget defaults
to the current Sleep setting; Backlog19 N overrides it at runtime
(0=Sleep, 50-250 ms, out-of-range clamped).

The sleep-blocking check in the main loop is corrected to test
pre-drain state: whether the fast queue has pending work or the
timed-lane timer has lapsed.

`Backlog19 N` - burst budget (runtime, non-persistent).

`Backlog20` extended: FastLane, FastDepth, FastDrained, FastBudget,
FastBudgetMs.

`Backlog21-29` extended: Lane field ("F"/"T") per entry.

---

### Memory impact (measured)

Ref: pr/backlog-diagnostics tip (a97fc92c)

| Target     | Flash |  RAM | IRAM |
|------------|-------|------|------|
| tasmota32  | +1100 |  +24 |    0 |
| tasmota-4M |  +860 |   +8 |    0 |

Cumulative PR6-PR10 vs development (5e8019dc):

| Target     | Flash |  RAM | IRAM |
|------------|-------|------|------|
| tasmota32  | +3500 |  +64 |    0 |
| tasmota-4M | +3744 |  +72 |    0 |

---

### Tested

Build-tested on ESP8266 (tasmota-4M) and ESP32 (tasmota32).
Target test on ESP8266 (ESP-12F): fast-lane drain confirmed under
sustained MQTT load; SetOption167 toggle at runtime verified; Backlog19
budget adjustment under load confirmed. A device in continuous
production use showed significant behavioral stabilization.

**Requires pr/backlog-diagnostics:** the fast-lane queue extends Backlog20/21-29 diagnostics and relies on the byte-limit infrastructure for queue-depth tracking.

## Checklist:

### DIFF
--- a/lib/default/TasmotaLList/src/LList.h
+++ b/lib/default/TasmotaLList/src/LList.h
@@ -58,6 +58,13 @@ public:
   LList() : _head(nullptr) {}
   ~LList() { reset(); }
 
+  // Structural size of one list node (sizeof LList_elt<T>, excluding heap-allocator bookkeeping).
+  // Callers that account for queue memory must not compute this from LList_elt<T> directly:
+  // the node layout is an implementation detail of LList and may change independently of T.
+  // Exposing it here keeps the per-element cost contract stable while enforcing the principle
+  // that only LList owns — and therefore discloses — the size of its own internal nodes.
+  static constexpr size_t element_size = sizeof(LList_elt<T>);
+
   // remove elements
   T * removeHead(void);      // remove first element
   void reset(void);           // remove all elements
--- a/tasmota/include/support_backlog.h
+++ b/tasmota/include/support_backlog.h
@@ -1,0 +1,46 @@
+/*
+  support_backlog.h - Public interface for the Backlog command queue
+
+  Copyright (C) 2021  Theo Arends
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _SUPPORT_BACKLOG_H_
+#define _SUPPORT_BACKLOG_H_
+
+namespace Backlog {
+  void Init();
+
+  bool     IsEmpty();
+  bool     IsNodelay();
+  uint32_t GetRemainingDelay_ms();
+
+  void SetNodelay(bool val);
+  void SetNoMqttResponse(bool val);
+
+  void OnCommandExecuted();
+  void ScheduleNow();
+  void ScheduleDelay(uint32_t ms);
+
+  void EnqueueCmd(const char* cmd);
+  void InsertCmd(const char* cmd, uint32_t position);
+  void Clear();
+
+  void Loop();
+}
+
+void BacklogLoop(void);
+
+#endif  // _SUPPORT_BACKLOG_H_
--- a/tasmota/include/support_backlog.h
+++ b/tasmota/include/support_backlog.h
@@ -21,6 +21,19 @@
 #define _SUPPORT_BACKLOG_H_
 
 namespace Backlog {
+
+  enum class NoDelay : uint8_t {
+    OFF,
+    ON,
+    NoChange
+  };
+
+  enum class NoMqttResponse : uint8_t {
+    OFF,
+    ON,
+    NoChange
+  };
+
   void Init();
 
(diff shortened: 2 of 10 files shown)


s-hadinger:
Hi, I understand the implementation, but I fear it looks a little bit as an overkill. Backlog is supposed to serve very minimalistic use cases, whereas full scripting with Berry (ESP32 and above) offer much more flexibility. Do you have concrete examples where this should be used?

I also fear that additional backlog options makes it hard to use and understand to most users that have basic needs.

@arendst any opinion?

s-hadinger:
Here are the findings found by gpt-5.5: can you double check please? (if we go for merging)

Findings:

1. **Rules IF-block `NoDelay` becomes sticky**
   - `xdrv_10_rules.ino:2014`
   - In `ExecuteCommandBlock()`, `Backlog::SetNodelay(true)` is set when a `NoDelay` token is seen, but it is never reset after inserting the next real command.
   - In `CmndBacklog()`, this is handled with `nodelay_oneshot`; the rules path lacks the equivalent.
   - Impact: inside `IF ... ENDIF` command blocks, `NoDelay` likely affects all subsequent inserted commands, not just the next one. It may also leave staged state true until another caller resets it.

2. **`NoDelay` timing semantics changed**
   - `support_command.ino:612`, `support_backlog.cpp:412-413`
   - Old behavior processed `NoDelay` as a queue entry at drain time. The delay before the `NoDelay` marker still occurred; the following command then ran in the same drain step and reset the next timer.
   - New behavior bakes `NoDelay` into the next command and uses lookahead to skip the delay before that command.
   - Impact: sequences like `Backlog cmd1; NoDelay; cmd2; cmd3` now run `cmd2` earlier than before. This can break scripts depending on inter-command settling time.

3. **Fast-lane mode appears to still modify the timed-lane timer**
   - `support_backlog.cpp:210`, `support_backlog.cpp:396`
   - The fast-lane comment says `_timer` is owned only by the timed lane, but each fast-lane `ExecuteCommand()` still enters `CommandHandler()`, which calls `Backlog::OnCommandExecuted()` and sets `_timer`.
   - Impact: when `USE_BACKLOG_FASTLANE` is enabled, fast-lane commands can delay timed-lane queue execution by `BACKLOG_EXT_DELAY`, contradicting the design comments and likely causing ordering/timing surprises.

4. **SetOption167 is defined but not properly exposed/named in settings**
   - `tasmota.h:225` defines `SO_BACKLOG_FASTLANE 167`.
   - `tasmota_types.h:204` still has `spare21`.
   - `tasmota_types.h:218` sets `MAX_SETOPTION_USED = 166`.
   - Impact: `SetOption167` can be used through generic decoding, but it is hidden from normal SetOption reporting and not named in the bitfield. If this option is intentional, `spare21` should be renamed and `MAX_SETOPTION_USED` should include 167; otherwise `SO_BACKLOG_FASTLANE` should not be advertised.

5. **`Backlog18` accepts negative payloads as huge unsigned limits**
   - `support_command.ino:593-596`, `support_backlog.cpp:164-166`
   - `Backlog18 -1` casts `-1` to `uint32_t`, setting the queue byte limit to `0xFFFFFFFF`.
   - Impact: the new heap safety net can be accidentally or deliberately disabled via a negative value, enabling memory pressure/DoS from large backlogs. This should reject negative values or clamp to a sane upper bound.

6. **Queue dump emits unescaped command strings into JSON**
   - `support_backlog.cpp:475-486`
   - `_DumpEntry()` inserts raw queued commands into JSON with `%s`.
   - Impact: commands containing quotes, backslashes, or control characters produce invalid JSON and can inject fake fields into diagnostic output. Use `EscapeJSONString(cmd).c_str()` or equivalent.

7. **Behavioral change in external-command delay**
   - `support_command.ino:470`, `tasmota.h:221-224`
   - Old `CommandHandler()` reset backlog delay to `SetOption34` for every command.
   - New code uses `BACKLOG_EXT_DELAY` default 100 ms, with `SetOption166` to disable.
   - Impact: existing configurations with non-default `SetOption34` may see backlog drain timing change when external commands arrive during a backlog. This may be intended, but it is a compatibility change and should be explicitly accepted/tested.

arendst:
As I noticed in PR #24881 The code size increase for edge-case functionality is just way to big.

If implemented it should be enabled with a compile time ``#define`` to be used during user debugging.

joluxer:
This is an opt-in compile-time feature. At least this is my understanding of the `#ifdef USE_BACKLOG_FASTLANE` sections, which get active only, if someone defines this macro in `user_config_override.h`. So it is by default off - from my point of view.

PR #24726, opened 2026-05-09
Add RX Enable GPIO support for Modbus

jziolkowski:
## Description:

**Related issue (if applicable):** fixes #<Tasmota issue number goes here>

## Checklist:
  - [X] The pull request is done against the latest development branch
  - [X] Only relevant files were touched
  - [X] Only one feature/fix was added per PR and the code change compiles without warnings
  - [ ] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [X] The code change is tested and works with Tasmota core ESP32 V.3.3.8 from Platform 2026.05.50
  - [X] I accept the [CLA](https://github.com/arendst/Tasmota/blob/development/CONTRIBUTING.md#contributor-license-agreement-cla).

_NOTE: The code change must pass CI tests. **Your PR cannot be merged unless tests pass**_

PR #24886, opened 2026-07-06
Fix ESP8266 Exception crash-loop on PZEM/Modbus energy monitors (regression from 15.5.0)

### LINKED ISSUE #24883: PZEM-004T / PZEM-016 (single phase) crash-loops on 15.5.0 — Exception 9, fine on 15.4.0
Been chasing this one for a while, so let me write it up properly.

I run two ESP8266 energy monitors that have worked fine for ages — one on a PZEM-004T v3, one on a PZEM-016, both single phase, plain PZEM Modbus at 9600. Bumped both from 15.4 to 15.5.0 and they instantly started boot-looping. Tasmota Device Manager just shows them as "EXCEPTION".

The pattern: device boots, joins wifi, looks totally normal for about 10 seconds, then throws an exception and reboots. Forever. BootCount keeps climbing.

Stack is identical on both boards (same 15.5.0 build):

```
RestartReason: Exception
Exception 9, EPC 40105f68, EXCVADDR 00000003, DEPC 00000000
```

Exception 9 is an alignment fault and 0x3 is an odd address, so something's doing a 32-bit access off an unaligned pointer. The ~10s timing matches the end of the PZEM stabilize window, so it's dying on the first real Modbus reply, not at init.

What I ruled out:
- Not telemetry/JSON — `TelePeriod 0` makes no difference, still crashes.
- Not the wiring/hardware — same boards, nothing physical changed.
- Pull the PZEM Tx/Rx GPIOs and it's rock solid (no driver = no crash), so it's clearly in the PZEM read path.
- Downgrade either board to 15.4.0 and it's stable and reads perfectly. So this is new in 15.5.0.

I poked at the changelog between 15.4.0 and 15.5.0. The two commits touching the energy/Modbus side (#24726 modbus rx-enable, #24790 multi-phase PowerTotal) both look guarded off for a single-phase device, so my guess is the actual culprit is something shared lower down (the serial/modbus lib?) rather than the PZEM driver itself, but I couldn't pin the line without the map for that build.

Happy to flash a test build or grab a proper decode if someone points me at what to try. Anyone else hitting PZEM crashes on 15.5.0?

```
15.5.0(release-tasmota)
Core: 2.7.8 / SDK: 2.2.2-dev(38a443e)
Build: 2026-06-22
Hardware: ESP8266EX, 4M flash (DOUT)
Module: Generic (18), GPIO1 = PZEM0XX Tx, GPIO3 = PZEM016 Rx
Drivers: 1,2,3,!4,!5,!6,7,!8,9,10,12,...,26,29,62
StatusSTK: {"Exception":9,"Reason":"Exception","EPC":["40105f68","00000000","00000000"],"EXCVADDR":"00000003","DEPC":"00000000"}
```

gionag:
## Description

Since v15.5.0, ESP8266 devices with a PZEM (PZEM-004T v3 / PZEM-016) or any Modbus energy monitor crash-loop a few seconds after boot with `Exception (9)` (load/store alignment) once the driver starts transmitting. It reboots roughly every ~10s (right at the end of the PZEM stabilize window), BootCount keeps climbing, and it never reads. 15.4.0 is fine.

Bisected to #24726 ("Add RX Enable GPIO support for Modbus", commit 0c85dbc) in `TasmotaModbus.cpp`. That change added an RX-enable block to `Send()`. The `digitalWrite`s are guarded by `if (mb_rx_enable_pin > -1)`, but the `delay(10)` was left outside the guard:

```c
  write(frame, framepointer);
#ifdef TASMOTA_MODBUS_RX_ENABLE
  delay(10); // runs on every Modbus Send, for every device
  if (mb_rx_enable_pin > -1) {
    digitalWrite(mb_rx_enable_pin, LOW);
  }
#endif
```

So every Modbus transmit now hits an unconditional `delay(10)`, even though almost nobody has an RX-enable pin configured. On ESP8266 that `delay()` yields into the SDK mid-send and faults — the crash EPC (`0x40105f68`) is in IRAM. It's ESP8266-only (ESP32 is unaffected) and reproduces with no PZEM even attached, just by assigning a `PZEM0XX Tx` pin (confirmed by a second user in #24883).

## Fix

Move the `delay(10)` inside the existing `mb_rx_enable_pin > -1` guard. For any setup without an RX-enable pin — i.e. every current PZEM/Modbus user — this makes `Send()` byte-for-byte identical to 15.4.0. RX-enable timing for people who do set that pin is unchanged.

## Verified

Built the standard `tasmota` env (ESP8266 core 2.7.8) with this change and OTA-flashed it to two boards that crash-loop on stock 15.5.0:
- PZEM-004T v3 (hardware serial)
- PZEM-016 (software serial)

Both went from Exception crash-loops (uptime never past ~10s) to fully stable — uptime climbing for minutes, RestartReason "Software/System restart", reading voltage/current/power normally (~232 V / ~900 W / addr 1). No change to RX-enable behaviour for setups that use it.

**Related issue:** fixes #24883

## Checklist:
  - [ ] The code change is tested and works with Tasmota core ESP32 V.3.3.8 from Platform 2026.05.50 (bug is ESP8266-only; the change is a trivial guard that compiles for ESP32 but I did not flash ESP32 hardware)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - BLE EQ3 float output in mqtt messages regression from v15.4.0.2 (#24869)
 - Default button/switch actions on builds without rules regression from v15.4.0.2 (#24871)
+- PZEM/Modbus energy monitor Exception crash-loop on ESP8266 regression from v15.5.0 (#24883)
 
 ### Removed
 
--- a/lib/lib_basic/TasmotaModbus-3.6.0/src/TasmotaModbus.cpp
+++ b/lib/lib_basic/TasmotaModbus-3.6.0/src/TasmotaModbus.cpp
@@ -213,8 +213,8 @@ uint8_t TasmotaModbus::Send(uint8_t device_address, uint8_t function_code, uint1
   write(frame, framepointer);
 
 #ifdef TASMOTA_MODBUS_RX_ENABLE
-  delay(10); // add a delay to allow write buffers to empty
   if (mb_rx_enable_pin > -1) {
+    delay(10); // allow write buffer to drain before re-enabling Rx
     digitalWrite(mb_rx_enable_pin, LOW);
   }
 #endif  // TASMOTA_MODBUS_RX_ENABLE


s-hadinger:
Thanks, good for me. @arendst ok to merge?

PR #24931, opened 2026-08-01
Adapt `BLEDebug` to common command syntax

SteWers:
## Description:
Enhance `BLEDebug` command and adapt to common command syntax:
- `BLEDebug0` - show debug logging status
- `BLEDebug1 <loglevel>` - Set log level for errors - default 0 -> off
- `BLEDebug2 <loglevel>` - Set log level for infos - default 0 -> off
- `BLEDebug3 <loglevel>` - Set log level for debug - default 0 -> off
- `BLEDebug4 <loglevel>` - Set log level for more debug - default 0 -> off

(for all details compile with `#define BLE_ESP32_DEBUG` and `#define EQ3_DEBUG`)

Code cleanup without functional changes.
## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
@@ -92,10 +92,13 @@
         BLEName <mac|alias> - read the name
         BLEName <mac|alias> <name> - write the name - few devices support this
       BLEDebug
-        enable more debug logging
-        BLEDebug - show debug logging status
-        BLEDebug 0 - turn off
-        BLEDebug 1 - turn on
+        Set log levels more debug logging
+        BLEDebug0 - show debug logging status
+        BLEDebug1 <loglevel> - Set log level for errors     - default 0 -> off
+        BLEDebug2 <loglevel> - Set log level for infos      - default 0 -> off
+        BLEDebug3 <loglevel> - Set log level for debug      - default 0 -> off
+        BLEDebug4 <loglevel> - Set log level for more debug - default 0 -> off
+        * for all details compile with #define BLE_ESP32_DEBUG and #define EQ3_DEBUG
       BLEDevices
         display or clear the devices list
         BLEDevices0 - clear list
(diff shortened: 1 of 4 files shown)


tasmota/tasmota_xdrv_driver/xdrv_79_esp32_ble.ino
```diff
@@ -1162,7 +1163,7 @@ void ReverseMAC(uint8_t _mac[]){
 #ifdef BLE_ESP32_FILTER_BY_NAME
 bool isDeviceInFilter(const String& deviceName) {
 #ifdef BLE_ESP32_DEBUG
-    if (BLEDebugMode > 0) AddLog(LOG_LEVEL_DEBUG,PSTR("BLE: Device chcked in filter %s"), deviceName);
+    if (BLEDebugMode) AddLog(LOG_LEVEL_DEBUG,PSTR("BLE: Device chcked in filter %s"), deviceName);
```
arendst: Instead of the using ``if (BLEDebugMode) Addlog(LOG_LEVEL_DEBUG,...`` you might want to consider the solution used in influxdb. There I replace the define LOG_LEVEL_DEBUG with a variable controlled by, in your case, command `BLEDebug`. 

So with command `BLEDebug 3` which sets variable `BLEDebugMode` to 3 and `AddLog(BLEDebugMode, PSTR("BLE: Device chcked in filter %s"), deviceName);` the log will beshown if the user sets it's log level to 3. Setting `BLEDebug` higher than max log level will disable showing of log levels.

This way all `if (BLEDebugMode) ` can be removed.

SteWers: Sounds interesting. I'll have a look at that. Thanks for the tip.

SteWers: In general, it is a good idea, but when `BLEDebug` is on, there are additional log messages produced with different log levels (DEBUG and INFO). So, there is a second differentiation possible, which messages should be shown for debugging.
At this point I would like to do not change the logging logic for the BLE driver. There should be a bigger rework in future.

SteWers: I implemented the possibility to set the log levels for debugging separately.

PR #24949, opened 2026-08-10
Fix/dsi dma2d psram buffers

LesNewell:
## Description:

Higher resolution displays on DSI can end up consuming considerable amounts of RAM. By default these buffers live in IRAM. To reduce memory footprint you need to specify a relatively small flush lines buffer size (e.g :B,60,01 in display.ini). 
After a lot of testing I found that moving LVGL buffers to  PSRAM actually improves performance. As it turns out the CPU overhead of rendering a full screen refresh such as a page switch in small blocks is significant.  Although PSRAM is slow this is far outweighed by the reduced CPU cycles to render a page in one hit. Setting flush lines to the height of the screen and moving all buffers to PSRAM results in significantly reduced page draw times and saves a lot of IRAM. For my 1024x600 display this saving was ~120K.

## Checklist:
  - [x ] The code change is tested and works with Tasmota core ESP8266 V.2.7.8
  - [x ] The code change is tested and works with Tasmota core ESP32 V.3.3.8 from Platform 2026.05.50

### DIFF
--- a/lib/lib_display/UDisplay/src/uDisplay_DSI_panel.cpp
+++ b/lib/lib_display/UDisplay/src/uDisplay_DSI_panel.cpp
@@ -10,6 +10,7 @@
 #include "esp_ldo_regulator.h"
 #include "driver/gpio.h"
 #include <rom/cache.h>
+#include "esp_rom_sys.h"
 
 extern void AddLog(uint32_t loglevel, const char* formatP, ...);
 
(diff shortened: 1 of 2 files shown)


Jason2866:
@arendst please hold on. We (@Staars and i) see negative side effects for other mcus

LesNewell:
My apologies. I have an S3 based board and I should have tested that. What issues are you seeing?

Staars:
1. The "solution" to simply wait in a loop with a hard lock of the CPU to wait for completion of the DMA operation is a bit hacky. Maybe we will do it like that, but maybe working with a completion handler is the way to go here, although the latter will add more code complexity. I will test both.

2. You changed the PSRAM preference for all displays and SOC's. This is definitely wrong. It is very likely for the P4 with DSI, that PSRAM is the preferred choice, but even here we could have special use cases.
We do not have a mechanism for this yet, but I would prefer to add explicit user choice for preferred RAM type in the display.ini, i.e. by using a yet unreserved bit of `lvgl_param`.  This could be done optional and backwards compatible, like  bit 5 of the packed `:B` parameter now means “prefer PSRAM for the LVGL draw buffer.

LesNewell:
To give some context on the hard lock the worst case I saw in testing was ~900us on the P4 for a full screen refresh, 600 flush lines. 
Making PSRAM optional sounds good to me.

LesNewell:
Looking through the source there is a flag #define USE_LVGL_PSRAM in my_user_config.h that is not referenced anywhere. It looks like someone else played with this before, or at least intended to.
I reverted the PSRAM change and ran some more tests. If I set my flush lines to a small number the buffers get allocated in internal ram and my free ram drops. However if  I set my flush lines to 600(display height) the buffers are automatically allocated in PSRAM. 
A simple option is to scrap my patch and add a note to the display driver docs that for best performance using flush lines=display height is probably best if the hardware can support display buffers in PSRAM. Otherwise it is a good idea to check free ram and adjust flush lines to suit. In my testing 50K seemed a reasonably safe minimum.

As for the blocking issue I'm not sure about the best way to go. As another data point, on my S3 with a 800x480 display a full screen refresh with flush lines=480 took about 36ms. Most updates were in the 1-2ms range.

s-hadinger:
Bump, sorry I couldn't find time to explore this yet

PR #24951, opened 2026-08-11
Matter: add remote devices via MQTT

Staars:
## Description:

While I was migrating some of my existing ESP8266 devices from "Homebridge" to Tasmotas own Matter bridge I had to learn, that performance under my local network conditions is horrible, which may be related to network congestion in a dense environment.
So I tried the MQTT way (remember the M in Tasmota) as by design it is much more efficient than HTTP polling for the weak end devices. In my local scenario this was a complete game changer.

This PR does add optional bridging of remote devices to Matter, when both (bridge and remote) are connected to the same MQTT broker. No change for HTTP, this can be used as before.
I tried to preserve the HTTP code path as good as possible, but code parts could be merged in the future.

Setup:
1. Configure all involved ESP devices in the "usual Tasmota way" to connect to the same MQTT broker.

3. In the existing Matter web config page, there is a new entry with discovered MQTT devices:
matter_mqtt_remote

4. Click on "add" for the desired device and you should see the familiar dialogue for setting it up, that was already existent for HTTP remote devices.

6. Then add the bridge to the Matter network and configure devices as it was done for HTTP remotes.

## Checklist:

### DIFF
--- a/lib/libesp32/berry_matter/src/be_matter_module.c
+++ b/lib/libesp32/berry_matter/src/be_matter_module.c
@@ -188,6 +188,7 @@ extern const bclass be_class_Matter_TLV;   // need to declare it upfront because
 #include "solidify/solidified_Matter_TCP_async.h"
 #include "solidify/solidified_Matter_HTTP_async.h"
 #include "solidify/solidified_Matter_HTTP_remote.h"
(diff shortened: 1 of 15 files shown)


sfromis:
Sounds good. I was already wondering a bit why MQTT was not the primary binding, with http polling as a "fallback" option.

PR #24959, opened 2026-08-14
Add support for TFA Marbella pool thermometer

pschroee:
## Description:

Adds support for the **TFA Dostmann Marbella (30.3066.01)** floating pool thermometer, received on 868 MHz with a CC1101 module. The driver reports water temperature, sensor id, rolling counter and battery state as regular sensor telemetry.

The packet layout, the checksum algorithm and the temperature formula come from the [rtl_433](https://github.com/merbanan/rtl_433) project (`src/devices/tfa_marbella.c` and the reflected reverse Galois digest in `src/bit_util.c`), Copyright (C) 2021 Benjamin Larsson, GPL-2.0-or-later, which is compatible with the GPL-3.0 of Tasmota.

### Wiring

The shared SPI bus plus two pins that belong to the module:

| Signal | GPIO role | example |
|---|---|---|
| SCK / MISO / MOSI | `SPI CLK` / `SPI MISO` / `SPI MOSI` — shared bus | GPIO14 / 12 / 13 |
| CSN | **`CC1101 CS`** | GPIO15 |
| GDO0 | **`CC1101 GDO0`** | GPIO4 |

`CC1101 CS` is added as a role of its own, the way every other SPI device has one (`LORA_CS`, `RC522_CS`, `MCP2515_CS` …). The three bus lines stay generic because they are shared — a display and this receiver hang on the same wires, and only chip select tells them apart. The name string for it already existed in the language files, unused.

### Commits

1. **Fix GPIO name list falling out of sync without `USE_MODBUS_RELAY`** — the enum guards four entries with `#ifdef`, the name list does not, so in any default build the two run four entries apart and the next role added inherits a foreign name. Independent of this driver; happy to split it off if preferred.
2. **The driver itself**, including the new `CC1101 CS` role.
3. **`SetOption166`** — publish readings as they arrive instead of on `TelePeriod` only. Off by default.
4. **Build configuration** — `USE_TFA_MARBELLA`, off by default, +12k6 on ESP8266 and +5k7 on ESP32.

### Settings that are not obvious

Established against hardware, and documented in the source:

- The **bit rate is 9.5238 kBit/s**, not 9.6. rtl_433 documents a symbol width of 105 µs; the 0.8 % difference accumulates to about 0.7 bit over an 88 bit frame, so the first bytes arrive intact and the last ones come out shifted.
- The **sync word is 0xAAD2**, although the frame begins with `AA 2D D4`. The radio matches in hardware and strips what it matched, so the two values describe the same transition from different sides.
- **The preamble quality threshold has to be switched off.** `begin()` derives it from the preamble length - `setPreambleLength(len, len - 4)` - and anything from 32 bits upwards clamps it to 7, the hardest setting the chip has. The receiver then only accepts a sync word after counting enough clean preamble transitions immediately before it, and a cheap sensor does not deliver that reliably. What makes this expensive to find is how it fails: the sync word is dropped in silence. No interrupt, no error, no packet - the driver sees an idle radio while the sensor transmits happily. Measured here with the threshold at 7: not a single interrupt edge in 90 seconds, while the sensor was transmitting. The fix itself is not verified on hardware yet - see the note at the end.
- **Interrupt on the packet-received edge only.** `setPacketReceivedAction()` picks the rising edge, which with GDO0 mapped inverted is the end of the frame - the point where the FIFO holds it complete. Interrupting on both edges also fires on the sync word, some 9 ms earlier, and reading then races the chip filling the FIFO: the byte last read comes out twice with everything after it shifted along ([errata SWRZ020](https://www.ti.com/lit/pdf/swrz020)). Such a frame fails the checksum and looks like interference, while the LQI reports a clean link.
- **The re-arm timer has to be measured against the last valid frame.** The CC1101 can sit in RX without ever toggling GDO0 again ([errata SWRZ020](https://www.ti.com/lit/pdf/swrz020)); no interrupt arrives and the reading simply stops, which is the hardest failure to notice. The driver restarts reception when it has been quiet for too long - but a timer fed by interrupts is never reached, because `startReceive()` puts GDO0 back to its idle level, which is an edge in itself. Measured on hardware: the receiver went silent for seven minutes while that timer, restarted by every one of those edges, never ran out. A safety net that never fires looks exactly like one that is not needed.
- **`disableAddressFiltering()` is required.** RadioLib leaves address checking on, and the chip then discards every frame whose first byte does not match the node address — always `0xAA` for this sensor. Without it reception runs and nothing ever arrives.

Built on RadioLib rather than `lib/lib_rf/cc1101`: that library declares `"platforms": "espressif8266"` because on ESP32 it collides with RadioLib's own `CC1101` class. Using RadioLib keeps one driver for both platforms and leaves the older library untouched.

`LQI` is the link quality of the frame, 0 to 127, lower being better - the opposite of the LQI Zigbee transceivers report, hence the note in the documentation. The RSSI the chip appends to each packet is deliberately absent: RadioLib keeps it in a private field, and `getRSSI()` reads the register instead, which holds the noise floor once the frame is over. Measured -84 dBm for a sensor lying next to the antenna - a number that looks like a reading and is none.

The sensor reports its battery as a single bit, so the payload carries `BatteryLow` rather than `Battery` - the latter is a percentage everywhere else in Tasmota, and `"Battery":1` would read as one percent while meaning the opposite.

### Status

Ready for review. The driver has been running on the receiver since 14 August 2026: twenty minutes of consecutive reception with a gapless rolling counter, LQI between 0 and 19, and reception returning immediately after a restart. The MQTT chain up to ioBroker is live. The one failure mode left in the field was the sensor itself - it stopped transmitting at night and came back after a battery power-cycle - which no receiver firmware can prevent.

### Verified against the sensor

Frames decode with matching checksums, a disturbed frame is rejected by the checksum, and the rolling counter advances without gaps over consecutive receptions:

```
AA 2D D4 68 B9 4A 0E 2D EA AA F7   33.4 C, checksum ok
AA 2D D4 68 B9 4A 00 2D EA AA E2   33.4 C, checksum ok
AA 2D D4 68 B9 4A 02 2D EA AA E5   33.4 C, checksum ok
AA 2D D4 68 B9 4A 04 2D DA AA 5E   two bits flipped, rejected by the checksum
```

15 consecutive frames at one per minute, none of them shifted, the counter
advancing without a gap. The one frame above that failed carries flipped bits
rather than a shifted payload - the checksum rejects it and the reading simply
stays at its previous value until the next transmission.

Resulting telemetry:

```json
{"TFAMarbella":{"Id":"68B94A","Temperature":33.4,"Counter":1,"BatteryLow":0,"LQI":4},"TempUnit":"C"}
```

**Documentation:** tasmota/docs#1566 adds the sensor page, the entry in the peripherals list and the `SetOption166` description.

**Related issue (if applicable):** —

## Checklist:

> On the ESP32 box: the code **builds** for `tasmota32`, but I have no ESP32 with a CC1101 to run it on, so I cannot honestly tick that line. Everything reported above was measured on ESP8266. Happy to have someone with the hardware confirm, or to restrict the driver to ESP8266 if you prefer.


tasmota/tasmota_xsns_sensor/xsns_121_tfa_marbella.ino
```diff
@@ -0,0 +1,544 @@
...
+    Off by default, because telemetry on TelePeriod is what Tasmota does
+    everywhere else and a driver should not quietly behave differently.
+
+    Worth turning on for this sensor though: it transmits about once a minute,
+    so with the default TelePeriod of 300 seconds four out of five readings
+    never leave the device, and the one that does can be almost five minutes
+    old. Every reading then becomes an MQTT message - useful for rules, but
+    also more traffic on a busy broker, which is why it is a choice.
+  */
+  if (Settings->flag6.marbella_publish) {     // SetOption166 - (TFA Marbella) Publish as they arrive (1)
```
arendst: Pls use SetOption147 for this or, as it transmits only once a minute, let the user set Teleperiod to 60 instead of 300.

pschroee: Done, thanks - SetOption166 is gone and flag6 bit 20 is spare again. The driver now publishes readings as they arrive and SetOption147 suppresses that, matching the serial bridge, IR, LoRaWAN and WizMote drivers.

I went with publishing by default rather than pointing users at TelePeriod 60: the sensor transmits about once a minute, so the default 300 s drops four readings out of five, and a receiving driver that stays quiet by default looks broken to whoever wires one up for the first time.

Two other things came out of verifying this on hardware, both in the same push:

**The radio parameters were wrong.** Both were taken from the rtl_433 device description and never checked against a real sensor. Measured with an SDR from three recordings, decoded at every sampling phase with the sensor's own checksum as the verdict:

| | was | measured |
|---|---|---|
| bit rate | 9523.8 Bd (105 µs/symbol) | **9400 Bd** |
| deviation | 30 kHz | **34.5 kHz** |

All three recordings peak at 9400 Bd and none produces a single valid frame at 9523.8. Over the 112 bits of a transmission a 1.3 % rate error drifts past a full bit, so the tail of every frame arrives shifted - which is why 9.6 kBit/s delivered nothing at all and 9.5238 only worked occasionally.

**New setting `Marbella <MHz>`.** The CC1101 derives its frequency from a 26 MHz crystal, and the module I have sat 27 kHz below what it was told. That still fits inside the 135 kHz receive filter, but it spends the margin the signal needs. Since it is a property of the board and not of the driver, it is stored in Settings (int16, 100 Hz steps, out of the filler after marbella_serial) rather than being a constant somebody has to rebuild for. A decimal point tells a frequency from a sensor id; ids now require the full six hex digits, so `Marbella 868` is refused instead of quietly binding to sensor 000868.

One caveat so the numbers are not oversold: most of the lost frames on my unit turned out to be a broken antenna connection, found after these changes. The parameter corrections are measured against the transmitted signal and stand on their own, but I cannot offer a clean before/after reception rate for them alone.

tasmota/include/tasmota_types.h
```diff
@@ -832,7 +832,9 @@ typedef struct {
   uint8_t       web_color2[2][3];          // EA0  Needs to be on integer / 3 distance from web_color
   uint16_t      zcdimmerset[5];            // EA6
 
-  uint8_t       free_eb0[20];              // EB0  20 bytes
+  uint32_t      marbella_serial;           // EB0  TFA Marbella sensor the receiver is bound to
+  int16_t       marbella_frequency;        // EB4  TFA Marbella receiver tuning offset in 100 Hz steps
+  uint8_t       free_eb6[14];              // EB6  14 bytes
```
arendst: As settings area becomes full I suggest to use Keeloq overlay for TFA as they are mutually exclusive.
```
  uint16_t      pulse_counter_debounce_high;  // FBA
  union {
    struct {
      uint32_t  keeloq_master_msb;         // FBC
      uint32_t  keeloq_master_lsb;         // FC0
      uint32_t  keeloq_serial;             // FC4
      uint32_t  keeloq_count;              // FC8
    };
    struct {
      uint32_t  marbella_serial;           // FBC  TFA Marbella sensor the receiver is bound to
      int16_t   marbella_frequency;        // FC0  TFA Marbella receiver tuning offset in 100 Hz steps
      uint8_t   free_fc2[2];               // FC2
      uint32_t  wizmote_comm_id;           // FC4  WizMote ESP-Now remote control comm id
      uint8_t   free_fc8[4];               // FC8
    };
  };
  uint32_t      device_group_share_in;     // FCC  Bitmask of device group items imported
 ```

PR #24976, opened 2026-08-24
Keep the first panic in the ESP32 crash recorder

tasict:
## Description:

On ESP32 the crash recorder can lose the frame of the original exception. When the panic handler stalls (seen repeatedly on a Sonoff ZBBridge Pro running `tasmota32-zbbrdgpro` 15.5.0.2: the `callx8 panic_handler` in `xt_unhandled_exception` never completed), the interrupt watchdog fires, re-enters through `panicHandler()` and `custom_crash_recorder()` overwrites the RTC record with a frame that only shows the panic handler itself. `Status 12` then reports this for every crash:

```
{"StatusSTK":{"Exception":6,"Reason":"IntegerDivideByZero","EPC":"40082307","EXCVADDR":"3ffda0b4","CallChain":["40082304","40081017","401ef59f","7ffda0b1"]}}
```

`40082307` is `xt_unhandled_exception`+0xf, `40081017` is `__wrap_xt_unhandled_exception`, and exccause 6 on the `panicHandler()` path is `PANIC_RSN_INTWDT_CPU1`, not an integer division — `edesc[]` only knows the Xtensa causes. The original exception (an `InstrFetchProhibited` from `esp_vApplicationIdleHook`) could only be reconstructed from the stale `EXCVADDR` and the tail of the call chain.

Changes in `support_crash_recorder.ino`:

- record only the first panic of a boot (`static bool crash_recorded`), so a nested entry cannot replace the useful frame;
- remember whether the record came in through `panicHandler()` (`pseudo_excause` in the RTC struct) and decode `exccause` with the `PANIC_RSN_*` names in that case (Xtensa `CrashDump()`);
- drop the duplicated `crash_recorder.magic = crash_magic` in `__wrap_xt_unhandled_exception`, the recorder sets it.

Same guard applied to the RISC-V branch; its decoding is unchanged.

Not compiled locally (no toolchain on this machine), relying on CI for the build. ESP8266 is not affected (ESP32-only code path).

## Checklist:

### DIFF
--- a/tasmota/tasmota_support/support_crash_recorder.ino
+++ b/tasmota/tasmota_support/support_crash_recorder.ino
@@ -250,18 +250,16 @@ static const char *edesc[] = {
 
 // exccause values set by the interrupt based entry points (int wdt, cache error, double exception, ...)
 // before they call panicHandler(), see esp_private/panic_reason.h
-static const char *pseudo_edesc[] = {
-    "Unknown", "DebugException", "DoubleException", "KernelException",
-    "CoprocException", "IntWdtCpu0", "IntWdtCpu1", "CacheError"
-};
-#define NUM_PSEUDO_EDESCS (sizeof(pseudo_edesc) / sizeof(char *))
+const char kPseudoEdesc[] PROGMEM = "Unknown|DebugException|DoubleException|KernelException|CoprocException|IntWdtCpu0|IntWdtCpu1|CacheError";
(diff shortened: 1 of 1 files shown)


s-hadinger:
Thanks looks good to me

tasict:
Moved the names to PROGMEM as a pipe separated `kPseudoEdesc` string read through `GetTextIndexed()`, same pattern as the other lookup tables.

PR #24993, opened 2026-09-01
MiEL HVAC Modbus: length-based framing, queue writes, FC03 sensor mirror

grzegorz914:
## Description

MiEL HVAC Modbus RTU slave follow-ups to #24982 / #24992.

### 1. Robust RTU receiver

A `FUNC_LOOP` poll cannot measure the 3.5-char inter-frame gap reliably above 9600 baud,
so on a busy 19200 8E1 multi-drop bus the slave counted more CRC errors than requests.
The receiver now:

- frames requests by **length** (computed from the function code), dispatching as soon as
  the bytes are buffered - no dependency on poll timing;
- **resyncs one byte at a time** on a framing/CRC mismatch instead of discarding the whole
  assumed frame and waiting for the bus to fall idle;
- counts a CRC error **only when the frame claims our own address** (other slaves' traffic
  and mid-resync noise no longer inflate the counter);
- uses a **1 kB RX buffer** and a bulk `read()` so a late poll does not lose bytes to a
  UART FIFO overflow (`TasmotaSerial` cannot report the overflow on ESP32);
- **skips `flush()` when no DE pin is configured** - on ESP32 `flush()` also discards the
  RX buffer, which dropped the next request on a shared bus;
- **waits the RTU turnaround silence** (up to 3.5 char times after the last received byte)
  before answering, so the response is not clipped by the master's line turnaround on an
  auto-direction transceiver;
- **suppresses a late reply** once another request is buffered behind it - after a
  `FUNC_LOOP` stall makes the response miss the master's timeout (Siemens S7 `MB_MASTER`
  STATUS `0x80C8`), the stale request is still applied but its answer is not put on the bus
  where it would derail the master's next transaction.

The idle gap is now only used to drop a stale sub-minimal fragment.

### 2. Queue writes when the HVAC link is not up yet

A write before the driver connected to the AC returned exception `0x04`. It now returns a
normal response and queues the update, matching the `HVACSetMode` / `HVACSetTemp` / ...
console commands. A PLC can pre-load settings and bench testing works without the AC.

### 3. Mirror sensor values on FC03 holding registers

Some PLCs / HMIs read holding registers (FC03) only. Mirror a selected subset of the
FC04 input registers directly after the writable control block (no gap):

| PLC ref | protocol | type | meaning | input mirror |
|---|---|---|---|---|
| 40016 | 0x000f | INT | room temperature C x10 | 30033 |
| 40017 | 0x0010 | INT | compressor (0 off, 1 on) | 30036 |
| 40018 | 0x0011 | INT | instantaneous power (W) | 30038 |
| 40019 | 0x0012 | INT | stage operation | 30057 |
| 40020 | 0x0013 | INT | stage fan | 30058 |
| 40021 | 0x0014 | INT | stage mode | 30059 |
| 40022 | 0x0015 | INT | diagnostics: requests received | 30081 |
| 40023 | 0x0016 | INT | diagnostics: CRC errors | 30082 |
| 40024 | 0x0017 | INT | connected to unit (0 no, 1 yes) | 30002 |

Read-only (`0x02` on write). FC03 now covers `0x0000..0x0017` contiguously; the control
block (`0x0000..0x000e`) is unchanged.

Verified with a `tasmota32s3` build and against `mbpoll` on a shared RS485 bus.

## Checklist

  - [x] The code change is tested and works with Tasmota core ESP8266 V2.7.9
  - [x] The code change is tested and works with Tasmota core ESP32 V3.x
  - [x] The code change pass the local tests


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Support for GUI tooltip on touch media like phones and tablets
 - ESP32 MiEL HVAC Modbus RTU slave on a second RS485 port exposing all states and functions for PLC use with `#define USE_MIEL_HVAC_MODBUS_SLAVE` (#24982)
 - MiEL HVAC Modbus slave mirrors room temperature, compressor, power, stage, link status and diagnostic counters on FC03 holding registers 0x000f..0x0017
+- MiEL HVAC Modbus swingH register reports the wide-vane position only, without the i-See sensor bit
```
arendst: Considering the amount of changes AND the lenghty messages I suggest you put this information in the head of your driver. I will then add a single line to the changelog reporting important changes have been made to your driver.

PR #25004, opened 2026-09-04
MiEL HVAC: fix HVACSetProhibit, add it to the web panel, apply sent changes optimistically

grzegorz914:
## Summary

Three fixes/additions around the MiEL HVAC settings path and its web control panel.

### 1. `HVACSetProhibit` never reached the unit

The remote-controller lock (power / mode / temperature) was accepted and echoed but
ignored by the unit, from the console and from the Modbus prohibit register alike. Two
things were wrong in the `0x41` set-settings request:

| | before | after (per [muart-group docs](https://muart-group.github.io/developer/it-protocol/0x41-set-request/0x01-set-settings)) |
|---|---|---|
| update flag | `1 << 13` → `0x0020` on the wire | `1 << 14` → `0x0040` |
| lock byte offset in the packet | payload 8 (the offset the **GET response** uses) | payload 11 |

The `prohibit` field in `miel_hvac_msg_update_settings` moved from payload offset 8 to 11
(`_pad1[4]` → `_pad1[3]` + `prohibit` + `_pad2[1]`); the CTASSERT was updated. Verified on
hardware (an i-See unit): `HVACSetProhibit power` / `mode` / `all` now take effect and are
reported back in `SENSOR`. No reference implementation (SwiCago HeatPump, muart itp-packet,
ESPHome mitsubishi_itp) writes prohibit, so there was nothing to cross-check against.

### 2. Prohibit on the web panel

Added as a `<select>` below "Air direction" (`off` / `power` / `mode` / `mode + power` /
`temp` / …), with a state-table row and the same live sync as the other controls.
`miel_hvac_web_optlabel()` gained a `_`→space fallback so unmapped keywords (`mode_power`, …)
read cleanly in the menus.

### 3. Sent changes are applied optimistically

For ~1&nbsp;s after a change, until the unit confirms with the next `0x62 0x02`, `sc_settings`
still held the pre-change state. Anything reading it in that window saw the old value: the
web panel control (e.g. the Mode segment after pressing **Off**) snapped back, and a PLC
reading the state back over Modbus re-issued the command. After sending the `0x41` update
the driver now writes the sent fields into `sc_settings` straight away — the same pattern
the driver already uses for the `0x08` run-state options. If the unit rejects the change,
its next report wins.

## Testing

- `tasmota` (ESP8266) and `tasmota32s3` builds compile without warnings (CTASSERT for the
  new struct layout passes).
- Prohibit set / read and the no-revert behaviour verified on hardware from the console and
  the web panel.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_44_miel_hvac.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_44_miel_hvac.ino
@@ -22,10 +22,22 @@
  * Mitsubishi Electric HVAC (CN105) serial interface
  *
  * Speaks the Mitsubishi "IT protocol" on the indoor unit's CN105 connector and exposes it
- * through the console (HVACSet* commands), MQTT (SENSOR / HVACSettings) and a climate
- * control panel on the web UI main page.  Protocol reference:
+ * through the console (HVACSet* commands), MQTT (SENSOR / HVACSettings) and a web control
+ * panel on the main page.  Protocol reference:
  * https://muart-group.github.io/developer/it-protocol/
  * Compile with USE_MIEL_HVAC; GPIOs "MiEl HVAC Rx" / "MiEl HVAC Tx".
+ *
+ * --- Web control panel (USE_WEBSERVER) ---------------------------------------------------
+ * Full climate panel on the main page: mode (Auto/Heat/Cool/Dry/Fan/Off), target
+ * temperature, fan speed, vertical + horizontal vane, air direction and the remote lock
+ * (prohibit).  Which controls appear is gated by the 0x7B 0xC9 Base Capabilities, and the
+ * panel is refreshed in place on every web_refresh poll so it also follows changes made
+ * from the IR remote / MQTT / console.
+ *
+ * A control change is written to sc_settings straight away, before the unit confirms with
+ * the next 0x62 0x02, so the panel, the Modbus registers and SENSOR show the intent
+ * instead of the stale pre-change state for ~1s; the unit's next report wins if it
+ * rejects the change.
  *
  * --- Modbus RTU slave (USE_MIEL_HVAC_MODBUS_SLAVE, ESP32) ---------------------------------
  * Optional second RS485 port that mirrors every driver state as read registers and maps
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - MiEL HVAC Modbus and CC1101 GPIO names shown as `MbsRelay ...` due to a duplicate entry in the GPIO name table (#24992)
 - MiEL HVAC Modbus length-based framing, queue writes, FC03 sensor mirror (#24993)
+- MiEL HVAC `HVACSetProhibit` (and the Modbus prohibit register) never reached the unit - wrong 0x41 update flag and lock-byte offset (#25004)
```
arendst: Pls add these to the head of your driver and stop adding them to the changelog; There is just too many changes and additions to overflrow the change log.

grzegorz914:
Done - dropped the changelog lines and moved the description into the doc block at the head of `xdrv_44_miel_hvac.ino`.

mamrai1:
I think the recent optimistic sc_settings update has introduced a regression in HVACSETTINGS publication.

After an HVACSet* command is sent, the requested fields are now copied immediately into sc_settings. This is useful for Web UI and Modbus responsiveness and should probably remain.

However, when the Mitsubishi subsequently confirms the command, the confirmed settings normally match the already optimistically updated sc_settings, so the existing memcmp() change detection may no longer trigger an MQTT settings publication.

A manual TelePeriod immediately publishes the correct state, confirming that the unit and internal state are correct.

I suggest publishing the dedicated HVACSETTINGS topic immediately after applying the optimistic settings locally.

If the indoor unit later rejects or modifies the command, the confirmed `0x62 0x02` will differ from the optimistic `sc_settings`, and the existing comparison logic can still publish the corrected `HVACSETTINGS` state\.

This keeps the new optimistic Web/Modbus behavior while restoring deterministic MQTT climate\-state feedback with minimal extra MQTT traffic\.

grzegorz914:
Thanks @mamrai1, good catch - confirmed. Fix in #25017: after the optimistic `sc_settings` apply the driver now publishes `HVACSettings` + `SENSOR` straight away, and if the unit later rejects or modifies the command the confirming `0x62 0x02` still differs from `sc_settings` so the existing comparison republishes the corrected state. One publish per command, same as before #25004.

PR #25017, opened 2026-09-08
MiEL HVAC: publish HVACSettings after an optimistic settings apply

grzegorz914:
## Summary

Follow-up to #25004. That PR made `miel_hvac_tick()` copy the fields of an
outgoing `0x41` set-settings request into `sc_settings` straight away, so the web
panel, the Modbus registers and `SENSOR` show the intended state during the ~1 s
before the unit confirms with the next `0x62 0x02`.

Side effect: when the confirming `0x62 0x02` arrives it now normally equals the
already-updated `sc_settings`, so the `memcmp()` in `miel_hvac_input_settings()`
no longer fires and the `HVACSettings` tele topic (and `SENSOR`) are not
published after an `HVACSet*` command. A manual `TelePeriod` shows the correct
state, so only the push was lost. Reported by @mamrai1.

Fix: publish `HVACSettings` + `SENSOR` right after the optimistic apply, the same
messages `miel_hvac_input_settings()` would have sent. If the unit rejects or
modifies the request, its `0x62 0x02` still differs from `sc_settings` and the
existing comparison republishes the corrected state. One publish per command,
matching the pre-#25004 behaviour.

## Testing

- `tasmota` (ESP8266) and `tasmota32s3` compile without warnings.
- `HVACSetMode` / `HVACSetTemp` from console and web panel: `HVACSettings` is
  published immediately again; no duplicate publish when the unit confirms an
  unchanged value; a unit-modified value still triggers a second publish.

## Checklist:

### DIFF
--- a/tasmota/tasmota_xdrv_driver/xdrv_44_miel_hvac.ino
+++ b/tasmota/tasmota_xdrv_driver/xdrv_44_miel_hvac.ino
@@ -39,8 +39,8 @@
  *
  * A control change is written to sc_settings straight away, before the unit confirms with
  * the next 0x62 0x02, so the panel, the Modbus registers and SENSOR show the intent
- * instead of the stale pre-change state for ~1s; the unit's next report wins if it
- * rejects the change.
+ * instead of the stale pre-change state for ~1s, and HVACSettings / SENSOR are published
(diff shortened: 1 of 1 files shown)


mamrai1:
Thanks @grzegorz914 ! Working as expected

PR #25027, opened 2026-09-12
MiELHVAC: add Home Assistant MQTT discovery

grzegorz914:
## Description:

Publishes a single retained MQTT discovery "climate" entity for the MiELHVAC driver (mode, current/target temperature, fan speed, action, vertical swing, horizontal swing/widevane, air direction as a preset), matching a manually written `climate:` YAML config, so the device shows up in Home Assistant with no manual configuration.

- Fan/swing/air-direction options show capitalized display labels (e.g. "Up Middle") via Jinja templates while still sending the driver's lower_snake_case values, the web control panel's own labels are updated to match (it previously used "Up-Middle" style hyphens, inconsistent with the discovery config)
- Widevane's "I-See" position is only offered once the unit is known to support it, and selecting it now re-engages air direction at the last active direction instead of sending a widevane byte the unit ignores, the web control panel gets the same fix (it previously hid "I-See" unconditionally)
- Adds a computed `HAAction` SENSOR field (heating/cooling/drying/fan/idle/defrosting/preheating/off), using the Stage packet's OperationStage (standby/defrost/preheat) as well as HAMode, so the discovery config doesn't need a template to derive it and HA gets the more specific action whenever the unit reports one
- Raises `MQTT_MAX_PACKET_SIZE` to 4096 in `my_user_config.h` (the combined discovery payload doesn't fit under the 1200 byte default)
- Republishes discovery on every MQTT (re)connect and honors SetOption19 like the rest of Tasmota's HA discovery

## Checklist:


mamrai1:
Is there a way to compile firmware with this commit for testing?

grzegorz914:
Just compile it locally

mamrai1:
> Just compile it locally

Compiled the firmware with the new ino file and also mqtt_max_packet_size and use_home_assistant. Working as expected with setoption19 0 ( 2 devices : 1 with tasmota native discovery and 2nd with mqtt discovery -climate). Great work !

mamrai1:
Just noticed some bug regarding hotizontal vane and dublicate entries
IMG_6675
IMG_6674

grzegorz914:
I don’t see any duplicated options.

mamrai1:
> I don’t see any duplicated options.

Left middle is the same as left center, also right middle and right center. Left and right center I think must be removed. There are 5 stops plus split and swing, total 7

grzegorz914:
MSZ-LN units have 11 positions

mamrai1:
> MSZ-LN units have 11 positions
Yes also my unit correctly reports 7+2, but the remote control only has 5+2.
At least we can dynamically expose to mqtt supported fan modes and preset modes. Not all units have quiet speed and I-see sensor. done some additions: https://github.com/grzegorz914/Tasmota/pull/2

grzegorz914:
Pushed an update based on @mamrai1's suggestion in grzegorz914/Tasmota#2 (thanks for digging into this).

- `fan_modes` is now built from `cap_fan_auto` and the fan count reported in the C9 capabilities response, instead of a fixed `Auto/Quiet/1-4` list.
- Quiet is dropped from MQTT discovery entirely. On some CN105 units, a Quiet selection made from the IR remote reports back over CN105 as fan speed 1, so it can't be represented as a distinct, reliable HA fan mode.
- `preset_modes` (AirDirection) is only advertised on units that are `isee_capable` (vertical vane + observed i-See), instead of unconditionally.
- Discovery is now republished once C9 capabilities arrive and once i-See is first observed, not just on MQTT connect, so HA picks up the correct capabilities without a reconnect.

Built and tested on `tasmota32s3`.

grzegorz914:
Follow-up pushed after testing on real hardware:

- Web panel now uses the same capability gates as MQTT discovery (Fan speed, Vane horizontal I-See option, Air Direction control and its SENSOR JSON field, readout row) — previously they disagreed on defaults when capabilities were still unknown.
- Quiet fan mode: initially dropped it from MQTT entirely per the suggestion above, but testing showed units reporting 5 fan speeds (Auto/Quiet/1-4) report Quiet back correctly and reliably. Reverted to gating Quiet on `fan_count == 5` (the same threshold `SetFanSpeed` already enforces) instead of removing it outright, in both MQTT and the web panel.
- `swing_horizontal_modes` / the web panel's Vane horizontal list are now ordered left-to-right (left, left_middle, left_center, center, right_center, right_middle, right, split, swing) instead of raw protocol byte order.

Also looked into whether the C9 capabilities response encodes the number of horizontal vane positions (to address the duplicate-looking Left Middle/Left Center etc. on units with fewer physical positions) — checked the mUART project's reverse-engineered protocol docs and source, and confirmed there's no such field. Even their own capability parser hardcodes horizontal vane support to `true` with a "things that have to exist, but we don't know where yet" comment. Leaving that as-is since there's no reliable way to detect it.

grzegorz914:
Added the Prohibit select suggested in grzegorz914/Tasmota#2 — publishes as a separate \`homeassistant/select/<id>_prohibit/config\` entity (climate has no native lockout concept), mirroring the web panel's Prohibit dropdown, on the same HA device as the climate entity. Its retained config also gets cleared when SetOption19 disables discovery.

One note for anyone testing: it's published with \`entity_category: "config"\`, since it's a device-lockout setting rather than a primary control. That means Home Assistant deliberately keeps it off the Area dashboard's auto-generated card even once an Area is assigned to the entity — that's standard HA behavior for config/diagnostic entities, not a bug. It's still reachable from the device page and from Settings → Areas → [area] → Entities.

Tested on real hardware (tasmota32s3).

grzegorz914:
Follow-up on horizontal vane naming, tested on real hardware (dual-vane MSZ-LN unit):

- \`left\`/\`center\`/\`right\`/\`split\`/\`swing\` move both vanes together; \`left_middle\`/\`left_center\` move only the left vane, \`right_middle\`/\`right_center\` only the right vane.
- Fixed \`right\`/\`right_middle\` being backwards: \`right\` now means the furthest-right extreme (matching what \`left\` already meant), \`right_middle\` the lesser, synchronized position. This was confirmed by testing rather than assumed from the general CN105 protocol docs, which describe a different, single-vane-oriented scheme for these raw values that didn't match this unit's actual behavior.
- Documented in code that this per-side, dual-vane behavior is likely model-specific (no capability bit indicates vane count), so it shouldn't be assumed to generalize to other units.

Built and tested on `tasmota32s3`.

grzegorz914:
Fixed a display bug: the web panel's "Vane horizontal" selector masked \`widevane\` with \`MIEL_HVAC_SETTINGS_WIDEVANE_MASK\` (0x0f) before comparing it to the current value, but I-See is signaled by the high bit (0x80) or one of two odd raw values (0x28/0xaa) seen in the wild -- none of which survive that mask. So the dropdown could never show I-See as selected, even right after engaging AirDirection (which sets widevane to exactly 0x80). Fixed by checking for i-See before masking, reusing the same check already used correctly elsewhere (readout rows, Air Direction control).

Tested on real hardware.

grzegorz914:
Fixed a second instance of the same class of bug: the live-sync JS path (which pushes updates into the already-open web panel without a full reload, respecting focus so it never yanks a control you're actively using) computed the Vane horizontal sync value as NULL whenever i-See was active, so it silently skipped updating that selector instead of syncing it to "I-See".

Tested on real hardware.

mamrai1:
> Follow-up on horizontal vane naming, tested on real hardware (dual-vane MSZ-LN unit):
> 
> * `left`/`center`/`right`/`split`/`swing` move both vanes together; `left_middle`/`left_center` move only the left vane, `right_middle`/`right_center` only the right vane.
> * Fixed `right`/`right_middle` being backwards: `right` now means the furthest-right extreme (matching what `left` already meant), `right_middle` the lesser, synchronized position. This was confirmed by testing rather than assumed from the general CN105 protocol docs, which describe a different, single-vane-oriented scheme for these raw values that didn't match this unit's actual behavior.
> * Documented in code that this per-side, dual-vane behavior is likely model-specific (no capability bit indicates vane count), so it shouldn't be assumed to generalize to other units.
> 
> Built and tested on `tasmota32s3`.

 ‘left center’ is like this: left middle position for left vane and center for right vane.
Right center : right middle position for right vane and center for left vane.

left and right middle positions move both vanes .

mamrai1:
Excellent work @grzegorz914 ! Thanks

tasmota/my_user_config.h
```diff
@@ -118,6 +118,11 @@
 #define MQTT_SOCKET_TIMEOUT    4                 // [MqttTimeout] Number of seconds before Mqtt connection timeout
 #define MQTT_WIFI_CLIENT_TIMEOUT 200             // [MqttWifiTimeout] Number of milliseconds before Mqtt Wi-Fi timeout
 
+#undef  MQTT_MAX_PACKET_SIZE                     // ESP32 builds set this to 1200 via a -D build flag (platformio_tasmota32.ini)
```
arendst: Make this dependant on your enabled MIEL HVAC driver to keep default 1200.

grzegorz914:
@mamrai1 correction to my reply above — my first description of this unit's behavior was based on an incomplete initial test, and it was wrong. A full retest gives:

- `left`/`center`/`right`/`split`/`swing`: both vanes together
- `left_middle`/`right_middle`: both vanes together
- `left_center`: left vane to left-middle, right vane to center (asymmetric split)
- `right_center`: mirror of that (left vane to center, right vane to right-middle)

That matches what you reported almost exactly, not "a genuinely different pattern" as I said before. So this may not actually be two different behaviors, just my earlier report being wrong. Leaving the code documented as model-specific rather than assuming this generalizes either way, since we still only have two data points and no capability bit to tell units apart.

mamrai1 (commented):
Nice! Unfortunately can’t test it on both my units.. Econocool is shown as not supported on tasmota. Engaging it on the remote works but one unit reports only the +2 on temp and auto swing and the other only the auto swing, even though the vertical vane is oscillating up and down

grzegorz914:
@mamrai1 thanks for testing this. Since EconoCool leaves an observable side effect (temp offset, auto swing engaging) even without a genuine 0x42 confirmation, it might be detectable the same way we ended up handling horizontal-vane support (round-trip evidence from what actually changes, rather than relying on the unreliable capability bit).

Could you grab a log for one of your units? Set SerialLog 4 (or WebLog 4 in the console), then:

1. Capture a SENSOR reading (or the console log lines) right before engaging EconoCool from the remote
2. Engage EconoCool from the remote
3. Capture SENSOR again right after

If you can do this for both units that would help even more, since you mentioned they react differently (temp+2 and auto swing on one, only auto swing on the other). I want to diff SettingsHex/StatusHex/StageHex byte by byte to see exactly what moves, the same way we tracked down the horizontal-vane detection.

grzegorz914:
@mamrai1 merged your horizontal-vane support detection from grzegorz914/Tasmota#3, thanks for that (VaneHSupported starting on and getting cleared only after positive no-support evidence, since there is no reliable C9 bit for it).

Added one thing on top: a one-shot probe on boot. Right after the first genuine settings read, if i-See is not already active, the driver nudges the horizontal vane to a different position and immediately restores the original position once the round trip resolves. That way support gets established automatically instead of only after someone happens to send HVACSetSwingH by hand, and a device reboot no longer means the controls incorrectly reappear on a GE35VA-class unit until manually re-tested.

I don't have a unit without a horizontal vane to verify the no-support path, my own units both have working dual vanes. Could you test this on your GE35VA/GE71VA when you get a chance? Specifically interested in whether VaneHSupported correctly comes back off on its own after a reboot now, without you needing to touch SwingH manually.

mamrai1:
> @mamrai1 thanks for testing this. Since EconoCool leaves an observable side effect (temp offset, auto swing engaging) even without a genuine 0x42 confirmation, it might be detectable the same way we ended up handling horizontal-vane support (round-trip evidence from what actually changes, rather than relying on the unreliable capability bit).
> 
> Could you grab a log for one of your units? Set SerialLog 4 (or WebLog 4 in the console), then:
> 
> 1. Capture a SENSOR reading (or the console log lines) right before engaging EconoCool from the remote
> 2. Engage EconoCool from the remote
> 3. Capture SENSOR again right after
> 
> If you can do this for both units that would help even more, since you mentioned they react differently (temp+2 and auto swing on one, only auto swing on the other). I want to diff SettingsHex/StatusHex/StageHex byte by byte to see exactly what moves, the same way we tracked down the horizontal-vane detection.

Unfortunately both units don’t send anything on cn105, except auto swing. The behaviour with the +2 on temp is related with a specific remote control I use when selecting econocool.

grzegorz914:
@mamrai1 thanks for checking. That settles it then, no reliable CN105 signal to key off for EconoCool on your units, and auto swing alone is too generic to use as a stand-in (could be triggered by plenty of other things). Matches what we found independently on Purifier/NightMode/EconoCool for a different unit in this thread, some of these options just aren't wired up over CN105 on certain units even though the remote clearly has them.

Leaving EconoCool as-is (not_supported, hidden from HA/web panel until a genuine round trip confirms it, same as Purifier/NightMode). Not pursuing side-effect detection for it, there's nothing to detect.

mamrai1:
> @mamrai1 merged your horizontal-vane support detection from [grzegorz914#3](https://github.com/grzegorz914/Tasmota/pull/3), thanks for that (VaneHSupported starting on and getting cleared only after positive no-support evidence, since there is no reliable C9 bit for it).
> 
> Added one thing on top: a one-shot probe on boot. Right after the first genuine settings read, if i-See is not already active, the driver nudges the horizontal vane to a different position and immediately restores the original position once the round trip resolves. That way support gets established automatically instead of only after someone happens to send HVACSetSwingH by hand, and a device reboot no longer means the controls incorrectly reappear on a GE35VA-class unit until manually re-tested.
> 
> I don't have a unit without a horizontal vane to verify the no-support path, my own units both have working dual vanes. Could you test this on your GE35VA/GE71VA when you get a chance? Specifically interested in whether VaneHSupported correctly comes back off on its own after a reboot now, without you needing to touch SwingH manually.

Tested on both units. Working as expected on GE35VA with no horizontal vane

PR #25029, opened 2026-09-15
Dali2 input device events

murfett-au:
## Description:

**Related issue (if applicable):** n/a

Second of three DALI-2 input device changes. Depends on [#1](https://github.com/murfett-au/Tasmota/pull/1) (event message decoding); review only the last commit until that merges.

DALI-2 control devices (IEC 62386-103) such as push buttons, occupancy and light sensors use 24-bit frames and have their own short address space, separate from control gear. Until now they could only be commissioned with raw `DaliSend255` frames.

**New commands**

- `DaliDeviceScan 1|2[,<max_count>]` assigns short addresses with the 24-bit INITIALISE / RANDOMISE / COMPARE / PROGRAM SHORT ADDRESS sequence. Mode 1 clears all control device short addresses first. Mode 2 probes the bus and only addresses devices without one, so existing addresses are never reused. Returns the number assigned, like `DaliScan`.
- `DaliDevice` lists control devices that answer on short addresses 1..64, in the same form as `DaliGear`.
- `DaliDevice <device>` reports version, status and each instance's type, enabled flag and event scheme.
- `DaliDevice <device>,<scheme>` sets the event scheme on all instances of a device (0 = broadcast). Input devices default to the Instance scheme, which carries no short address, so scheme 2 (DeviceInstance) is what makes commissioning useful: every event then identifies the device that sent it.

**Implementation notes**

- The binary search shared with `DaliScan` gained a `device` flag defaulting to false, so control gear commissioning is unchanged.
- IEC 62386-103 constants are kept in the driver to leave the TasmotaDali library untouched.
- Control gear and control device short addresses are independent name spaces, so the two scans are deliberately separate commands.

**Tested with:** zencontrol zc-switch rotary (3 push button instances) on an ESP32-S3-Pico with the Waveshare Pico-DALI2. Sequence: `Backlog DaliDeviceScan 1; DaliDevice; DaliDevice 1; DaliDevice 1,2`, then confirmed rotation events arrive with `"Scheme":"DeviceInstance"` and an `Address` field.

Documentation for the Commands section of `docs/DALI.md` will follow in tasmota/docs.

## Checklist:


s-hadinger:
I don't have such devices, @arendst ok to merge?

PR #25044, opened 2026-09-20
Cap Berry GC threshold growth on PSRAM to avoid multi-second stalls

LesNewell:
## Description:

On boards with PSRAM with heavy Berry usage, the GC threshold-tightening logic added in #13295 is skipped entirely (`!UsePSRAM()` guard), so the default steprate allows Berry's heap to grow unbounded until a GC finally sweeps tens of thousands of objects out of a nearly-full pool in one go.

Measured on an ESP32-P4 board running a Berry/LVGL/HASPmota UI: a GC that fired after usage grew to 2.6MB took **4.4 seconds** to sweep ~25k freed objects, causing a multi-second stall of the main loop. Normal collections on the same device complete in 20-30ms.

After the fix, 36 GCs over ~7.5 minutes all completed in 8-31ms with usage capped around 1.4MB.

## Checklist:
  - [x] The code change is tested and works with Tasmota core ESP32 V.3.3.8 (ESP32-P4)
  - [x] I accept the CLA.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


s-hadinger:
Oh wow, this is a nasty bug

s-hadinger:
CI is fixed in later commits, so it would require a rebase. I'm merging since the change is harmless.

